PR #4148, opened 2014-10-28
Enable adapting to screen layout changes on Linux

moben:
use CheckTypeSize instead of CheckSymbolExists

From http://www.cmake.org/cmake/help/v3.0/module/CheckSymbolExists.html :
    If the symbol is a type or enum value it will not be recognized
    (consider using CheckTypeSize or CheckCSourceCompiles).

Also fixes issue4036 (https://github.com/synergy/synergy/issues/4036).
The code was there, the problem was just that someone fucked the cmake check up

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -200,7 +200,10 @@ if (UNIX)
 		set(CMAKE_INCLUDE_PATH "${CMAKE_INCLUDE_PATH}:/usr/local/include")
 
 		set(XKBlib "X11/Xlib.h;X11/XKBlib.h")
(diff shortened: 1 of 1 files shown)


XinyuHou:
@moben 

moben:
This should fix both issue 4036 and issue 4420 as it enables the code that handles those cases. It just doesn't get compiled without the define being set

nlyan:
Merged in v1.8.4

mwilck:
This change caused a regression for me, issue #5831.

PR #4194, opened 2014-11-07
System tray is unavailable on KDE5

cnsworder:
running when system tray is unavailable, but show message to user don't close window.


XinyuHou:
@cnsworder 

Could you attach a screenshot please?

nlyan:
Merged. Bad solution is better than no solution.

Issue #2866, opened 2014-10-13
High client CPU usage after being disconnected. CSocketMultiplexer null job buildup.


nbolton:
**Imported issue:**
- Author: Brian Vincent
- Date: 2011-08-11 23:19:43
- Legacy ID: 2935
- Version: 1.4.3

**Steps to reproduce:**  
1. Start up a synergy client that attempts to connect to a server that is refusing connections.  
2. Wait a day or longer.  
3. Connect to the client and move around the mouse.  

Expected: Normal smooth mouse movement with minimal CPU.  
Actual: Extremely laggy mouse movement and use a lot of CPU.  

Here is my assessment.  If you add a debug message to output the size of the list m_socketJobs inside of CSocketMultiplexer, you'll notice the list growing every second when it tries to connect to the server.  This is causing a memory leak and causing the CPU to rise every time it traverses the entire list with nextCursor.  

This appears to be the same bug as bug #65, but that was 2 years ago so maybe not. It certainly seems like the bug from this discussion:  
http://groups.google.com/group/synergy-plus/browse_thread/thread/b5ab3ff4c7449493  

Here is my patch to src/lib/net/CSocketMultiplexer.cpp that appears to fix it.  

--- CSocketMultiplexer.cpp.orig 2011-08-11 17:08:31.000000000 -0500  
+++ CSocketMultiplexer.cpp      2011-08-11 17:11:09.000000000 -0500  
`@ -255,6 +255,7 @`  
                for (CSocketJobMap::iterator i = m_socketJobMap.begin();  
                                                        i != m_socketJobMap.end();) {  
                        if (*(i->second) == NULL) {  
-                               m_socketJobs.erase(i->second);  
                              m_socketJobMap.erase(i++);  
                              m_update = true;  
                      }  

**Versions (Synergy, OS):**  

Server:  
Client 1: 1.4.3  

**Workaround:**  

Restart the client.

jrconlin:
Not sure if this is fully related, but I'm observing a similar behavior with my current configuration:

Synergy:1.7.3 configured as Client
Host: Windows 8.1 Pro

Server seems not to matter. However if the network is not responding or unavailable, procexp will report that a thread will consume a LARGE amount of CPU, to the point where the fan starts kicking on an i7 with 8 cores. This is particularly noticeable when coming out of sleep. 

[...]

PR #4228, opened 2014-11-17
High client CPU usage and laggy mouse

galkinvv:
I've encountered Issue #2866 in 1.6.1, so patch from tracker is still actual.

Detailed steps and investigation is in the issue: https://github.com/synergy/synergy/issues/2866#issuecomment-58849356

### DIFF
--- a/src/lib/net/SocketMultiplexer.cpp
+++ b/src/lib/net/SocketMultiplexer.cpp
@@ -243,6 +243,7 @@ SocketMultiplexer::serviceThread(void*)
 		for (SocketJobMap::iterator i = m_socketJobMap.begin();
 							i != m_socketJobMap.end();) {
 			if (*(i->second) == NULL) {
+				m_socketJobs.erase(i->second);
 				m_socketJobMap.erase(i++);
 				m_update = true;
 			}


nbolton:
Thanks for the PR. On what OS do you see the laggy mouse?

galkinvv:
I've encountered the laggy mouse on the Windows7 x64 machine running synergy client. My server is Debian Linux.

But the origin of the problem as described in https://github.com/synergy/synergy/issues/2866#issuecomment-58849356 seems to be OS-independent.

Though it may be toolchain-dependent how fast is iteration over huge filled m_socketJobs, so the acctual cpu usage and mouse slownes may be OS-dependent.

nbolton:
Thanks @galkinvv

nbolton:
Thanks for the pull request! Buildbot is now [building](http://buildbot.synergy-project.org/builders) this PR, which usually takes about 30 minutes. When done, it'll upload the packages to our [nightly builds page](http://synergy-project.org/nightly?filter=pr-galkinvv-mac-lag-fix). Happy testing!

nbolton:
Dang, this doesn't seem to solve the problem for Windows server and Mac client. I installed pr-galkinvv-mac-lag-fix-0eaaf53 on both server and client.

nbolton:
@XinyuHou @speaker I think the branch name `pr-galkinvv-mac-lag-fix` is probably misleading. Perhaps we should rename to `pr-galkinvv-cpu-usage-fix`?

galkinvv:
I don't think the problem with lags on Mac client is related to this. This problem never appears immediately but only after a some time client was working.

nbolton:
@galkinvv Do you think its a a Linux server CPU issue or a Mac client CPU issue?

galkinvv:
I doesn't encouter mac problem myself (I have no OS X). I've read comments and thought about how the behaviour described in comments can correspond to problem solved by this PR. The problem solved here internally is "resource leakage" that began worse and worse over the time and never becomes better until client is restarted. 

https://github.com/synergy/synergy/issues/4132 describes the behaviour that ir quite random, but doesn't goes worse during running client time.

galkinvv:
(close was my mistake, I thought "Close" means "Close window")

nbolton:
@galkinvv No problem :)

visit1985:
I can confirm this on Arch Linux too. We even have a [special build](https://aur.archlinux.org/packages/synergy-fixed/) of synergy just to solve this issue.

nbolton:
@visit1985 Does this pull request solve the issue for you?
http://synergy-project.org/nightly?filter=pr-galkinvv-mac-lag-fix

visit1985:
Sorry, I'm not able to test this. At the moment I only have one monitor.

But maybe somebody else can:

To reproduce this, you need to to connect a synergyc client to a server and then shutdown the servers synergyc for about one week. When the client connects back, you should see much higher cpu load and lagging mouse cursor on the client screen.

PR #4290, opened 2014-12-23
Remove auto Alt+Printscreen on Windows

foliveira:
As per issue https://github.com/symless/synergy/issues/4288

### DIFF
--- a/src/lib/platform/MSWindowsKeyState.cpp
+++ b/src/lib/platform/MSWindowsKeyState.cpp
@@ -1071,11 +1071,6 @@ MSWindowsKeyState::getKeyMap(synergy::KeyMap& keyMap)
 			}
 		}
 
-		// add alt+printscreen
-		if (m_buttonToVK[0x54u] == 0) {
-			m_buttonToVK[0x54u] = VK_SNAPSHOT;
-		}
-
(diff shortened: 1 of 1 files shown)


the-wes:
I don't know about this. I see the function has been changed to use the keycode for PrintScreen rather than the "snapshot" virtual key. I just wonder if it was like that for a reason, or if there is a better virtual key we could use instead. Or am I just worrying too much?

nbolton:
@the-wes Would you like us to use this one? Did you test it? @speaker

EvzenP:
Almost a year after a fix was committed... and this damned bug is still present and driving me nuts every single day :(.
Will this fix ever be released? Are we waiting for the-wes, who doesn't seem to care anymore?

shelwig:
This behavior is really frustrating. When will the fix be released?

EvzenP:
Hello?!?! Anybody reading it here?! Can this be FINALLY released?!

XinyuHou:
@foliveira 
I can't seem to create a branch for this PR. Any suggestion?

ryan-codingintrigue:
@XinyuHou Can't you just create a new PR and reference this one? It's only 5 removed lines and would be a valuable fix.

PR #4454, opened 2015-03-20
Fix #4349 - Bringing windows to front

### LINKED ISSUE #4349: Mouse click does not always bring window to front
**Dev notes:** Effects Windows 8.1 and Windows 10. Effects both server and client. Stopping Synergy solves the issue. Also known as the window "focus bug" (though it does actually focus the window, it just doesn't bring it to front).

--

Server: Windows 10 TP (Build 9926), Client: 9932
Synergy 1.6.2 - Admin / Elevated Mode (Compatibility settings/Admin settings do not affect the outcome)

Repro:

1. Run synergy on both machines (I'll call server and client)
2. Switch applications on server (EX: Lync to Chrome, etc)
3. Move mouse to client, try to click another application other than the one currently focused.(Example: I am using Chrome and I have outlook running but it's behind chrome, try running outlook)

Expected: Other application is brought to the front and gains focus
Actual: about 50% of the time, the previous application which had focus stays on top while the other app becomes focused by the input.  In other words, I cannot force the window to be brought to the front.

Current work around: Clicking the application which has focus then clicking back in to the other app usually fixes the issue; however, this becomes extremely annoying after time....  *(Note, the current work around does not work with all apps - IE and Chrome are not playing nicely and chrome refuses to allow IE to take focus)

(see attached image - in this screenshot, chrome is the active window yet Synergy refuses to allow chrome to be brought to the front)

untitled

legonigel:
Fixes issue #4349 in MSWindowsDesks.cpp by using HWND_TOP instead of HWND_TOPMOST. All credit for this fix goes to @gitscrubz and @dubuaych

### DIFF
--- a/src/lib/platform/MSWindowsDesks.cpp
+++ b/src/lib/platform/MSWindowsDesks.cpp
@@ -435,8 +435,7 @@ MSWindowsDesks::destroyClass(ATOM windowClass) const
 HWND
 MSWindowsDesks::createWindow(ATOM windowClass, const char* name) const
 {
-	HWND window = CreateWindowEx(WS_EX_TOPMOST |
-									WS_EX_TRANSPARENT |
+	HWND window = CreateWindowEx(WS_EX_TRANSPARENT |
 									WS_EX_TOOLWINDOW,
 								reinterpret_cast<LPCTSTR>(windowClass),
 								name,
@@ -596,7 +595,7 @@ MSWindowsDesks::deskLeave(Desk* desk, HKL keyLayout)
 			w = m_w;
 			h = m_h;
 		}
-		SetWindowPos(desk->m_window, HWND_TOPMOST, x, y, w, h,
+		SetWindowPos(desk->m_window, HWND_TOP, x, y, w, h,
 							SWP_NOACTIVATE | SWP_SHOWWINDOW);
 
 		// if not using low-level hooks we have to also activate the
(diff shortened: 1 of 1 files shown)


legonigel:
May need to add MSWindowsScreen.cpp changes from https://github.com/gitscrubz/synergy/commit/a64906d539c7b165763607107a280bbf04ea1624

raphendyr:
Interesting history about the arguments: http://blogs.msdn.com/b/oldnewthing/archive/2005/11/21/495246.aspx

dubuaych:
I definitely didn't (and don't) need any changes from MSWindowsScreen; I tried to find the minimum set of changes needed. I have been going for over three hours with this change and no problems encountered so far. Further testing might reveal the need; I guess time will tell.

Thanks for your help.

dubuaych:
@raphendyr; Yeah I saw that too. Plus I remember reading the same article many years ago. :)

nbolton:
@gitscrubz @dubuaych Aha! @legonigel created the PR for us, thanks!

@speaker @XinyuHou

nbolton:
Created branch `legonigel-focus_fix_gitscrubz` which is building now. When its finished it should appear here: http://synergy-project.org/nightly?filter=legonigel-focus_fix_gitscrubz

Thanks again to @gitscrubz and @legonigel for their work.

alexwhitman:
I've just installed `synergy-1.7.1-legonigel-focus_fix_gitscrubz-447169d-Windows-x64.msi` and so far it seems like it has fixed the issue.

XinyuHou:
@dubuaych @legonigel @gitscrubz @nbolton 
I confirm synergy-1.7.1-legonigel-focus_fix_gitscrubz-447169d-Windows-x64.msi solves bring to front issue on both my Windows 8 server and client.

Thank you for the contribution!

nbolton:
@dubuaych @legonigel @gitscrubz Please feel free to add me on Skype: **nickbolton2705**

PR #4496, opened 2015-03-31
Added support for not signing the application bundle

guyzmo:
- do not enforce mac identity usage on qt5.4 ;
 - when no mac identity is provided do not try to
   use the codesign argument of macdeployqt ;
 - if a mac identity is provided, just use it.

Signed-off-by: Guyzmo <guyzmo+github@m0g.net>

### DIFF
--- a/ext/toolchain/commands1.py
+++ b/ext/toolchain/commands1.py
@@ -788,7 +788,10 @@ class InternalCommands:
 
 		print output[1]
 		if "ERROR" in output[1]:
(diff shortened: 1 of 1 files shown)


guyzmo:
this patch is mostly to enable builds without code-signing at all (which work fine) on 10.10 with qt5.4. When running it through brew, I get tons of `test: identity not found`, and without identity at all, it's working just fine.

I'm pushing the patch for the 1.6.3 homebrew build until integration and validation!

PR #4504, opened 2015-04-05
Improve Korean language description

conr2d:
correct Language.xml indent and replace "한국의" with "한국어" in the meaning of "Korean Language", not "of Korea"


XinyuHou:
@conr2d 
Please test this nightly.
http://synergy-project.org/nightly?filter=synergy-pr4504-improve-korean-language-description-beta-45efb6f

Issue #4519, opened 2015-04-13
out-of-sync modifier state on screen lock 


jtbraun:
Client: Win7, with elevated mode enabled
Server: Ubuntu 14.04

On the client, I created a shortcut.lnk with the Target:
    C:\Windows\System32\rundll32.exe user32.dll,LockWorkStation
And assigned a shortcut key "Ctrl+Alt+L"

Repro:
* Bring up the onscreen keyboard (so you can view Window's view of the modifier keys)
* press and hold ctrl+alt (keep them pressed until later)
* tap the L key (continue holding the modifiers)
* Observe the screen lock on the client.  Also note that synergy becomes (temporarily) disconnected while the screen locks. The onscreen keyboard shows Ctrl and Alt continue to be pressed. Release Ctrl and Alt.  The on-screen keyboard still shows ctrl-alt pressed. Even if you wait a bit for synergy to reconnect, move the mouse back to the client, release Ctrl-Alt, it still stays pressed.
* From now on, windows and Synergy disagree about the state of the modifier keys. Sometimes you can unwedge things by pressing/releasing Ctrl and Alt on the physical keyboard on the client system and synergy. Typically (while viewing the onscreen keyboard), pressing and releasing ctrl on the server system will cause the Windows client to think Alt is pressed, and pressing/releasing Alt will cause Windows to think Ctrl is pressed. Sometimes it seems to get stuck such that only a reboot of the client system will unstick things.

Workarounds:
* Releasing the modifier keys before the screen has a chance to lock appears to work (assuming you're fast).
* Setting up a different keystroke (Ctrl-Alt-K) on the server, and having that send Ctrl-Alt-L to the client works, as long as the mouse cursor is not on the affected client screen at the time of the locking (so that the synergy client doesn't have any modifier keys held down at the time of the screen lock).
* Generally, I've found pressing the client's local keyboard ctrl-alt keys causes the keys to unstick.  This doesn't always "work" though, as there's still stored state in the synergy client/server that causes confusion with later keypresses.

From the client log (attached shortly), it seems like what happens is:
* line 8: the synergy client detects a switch to Winlogon.
* line 12: before shutting down, it sends 01d and 038 up (left Ctrl and Alt I assume).
* line 94-101: at startup of the client on the Winlogon session, it detects: DEBUG1: modifiers on update: 0x0006
[...]

Issue #4449, opened 2015-03-18
Arrow keys and ctrl+arrow keys cause screen rotation on Windows 7 Enterprise


dckornbluh:
I'm using Synergy on my Dell laptop running Windows 7 (Synergy server) and a Dell desktop running Ubuntu 12.04 LTS.

Since upgrading to 1.6.2 I have twice had the problem where pressing a key on the laptop causes the screen to rotate either 90 or 180 degrees. In the first incident, I was editing an email. When I pressed ctrl+left arrow, the screen rotated 90 degrees to the left. The other ctrl+<arrow> combinations also caused screen rotation. In the second incident, it was  the arrow key alone that cause the screen to change orientation.

My gut feel is that undocking and redocking may have caused the software to lose its mind.

Server specs:
Dell Latitude E5440 i5-4300U CPU @ 1.90GHz 2.50 GHz
4 GB RAM
connected to Planar PLL 2410W monitor 
Windows 7.1 Enterprise 64-bit with Service Pack 1

Client specs:
Dell desktop
Ubuntu 12.04 LTS

beporter:
I bet this is actually the "sticky modifiers" issue (originally #9, now #3823 and/or #4057). Normally `Ctrl-Arrow` should only move your cursor, but I bet the configured hotkey on your system for screen rotation is `Ctrl-Shift-Arrow` or `Ctrl-Alt-Arrow`.

Synergy has a problem where sometimes the Shift, Alt, or Ctrl keys will act as if they are being held down even when the keys aren't being physically pressed on any of the keyboards. This causes the system to register plain key press differently since there's an "extra" modifier key that's active at the time. Typing `p` can surprisingly become `Ctrl-p`, for example.

Usually it works to just press each modifier a few times to "unstick" them and get Synergy back into correct shape.

Unless you use the screen rotation features regularly, probably the best thing to do is disable those hotkeys in your graphics settings since this is a difficult situation to recover from. If you continue to have issues with modifiers getting stuck down beyond that, your only real options are to downgrade Synergy back to the version where you didn't have the problem, or learn to live with it (this bug has been around for a **very** long time.)

beporter:
Signal boosting: There is [a new nightly build available](https://github.com/synergy/synergy/pull/4564#issuecomment-95926896) that may help fix this issue. If anyone monitoring this thread is able to test it, it would be appreciated.

Issue #4057, opened 2014-10-13
Command key stuck (Mac client, Windows server)


nbolton:
**Imported issue:**
- Author: Mike Wyant
- Date: 2014-08-14 14:56:50
- Legacy ID: 4127

**Steps to reproduce:**
1. _Important:_ Provide numbered steps to reproduce the bug.
2. If you don't provide steps, your issue will be marked as invalid.
3. Replace this example text with your own (or you will look silly).
4. You should also provide: expected, actual, versions, workarounds, etc.

Expected: Daily functioning continues unabated

Actual: After several minutes to an hour of use, the remote machine will begin acting like the Command key is held down.  Sometimes using Command-Tab to switch through apps will reset this, but it's about a 50%% chance of success.

**Versions and operating systems:**

Server: Windows 7 Home Premium SP1
Client: Mac OS X 10.9.4

**Temporary workarounds:**
As mentioned, sometimes using Command-Tab will reset this functionality; otherwise I need to close down the connection (either from client or server) to get full functionality back,

**Similar bugs:**

There are some others related to modifier keys sticking, but this one is very specific and happens whether Office or Photoshop is open at all

**Additional comments:**

TODO

128keaton:
Same problem here, (Win8.1 x64, 10.10.2) really sad :(

cra5370:
I'm experiencing this issue as well on a consistent enough basis that I've gotten used to spam pressing the command key after I use it so that it clears the stickiness (which only works half the time). Would love to see this fixed.

adamnovak:
I can reproduce this with an Ubuntu 12.04 host on Synergy 1.6.3 and an OS X Yosemite client on 1.6.2.

If I am moving the mouse when I press or release a modifier key, that modifier key does not change state. I have to hold the mouse still and tap the relevant key to reset it.

mengesb:
I have the same issue with 1.6.3 server on ubuntu and my windows 7 client only. My other ubuntu machines operate just fine, no ctrl stuck.

ryanwillis:
I have had this issue for a year, it persists through each update. Windows 7 server, mac client. I never experience this on the server, only on the client.

indoctrine:
1.6.3, Windows 7 64 bit server and Mac OSX Yosemite client. Getting the sticking on the client only, it's a real pain, especially when trying to paint. It started a few minutes after I opened the connection, I thought it was lag at first.

ematthews:
[...]

Issue #3823, opened 2014-10-13
Left mouse button gets stuck in right click mode on OS X client


nbolton:
**Imported issue:**
- Author: Mark Petty
- Date: 2014-01-21 12:52:40
- Legacy ID: 3893

**Steps to reproduce:**

1) Setup Windows Synergy server
2) Plug in standard PC mouse to server
3) Setup OS X Synergy client
4) Move focus to client
5) In Xcode (I've had most success with this application for some reason) press and hold the left CTRL key
6) Then release the left CTRL key and the exact same time as you press and release the left mouse button (I think that you need to do this step quickly enough so that the inputs are captured in the same sampling frame)
7) The left mouse button will get stuck in right click mode

Expected:

The left mouse button does not behave as if it's the right mouse button

Actual:

The left mouse button behaves as id it's the right mouse button

**Versions and operating systems:**

Synergy: 1.4.15
Server: Windows 7 Professional 64 bit
Client: Mac OS X 10.8.5

**Temporary workarounds:**

Press and release the left CTRL key to trigger the key release event

**Similar bugs:**

Loads of sticky key bugs since 2009

**Additional comments:**

(square brackets denote sample frames)

[1:][2:][3:LEFT-CTRL-DOWN][4:LEFT-CTRL-DOWN][5:LEFT-CTRL-DOWN][6:LEFT-MOUSE-BUTTON-DOWN,LEFT-CTRL-UP,LEFT-MOUSE-BUTTON-UP][7:][8:][9:][10:]

I think that in frame 6 something goes wrong that leaves Synergy with the left mouse button mapped to the right mouse button.

gtd:
Just as an FYI, I am experiencing this on OS X Yosemite with no Synergy in sight (this was the top Google hit).  I do have a fair number of applications installed and I'm running a Logitech mouse in addition to my trackpad in my MBP, so there could be some interaction issues there, or it could be a native issue... I haven't tested on a stock system.

stevenfrancisco:
Confirming the issue on synergy 1.5.1 with a Linux Server and a Mac Client.

The Control Key workaround seems to work for stopping the issue as well.

beporter:
Signal boosting: There is [a new nightly build available](https://github.com/synergy/synergy/pull/4564#issuecomment-95926896) that may fix this issue. If anyone monitoring this thread is able to test it, it would be appreciated.

PR #4564, opened 2015-04-22
Modifier keys often stuck down on Mac client

jpmcmu:
On MacOSX modifier keys will often stick. I was able to verify that modifier state in the Synergy client is correct and the system modifier state is incorrect. Furthermore, I was able to verify that an event is posted to the operating system for the key up event. 

After determining that Synergy is in the correct state, I attempted to fix this issue by asking the operating system what its modifier state was before every event was sent and then sending extra events to correct the systems modifier state if it was incorrect. This solution worked but after looking at the log of corrections, it became evident that this code was only required to correct mouse events. This was due to the fact that all key events pass the modifier state along with them. So the final solution was to remove the consistency check and add the modifier state to all mouse events.

This seems like an odd fix, but I have been unable to cause keys stick after these changes. 

Steps to cause a modifier key stick before these changes:
1. Hold down a modifier key (Shift key, etc)
2. Move mouse as rapidly as possible without stopping
3. While continuing to move the mouse rapidly, release the modifier key
4. Continue moving the mouse for another few seconds

Following the above steps, without the changes, I was able to cause a key stick around once in every three tries. With these changes I have not been able to get a key to stick.

### DIFF
--- a/src/lib/platform/OSXKeyState.cpp
+++ b/src/lib/platform/OSXKeyState.cpp
@@ -189,7 +189,7 @@ OSXKeyState::OSXKeyState(IEventQueue* events) :
 OSXKeyState::OSXKeyState(IEventQueue* events, synergy::KeyMap& keyMap) :
 	KeyState(events, keyMap)
 {
-	init();
+	init();a
 }
 
 OSXKeyState::~OSXKeyState()
@@ -388,6 +388,34 @@ OSXKeyState::fakeCtrlAltDel()
 {
 	// pass keys through unchanged
 	return false;
+}
+
+CGEventFlags
+OSXKeyState::getModifierStateAsOSXFlags()
+{
+    CGEventFlags modifiers = 0;
+    
+    if (m_shiftPressed) {
+        modifiers |= kCGEventFlagMaskShift;
+    }
+    
+    if (m_controlPressed) {
+        modifiers |= kCGEventFlagMaskControl;
+    }
+    
+    if (m_altPressed) {
+        modifiers |= kCGEventFlagMaskAlternate;
+    }
+    
+    if (m_superPressed) {
+        modifiers |= kCGEventFlagMaskCommand;
+    }
+    
+    if (m_capsPressed) {
+        modifiers |= kCGEventFlagMaskAlphaShift;
+    }
+    
+    return modifiers;
 }
 
 KeyModifierMask
--- a/src/lib/platform/OSXKeyState.cpp
+++ b/src/lib/platform/OSXKeyState.cpp
@@ -189,7 +189,7 @@ OSXKeyState::OSXKeyState(IEventQueue* events) :
 OSXKeyState::OSXKeyState(IEventQueue* events, synergy::KeyMap& keyMap) :
 	KeyState(events, keyMap)
 {
-	init();a
+	init();
 }
 
 OSXKeyState::~OSXKeyState()
--- a/src/lib/platform/OSXKeyState.h
+++ b/src/lib/platform/OSXKeyState.h
@@ -100,6 +100,7 @@ public:
 	virtual SInt32		pollActiveGroup() const;
 	virtual void		pollPressedKeys(KeyButtonSet& pressedKeys) const;
 
+    CGEventFlags getModifierStateAsOSXFlags();
 protected:
 	// KeyState overrides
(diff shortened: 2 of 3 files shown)


src/lib/platform/OSXKeyState.cpp
```diff
@@ -189,7 +189,7 @@ OSXKeyState::OSXKeyState(IEventQueue* events) :
 OSXKeyState::OSXKeyState(IEventQueue* events, synergy::KeyMap& keyMap) :
 	KeyState(events, keyMap)
 {
-	init();
+	init();a
```
interfect: This doesn't look particularly compile-able.

interfect:
Thanks for working this out. Apparently if it works you will have fixed six years of annoyingness.

nbolton:
@speaker Please create PR branch. You may need to fix OSXKeyState.cpp:192 (as @interfect points out, doesn't look compile-able).

jpmcmu:
Removed the stray character. Should be good to go now.

beporter:
List of issues that may be on the verge of being solved:

* Left mouse button gets stuck in right click mode on OS X client #3823
* Ctrl or shift keys randomly simulated while in IE #3194
* Command key stuck (Mac client, Windows server) #4057
* Arrow keys and ctrl+arrow causes screen rotation on Windows #4449
* Modifier state becomes out of sync on screen lock #4519
* _(And the grand-pappy of them all: Sticky modifier/meta keys (shift, ctrl, alt) #9)_

If this really works, somebody needs to give @jpmcmu an **actual** medal. I may be in shock right now. :scream:

XinyuHou:
Could someone test this pull request from our nightly build?

http://synergy-project.org/nightly?filter=a597596

nbolton:
@beporter Thanks for signal boosting. Looks like you got a hit with @hidden72 and @timpokorny but a miss with @jtbraun. We're considering if we should merge this fix in... looks fairly strong, what do you think? @XinyuHou

jpmcmu:
This fix would only work for MacOSX clients (The fix itself is in a Mac specific portion of the code base). Based on the issue report, https://github.com/synergy/synergy/issues/4519, I think @jtbraun was testing this fix on a Win7 client / Linux server setup.

jtbraun:
Confirmed that I was testing on a Win7 client, Ubuntu 14.04 server.

beporter:
I can also confirm that I've had stuck keys occur on my Windows **host** machine (with a Mac Synergy client connected). The fix should ideally be cross-platform.

jpmcmu:
The key sticking issues are caused by a number of different bugs. This pull request only fixes one of those bugs, which is either caused by Apple not correctly documenting how MacOSX tracks modifier state or a bug in their code. ( I am leaning towards a bug, it looks a lot like a race condition to me)

As an example I found another bug that causes the Ctrl key to stick off when Synergy switches between different clients. I was able to verify that is a bug in the server code, but that is as far as I have gone towards correcting that particular bug. 

What I am getting at is the "sticky key" issue is caused by a number of different bugs that will require multiple solutions/pull requests to fix.

beporter:
I take back what I said about giving @jpmcmu a medal-- just one isn't going to be enough.

nbolton:
@beporter Harsh ;) Sounds to me like @jpmcmu is a hero!

nbolton:
OK, based on @jtbraun's comment (his issue is about Windows client), I think we should merge this PR into master.

nbolton:
Either @XinyuHou or @speaker will do the merge.

franciozzy:
I'm SO HAPPY to see this issue fixed. Testing 1.7.2 now.

apiliguzov:
This bug was a major inconvenience for me. I am so glad to hear it is fixed. Going to upgrade and test today!

chrishas35:
This is incredible!  Thank you for fixing this long standing annoyance!!

scosant:
This... is... awesome! Thank you for fixing this bug!

oriolrius:
I updated to Synergy 1.7.2 in server and client and I have the same problem. My server is a Fedora 20 and my Client is a Windows 7. Both of them using 2 screen in full-HD. So far today everything was working Ok I updated to last version and everything started failing. What can I do?

Thanks for your help.

jsnyder:
Just chiming in to say thank you on this and the triple click fix.  Both of these have been driving me slightly insane for months.

PR #4595, opened 2015-04-29
Typo 'occurred' in gui.ts

joweiser:
Corrected a small typo concerning an error message I received myself while using synergy.


nbolton:
Actually this file is generated using `lang.cmd` and `langbuild.cmd` so it might be worth checking to see if running that is better.

nbolton:
Fixed in 9e9e1362ac2eb1056dcd302881863ccb7c74fa06

PR #4610, opened 2015-05-03
Add TLS Ciphersuite selection

### LINKED ISSUE #4598: TLS ciphersuites are not specified
The choice of [TLS ciphersuite](https://en.wikipedia.org/wiki/Transport_Layer_Security#Cipher) is critical to the security of a TLS connection. As best as I can tell, Synergy does not ever specify a ciphersuite list, which means it uses [the default](https://www.openssl.org/docs/apps/ciphers.html) of `ALL:!aNULL:!eNULL` (depending on the distro). Since Synergy turns off SSLv3, the worst ciphersuites are disabled, but there are still some *very* insecure ciphersuites that are supported, such as `DES-CBC-SHA`, which has a blistering 56 bits of security - trivially breakable. 

I've confirmed this by building the project, starting up the Synergy server with the command

    debug/synergys -a 127.0.0.1 -f --enable-crypto --plugin-dir ./plugins

and using the command

    openssl s_client -connect 127.0.0.1:24800 -cipher "DES-CBC-SHA"

to connect to the server. Since the connection succeeds, `DES-CBC-SHA` must be supported, which means the connection offers no confidentiality. 

To fix this, `SSL_CTX_set_cipher_list` needs to be called on the SSL context. For example, in `initContext` in `src/lib/plugin/ns/SecureSocket.cpp`, one could write

    SSL_CTX_set_cipher_list(m_ssl->m_context, "ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-DSS-AES128-GCM-SHA256:kEDH+AESGCM:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA:ECDHE-ECDSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-DSS-AES128-SHA256:DHE-RSA-AES256-SHA256:DHE-DSS-AES256-SHA:DHE-RSA-AES256-SHA:!aNULL:!eNULL:!EXPORT:!DES:!RC4:!3DES:!MD5:!PSK");

to set the ciphersuite list to a secure one. (This list was taken from the [Mozilla SSL configuration generator](https://mozilla.github.io/server-side-tls/ssl-config-generator/?1) with nginx/modern settings). With this call added, a connection attempt with DES-CBC-SHA fails - as it should.

legonigel:
Closes #4598
See #4598

Note: Untested
I am unable to test this because I have not paid for synergy.

### DIFF
--- a/src/lib/plugin/ns/SecureSocket.cpp
+++ b/src/lib/plugin/ns/SecureSocket.cpp
@@ -1,11 +1,11 @@
 /*
  * synergy -- mouse and keyboard sharing utility
  * Copyright (C) 2015 Synergy Si Ltd.
- * 
+ *
  * This package is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
  * found in the file COPYING that should have accompanied this file.
- * 
+ *
  * This package is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
@@ -110,10 +110,10 @@ SecureSocket::secureRead(void* buffer, UInt32 n)
 	if (m_ssl->m_ssl != NULL) {
 		LOG((CLOG_DEBUG2 "reading secure socket"));
 		r = SSL_read(m_ssl->m_ssl, buffer, n);
-		
+
 		bool fatal, retry;
 		checkResult(r, fatal, retry);
-		
+
 		if (retry) {
 			r = 0;
 		}
@@ -129,7 +129,7 @@ SecureSocket::secureWrite(const void* buffer, UInt32 n)
 	if (m_ssl->m_ssl != NULL) {
 		LOG((CLOG_DEBUG2 "writing secure socket"));
 		r = SSL_write(m_ssl->m_ssl, buffer, n);
-		
+
 		bool fatal, retry;
 		checkResult(r, fatal, retry);
 
@@ -205,7 +205,7 @@ SecureSocket::initContext(bool server)
 	SSL_library_init();
 
 	const SSL_METHOD* method;
- 
+
 	// load & register all cryptos, etc.
 	OpenSSL_add_all_algorithms();
 
@@ -219,13 +219,17 @@ SecureSocket::initContext(bool server)
 	else {
 		method = SSLv23_client_method();
 	}
-	
+
 	// create new context from method
 	SSL_METHOD* m = const_cast<SSL_METHOD*>(method);
 	m_ssl->m_context = SSL_CTX_new(m);
 
 	// drop SSLv3 support
 	SSL_CTX_set_options(m_ssl->m_context, SSL_OP_NO_SSLv3);
+
+	// Set ciphersuite list
+	// drops some ciphersuites
+	SSL_CTX_set_cipher_list(m_ssl->m_context, "ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-DSS-AES128-GCM-SHA256:kEDH+AESGCM:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA:ECDHE-ECDSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-DSS-AES128-SHA256:DHE-RSA-AES256-SHA256:DHE-DSS-AES256-SHA:DHE-RSA-AES256-SHA:!aNULL:!eNULL:!EXPORT:!DES:!RC4:!3DES:!MD5:!PSK");
 
 	if (m_ssl->m_context == NULL) {
 		showError();
@@ -249,10 +253,10 @@ SecureSocket::secureAccept(int socket)
 
 	// set connection socket to SSL state
 	SSL_set_fd(m_ssl->m_ssl, socket);
-	
+
 	LOG((CLOG_DEBUG2 "accepting secure socket"));
 	int r = SSL_accept(m_ssl->m_ssl);
-	
+
 	bool fatal, retry;
 	checkResult(r, fatal, retry);
 
@@ -278,10 +282,10 @@ SecureSocket::secureConnect(int socket)
 
 	// attach the socket descriptor
 	SSL_set_fd(m_ssl->m_ssl, socket);
-	
+
 	LOG((CLOG_DEBUG2 "connecting secure socket"));
 	int r = SSL_connect(m_ssl->m_ssl);
-	
+
 	bool fatal, retry;
 	checkResult(r, fatal, retry);
 
@@ -314,7 +318,7 @@ SecureSocket::showCertificate()
 {
 	X509* cert;
 	char* line;
- 
+
 	// get the server's certificate
 	cert = SSL_get_peer_certificate(m_ssl->m_ssl);
 	if (cert != NULL) {


legonigel:
I finally got SSL to work without a premium account (yay!) but it appears that with this fix, no ciphers work. I cannot connect to the server at all. I just get back an `sslv3 alert handshake failure` in the response. 
I tried changing the cipher list to `HIGH:!aNULL:!kRSA:!PSK:!SRP:!MD5:!RC4` and that seems to work, but sadly I do not fully understand the full security difference, or what is causing the errors.

nbolton:
I really appreciate you putting in the time to test the chiphersuite setting, @legonigel!

src/lib/plugin/ns/SecureSocket.cpp
```diff
@@ -227,6 +227,10 @@ SecureSocket::initContext(bool server)
 	// drop SSLv3 support
 	SSL_CTX_set_options(m_ssl->m_context, SSL_OP_NO_SSLv3);
 
+	// Set ciphersuite list
+	// drops some ciphersuites
+	SSL_CTX_set_cipher_list(m_ssl->m_context, "ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-DSS-AES128-GCM-SHA256:kEDH+AESGCM:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA:ECDHE-ECDSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-DSS-AES128-SHA256:DHE-RSA-AES256-SHA256:DHE-DSS-AES256-SHA:DHE-RSA-AES256-SHA:!aNULL:!eNULL:!EXPORT:!DES:!RC4:!3DES:!MD5:!PSK");
```
azet: why would you hard-set this instead of making it configurable?

src/lib/plugin/ns/SecureSocket.cpp
```diff
@@ -227,6 +227,10 @@ SecureSocket::initContext(bool server)
 	// drop SSLv3 support
 	SSL_CTX_set_options(m_ssl->m_context, SSL_OP_NO_SSLv3);
```
azet: also consider `SSL_OP_ALL`, see: https://www.openssl.org/docs/ssl/SSL_CTX_set_options.html

src/lib/plugin/ns/SecureSocket.cpp
VeNoMouS: Valid point, but perhaps they can have this as a default, with an overwritable option? as Joe user wont know what ciphers are still valid/secure... etc..

legonigel:
I spent some time on IRC with jnpplf. We believe we know why this won't work. ECDH (and all PFS algorithms) require some setup of temporary values in the code, and all the ciphersuites that were in the PR have PFS. This means that we can't use these ciphersuites without proper setup. I am just beginning to understand this. I have yet to figure out how this is implemented in code, but I will figure it out. 

jnpplf suggested that since we control the version of openSSL (since it is included in install) we could set specific ciphers. We know that the version of OpenSSL with synergy will support the cipher we list. He specifically suggested AES256-GCM-SHE384. I plan to create another PR to implement this temporary fix for higher security.

src/lib/plugin/ns/SecureSocket.cpp
tcely: Please don't use SSL_OP_ALL. Synergy is both the client and server and should not require any workarounds for Netscap/Safari/Microsoft anyway. http://daniel.haxx.se/blog/2012/01/27/sloppily-using-ssl_op_all/

src/lib/plugin/ns/SecureSocket.cpp
tcely: The only advantage I see for having a configurable cipher string and sane defaults is that if a particular cipher is discovered to be weak (e.g. RC4) the user can change their list to mitigate the problem without a new release.

azet:
I think you guys underappreciate the importance of (perfect) forward secrecy. Forward secure communication ensures that an attacker or someone that gets hold of your key material can't decrypt previous passively collected traffic.

Twitter were one of the first to implement it on their front-end infrastructure and, back in 2013, wrote this nice article explaining the importance of it: https://blog.twitter.com/2013/forward-secrecy-at-twitter
It has since become the best practice for web-traffic and TLS communication in general. A recent RFC on TLS highlights this as well: https://tools.ietf.org/html/rfc7525

```
6.3.  Forward Secrecy

   Forward secrecy (also called "perfect forward secrecy" or "PFS" and
   defined in [RFC4949]) is a defense against an attacker who records
   encrypted conversations where the session keys are only encrypted
   with the communicating parties' long-term keys.  Should the attacker
   be able to obtain these long-term keys at some point later in time,
   the session keys and thus the entire conversation could be decrypted.
   In the context of TLS and DTLS, such compromise of long-term keys is
   not entirely implausible.  It can happen, for example, due to:

   o  A client or server being attacked by some other attack vector, and
      the private key retrieved.

   o  A long-term key retrieved from a device that has been sold or
      otherwise decommissioned without prior wiping.

   o  A long-term key used on a device as a default key [Heninger2012].

   o  A key generated by a trusted third party like a CA, and later
      retrieved from it either by extortion or compromise
      [Soghoian2011].

   o  A cryptographic break-through, or the use of asymmetric keys with
      insufficient length [Kleinjung2010].

   o  Social engineering attacks against system administrators.

   o  Collection of private keys from inadequately protected backups.

   Forward secrecy ensures in such cases that it is not feasible for an
   attacker to determine the session keys even if the attacker has
   obtained the long-term keys some time after the conversation.  It
   also protects against an attacker who is in possession of the long-
   term keys but remains passive during the conversation.
[...]
```

tcely:
I'm sympathetic about the forward secrecy. Synergy should certainly setup the ECHDE parameters to use these ciphers.

However, your notion of updating OpenSSL is not going to wash. To update the bundled library you'd have to update Synergy anyway. Given this and the average user's total unfamiliarity with which ciphers are secure and which aren't, having a static list of ciphers is reasonable. If for some reason you needed to update the bundled OpenSSL library and/or the cipher list I'd sooner trust the Synergy authors to get this correct. Either way you'd be waiting for a new release. As I said earlier, the only case where the benefit seems worth it is when a previously trusted cipher is no longer secure. You can't add to the supported ciphers in any particular version of OpenSSL but you could disable one. Moving the cipher list to a configuration file where any naive person can muck with it doesn't preset enough benefit to justify that decision in my view.

azet:
> However, your notion of updating OpenSSL is not going to wash. To update the bundled library you'd have to update Synergy anyway. Given this and the average user's total unfamiliarity with which ciphers are secure and which aren't, having a static list of ciphers is reasonable. If for some reason you needed to update the bundled OpenSSL library and/or the cipher list I'd sooner trust the Synergy authors to get this correct. Either way you'd be waiting for a new release. As I said earlier, the only case where the benefit seems worth it is when a previously trusted cipher is no longer secure. You can't add to the supported ciphers in any particular version of OpenSSL but you could disable one. Moving the cipher list to a configuration file where any naive person can muck with it doesn't preset enough benefit to justify that decision in my view.

I violently disagree. This is done by all modern daemons, be it web or otherwise (nginx, apache, dovecot, exim, postfix,..). What's the issue with shipping a config file that has the developer preferred ciphers in them? If it turns out that a particular part of the cipher-suite is insecure, any user of the software can switch to a suite that has it disabled (e.g. by adding `+!RC4` - in OpenSSL notation), without having to wait for an update by the project maintainers. Sure people make mistakes and mis-configure software. But with all the media attention to TLS bugs, it seems admins and users have begun to listen to guides and articles. Mozilla maintains a similar one to bettercrypto, which also seems to be heavily used by sysadmins, developers and engineers, over here: https://wiki.mozilla.org/Security/Server_Side_TLS

I can't quite follow your argument. If anything -  hardcoding generates more issues with update cycles and vulnerability response. It's not only cipher-suites but having booleans on protocol version support, extensions and tuning parameters on caches et cetera which might be of interest to users of the software.

azet:
BTW: If there's no interest to develop and maintain a proper TLS implementation for synergy, it might make more sense to refer users to something alike to stunnel (https://www.stunnel.org) for transport layer security.

tcely:
The examples you cite all fall into a use-case that doesn't apply to Synergy. That is, they only control the server side of the communication. Stunnel falls into both use-cases, but because it tries to be compatible it's also more like the others. Synergy doesn't have to be compatible with random web browser, or email clients so I'm arguing it shouldn't add knobs and hooks for things we know shouldn't be turned on. Why let someone turn on SSLv2, SSLv3, TLS1, RC4, DES, etc? We have no need to be compatible with any out-of-date software, so allowing those configurations makes no sense to me.

Can you give me a reason to make any of these TLS/SSL related settings configurable other than disabling a previously used cipher (the reason I first put forward) that doesn't rely on the argument that everyone else does things this way? What are the "more issues" you see occurring from preventing users from throwing bad ciphers into the mix? I've already laid out my view about update cycles, etc. Synergy is already in a position where users will have to wait for the next release before changes to the library can be made, we wouldn't be adding any extra delay by making cipher list changes in that release.

I'd say there is more interest to develop a proper TLS implementation here than for most projects I'm familiar with. That's why these discussions are happening. I'm arguing for definition of "proper" that doesn't involve including a knob for everything under the sun and you seem to be arguing for those knobs from what I can see. You are welcome to try using stunnel if you prefer, but good luck getting the wide platform support you'd want with Synergy going in a secure configuration with that solution. You will have your work cut out for you.

speaker:
@legonigel , I started a standard pull request but the merge was very messy, so I backed out and just inserted in the SSL_CTX_set_cipher_list() by hand. I'm not sure this is a change we want to use, at least not as it is, but it really points out a security consideration we need to take into account.

Thanks for the pull request! I created a branch for you (https://github.com/synergy/synergy/tree/pr-legonigel-tls-cipher-select). Buildbot is now [building](http://buildbot.synergy-project.org/builders) (it usually takes about 30 minutes). When done, the packages will be automatically uploaded (http://synergy-project.org/nightly?filter=pr-legonigel-tls-cipher-select). Happy testing!

azet:
Sorry for the delay, have been on the road for the last week or so.

@tcely:
> Why let someone turn on SSLv2, SSLv3, TLS1, RC4, DES, etc? We have no need to be compatible with any out-of-date software, so allowing those configurations make no sense to me.

I don't think you (want to) understand my comment. I'm absolutely against enabling deprecated protocols. This, however, can be ensured by setting `SSL_OP_NO_SSLv2`, `SSL_OP_NO_SSLv3` et cetera. My argument is specific to asymmetric and symmetric algorithms in the cipher-suite that may become obsolete or a security concern. As you point out - you also put this argument forward earlier. 

That being said the list specified by @speaker  in 2336c711430c39e9fc3cfba9d9665ff2854595d1 is rather solid. 

> I've already laid out my view about update cycles, etc. Synergy is already in a position where users will have to wait for the next release before changes to the library can be made, we wouldn't be adding any extra delay by making cipher list changes in that release.

Why is that exactly? If there's not an implementation bug in OpenSSL but a generic attack on one of the algorithms currently hardcoded any user/admin can easily switch them off. I don't see people messing with this configuration without a heads-up.


> I'd say there is more interest to develop a proper TLS implementation here than for most projects I'm familiar with. That's why these discussions are happening. I'm arguing for a definition of "proper" that doesn't involve including a knob for everything under the sun and you seem to be arguing for those knobs from what I can see. You are welcome to try using stunnel if you prefer, but good luck getting the wide platform support you'd want with Synergy going in a secure configuration with that solution. You will have your work cut out for you.

stunnel was just an example. I'd be far more happy to have this in synergy directly. I can't force you guys to do this though, and this discussion is somewhat getting old as we repeat ourselves. My opinion is on the record in this PR and I stand by that.


Another thought more into the direction you're headed - if you want to limit these and take configuration away from users, ship a certain library version and not worry about all that a lot - why are you using OpenSSL anyway? Have you looked at NaCl/libsodium? That might be an interesting choice:

- http://nacl.cr.yp.to/
- https://download.libsodium.org/doc/

tcely:
@azet 

>That being said; the list specified by @speaker in 2336c71 is rather solid. I would remove any suites containing DSS though (https://people.csail.mit.edu/rivest/pubs/RHAL92.pdf).

Removing DHE ciphers is more important for that list. DSS won't be used with a RSA certificate anyway. Consider the age and source of that publication.

>Why is that exactly? If there's not an implementation bug in OpenSSL but a generic attack on one of the algorithms currently hardcoded any user/admin can easily switch them off. I don't see people messing with this configuration without a heads-up.

Experience shows me people do unexpected and unfortunate things often. I'd suggest a hardcoded cipher list and a blacklist for the user.

>Another thought more into the direction you're headed - if you want to limit these and take configuration away from users, ship a certain library version and not worry about all that a lot - why are you using OpenSSL anyway? Have you looked at NaCl/libsodium? That might be an interesting choice:
>
- http://nacl.cr.yp.to/
- https://download.libsodium.org/doc/

Agreed.

azet:
@tcely 

> Removing DHE ciphers is more important for that list. DSS won't be used with a RSA certificate anyway. Consider the age and source of that publication.

The age does not matter as the DSS standard is still used in the same way. The source is Ron Rivest and the argument is sound and holds today. I'd just remove it if it's not desired to be used anyway. Why are you so keen on removing DHE specifically? Performance when compared to ECDHE? Or the recent "LogJam" paper (it does not conclude that DHE should be abolished, there's a working draft in the IETF that has been worked on for a long time now that would solve a lot of these issues. It's been accepted by the TLS-WG as an item recently: https://tools.ietf.org/html/draft-ietf-tls-negotiated-ff-dhe).

tcely:
@azet 
>The age does not matter as the DSS standard is still used in the same way. The source is Ron Rivest and the argument is sound and holds today. Just remove it if it's not desired to be used anyway, right?

Absolutely. The source being Ron Rivest, it's not surprising that he's arguing against his competition. I'd be surprised if none of the weaknesses pointed out have been addressed since then. Does it still use the same key length for instance? Anyway, I won't argue for keeping DSS.

>Why are you so keen on removing DHE specifically? Performance when compared to ECDHE? Or the recent "LogJam" [https://weakdh.org] paper? It does not conclude that DHE should be abolished, to the contrary; there's a working draft in the IETF that has been worked on for a long time now that would solve a lot of these issues. It's been accepted by the TLS-WG as an item recently: https://tools.ietf.org/html/draft-ietf-tls-negotiated-ff-dhe - some of the "LogJam" authors even voiced support in the working group.

The possibility of pre-computing the (default / shared) DH parameters is objectionable. Better performance from ECDHE is also a good reason to avoid the DHE ciphers too. I was tempted to suggest minding the DH parameters when including those DHE ciphers, but decided against that. It'd be better to just move to the ECDHE ciphers anyway as nothing will be demanding DHE support from Synergy.

I'm sure the working group will eventually get it all sorted for TLS eventually, but hopefully the number of TLS clients supporting DHE and not ECDHE will be smaller by then.

azet:
@tcely 

> Absolutely. The source being Ron Rivest, it's not surprising that he's arguing against his competition. I'd be surprised if none of the weaknesses pointed out have been addressed since then. Does it still use the same key length for instance? Anyway, I won't argue for keeping DSS.

http://blog.cr.yp.to/20140323-ecdsa.html

> The possibility of pre-computing the (default / shared) DH parameters is objectionable. Better performance from ECDHE is also a good reason to avoid the DHE ciphers too. I was tempted to suggest minding the DH parameters when including those DHE ciphers, but decided against that. It'd be better to just move to the ECDHE ciphers anyway as nothing will be demanding DHE support from Synergy.

As synergy controls server and client I don't see why using only ECDHE would be a problem. I'm in support.

tcely:
@azet 

>Why is that 'not surprising'? I don't follow.

Even if DSS were the best thing since sliced bread, he'd have personal and financial reasons to argue against it. I'm fine with pulling DSS/DSA/ECDSA ciphers from the list.

>I'd say that pre-computing for FFDHE is quite impossible today.

Even in current OpenSSL versions the minimum DH parameter length (768 bit) can be pre-computed in a week or less leading to the possibility of breaking the TLS handshake in real time. The logjam paper shows that even 1024 bit DH parameters are likely within the reach of any group with enough resources.

This won't be a concern if we're not using DHE ciphers as ECDHE is going to be stronger and faster, but is the reason the suggested length of DH parameters was bumped up to 2048 recently.

azet:
> Even if DSS were the best thing since sliced bread, he'd have personal and financial reasons to argue against it. I'm fine with pulling DSS/DSA/ECDSA ciphers from the list.

Sounds conspiratorial. I can't really follow your argument, but this seems off-topic anyhow.

> Even in current OpenSSL versions the minimum DH parameter length (768 bit) can be pre-computed in a week or less leading to the possibility of breaking the TLS handshake in real time. The logjam paper shows that even 1024 bit DH parameters are likely within the reach of any group with enough resources.

768bit and 1024 are still a large difference. I agree that 2k+ parameters should be used, we've advised that in bettercrypto since the inception of the project back in 2013.

I'd agree that going with ECDHE is a good choice.

XinyuHou:
@legonigel 
This is conflicting now. Could you please fix the conflict?

PR #4879, opened 2015-07-17
Memory leak caused by bad delete in IpcReader

cym13:
src/gui/src/IpcReader.cpp line 68, data is an array of char, but only delete data is called.

### DIFF
--- a/src/gui/src/IpcReader.cpp
+++ b/src/gui/src/IpcReader.cpp
@@ -65,7 +65,7 @@ void IpcReader::read()
 			char* data = new char[len];
 			readStream(data, len);
 			QString line = QString::fromUtf8(data, len);
-			delete data;
+			delete[] data;
 
 			readLogLine(line);
 		}


XinyuHou:
@cym13 
We need to know exactly what issue this solves for the user. Please could you add steps to reproduce to issue #5295? Once you do this we'll merge and release.

cym13:
As it stands this bug is in dead code so I cannot provide a way to reproduce the bug using synergy as it is. Otherwise, any call to the IpcReader::read() method that actually reads something (that is m_Socket->bytesAvailable() isn't 0) triggers the bug.

Freeing an array allocated with new[] using delete instead of delete[] is undefined behaviour so the exact consequences are runtime-dependant.

In the best case it causes a memory leak of a maximum of 4GB per call, in the worst it can corrupt memory or just make the whole program crash. If the corruption happens to be deterministic enough (again, runtime-dependant) one may use it to take control of the program flow using technics similar to double-free exploitation.

I don't mean to be alarmistic,I'm sure in most cases it would just be a memory leak and it's still in dead code for the moment, but I really see no good reason not to fix a bug as clear as this one especially since it has no direct impact on the users.

nlyan:
Fixed in v1.8.3

PR #4883, opened 2015-07-20
XWindowsClipboard::m_cacheTime used before being initialized

Maeyanie:
XWindowsClipboard::m_cacheTime could be used before being initialized.

It would rarely cause a problem as the uninitialized memory would have to randomly end up matching the new time to be an issue, but it should still be fixed.


XinyuHou:
@Maeyanie 

XinyuHou:
Already in v1.8.6

PR #4915, opened 2015-08-04
Large debug message causes buffer overrun 

DALDEI:
With Linux + Windows connection, debug build and debug mode printing the message buffer used is too small.   A growable buffer is used for the message bug in debug mode a fixed size stack buffer is used to append the message causing crashes with large messages, easy to reproduce with a large clip buffer.

This doesn't fix a more subtle crash related to clipbuffers but atleast it keeps the debug message itself from crashing


XinyuHou:
@DALDEI 

XinyuHou:
Problem resolved.

PR #4974, opened 2015-08-18
Use python2 explicitly if available

moises-silva:
I was trying to build master on arch Linux to debug some latency issues I'm experiencing and I had to do the following adjustments to build it:

* Add QObject include in CommandProcess.h
  This solves compilation error:

  'src/CommandProcess.h:25:2: error: ‘Q_OBJECT’ does not name a type'

  This is probably due to newer Qt version used by arch linux
  The QtObject header seems available at least since Qt 4, if older
  versions need to be supported this could be changed to qobject.h
  depending on the Qt library version detected

* Improve hm.sh to detect if python2 is explicitly available
  and use it to try to avoid using python3, which is unsupported
  by the hm.py script

### DIFF
--- a/hm.sh
+++ b/hm.sh
@@ -1,3 +1,11 @@
 #! /bin/bash
 
-python hm.py "$@"
+pybin=python
+# Use python2 explicitly if available
+# otherwise things break on arch or any
+# distro that has python3 by default
+py2=$(which python2 2> /dev/null)
+if [ $? = 0 ]; then
+	pybin=$py2
+fi
+$pybin hm.py "$@"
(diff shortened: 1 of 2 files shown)


jakepetroules:
This fixes build issues on all OSes -- QStringList simply no longer transitively includes QObject.

moises-silva:
yeah I figured this would most likely be a fix elsewhere as well. Thanks for the confirmation.

If I should amend the commit message to reflect that this is not only for arch let me know.

XinyuHou:
@moises-silva 
Could you raise an issue about python conflict and commit the changes only related to that issue?

There is another PR fixed the QObject issue.

moises-silva:
@XinyuHou Sure, will do so later today. Are you saying another PR already fixed the QObject issue? if so, can you link it here?

moises-silva:
@XinyuHou I created PR #5301 for the change in hm.sh

PR #4982, opened 2015-08-20
Adding more information on VS version error

ilbolzan:
Adding a user friendly message in case you have a Visual Studio version that is not suported.


ilbolzan:
Sorry, someone has already fixed that.

Issue #4991, opened 2015-08-31
Oversized Bonjour protocol name could conflict


nyetwurk:
Bad service type in x.x.x.x._synergyServerZeroconf._tcp.local. Application protocol name must be underscore plus 1-15 characters. See http://www.dns-sd.org/ServiceTypes.html

Bad service type in ._synergyClientZeroconf._tcp.local. Application protocol name must be underscore plus 1-15 characters. See http://www.dns-sd.org/ServiceTypes.html

nyetwurk:
Service name cannot have more than 15 chars after the initial _

````patch
diff --git a/src/gui/src/ZeroconfService.cpp b/src/gui/src/ZeroconfService.cpp
index 165912d..5412944 100644
--- a/src/gui/src/ZeroconfService.cpp
+++ b/src/gui/src/ZeroconfService.cpp
@@ -33,8 +33,8 @@ static const QStringList preferedIPAddress(
                                "10." <<
                                "172.");

-const char* ZeroconfService:: m_ServerServiceName = "_synergyServerZeroconf._tcp";
-const char* ZeroconfService:: m_ClientServiceName = "_synergyClientZeroconf._tcp";
+const char* ZeroconfService:: m_ServerServiceName = "_synergyServer._tcp";
+const char* ZeroconfService:: m_ClientServiceName = "_synergyClient._tcp";

 ZeroconfService::ZeroconfService(MainWindow* mainWindow) :
        m_pMainWindow(mainWindow),
````

nyetwurk:
Sigh

nbolton:
@nyetwurk Do you use Bonjour?

nyetwurk:
Yes, and unfortunately the PR breaks old/new synergy client/server discovery due to the renaming. Was hoping this PR would be included sooner rather than later so less people would be affected.

If I knew more about Bonjour I could work on forward/backwards compat, but I don't have any dev experience

nbolton:
@nyetwurk Have you tried Synergy 2 Beta? It has an alternative solution to Bonjour. https://symless.com/synergy/features/s2

nyetwurk:
I was LITERALLY only using bonjour to test my PR. I assumed it was refused because of the backwards/forwards compatibility issue between old/new client/servers

nbolton:
I ask because I'm considering disabling it by default. It seems to cause more issues than it solves.

nyetwurk:
I vote nuke it from orbit.

[...]

PR #4994, opened 2015-09-02
Fixes for clipboard bugs #4735, #4991, #5041 

nyetwurk:


ctlajoie:
I can vouch for this PR. Before I started using this code synergy (the client specficially) would segfault rather frequently. It hasn't happened at all since I started using it. In addition my clipboard now works in one direction (server->client) whereas it did not work at all before. It probably also fixes client->server clipboard but I have not been able to compile a functional build on Windows (which is where I run my server).

nyetwurk:
BTW I think the same thing is happening server side (sometimes) in Server::switchScreen().

If you call m_active->leave() before getClipboard() (in onClipboardChanged), getClipboard() sometimes gets an empty clipboard.

nyetwurk:
So we can call m_active->leave() first, or we have to split onClipboardChanged into two pieces, getClipboard() and marshall(), setClipboard(), and call m_active->leave() after getClipboard().

Testing now to see if just moving ->leave() is ok. I know we split the client side into two pieces... I don't even know if that was strictly necessary.

nyetwurk:
False alarm. That isn't the problem either. For some reason, firefox and thunderbird clipboards are broken after some time (screen saver?). Re-ordering the leave doesn't fix it.

src/gui/src/ZeroconfService.cpp
```diff
@@ -33,8 +33,8 @@ static const QStringList preferedIPAddress(
 				"10." <<
 				"172.");
 
-const char* ZeroconfService:: m_ServerServiceName = "_synergyServerZeroconf._tcp";
-const char* ZeroconfService:: m_ClientServiceName = "_synergyClientZeroconf._tcp";
+const char* ZeroconfService:: m_ServerServiceName = "_synergyServer._tcp";
+const char* ZeroconfService:: m_ClientServiceName = "_synergyClient._tcp";
```
nyetwurk: #4991

ceribik:
Do both the client and server need these patches? Particularly in the case where the server is running linux, and the client is on Windows?

nyetwurk:
@ceribik No. The client changes are to the XWindows client, the server changes are to the win32 server.

knowncitizen:
This works for me on both Fedora 22 and Mac OS X El Capitan (10.11) after building from source on both.

XinyuHou:
Thanks to @nyetwurk 's work.

Can any one test this nightly builds?

http://synergy-project.org/nightly?filter=nyetwurk-master-alpha-3eaefc

webdawg:
Can someone tell me where to get the source for this.........

knowncitizen:
@webdawg you're on the site here, are you familiar with github?

nyetwurk:
Root cause is regression introduced in commit cf5347c8f6cae16559678838d280433f97941908

nyetwurk:
Also see #5045 - after reading that and galkinvv's comments to cf5347c I think this approach is wrong.

galkinvv:
I think that though approach is maybe too complex, but ot may worked fine for client...
if clipboard transfer is fast - race conditions are avoided. If it is slow - it timed out and becomes racy. But it is improvement to the current situation when it is always racy.

From the user point of view the only real problem with this PR is that it introduces syncronization in client.cpp, but not on server.cpp so #5045 is not fixed. It seems it can be fixed by similar changes in server

baston85:
@XinyuHou I'm running the x86_64.deb nightly build now and it has not locked up or crashed yet. Client OS is Mint 17.2. It's only been ~5 hours, but it would have frozen many times by now on 1.7.4. I also tested my repro steps (#5090) and it never froze and it doesn't even lag when I switch screens rapidly. Thank you for the suggestion and to @nyetwurk for your work on the update

XinyuHou:
@nyetwurk 
I have been testing your changes these days. Copying from client is more stable.

But I still have #5041 even after I restart both server and client.

Is there any repro steps to prove your fix has fixed #5041?

Is MSWindowsClipboard::get function the only function you changed to fix $5041? The other changes look like a refactoring.

nyetwurk:
@XinyuHou
I have NO real world experience with #5041 other than the fact the two bugs seem related. I have not been able to reproduce #5041 in my particular set up but that means nothing.. I can try to reproduce it on my set up and see if I can fix it for good but no guarantees.

nyetwurk:
@XinyuHou 
I have a separate branch that adds a ton of #5041 debugging messages... I can open another PR for that and if you can reproduce #5041 and send me the output perhaps we can figure it out together.

XinyuHou:
@nyetwurk 
Thanks for the reply.

I can reproduce copying from chrome doesn't work issue, but I can't approve it 's related to screensaver. Because my Windows refuses to bring screensaver on even I set it to 1min. 

If your another pr is about debugging copying from browser,  sure I'm more than happy to send you my log. If it's for debugging screensaver related, I don't think my setup would give you much helpful information. But still worth a try.

XinyuHou:
@nyetwurk 
Could you create a new pr for the logging. So we can work on that.

XinyuHou:
@nyetwurk 
Sorry I revert your changes for #5041.

I found it didn't make sense to check if we counld find a matching format in system clipboard enum, because to reach into get(), there is always a has() before it to check if the format is available. Plus, the work around is ignoring the mismatch and insisting on the format we think that is correct.

commit: 52f8f40f7bd39c6e9b7d8820448a907d18d32c02

The reason that it failed to find the matching format in system is the process is started with elevated mode, either by tick elevate mode in GUI or screensaver/UAC triggers it. I would assume when the desktop switches back up default, services would restart a new process, but this is not the case, it keeps the elevated process until you click apply or restart.

nyetwurk:
@xinyouHou
Makes sense to me. The get() above was a straight copy/paste; your refactor looks fine.. assuming it works for you in the case where the services do NOT restart a new process when the server goes from screensaver -> default.

The code there is very confusing, btw.

XinyuHou:
@nyetwurk 
It works for me in some degree. When it doesn't work it is because of some other clipboard issue.

I only tested between a Windows server and a Mac client. I'm gonna test the rest combination today.

nyetwurk:
Also see #5139 - definitely related.

XinyuHou:
Originally merged in 1.7.6.

PR #5015, opened 2015-09-10
Add a hack for #2765

dguerri:
I am not a obj C developer and this is definitely not a solution. I couldn’t go any further in my investigation but the problem could be in the Apple CoreServices Framework or in the way Synergy uses it.
Even if this workaround could theoretically break some keyboard layout, it should work for most of them and it could be used while a proper fix is implemented. The current Mac Version of Synergy is not usable so you might want to distribute this workaround (also because it only impacts OsX).
This patch has only been tested on OsX 10.10.5


dguerri:
Here you go!

XinyuHou:
@dguerri 

We are working on a proper fix.

dguerri:
Cool! Thanks for fixing this!

PR #5021, opened 2015-09-11
Update mac instructions to install zip distribution

beefnog:
I updated the directions such that a clear installation path is provided that does not require the direct use of root.


XinyuHou:
@beefnog 
What is the advantages of doing as user then change permission over doing as root in the first place?

beefnog:
The advantage is that performing the operations using sudo actually requires fewer steps than activating the root user to perform the installation. Additionally, Installing the software using sudo does not require making persistent user account changes to the system.

dgentry:
It is still generally considered better practice to make changes using sudo than to become root, so this change would still be helpful.

dgentry:
Thanks!

XinyuHou:
Merged into v1.8.6

PR #5025, opened 2015-09-14
Update link to build instructions in README

sas:


sas:
Ah, someone already created a PR for the same thing. See PR#5022.

PR #5026, opened 2015-09-15
Fix incorrect check of return code from dup

DALDEI:
Return code from dup() is incorrectly checked and debug message printed on success.
dup() returns -1 on failure,  fd on success.


XinyuHou:
@DALDEI 

XinyuHou:
Merged into v1.8.6

Issue #4913, opened 2015-08-03
Future-proof GUI login by using newer auth URL


the-wes:
Problem: Synergy uses https://synergy-project.org/premium for authentication. This is deprecated and has been replaced by /account.
Solution: Update Synergy code base to use /account instead.

PR #5043, opened 2015-10-02
Fix auth URL

ivebeenlinuxed:
Fir for #4913


XinyuHou:
Changes are merged into https://github.com/symless/synergy/tree/v1.8.2

PR #5067, opened 2015-10-17
Fix wrong array deallocation

noshbar:
An array is being allocated, so should use the corresponding delete[] operator.


XinyuHou:
Duplicate to https://github.com/symless/synergy/pull/4879

PR #4665, opened 2015-05-13
Add support for Visual Studio 11

wadetb:
Works around various compatibility issues with the newer compiler.

(Just noticed the comment about variadic macros is wrong, it's actually variadic templates, you might want to fix that)

PR #5117, opened 2015-11-20
Recent Visual Studios (VS2012#11,2013#12,2015#14) and QT5.5 support

sheavner:
I've built upon the original abandoned pull request #4665 by @wadetb to allow hm conf, build, and dist to work with VisualStudio 2015 free and QT5.5.

Changes:

1. Dynamically determine VS parameters for versions 11, 12, 14.   I've really only tested 14-32bit.  Others should be similar, assuming @wadetb's port was working.
2. Dynamically determine QT path and version.  Wix install is still pretty hacky.
3. Added a "hm distclean" to wipe out any files generated by previous hm runs.  It decided what to clean based on the contents of .gitignore.  Early lines in the file (those above the marker) will be deleted.   If you never invoke run hm distclean this change won't affect anything.
4. Minor code /  #defines as needed.

Build worked with minimal arguments:
```
hm config -g1
hm build
hm dist win
```

It's up to date with synergy/master as of this morning.  Please consider merging it soon :)


XinyuHou:
@sheavner 
I have tried your pr. It compiled under release mode but failed to compile on debug mode. I the GUI crashes when I run it after a release build. Which version of wix are you using?

sheavner:
I'm using C:\Program Files (x86)\WiX Toolset v3.10\

sheavner:
@XinyuHou I fixed debug build under VS2015, which VS were you testing with?  

I see that VS2015 community supports 64 bit output.  I'm testing that now.  I see my QT builds are always 32 bit though.   Is that normal?

XinyuHou:
@sheavner 
VS 2015 professional, I need to update my license. I will try out you pr. Thanks for your work!!

The GUI is always 32bit.

sheavner:
The 64 bit packaging is still broken.   It's complaining about some 32 bit CRT libs.

sheavner:
@XinyuHou 64 bit crashes appear to be a legacy bug, maybe only exposed with VS2015?   GUI + other apps should run now.

sheavner:
@XinyuHou I've built and packaged 64 bit with VS2013 and it works.   There's some extra checks maybe in the VS2015 redist that trigger a failure.

My proposal:  

1. Verify things still work with the old QT + old Visual Studio that you've been using for builds.   My intent was always to do no harm.
2. If that works, merge things to the main branch.  
3. Eventually add a solution for VS2015 64-bit packaging - a good solution would be to use VS based QT apps instead of mingw.   QT5.5. doesn't yet prebuild for VS2015 though you could build your own.

It's your call if you want to do some testing on VS2013 and move the build system to VS2013.   I'd probably look for a QT solution and go to 2015 rather than do two migrations.

XinyuHou:
@sheavner 

Thank you for all your work.

Recently we have some enterprise problems that are needed to be solved urgently. But I will try out your PR in my spare time.

Unfortunately we have a strict rule about putting non-working code into code base. I can keep pulling your changes into the corresponding pr branch and let the others test the nightly builds.

> a good solution would be to use VS based QT apps instead of mingw. QT5.5. doesn't yet prebuild for VS2015 though you could build your own.
Sorry I don't understand this bit. Could you explain a bit more?

Thanks agian

aschaerorange:
hello, I installed a fresh machine from start, following the "compiling section". I can assure you this DOES NOT compile out of the box with
- QT 5.6 
  => a lot of classes have moved and thus are not found (need to change code to reflect the change) (I ended up moving QT5.6 parts to QT2010 subdirectories ... uggly)
  => the code is still bound to QT 2010 (in python compiled code there are references to path that is hardcoded)
  => if solving those issues, you get a warning that the expected QT version is 4.x (can live with it)
- visual studio community (2015)
- windows 7 x64

after a lot of attempts, I could get the code to compile, even build the installer. However, the syngerys process will not launch even if the log says "command synergys --xxx". If I launch it manually, it takes 50% CPU (one full core) ...

heck. would have loved to get it compiled properly (and working, especially with a visual studio available to eveyone), but I'm sorry this is not the case with the P.R. here

extrealm:
Just 2-cents, Better to separate PR as QT updates and VS stuff.
also, MS recommends vcredist exe instead of merge modules currently used in wix.

nlyan:
Equivalent changes merged in to v1.8.3

nickoe:
@nlyan  What do you mean with equivavalent changes being merged? I don't see support for anything newer than support for VS10 in the master branch.

rootkow:
@nickoe use the v1.9.0 branch.

QuAzI:
Please describe in https://github.com/symless/synergy/wiki/Compiling#Windows_x86 how **VS2013 Community Edition** can be added into `hm genlist`

Current list for v1.9.0 branch
```
c:\Projects\synergy>hm genlist
1: Visual Studio 14
2: Visual Studio 14 Win64
3: Visual Studio 10
4: Visual Studio 10 Win64
5: Visual Studio 9 2008
6: Visual Studio 9 2008 Win64
7: Visual Studio 8 2005
8: Visual Studio 8 2005 Win64
```

yupi2:
@QuAzI https://cmake.org/cmake/help/v3.7/manual/cmake-generators.7.html#visual-studio-generators

QuAzI:
@yupi2 thanks
But I still can't built something :(

```
c:\Projects\synergy>cmake -G "Visual Studio 12 2013 Win64"
-- The C compiler identification is MSVC 18.0.40629.0
-- The CXX compiler identification is MSVC 18.0.40629.0
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio
12.0/VC/bin/x86_amd64/cl.exe
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio
12.0/VC/bin/x86_amd64/cl.exe -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studi
o 12.0/VC/bin/x86_amd64/cl.exe
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studi
o 12.0/VC/bin/x86_amd64/cl.exe -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: C:/Projects/synergy

c:\Projects\synergy>hm.cmd build
Error: [Error 2] The system cannot find the file specified
```

yupi2:
@QuAzI You'll need to add something like
`9 : VisualStudioGenerator('12 2013 Win64'),`
[in here](https://github.com/symless/synergy/blob/v1.9.0/ext/toolchain/commands1.py#L270) then use
`hm conf -g9` then `hm build` should work

PR #5130, opened 2015-11-25
Enable compilation on El Capitan

dim13:


XinyuHou:
@dim13 
Could you give an explanation about this PR?

dim13:
Honestly, since it's so far away from now, I don't really recall all details. IIRC, I got configuration errors, as I tried to compile Synergy for El Captain (10.11.0 then) because of version mismatch.

However, since I'm not a OSX developer, I'm not sure about OSX_TARGET_MAJOR variable. I seemed to me, that it refers to x.11.x and not to 10.x.x, as one would expect. This patch fixed compilation problem for me.

XinyuHou:
You code would allow the source code compile on 11.01, but right now we are restricting it to 10.x

Issue #3241, opened 2014-10-13
Redundant restart for elevation when already elevated


nbolton:
**Imported issue:**

* Author: Nick Bolton
* Date: 2012-07-31 21:33:12
* Legacy ID: 3311
* Version: 1.4.15

1. Check the "Elevate" option
2. Click Apply
3. Cause a UAC dialog to display

Expected: Mouse is usable immediately

Actual: Mouse can't be used for ~5 secs while server/client restarts for elevation (when already elevated)

dpolivy:
Is this fix still relevant, and can in be incorporated into the latest builds? It really seemed to help back when I was using 1.4.x, and it seems there are a bunch of other issues here that would be solved by adding this in for all.

dpolivy:
Quick update, I pulled the latest commits into my fork, and then re-applied this diff, and sure enough, if I run Synergy in elevated mode, then I don't lose the connection when UAC prompts appear!

The updated diff is:
````
@@ -527,7 +527,9 @@ void MainWindow::startSynergy()
		// is switched; this is because we may need to elevate or not
		// based on which desk the user is in (login always needs
		// elevation, where as default desk does not).
-		args << "--stop-on-desk-switch";
+		if (!appConfig().elevateMode()) {
+			args << "--stop-on-desk-switch";
+		}
#endif
	}
````

I don't know what the current issues are with always running in an elevated mode, but if they're minimal or have been addressed since this was originally opened, I'd strongly suggest pulling this change in.

ktiedt:
Subscribing, plus curious, how has this gone over a year w/o being fixed when there is a patch!?!?!? Are you waiting for a pull request? Especially considering how annoying this problem is (I actually have to restart my server to reconnect it seems -- or maybe I just dont wait long enough...).

pmkane:
Still seeing this behavior with Synergy Pro 1.7.6.

XinyuHou:
Can someone test this nightly?
http://synergy-project.org/nightly?filter=synergy-pr5139-fix-redundant-restart-for-elevation-beta-06c853a

babist1st:
I have been using it for about 2 days on my windows 10 laptop.
The server is an Arch linux machine with the latest stable release.

The UAC popup doesn't disconnect me anymore and it also seems to fix a similar disconnect when unlocking windows after having locked windows with win key+l

This might be a different problem.
There is a slight freeze if you move the mouse when unlocking but no disconnect.
[...]

Issue #4683, opened 2015-05-18
Synergy in elevated mode still disconnecting upon UAC popup


xorinzor:
When running both the server and the client (version 1.7.2) in elevated mode, whenever either gets a UAC message it will cause a disconnect.

Nothing to see in the log screen except for the regular disconnect message.

Running a fresh windows 10 build 10074 x64 install on both the client and the server.

IanSmithers:
I just came here to report this. I am running Windows 8.1 x64 and Windows 7 Ult x64 with Synergy clean install 1.7.3

IanSmithers:
I just came here to report this. I am running Windows 8.1 x64 and Windows 7 Ult x64 with Synergy clean install 1.7.3

speaker:
Unfortunately this functionality is working as designed. The problem comes down to what's happening on a server and client when the UAC window pops up. On a windows machine, the UAC window pops up in a new desktop (not the one synergy was in) that requires special privileges to operate. So on the server the KB and mouse still work because synergy is effectively turned off for the duration. On the client, it's more problematic, because the new desktop doesn't contain a working synergy client that has the privileges to operate. In order to allow interaction on the new desktop, the old client needs to disconnect so a new one with the necessary privileges can interact with the UAC window. Once this is done, the new desktop that contained the UAC window goes away, and the old client needs to connect again. 

There might be ways to make Synergy a bit smarter about when to disconnect (server vs, client for example), however, we've tried having synergy run with privileges in the past with very limited success.

@XinyuHou , any thoughts on this?

xorinzor:
@speaker thanks for clarifying, I did not know that yet. I knew UAC was running in a sandboxed environment, but not a new desktop, which then makes sense as to why it doesn't work.

I'm wondering if there are any workarounds for this

dakotak:
I have this issue with my setup (ubuntu server, windows 7 client), I know that some software has found ways around this. I don't believe this was an issue using inputdirector (windows only). Perhaps there is a way to change the timeout for the reconnect; that would be a good workaround for now.

ktiedt:
Came here looking for a better solution to this... currently I just have to open the server cpanel and "apply" after UAC opens, and again after I tell it "yes" to force reconnects on my Win 10 box... Hope that helps someone else.

dpolivy:
[...]

PR #5139, opened 2015-12-03
Don't disconnect on elevation if already elevated

dpolivy:
Port the fix from #3241 into the latest codebase to prevent disconnects
due to elevation when already elevated.

This seems to be a hot issue for many users, and I've been successfully using this patch on my old 1.4.x fork for a few years. The code has changed slightly due to some refactoring, so I've ported the patch over to the latest codebase.

It's been working well for me so far -- no disconnects on UAC prompts on clients, and so far no other unintended issues from running elevated all of the time.

### DIFF
--- a/src/gui/src/MainWindow.cpp
+++ b/src/gui/src/MainWindow.cpp
@@ -527,7 +527,9 @@ void MainWindow::startSynergy()
 		// is switched; this is because we may need to elevate or not
 		// based on which desk the user is in (login always needs
 		// elevation, where as default desk does not).
-		args << "--stop-on-desk-switch";
+		if (!appConfig().elevateMode()) {
+			args << "--stop-on-desk-switch";
+		}
 #endif
 	}
 


dpolivy:
Also addresses #4683.

XinyuHou:
@dpolivy 
Can someone test this nightly?
http://synergy-project.org/nightly?filter=synergy-pr5139-fix-redundant-restart-for-elevation-beta-06c853a

dpolivy:
@XinyuHou I'm travelling at the moment so won't be able to get to this for a week and a half, sorry. I can look at it then, though.

XinyuHou:
@dpolivy 
I assume this fix is only target for users who tick elevate mode, is this correct?

I would think a proper fix is something in the service.

dpolivy:
@XinyuHou Yes, it only takes effect if you are already elevated -- it basically eliminates an unnecessary restart.

I don't know if there's another fix that can be done without being elevated.

dpolivy:
@XinyuHou I had a chance to play with this build a little bit, and when I enable Elevated mode, then it does work as expected on Windows 10 and I can correctly interact with the UAC dialog with no lag/delay. Prior to enabling the Elevated mode, it did not work. This looks good to ship. :+1:

The UI looks terrible on my new 4K laptop, but that's a separate issue.

XinyuHou:
Changes are in https://github.com/symless/synergy/tree/v1.8.3

spronkey:
After upgrading to 1.8.4 today, Synergy server crashes whenever I open certain apps (noticed with notepad.exe), immediately after the UAC dialog pops up.

Situation is Windows 7 server, UAC dialog popup on same server. Have duplicated issue with notepad.exe and calc.exe.

Windows 'application has crashed' popup appears, and local mouse is extremely jittery until dismissing said popup. Synergy does restart itself after you dismiss the popup, but it takes a little while.

PR #5140, opened 2015-12-04
Fix build with Qt 5.5 by adding #include of <QObject>

jpwhiting:
I think Qt 5.5 QStringList no longer #includes <QObject> without this synergy fails to build from git on debian linux stretch. With this change it builds again.


XinyuHou:
Duplicate to PR #5117

Issue #451, opened 2014-10-13
Fast cursor on any client with Mac server


nbolton:
Keyword: accelerated cursor

**Imported issue:**

* Author: thatjunkaccount (Google user)
* Date: 2010-04-29 00:00:00
* Legacy ID: 461

> What list of steps will reproduce the problem?  
  
Running Synergy+ and moving the cursor from the Server screen to the Client screen  
  
> What version of operating system, and Synergy+ are you using?  
  
Server: Snow Leopard  
Client: Windows 7 Home Premium  
Synergy +: 1.3.5  
  
> What is the expected behaviour, and what happens instead?  
  
Normal mouse speed on the Server (Snow Leopard) is unusably fast on the Client (Windows 7).   
Slowing down the server mouse speed to the lowest setting makes the Client speed usable.   
Altering the mouse speed on the Client has no effect. I have also experienced the 'infinite   
acceleration' problem during FPS games on the Windows 7 Client.  
  
> Is there a way of temporarily working around your problem?  
  
Turning down the Server mouse speed solves the client-too-fast-to-use problem, but generates   
a server-too-slow-to-be-practical  
  
> Would you like to solve the problem yourself and submit a patch?  
  
I have no experience in coding, or applying patches

JoshMcCullough:
Note that if I hit `ScrLk` on my Win8.1 client (Win7 server), the mouse cursor speed doubles (roughly). It's not just an OSX issue.

robnasby:
Glad to see this is moving up the priority list.  Thanks!

zeroasterisk:
Same problem:

* Synergy : 1.6.3
* Server OS: 10.10.2
* Server Machine: MacBook Pro 17" Early 2011 @ 1920x1200 
 * with a second HDMI display @ 1920x1080
 * fresh OS / Synergy install
* Client Machine: Ubuntu 14.04 64-bit running DVI display @ 1680x1050 (highest)
 * pointer speed on slowest setting
 * fresh OS / Synergy install

Cursor just zooms around when on client... working great otherwise.

SynVisions:
+1 on this, have to restart my OS X Synergy server on a regular basis to fix.

Thanks

rogerdrummond:
FYI- The Client (Win8.1) does NOT have to even be powered on to create this behavior using a Mac Server.  

For verification:
* I started Synergy on the MacbookPro Client (Yosemite), and slid over to the Win8.1 desktop (mouse performed normally). 
[...]

PR #5148, opened 2015-12-10
Fix mouse speed increase with sleep on Mac server

ErikSwan:
As reported in #451, with the Mac server, sleeping and resuming (on the server) causes the mouse speed to double on the client upon resume.

Fix by removing the CFRunLoopSource from the run loop on screen disable before releasing it. CFRunLoopAddSource in enable() retains the CFRunLoopSource, so even though the source is released in disable(), the run loop still has a copy. When the server comes out of sleep and the screen is enable()'d, another event tap and run loop source are created and added to the run loop, so the callback is now being called twice for every mouse movement, and so on for every additional time the server sleeps.

This is a better approach than the fix in 267f3ac41f7, because although that fixes the issue by disabling the event tap before the event tap and run loop source are released, a memory leak still occurs since they are retained by the run loop.

Additional references on the behavior of CFRunLoopAddSource:
https://developer.apple.com/library/mac/documentation/CoreFoundation/Reference/CFRunLoopRef/index.html#//apple_ref/c/func/CFRunLoopAddSource
http://www.cocoabuilder.com/archive/cocoa/242438-trouble-with-event-taps.html

I have compiled and verified the fix.

For anyone wishing to verify, I had to make several other changes to get Synergy to compile on El Capitan with Qt 5.5. I've compiled those changes into a single commit 4c7a70b for easy reference, but they should probably be pulled in from their corresponding PRs referenced in that commit.

### DIFF
--- a/src/lib/platform/OSXScreen.cpp
+++ b/src/lib/platform/OSXScreen.cpp
@@ -841,6 +841,7 @@ OSXScreen::disable()
 	// FIXME -- stop watching jump zones, stop capturing input
 	
 	if (m_eventTapRLSR) {
+		CFRunLoopRemoveSource(CFRunLoopGetCurrent(), m_eventTapRLSR, kCFRunLoopDefaultMode);
 		CFRelease(m_eventTapRLSR);
 		m_eventTapRLSR = nullptr;
 	}


XinyuHou:
Please test this nightly

http://synergy-project.org/nightly?filter=v1.7.6-rc3-51705f7

XinyuHou:
Merged

PR #5178, opened 2016-01-09
Fix compile warnings

ckorn:
Warnings were
- unused variables: Just "use" it
- members initialized in wrong order: reorder them
- variable used twice in initialisation: remove one occurrence


src/lib/arch/unix/ArchNetworkBSD.cpp
```diff
@@ -332,7 +332,9 @@ ArchNetworkBSD::pollSocket(PollEntry pe[], int num, double timeout)
 		do {
 			ignore = read(unblockPipe[0], dummy, sizeof(dummy));
 		} while (errno != EAGAIN);
-
+		// without ignore there is a warning that the return value of read is ignored
```
XinyuHou: Why don't we put ignore in the do scope and do some error checking?

XinyuHou:
Closed because of insufficient information.

Issue #5183, opened 2016-01-15
Slowly moving the cursor has no effect on high DPI clients


aragonsr:
Server: Windows 10 laptop
Client: Windows 10 desktop
version 1.7.5

When I move the cursor onto the client machine, if I move the cursor really slowly to the left or up it works fine. If I move the cursor really slowly to the right or down, the cursor does not move. I have to flick the mouse to get the cursor to move.

It looks like the cursor won't move unless the velocity is above some threshold.

This causes issues when I'm trying to select a tab in chrome, among other frustrations.

AMagill:
I have the same problem, and it's very annoying.  My Windows 10 server has a pair of 4K screens scaled to 150%, and the client has a single 1080p screen scaled to 100%.

This appears to happen because DPI scaling is done by integer division.  See [MSWindowsScreen::onMouseMove](https://github.com/synergy/synergy/blob/master/src/lib/platform/MSWindowsScreen.cpp#L1360-L1361) and [Server::onMouseMoveSecondary](https://github.com/synergy/synergy/blob/master/src/lib/server/Server.cpp#L2010-L2011).  If the DPI-scaled mouse movement in a single update is less than one pixel in the positive direction, it gets truncated to zero.  A sub-pixel movement in the negative direction also gets truncated, but in this case always results in a movement of one pixel.  The server needs to accumulate these fractional movements across updates.

This looks to be an easy fix.  If I can manage to get the synergys building on my machine, I'll see about make a pull request.

the-wes:
seeing a lot of reports of this since 1.7.5. if they uninstall and move back to 1.7.4 (and disable DPI scaling on synergys.exe), it doesn't happen. confirmed new bug!

AMagill:
That's because DPI scaling of mouse movement was only added in commit a09bfc5.  Which is the right thing to do, but implementing it is a bit trickier than it appears at first glance.  I've been using the fix I wrote (PR #5190) over the last eleven days and it's worked pretty well for me.

XinyuHou:
Could someone test this nightly?
http://synergy-project.org/nightly?filter=pr5190-accumulate-fractional-moves-beta-f240faf

mike-rsi:
[...]

PR #5190, opened 2016-01-19
Accumulate fractional moves across updates.  #5183

AMagill:
This change allows fractional pixel movements to accumulate between updates.

For example, if you have a DPI setting of 150% and move the mouse slowly to the right so that there is a one pixel movement every update, those movements will be scaled to 0.66 pixels.  Right now that value is always truncated to zero, making it difficult to move the mouse slowly down or right.

A more correct approach is to accumulate the fractional part of those movements.  For example:
Pixel movements:  +1, +1, +1, +1, +1, +1, +1, +1
DPI scaled movements:  +0.66, +0.66, +0.66, +0.66, +0.66, +0.66, +0.66, +0.66
Old movements sent to client:  0, 0, 0, 0, 0, 0, 0, 0
New movements sent to client: 0, 1, 1, 0, 1, 1, 0, 1


src/lib/platform/MSWindowsScreen.h
```diff
@@ -266,6 +270,9 @@ class MSWindowsScreen : public PlatformScreen {
 	// last mouse position
 	SInt32				m_xCursor, m_yCursor;
 
+	// accumulated fractional pixel moves
+	float					m_xFractionalMove, m_yFractionalMove;
```
XinyuHou: Indentation to line 25 please

XinyuHou:
@AMagill 
Could someone test this nightly?
http://synergy-project.org/nightly?filter=pr5190-accumulate-fractional-moves-beta-f240faf

AMagill:
@XinyuHou Could you re-run that nightly build?

XinyuHou:
Merged in 1.8.4

Issue #3286, opened 2014-10-13
Problems with Korean chars (T^T)


nbolton:
**Imported issue:**

* Author: onesun archiForum
* Date: 2012-08-28 01:59:36
* Legacy ID: 3356

**Steps to reproduce:**

1. choose Korean input source. (Eng -> Kor)
2. i cannot input Korean.
3. the characters i wrote by Korean input source are look English on my monitor even though i choose Korean input source.
4. sometimes when i write Korean i can see Korean.
5. but sometimes T^T

Thank you for synergy.
synergy is unbelievable software!!

**Versions and operating systems:**

Server : Synergy 1.4.10 nightly with Windows 7 Professional Korean

Client : Segergy 1.4.10 nightly with Mac OS X Mountain Lion 10.8.1

**Temporary workarounds:**

i knew a temporary solution.
that's a clicking input source icon on menu bar.
when i cannot write korean on Korean input source, i can write korean by clicking the icon.

**Additional comments:**

None

the-wes:
There is some info here on a possible fix for Korean characters:

http://b4you.net/blog/265

nlyan:
Apparently fixed by PR #5196

Issue #3071, opened 2014-10-13
Hangul/English toggle and hanja function key for Korean Keyboard is not working


nbolton:
**Imported issue:**
- Author: Chung Whan Lee
- Date: 2012-03-31 19:56:03
- Legacy ID: 3141

**Steps to reproduce:**  
1. Use the Korean keyboard.  
2. press Hangul/English toggle key.  
3. press Hanja Key.  

Expected: Pressing Hangul/English key have to toggle between Hangul and English. pressing Hanja Key have to convert Hangul to Hanja  
Actual: Pressing Hangul/English key don't any function. Also Hanja key don't  

**Versions (Synergy, OS):**  

Server: repository revision 1309, windows 7 64bit  
Client 1: repository revision 1309, windows 7 64bit  
Client 1: repository revision 1309, ubuntu 10.04 lucid 32bit  

**Workaround:**  

This problem issued many times like these  
http://synergy-foss.org/osqa/questions/1047/korean-keyboard-support  
issue #redmine:662  
issue #redmine:684  
issue #redmine:705  

I made the simple patch that is not interfering other one, and it works perfectly.   

**Comments:**  

This problem is critical to korean because can't type a korean. So when each times use other version of synergy, korean must apply patch.

the-wes:
A user submitted a patch which may help with this issue:

diff -Naur synergy-1.2.7.org/lib/platform/CMSWindowsKeyState.cpp synergy-1.2.7/lib/platform/CMSWindowsKeyState.cpp 
--- synergy-1.2.7.org/lib/platform/CMSWindowsKeyState.cpp 2005-12-19 03:41:33.000000000 +0900 
+++ synergy-1.2.7/lib/platform/CMSWindowsKeyState.cpp 2006-01-27 00:55:25.000000000 +0900 
@@ -56,7 +56,7 @@ 
/* 0x012 */ { kKeyAlt_L }, // VK_MENU 
/* 0x013 */ { kKeyPause }, // VK_PAUSE 
/* 0x014 */ { kKeyCapsLock }, // VK_CAPITAL 
- /* 0x015 */ { kKeyNone }, // VK_KANA 
+ /* 0x015 */ { kKeyShiftSpace }, // VK_KANA 
/* 0x016 */ { kKeyNone }, // VK_HANGUL 
/* 0x017 */ { kKeyNone }, // VK_JUNJA 
/* 0x018 */ { kKeyNone }, // VK_FINAL 
[...]

PR #5194, opened 2016-01-23
Fix Korean toggle key issue

ommokazza:
Korean and Japanese keyboards have same keycode in widnows
(VK_HANGUL == VK_KANA and VK_HANJA == VK_KANJI)
But They have different X11 keysym.


ommokazza:
This patch is for issue #3071, #3286

ommokazza:
This does not consider various case..

PR #5245, opened 2016-03-18
Add an option to ignore VirtualBox network adapters

### LINKED ISSUE #4251: Auto config selects the wrong IP address
I'd like to use autoconfiguration, but I have multiple interfaces on my PC (server).  One interface is a physical network, and the other is a VM network that is internal-only for virtualbox instances.

After the upgrade to 1.6.2, autoconfiguration advertises both addresses, and my mac (client) only tries to connect to the internal-only network, which fails.

I suspect that there is a configuration option to fix this, but I have only used the interactive configuration so far, and I think it should be possible to mask out an interface via the interactive configuration.

sometoby:
This adds an option to ignore network interfaces starting with "vboxnet" to prevent VirtualBox network adapters from interfering with autoconfiguration. It should partially fix #4251.

The added option could perhaps be extended to include virtual network adapters from other virtualization software, too.

### DIFF
--- a/src/gui/src/MainWindow.cpp
+++ b/src/gui/src/MainWindow.cpp
@@ -1059,7 +1059,7 @@ bool MainWindow::on_m_pButtonBrowseConfigFile_clicked()
 	return false;
 }
 
-bool  MainWindow::on_m_pActionSave_triggered()
+bool MainWindow::on_m_pActionSave_triggered()
 {
 	QString fileName = QFileDialog::getSaveFileName(this, tr("Save configuration as..."));
 
--- a/src/gui/src/ServerConfigDialog.cpp
+++ b/src/gui/src/ServerConfigDialog.cpp
@@ -25,9 +25,10 @@
 #include <QtGui>
 #include <QMessageBox>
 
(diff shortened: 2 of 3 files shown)


XinyuHou:
@sometoby 
Why do you want to add this option only for Unix?

sometoby:
Network device names aren't human-readable on Windows ([source](https://doc.qt.io/qt-4.8/qnetworkinterface.html#name)), which means we can't easily check if the interface is provided by VirtualBox.

It might actually be a better approach to favor whatever interface the default gateway uses, but I'm not sure how difficult that would be.

XinyuHou:
What is the VirtualBox interface name on Windows?

sometoby:
I don't know, and I'm unable to test that right now. I'll install VirtualBox on my Windows box later and see if I can get something useful back.

krenn08:
VirtualBox Host-Only Network X (where X=1..9)

Additionally: VMware Network Adapter VMnetX (where X=1..9)

XinyuHou:
@krenn08 
How did you find out that information?

XinyuHou:
Related to #5765

PR #4968, opened 2015-08-15
Add --enable-crypto argument to help text

legonigel:
Does what is says. Doesn't need review, just a text change.

PR #5270, opened 2016-04-07
Add "--enable-crypto" to "HELP_COMMON_INFO_1" DEFINE

justin531:
added "--enable-crypto" to the "HELP_COMMON_INFO_1" Define, so users can find this command parameter as it is not documented anywhere else.
I had to put in an issue request to find it my self.


legonigel:
This is duplicate to my pr #4968.

PR #5301, opened 2016-04-25
Fix hm.sh to prefer python2 if available

moises-silva:
The build scripts only support python2 at the moment
and distributions that already switched to python3
require explicit use of python2 when using hm.sh

### DIFF
--- a/hm.sh
+++ b/hm.sh
@@ -1,3 +1,11 @@
 #! /bin/bash
 
-python hm.py "$@"
+pybin=python
+# Use python2 explicitly if available
+# otherwise things break on arch or any
+# distro that has python3 by default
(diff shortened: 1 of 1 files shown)


XinyuHou:
@moises-silva

moises-silva:
@XinyuHou No. Do you need one? this seems pretty straight forward.

XinyuHou:
@nlyan 
Could you test on your Arch with python 3 installed?

nlyan:
This PR will be closed soon.

All PRs targetting the old build scripts will be closed upon release of v1.9.0-beta, which won't be using them, and will ship with new build instructions.

If there's anything in here build-system agnostic that needs fixing, please shout loudly now.

PR #5315, opened 2016-05-03
Default install path on Ubuntu is /usr/include/...

johnweldon:
When using apt to install dependencies; include path is /usr/include/...
rather than /usr/local/include/...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,7 +202,7 @@ if (UNIX)
 	else() # not-apple
 
 		# add include dir for bsd (posix uses /usr/include/)
-		set(CMAKE_INCLUDE_PATH "${CMAKE_INCLUDE_PATH}:/usr/local/include")
+		set(CMAKE_INCLUDE_PATH "${CMAKE_INCLUDE_PATH}:/usr/local/include:/usr/include")
 
 		set(XKBlib "X11/Xlib.h;X11/XKBlib.h")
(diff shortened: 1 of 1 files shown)


XinyuHou:
@johnweldon 

johnweldon:
No, I looked for one and didn't find one.  It's just a little thing I ran
into when compiling for ubuntu.




John Weldon
m) 503-941-0825

XinyuHou:
@johnweldon 
Without the fix, does it stop you from compiling?

johnweldon:
IIRC yes, but it's been a while :)

XinyuHou:
@johnweldon 
Could you tell us your ubuntu version?

johnweldon:
I don't recall what version I was using in May; probably still 14.04 though.  I see that in 16.04 which I'm using now that the default include path is still /usr/include/... not /usr/local/include/...

nlyan:
This PR makes no sense to me. /usr/include is always in the default include path

johnweldon:
This PR is from a year ago... I don't really have any recollection.  I should have filed an issue first with all the details.  Feel free to just close this PR.  If no-one else can reproduce it with a default Ubuntu install then it's probably just something weird in my setup.

nlyan:
@johnweldon Thanks for the contribution anyway. Sorry this was dragged out so long.

PR #5318, opened 2016-05-06
Documentation

virtualdxs:
I've just done a couple things to make the documentation nicer. I do plan on contributing more in the future, but the first thing I wanted to do was make the main Github page look nicer.


virtualdxs:
Could you please at least say why you rejected this?

Issue #5204, opened 2016-02-03
Drag and Drop for Files is not working


anubhavranjan:
Hi,

I am using Windows 10 as the Server and Mac OSX as the Client.
Everything works great except for the **drag and drop** of files from both Windows and Mac.

Everytime I drag a file from Windows, the moment I reach the edge, the file is no longer getting dragged and I get a Window in the Taskbar with title **DropWindow**. After repeating the same for couple of times, I could see the application is **synergys.exe** in the Task Manager.

I have recorded the screen as well for the activity:
http://screencast.com/t/fkTP8kSG

XinyuHou:
@anubhavranjan 
Could you give us some logging information?

rockking1379:
i can't seem to get drag and drop working either. if it helps any i complied synergy from source as opposed to @anubhavranjan who it looks like paid for the pro version

wywzxxz:
@XinyuHou 
same problem here.
OS: windows 10 x64
version: compiled from the latest master branch, build for "windows 10"(not "windows 10 x64")
when I drag the file cross the border from computer A to computer B. the file was no longer dragged in computer A and a new window  flash across. nothing happened in computer B, the mouse comes without the file and the release of mouse cause nothing.



here is log(DEBUG1) of computer A:
computer A:DESKTOP-SBHSR1Q
computer B:S3Pwywzxxz


[2016-04-28T19:07:40] DEBUG1: try to leave "DESKTOP-SBHSR1Q" on right
[2016-04-28T19:07:40] DEBUG1: no neighbor right
[2016-04-28T19:07:40] NOTE: accepted client connection
[2016-04-28T19:07:40] DEBUG1: registered event type IStreamEvents::outputError as 44
[2016-04-28T19:07:40] DEBUG1: registered event type IStreamEvents::inputShutdown as 45
[2016-04-28T19:07:40] DEBUG1: registered event type IStreamEvents::outputShutdown as 46
[2016-04-28T19:07:40] DEBUG1: saying hello
[2016-04-28T19:07:40] DEBUG1: registered event type ClientProxyUnknownEvents::success as 47
[2016-04-28T19:07:40] DEBUG1: registered event type ClientProxyUnknownEvents::failure as 48
[2016-04-28T19:07:40] DEBUG1: parsing hello reply
[2016-04-28T19:07:40] DEBUG1: querying client "SP3wywzxxz" info
[...]

PR #5332, opened 2016-05-14
Unicode filename support for Drag and Drop

wywzxxz:
For now, Drag and Drop any file which name contains unicode will fail,because the  conversion method beturn wchar to char are wrong. 
i.e: a file named "腾讯QQ" will cause a "failed to get drag file name from OLE" error,and probably crash synergys(because there's a stack overflow). It is not because OLE truely failed to get filename---actually it do get the name,but mistakly  converse it.
As a chinese User,It's great chance that I Drag a chinese-word-contained file. It will fail, what's worse is it will crash synergys(which I think is the file transfer system),make pure english file transfer unavaliable.So besicly it looks like file transfor weren't work on chinese environment.

This pull should fix above problem.


wywzxxz:
motivied by this report:#5204

virtualdxs:
Note: synergys is the entire Synergy server. The GUI just manages starting it (or synergyc for the client) in the background.

XinyuHou:
@wywzxxz 
Please test this nightly.
http://symless.com/nightly?filter=synergy-issue4376-unicode-drag-filename-rc2-86aa690

XinyuHou:
@wywzxxz 
I have tested with your changes. It still has the "failed to get drag file name from OLE" error.

Issue #4323, opened 2015-01-15
Make auto-elevate optional


the-wes:
There are some corner cases where auto-elevate doesn't make the most sense. Please add an option to disable it so that it will either be always unelevated, or always elevated.

Sometimes Synergy tries to restart with elevation when it doesn't need to, and sometimes it takes a loooong time to reconnect after restarting, making it more practical for the user to reach over and use the client's mouse to click Yes to the UAC prompt than to wait for Synergy.

Also, when Synergy is set to Elevated mode, auto-elevate still restarts Synergy when a UAC prompt comes up even though it is not needed.

PR #5371, opened 2016-06-10
v1.8.3 Make auto-elevate optional

nlyan:
This pr expands the boolean elevation option to a tristate preference, closing #4323 and #3241


src/gui/src/IpcClient.h
```diff
@@ -21,6 +21,8 @@
 #include <QObject>
 #include <QAbstractSocket>
 
+#include "AppConfig.h"
```
nlyan: Separate header file would be more clear.

PR #5506, opened 2016-07-27
Add exemption for linking GPL code with OpenSSL

epakai:
Per OpenSSL FAQ https://www.openssl.org/docs/faq.html#LEGAL2
Please include a blanket exception for linking with OpenSSL.


nlyan:
Merged in 1.8.3

PR #5551, opened 2016-08-19
Fix typo: s/paclage/package/

elopio:


nlyan:
Thanks, fixed in 1.8.3

PR #5552, opened 2016-08-19
Fix hangul multi ime issue

ommokazza:
When we have 2 or more IMEs for different languages in Windows OS,
virtaual key table is mapped for all IMEs not just active IME.

And this cause wrong modifier key pressed issue.

For example, if I use Korean and Japanese IMEs, 
pressing hangul key makes alt key pressed.
So when I press just 'a', client interpretes that 'alt-a'.


XinyuHou:
@ommokazza 

Thank you for creating this PR.

Is this PR related to any issue in our issue tracker?

Why there are 49 commits while there is only 1 change?

ommokazza:
@XinyuHou 
I can't find any issue about this.

And my branch is based from tags/v1.8.2-stable, so this shows many commit logs. (I think)
If this is not proper RR, I'll make a new RR.

XinyuHou:
@ommokazza 
Could you create an issue in our issue tracker and associate it with your PR?

Em, it's weird. It says 49 commits even though there is only just 1 change.

Could you try to create another PR based from our master?

Thanks

nlyan:
Thanks @ommokazza, we'll get to this soon.

ommokazza:
I have made a new PR on #5578

Issue #5577, opened 2016-08-31
Multi IMEs issue on Windows clients


ommokazza:
When I have 2 or more IMEs for different languages in Windows OS,
virtaual key table is mapped for all IMEs not just active IME.

And this cause wrong modifier key pressed issue.

For example, if I use Korean and Japanese IMEs, 
pressing hangul key makes alt key pressed.
So when I press just 'a', client interpretes that 'alt-a'.

But I testing this on just 2 cases. (Korean and English / Korean and Japanese)

ommokazza:
I have made a fix for is issue - #5578

ommokazza:
@XinyuHou 
This issues is invalid if hangul key patch is not applied.
I'll make the issue and pull request later.

Thanks

PR #5196, opened 2016-01-25
Korean and Japanese keyboards have same keycode

ommokazza:
(VK_HANGUL == VK_KANA and VK_HANJA == VK_KANJI)
But They have different X11 keysym.

So VK_HANGUL and VK_HANJA do not working on 103/106 korean keyboard layout in windows server and linux client

PR #5578, opened 2016-08-31
Fix muti imes issue in Windows OS,

ommokazza:
virtaual key table is mapped for all IMEs not just active IME.

And this cause wrong modifier key pressed issue.

For example, if I use Korean and Japanese IMEs,
pressing hangul key makes alt key pressed.
So when I press just 'a', client interpretes that 'alt-a'.

### DIFF
--- a/src/lib/platform/MSWindowsKeyState.cpp
+++ b/src/lib/platform/MSWindowsKeyState.cpp
@@ -1077,7 +1077,7 @@ MSWindowsKeyState::getKeyMap(synergy::KeyMap& keyMap)
 		}
 
 		// set virtual key to button table
-		if (GetKeyboardLayout(0) == m_groups[g]) {
+		if (activeLayout == m_groups[g]) {
 			for (KeyButton i = 0; i < 512; ++i) {
 				if (m_buttonToVK[i] != 0) {
 					if (m_virtualKeyToButton[m_buttonToVK[i]] == 0) {


ommokazza:
This is fix for issue #5577

XinyuHou:
@ommokazza 
Thank you for creating this PR.

Could you give us some explanations about your change?

src/lib/platform/MSWindowsKeyState.cpp
```diff
@@ -1077,7 +1077,7 @@ MSWindowsKeyState::getKeyMap(synergy::KeyMap& keyMap)
 		}
 
 		// set virtual key to button table
-		if (GetKeyboardLayout(0) == m_groups[g]) {
+		if (activeLayout == m_groups[g]) {
```
XinyuHou: Line 931 initiate this variable as HKL activeLayout = GetKeyboardLayout(0); So why do we need to call the function again?

ommokazza: Because line 931 is used to remember original active keyboard layout.
Please see a few lines later, line 941.

ActivateKeyboardLayout(m_groups[g], 0);
This make active keyboard layout different one.
So, (I think) "if (GetKeyboardLayout(0) == m_groups[g])" will be always true.

XinyuHou:
@ommokazza 
Please test this nightly.
http://symless.com/nightly?filter=f38e2fa

ommokazza:
@XinyuHou 
ok, I'll test that build and share it on my korean blog.

ommokazza:
@XinyuHou 
I installed this version, but I can't test.
Because this version of synergy ignore hangul-key.

When I pressed hangul-key, it shows like below log

[2016-12-08T10:52:00] DEBUG1: hook: 0x00000015 0x01f20001
[2016-12-08T10:52:00] DEBUG1: hook: 0x06000015 0x01f20001
[2016-12-08T10:52:00] DEBUG1: hook: 0x07000015 0x01f20001
[2016-12-08T10:52:00] DEBUG1: event: Key char=0, vk=0x15, nagr=0, lParam=0x01f20001
[2016-12-08T10:52:00] DEBUG1: new mask: 0x2000
[2016-12-08T10:52:00] DEBUG1: new mask: 0x2000
[2016-12-08T10:52:00] DEBUG1: cannot map key


I didn't see the logic of master branch, so i don't know what is problem right now.
I had tested this on v1.8.x version.

XinyuHou:
@ommokazza 
I know you mentioned ignoring this comment from another PR.

Is this PR still valid?

ommokazza:
@XinyuHou 
Is the nighly build based 1.8.6-rc2?
If so, I have to check this PR a bit more.

ommokazza:
@XinyuHou 
This PR is invalid if hangul key patch is not applied.
I'll make the issue and pull request later.

Thanks

nlyan:
@ommokazza Is this PR valid if we merge #5196?

ommokazza:
@nlyan
Actually, I think this PR is valid if #5196 is not merged.
But I can't find any way to reproduce this issue.
So I wrote that this was invalid now, and I would request this later.

If #5196 is merged, this will be also merged as posible.
I was reported this issue on my blog.

nlyan:
@ommokazza 

Can you test this build?

http://symless.com/nightly?filter=4b913b5

ommokazza:
@nlyan 
of course. i'll install and share it on my blog.

PR #5581, opened 2016-09-02
add missing dll (libgcc_s_dw2-1.dll)

insidexino:
Removing libgcc_s_dw2-1.dll causes missing DLL error.
I added it back to Product.wxs


nlyan:
Our intention is to phase out the GCC compilation stage so the Windows GUI is compiled with VS2015. This library will not be required by the time 1.9.0 is ready for beta.

PR #5600, opened 2016-09-16
v1.8.3-Enable media key support on macOS servers

nlyan:


src/lib/platform/OSXKeyState.cpp
```diff
@@ -117,7 +119,9 @@ static const KeyEntry	s_controlKeys[] = {
 	{ kKeyCapsLock,		s_capsLockVK },
 	
 	{ kKeyMissionControl, s_missionControlVK },
-	{ kKeyLaunchpad, s_launchpadVK }
```
XinyuHou: We should add a new code style for this, some of them are aligned vertically while others just one space.

src/lib/platform/OSXScreen.cpp
```diff
@@ -1333,6 +1334,27 @@ OSXScreen::onKey(CGEventRef event)
...
+	bool down;
+	bool isRepeat;
+
+	if (!getMediaKeyEventInfo (event, &keyID, &down, &isRepeat)) {
+		LOG ((CLOG_ERR "Failed to decode media key event"));
+		return;
+	}
+
+	LOG ((CLOG_DEBUG2 "Media key event: keyID=0x%02x, %s, repeat=%s",
+						keyID, (down ? "down": "up"),
```
XinyuHou: Indentation doesn't look right

src/lib/platform/OSXKeyState.cpp
nlyan: It was already inconsistent. Changing it would have just muddied the commit log with whitespace fixes.

PR #5671, opened 2016-10-16
Accumulate scrolls less than supported scroll on Linux (fixes #5670)

### LINKED ISSUE #5670: Linux client ignores scrolls of less than 120 points
### Operating Systems

Client: Ubuntu Linux 16.04
Server: OSX 10.12 Sierra
### Synergy Version

1.8.4 stable
### Steps to reproduce bug
1. Scroll slowly on the server
2. Observe that the client ignores the scroll
3. Observe that the log is filled with "WARNING: Wheel scroll delta (22) smaller than threshold (120)"
### Other info

Will do a PR with a fix shortly

p12tic:
This PR resolves https://github.com/symless/synergy/issues/5670 issue and the issue underlying https://github.com/symless/synergy/issues/3657 feature enhancement request.

Previously synergy ignored scrolls less than 120px which is the default value of --yscroll. The problem is that this value in certain cases (such as when using OSX as the server) larger than most of the scroll events during normal use which are then lost. This results in poor usability because it's only possible to do fast scrolls and the user can't e.g. reasonably perform a single scroll of one or several lines.

The fix is very similar to https://github.com/AMagill/synergy/commit/cfb784fa324562aa8b8d185569297a6f72fbe073. We accumulate scroll deltas that are less than 120 pixels, issue additional scroll whenever the accumulated scroll is outside the range (-120 .. 120) and reduce the accumulated scroll by the scrolled amount.

This results in a much better user experience even in this case when the platform does not support fine scrolls at all by design.

The fix has been tested with Ubuntu Linux 16.04 as client and MacOSX Sierra 10.12 as the server.

### DIFF
--- a/src/lib/platform/XWindowsScreen.cpp
+++ b/src/lib/platform/XWindowsScreen.cpp
@@ -99,6 +99,7 @@ XWindowsScreen::XWindowsScreen(
 		IEventQueue* events) :
 	m_isPrimary(isPrimary),
 	m_mouseScrollDelta(mouseScrollDelta),
+	m_accumulatedScroll(0),
 	m_display(NULL),
 	m_root(None),
 	m_window(None),
@@ -868,9 +869,11 @@ XWindowsScreen::fakeMouseWheel(SInt32, SInt32 yDelta) const
 		return;
 	}
 
+	int numEvents = accumulateMouseScroll(yDelta);
+
 	// choose button depending on rotation direction
 	const unsigned int xButton = mapButtonToX(static_cast<ButtonID>(
-												(yDelta >= 0) ? -1 : -2));
+												(numEvents >= 0) ? -1 : -2));
 	if (xButton == 0) {
 		// If we get here, then the XServer does not support the scroll
 		// wheel buttons, so send PageUp/PageDown keystrokes instead.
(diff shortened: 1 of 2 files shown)


XinyuHou:
Please test this nightly.
http://symless.com/nightly?filter=synergy-issue5670-accumulate-scrolls-rc2-02b9a32

XinyuHou:
Please try this nightly which merged pr 5671 and 5673
http://symless.com/nightly?filter=synergy-issue5604-choppy-scrolling-rc2-f03a03a

p12tic:
Hi, I've verified that the nightly build fixes the observed issues.

XinyuHou:
@p12tic 
Thanks for the feedback. Did you test the nighty ending with f03a03a?

p12tic:
Yes, I've tested this revision f03a03a.

p12tic:
Hi, what prevents this being merged into master or some other release-targeted branch? It looks like that other people are interested into these fixes too and I can only refer them to my private builds instead of the official ones. I don't want to maintain my own set of fixes separately, because over time this could become a fork of the project.

p12tic:
Closing in favor of https://github.com/debauchee/barrier/pull/68.

PR #5673, opened 2016-10-16
Improve precision of grabbed scroll events on OSX server (fixes #5672)

### LINKED ISSUE #5672: The smallest scroll size on OSX is 76 pixels
### Operating Systems ###

Client: Ubuntu Linux 16.04
Server: OSX 10.12 Sierra

### Synergy Version ###

1.8.4 stable

### Steps to reproduce bug ###

1. Scroll slowly on the server
2. Observe that the client ignores the scroll
3. Scroll faster until the client log starts to show  "WARNING: Wheel scroll delta (76) smaller than threshold (120)" messages
4. Observe that the scrolling is very jumpy.

### Other info ###

Will do a PR with a fix shortly

p12tic:
On OSX synergy uses old API to retrieve the scroll event delta. This results in retrieved scroll sizes being multiple of 76 points. Effectively, any scroll event that results in delta that is less that this threshold is ignored and additionally, even fast scrolling is very jumpy.

This PR fixes this by using a different API that returns fractional scroll event deltas too.

The new API is available since OSX 10.8 (https://developer.apple.com/library/content/releasenotes/General/APIDiffsMacOSX10_8/CoreGraphics.html) so should be no issues with backwards compatibility. The old API was introduced in OSX 10.4.

The fix was tested on OSX 10.12 server and Ubuntu Linux 16.04 client. 

Together with https://github.com/symless/synergy/pull/5671 this fix results in very smooth scrolling in this client/server combination, the feel is almost indistinguishable from native behavior.

### DIFF
--- a/src/lib/platform/OSXScreen.cpp
+++ b/src/lib/platform/OSXScreen.cpp
@@ -1416,7 +1416,7 @@ OSXScreen::mapMacButtonToSynergy(UInt16 macButton) const
 }
 
 SInt32
-OSXScreen::mapScrollWheelToSynergy(SInt32 x) const
+OSXScreen::mapScrollWheelToSynergy(float x) const
 {
 	// return accelerated scrolling but not exponentially scaled as it is
 	// on the mac.
@@ -1425,7 +1425,7 @@ OSXScreen::mapScrollWheelToSynergy(SInt32 x) const
 }
 
 SInt32
-OSXScreen::mapScrollWheelFromSynergy(SInt32 x) const
+OSXScreen::mapScrollWheelFromSynergy(float x) const
 {
 	// use server's acceleration with a little boost since other platforms
 	// take one wheel step as a larger step than the mac does.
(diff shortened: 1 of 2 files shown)


ssreekanth:
@p12tic it would be nice if you could have a check for OSX version, and call the api appropriately. Something like -
```
if (osx_version >= 10.8) then
  // call new api
else
  // call old api
```
That would leave Synergy app backward compatible as well for guys like me, who still try to keep the app running on Snow Leopard.

p12tic:
I assumed that since the lowest supported OSX version that is supported by Synergy is 10.9, that additional code would be unnecessary cruft and to be avoided.

XinyuHou:
Please test this nightly.
http://symless.com/nightly?filter=synergy-issue5604-choppy-scrolling-rc2-d68bc41

XinyuHou:
Please try this nightly which merged pr 5671 and 5673
http://symless.com/nightly?filter=synergy-issue5604-choppy-scrolling-rc2-f03a03a

p12tic:
Hi, I've verified that the nightly build fixes the observed issues.

p12tic:
@nbolton Hey, is there any chance for this PR to be merged? I've been using the changes in a custom-built version of synergy for more than a year and have not seen any issues. If you think the PR makes sense, I will rebase the changes to the current master and retest.

p12tic:
Closing in favor of https://github.com/debauchee/barrier/pull/64.

PR #5735, opened 2016-11-08
force XSync to help key repeat detection

Unknow0:
this avoid miss in key repeat detection.


XinyuHou:
@Unknow0 
Could you give us some insight of your changes?

Unknow0:
I noticed some time the detection of key repeat fail.

When i looked into it the way you detect the key repeat should works.
The only way to miss the key repeat is if the KeyPress isn't in the event queue.

So i added a call to XSync to retreive all event before looking for the KeyPress event.

XinyuHou:
@Unknow0 
Please test this nightly
http://symless.com/nightly?filter=linux-key-repeat-rc2-03ec531

XinyuHou:
@Unknow0 
Sorry, it seems this PR doesn't solve the problem.
https://github.com/symless/synergy/issues/4658#issuecomment-265192845

PR #5746, opened 2016-11-20
Adjust showSecureCipherInfo() to build with newer openssl versions

epakai:
Allows synergy to compile against newer versions of openssl (tested by building against Debian's openssl 1.1.0c) while maintaining compatibility with older versions

### DIFF
--- a/src/lib/net/SecureSocket.cpp
+++ b/src/lib/net/SecureSocket.cpp
@@ -805,10 +805,15 @@ SecureSocket::showSecureCipherInfo()
 		showCipherStackDesc(sStack);
 	}
 
-	// m_ssl->m_ssl->session->ciphers is not forward compatable, In future release
-	// of OpenSSL, it's not visible, need to use SSL_get_client_ciphers() instead
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+	// m_ssl->m_ssl->session->ciphers is not forward compatable,
+	// In future release of OpenSSL, it's not visible,
 	STACK_OF(SSL_CIPHER) * cStack = m_ssl->m_ssl->session->ciphers;
-		if (cStack == NULL) {
+#else
+	// Use SSL_get_client_ciphers() for newer versions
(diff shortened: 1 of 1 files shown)


XinyuHou:
@epakai 
Sorry we can't merge this pr for now. As we are not bundle OpenSSL lib with our binaries. So there would be a mismatch between OpenSSL calls no matter which version we compiled against. Let let me know if I'm wrong.

yupi2:
@XinyuHou Since any Linux distros would be compiling and packaging Synergy in their own package managers it would be tied to the packaged OpenSSL version so that wouldn't be a problem would it? And distributing compiled Linux binaries are going to be problematic anyway with multiple versions of OpenSSL.

There's also an option of statically linking something. LibreSSL works for this and can be used in CMake as an external-project.

anthraxx:
I agree that a proper supported switch should be the way to go as different distributions want to compile from source and may have different versions of openssl (1.0 or 1.1).

Statically linking openssl should only be considered for the binaries provided by synergy itself, its certainly nothing that most distributions want to do, using the system openssl is the preferred way to build from source. We (Arch Linux) are looking forward for a support of 1.1 without static linking and/or bundled openssl stuff.

HorreC:
while this isnt the best patch job I did take the diff and stripped out some of the debug or multi version of openssl and got it to run on debian with OpenSSL 1.1.0e . 
starting at line 808 in src/lib/net/SecureSocket.cpp
	if (OPENSSL_VERSION_NUMBER < 0x10100000L) {
	STACK_OF(SSL_CIPHER) * cStack = SSL_get_client_ciphers(m_ssl->m_ssl);
	showCipherStackDesc(cStack);
	}
	return;
}  //this should be line 813

And this compiled and ran fine for me. And those that wanna just compile till its pushed this might work for you.

yupi2:
@HorreC This pull-request handles different versions of OpenSSL already and there's no reason to remove the DEBUG1 messages.

bengecko:
The above fix from @HorreC worked for me as well. When is this scheduled for release?

nlyan:
Merged in to v1.9. Thanks for your contribution

PR #5785, opened 2016-12-29
fixed vertical screen switch problem when cursor is in a corner

hobb0001:
I have my screens configured such that there is a screen below my primary screen, but nothing to the left of my primary screen. I noticed that when the cursor is in the bottom-left corner of the primary screen, it wouldn't switch to the lower screen. The problem is that the code in `Server::onMouseMovePrimary()` only calls `isSwitchOkay()` for a single direction. When the cursor is in the bottom-left corner, the code calls `isSwitchOkay()` for the `kLeft` direction, but then never checks the `kBottom` direction when that fails. The code in this PR checks `isSwitchOkay()` in both the horizontal and vertical directions when the cursor is in a corner.

### DIFF
--- a/src/lib/server/Server.cpp
+++ b/src/lib/server/Server.cpp
@@ -1766,52 +1766,66 @@ Server::onMouseMovePrimary(SInt32 x, SInt32 y)
 	}
 
 	// see if we should change screens
-	EDirection dir;
+	// when the cursor is in a corner, there may be a screen either
+	// horizontally or vertically.  check both directions.
+	EDirection dirh = kNoDirection, dirv = kNoDirection;
+	SInt32 xh = x, yv = y;
 	if (x < ax + zoneSize) {
-		x  -= zoneSize;
-		dir = kLeft;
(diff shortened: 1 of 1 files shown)


XinyuHou:
@hobb0001 
Are primary and secondary screens the same resolution and orientation?

hobb0001:
@XinyuHou No, the top (primary) screen is 1920x1080 and the bottom (secondary) screen is a laptop with 1366x768 resolution. Both are in standard widescreen/landscape orientation. The config is very basic; I've attached it for reference. Thanks.

[synergy.conf.txt](https://github.com/symless/synergy/files/678966/synergy.conf.txt)

```
section: screens
	laptop:
	desktop:
end
section: links
	laptop:
		up = desktop
	desktop:
		down = laptop
end
```

nlyan:
@hobb0001 

Can you see if this build works for you? 

http://symless.com/nightly?filter=4b913b5

hobb0001:
@nlyan Yep, it works great. Thanks!

Issue #3992, opened 2014-10-13
Mac OSX: Dragging object in Unity 3D does not work due to lack of mouse deltas


nbolton:
**Imported issue:**
- Author: James McMullan
- Date: 2014-05-23 18:22:40
- Legacy ID: 4062

**Steps to reproduce:**
1. Open Unity 3D
2. Open an existing scene or create a new one
3. Attempt to move a Game Object within the scene by selecting its 3D cursor and dragging it

Expected: Dragging an object in Unity 3D should move the object.

Actual: Dragging an object does nothing/

**Versions and operating systems:**

Any version of Synergy running on any version of Mac OSX

**Solution**

The issue is caused because the CPlatformSceen implementation on Mac OSX does not correctly emulate mouse dragging events. Mouse dragging events on Mac OSX require not only the new position to be set in the event but also the mouse movement delta. Adding the following code to the postMouseEvent function in COSXScreen fixes the issue:

```
SInt64 deltaX = pos.x;
deltaX -= m_xCursor;

SInt64 deltaY = pos.y;
deltaY -= m_yCursor;

CGEventSetIntegerValueField(event, kCGMouseEventDeltaX, deltaX);
CGEventSetIntegerValueField(event, kCGMouseEventDeltaY, deltaY);

double deltaFX = deltaX;
double deltaFY = deltaY;

CGEventSetDoubleValueField(event, kCGMouseEventDeltaX, deltaFX);
CGEventSetDoubleValueField(event, kCGMouseEventDeltaY, deltaFY);
```

**Temporary workarounds:**

 No work around

**Similar bugs:**

**Additional comments:**

XinyuHou:
Reverted

thediamondgames:
Hmm... Is the fix going to land to the mainstream? Because I see it has been reverted. I just added the lines from the solution to the source code, compiled for El capitan and now unity works just fine without any downsides I could think of.

cbrisebois42:
I would really love this to be fixed.  Having a second mouse just for Unity development is annoying.

XinyuHou:
> Maybe problem is in capture mouse/keyboard hold button states.
> I have this issue in Unity, UnrealEngine, Blender.
> I can grab short video on both machines with full description for more explanation of that problem, if you want.

PR #5816, opened 2017-01-25
#3992 macOS: Dragging broken in Unity

nlyan:
OSXScreen was not adding mouse movement deltas to mouse events while
dragging. Some 3D applications rely on these deltas to implement
dragging. Adding the mouse deltas to the mouse event fixes dragging in
these applications. Ex: Unity3d


nlyan:
Fixes #3992

Issue #5831, opened 2017-02-05
Screen size misdetected on multi-monitor display with XRandR (1.8.4+)


mwilck:
### Problem ###

My synergy client has 3 monitors: two 1920x1080 monitors standing vertically, and my horizontal 1920x1080 laptop screen on the right. This gives a total horizontal screen width of 1080+1080+1920 = 4080 (height is 1920).

```
mwilck@apollon:~> xrandr
Screen 0: minimum 320 x 200, current 4080 x 1920, maximum 8192 x 8192
eDP-1 connected 1920x1080+2160+840 (normal left inverted right x axis y axis) 309mm x 174mm
   1920x1080     60.02*+  47.99  
DP-1-1 connected 1080x1920+1080+0 left (normal left inverted right x axis y axis) 527mm x 297mm
   1920x1080     60.00*+
DP-1-3 connected primary 1080x1920+0+0 left (normal left inverted right x axis y axis) 518mm x 324mm
   1920x1080     60.00* 
```

**After update of synergy client to 1.8.7, I can't move the mouse pointer beyond horizontal position 1200**.  This can be verified by the log messages of the client. The client messages on my setup with 1.8.2 (good):
```
Synergy 1.8.2: [2017-02-05T21:04:29] DEBUG: screen shape: 0,0 4080x1920 (xinerama)
```
and with 1.8.7 (bad):
```
Synergy 1.8.7: [2017-02-05T20:55:55] DEBUG: screen shape: 0,0 1200x1920 (xinerama)
```

The number 1200 is particularly strange because this size doesn't appear anywhere in my configuration. I'm also not using Xinerama (at least not conciously) but that doesn't seem to matter here.

### Preliminary analysis ###

This was apparently cased by the changes made for issue #4420, pull request #4148. These changes enabled XRandR support, but apparently this support is broken. synergy uses calls to XRRSizes() and XRRRotations() to determine the screen size. I am not sure if this is still state of the art. The Linux xrandr client, for example, uses XRRGetScreenResources(). It seems to me that XRRSizes() returns the size of one monitor rather than the whole screen. 

The xrandr library on OpenSUSE Tumbleweed is libXrandr-1.5.1-1.1. It's possible (although not very likely IMO) that XRRSizes() is broken in this library version.

### Operating Systems ###

Server: Fedora 24
Client: openSUSE Tumbleweed (20170203)

### Synergy Version ###

client: 1.8.7 (bad), 1.8.2 (good)
server: 1.7.6

[...]

PR #5832, opened 2017-02-06
XRandR: fix screen size calculation

mwilck:
XWindowsScreen::saveShape() using XRRSizes / XRRRotations to
calculate screen dimensions when XRandR and a rotated screen
was detected. This is wrong. The screen dimensions in the
display properties already reflect rotation.

Moreover, on servers supporting XRandR >= 1.2, the XRRSizes()
and XRRRotations calls from XRandR 1.1 will return the
properties of the "primary output" in XRandR 1.2 terms rather than
the properties of the entire screen.

### DIFF
--- a/src/lib/platform/XWindowsScreen.cpp
+++ b/src/lib/platform/XWindowsScreen.cpp
@@ -974,22 +974,6 @@ XWindowsScreen::saveShape()
 	m_w = WidthOfScreen(DefaultScreenOfDisplay(m_display));
 	m_h = HeightOfScreen(DefaultScreenOfDisplay(m_display));
 
-#if HAVE_X11_EXTENSIONS_XRANDR_H
-	if (m_xrandr){
-	  int numSizes;
-	  XRRScreenSize* xrrs;
-	  Rotation rotation;
-	  xrrs = XRRSizes(m_display, DefaultScreen(m_display), &numSizes);
-	  XRRRotations(m_display, DefaultScreen(m_display), &rotation);
-	  if (xrrs != NULL) {
-	    if (rotation & (RR_Rotate_90|RR_Rotate_270) ){
-	      m_w = xrrs->height;
-	      m_h = xrrs->width;
-	    }
-	  }
-	}
-#endif
-
 	// get center of default screen
 	m_xCenter = m_x + (m_w >> 1);
 	m_yCenter = m_y + (m_h >> 1);


mwilck:
This pull request fixes issue #5831 (verified on my openSUSE installation).

nlyan:
Yeah, the removed code looks like legacy crap. Given the age of RandR 1.2 (late 2007?) I'm wondering if there's even a case for not requiring XRandR. Future versions will certainly require it...

I'm going to push this in. The bottom line is it's a partial revert, which seems low risk, and my testing also shows the shape adaption code still works.

nlyan:
@mwilck 

Are you able to try our build? http://symless.com/nightly?filter=fc3cc78

nlyan:
@mwilck 

By the way, SuSE is one of the distros we currently don't have in our new build infrastructure (which will be used for v1.9 and up). Would you recommend Leap to target the most desktop users?

sleep-walker:
@nlyan Good question. IMHO both Leap and Tumbleweed targets desktop users. Leap may attract more server users as maturational distro :)

mwilck:
@nlyan, the build you provide isn't installable on openSUSE.

```
Problem: nothing provides libcrypto.so.10()(64bit) needed by synergy-1.8.8-1.x86_64
```
openSUSE tumbleweed has `libcrypto.so.1.0.0` from openssl-1.0.2k. Apparently Fedora and openSUSE use different library versioning for openssl, I'm wondering why Fedora uses "10" for version 1.0. 

But I tested my own build which is identical with yours code-wise, so I've high confidence that your build would fix my problem, too.

Not knowing much about your build infrastructure, if you want to build synergy cleanly for openSUSE and/or SLES/SLED (SUSE's enterprise server / desktop products), I'd recommend to have a look at [openSUSE build service (OBS)](https://build.opensuse.org/). The [OBS devel project for synergy](https://build.opensuse.org/package/show/X11:Utilities/synergy) is where the latest code for inclusion in various official openSUSE repos is maintained. It's possible to setup semi-automatic builds from github in OBS (the current synergy package isn't set up that way, but it'd be easy to do).

nlyan:
@mwilck Many thanks for this. We really appreciate it. I've spun up a Leap 42.2 VM buildbot so we're going to make sure it's represented in v1.9

mwilck:
@nlyan, thanks for your kind words and for your excellent product.

PR #5840, opened 2017-02-12
Fix ClipboardChunkTests unit test on big endian archs.

epakai:
This only fixes the test. I have not verified the code behind the test is functioning as
designed. It was relayed I should submit a pull request for review.

This is related to: https://github.com/symless/synergy/issues/5838


nlyan:
The unit test is indeed bogus. It works by chance even on little endian archs (because the sequence number used is 0, chunk[1] is also always 0).

PR #5943, opened 2017-03-27
Compile time argument to skip building tests

### LINKED ISSUE #6725: Disable the building of tests with CMake option
### Problem ###

Currently the CMake will always build the tests some of which have headers/test code in non test files. 

### Solution ###

Add an option to turn off the test and #ifdef any code for tests that are not in a dedicated test code file

pieterjanc:
Resolves #6725 
Signed-off-by: Pieterjan Camerlynck <pieterjan.camerlynck@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,8 @@ endif()
 # First, declare project (important for prerequisite checks).
 project(synergy C CXX)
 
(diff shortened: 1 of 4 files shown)


Jnewbon:
This is good work, however its targeting V2 branch which is now stale. 

I have cherry picked your change and put it on the current branch heading to master, #6726

PR #5944, opened 2017-03-27
Remove Pro License Check for Encryption Setting

### LINKED ISSUE #5810: Encryption should not be a Pro feature
I'm opening this issue to start a discussion among the developers of this app regarding something I feel very strongly about: in this day and age, secure operation for any application should absolutely not be an extra/premium/pro feature.

Last year, cybersecurity came to the forefront of the minds of the public who began to feel more than ever its huge rippling effects in global politics, freedom, and personal safety. The push towards HTTPS for all websites continued, while end-to-end encrypted chat came into the mainstream with Whatsapp, Signal, and others. Apple introduced a new policy of no longer even accepting apps in its app stores that don't use TLS encryption to connect to external services. Users now have an expectation of privacy and security by default for anything they use (free, paid, or "pro"), and they are right to.

When you have tiered pricing plans where one of your "extra" features is encryption, effectively what you're saying is that users can pay $19 for your software, but it's $10 extra for that software not to become a giant gaping hole in their computer's security that lets anyone sitting next to them in a cafe log their keystrokes and steal their identity.

Not only is that dangerous for anyone using your "Basic" version, but I don't think it sends the message you want to your users.

deckar01:
Fixes #5810

### DIFF
--- a/src/gui/src/AppConfig.cpp
+++ b/src/gui/src/AppConfig.cpp
@@ -286,7 +286,7 @@ void AppConfig::setCryptoEnabled(bool e) {
 }
 
 bool AppConfig::getCryptoEnabled() const {
-	return (edition() == kPro) && m_CryptoEnabled;
+	return m_CryptoEnabled;
 }
 
 void AppConfig::setAutoHide(bool b) { m_AutoHide = b; }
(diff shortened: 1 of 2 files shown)


cardonator:
Thanks, maybe I should have submitted a PR :)

nlyan:
Plans with regard to SSL etc will become apparent in v1.9

deckar01:
@nlyan Can you add a priority label to #5810 since you are planning to resolve it in v1.9?

PR #5968, opened 2017-04-02
Do not put debug files in a separate dir on unix

pieterjanc:
Signed-off-by: Pieterjan Camerlynck <pieterjan.camerlynck@gmail.com>


XinyuHou:
@pieterjanc 
Thank you for contribution. Actually we have done the same changes in v1.9 branch which would highly likely to be the next release unless there is something urgent poping up to be fixed in 1.8.

I will close this PR. Appreciate your help again.

PR #5980, opened 2017-04-07
Add the packaging metadata to build the synergy snap

elopio:
This package will let you publish the latest synergy in the Ubuntu store, and from there reach many users on all the supported Ubuntu versions, and more Linux distributions in progress.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,42 @@
+name: synergy
+version: 'v1.9'
+summary: Share one mouse and keyboard between multiple computers
+description: |
+  Synergy combines your desktop devices together in to one cohesive experience.
+  It's software for sharing your mouse and keyboard between multiple computers
+  on your desk. It works on Windows, macOS and Linux.
+
+grade: devel # must be 'stable' to release into candidate/stable channels
+confinement: strict
+
+apps:
+  synergy:
+    command: desktop-launch synergy
+    plugs: [x11]
+    desktop: share/applications/synergy.desktop
+  client:
+    command: synergyc
+    plugs: [network, x11]
+    aliases: [synergyc]
+  server:
+    command: synergys
+    plugs: [network-bind, x11]
+    aliases: [synergys]
+
+parts:
+  synergy:
+    source: .
+    plugin: cmake
+    configflags: ['-DCMAKE_CXX_FLAGS=-std=c++11']
+    build-packages:
+      - g++
+      - libx11-dev
+      - libxtst-dev
+      - libxinerama-dev
+      - libxrandr-dev
+      - libsm-dev
+      - libavahi-compat-libdnssd-dev
+      - libcurl4-openssl-dev
+      - libssl-dev
+    install: mv bin/synergy $SNAPCRAFT_PART_INSTALL/bin/
+    after: [desktop-qt5]


elopio:
Hello!

As promised during our meeting yesterday, here's an initial snap package. I tried it in my two machines, but it of course needs more testing. More information about snapcraft here: http://snapcraft.io/

To test it in an Ubuntu 16.04 machine:

    $ sudo apt install git snapcraft
    $ git clone https://github.com/elopio/synergy
    $ cd synergy
    $ git checkout snapcraft
    $ snapcraft
    $ sudo snap install *.snap --dangerous

(dangerous because the snap you build doesn't come signed from the store. Once you push it to the store, your users will just have to run: $ sudo snap install synergy, or install it with a couple of clicks in the software center)

A couple of important details that you will notice due to the security model of the snaps:

 * There are no maintainers, with this snap you can push directly to the store whenever you want. But that means that we need to carefully think of a new model for name registrations and to solve name conflicts. Currently, you can just register one name per snap, but you have 3 binaries in there. So your commands will be `synergy`, `synergy.server` and `synergy.client`. You will see in the yaml that I defined aliases, to allow you to call `synergys` and `synergyc` as on the deb. That's an experimental feature, currently being discussed in our development forum: https://forum.snapcraft.io/t/improving-the-aliases-implementation/18 Any feedback you can give us there will be highly appreciated.

 * Snaps are very very confined by default. Your default config is in /home/$USER/.synergy.conf, but that file can't be accessed by this snap. You extend the permissions of your snap plugging it to  [interfaces](https://github.com/snapcore/snapd/wiki/Interfaces), like network and network-bind in this case. We have a temporary interface that lets snaps access files in /home/$USER, but that one still doesn't let them access hidden . files. So, I started my server using a path in the private space of the snap:

    $ synergy.server -c ~/snap/synergy/current/.synergy.conf

Personally, I would prefer `~/snap/synergy/current/.synergy.conf` to be the default config path for the synergy snap, because that way I don't have to give synergy access to all my files. But if you need a different solution, let us know and we will make sure to take it into account while designing the replacement for the temporary home interface.

 * I've never used the synergy UI before. You will have to give me a hand there. I checked that it opens correctly, but it throws an error when it tries to access zeroconf, because that's not something a snap can do by default. We have a tool to debug the apparmor denials when a snap tries to exit it's confinement, and it records:

```
= AppArmor =
Time: Apr  7 00:28:10
Log: apparmor="DENIED" operation="create" profile="snap.synergy.synergy" pid=701 comm=517420626561726572207468726561 family="netlink" sock_type="raw" protocol=0 requested_mask="create" denied_mask="create"
Suggestion:
* add one of 'account-control, network-control' to 'plugs'
```

You can add the account-control or network-control plugs, but they are not autoconnected because they can be dangerous. So we leave the user to manually connect them after the snap is installed and give them full control of his machine. The other plugs we are using here, network and network-bind are not dangerous, so they are autoconnected, but the user is still in control and he can manually disconnect them. We have a whole interesting topic here with cryptographic assertions that can be signed by the provider of the operating system image and the developer, to grant some autoconnection for dangerous things to applications that are trusted. Here again, your input would be awesome.

Ok, a very long text for such a simple yaml. I'll let you review it and try it. Please let me know if you have any questions or comments.

I will be on holidays next week, but I will make sure that one of my friends takes a look here once in a while so you don't have to wait for my return.

pura vida.

ps: I've been working a lot in the past months with a snap called [syncthing](https://forum.snapcraft.io/t/call-for-testing-syncthing/188/3). You have no idea how hard it is for me to not mix the two names :) I've just made a search and replace here, and when talking I can hear my brain really confused trying to choose the right name...

XinyuHou:
@elopio 
Thanks for all the detailed instructions, Leo.

XinyuHou:
@elopio 
I just tested out the snap build on an Ubuntu vm. It seems quite straight forward.

Like you mentioned, our default config file path in a hidden file. And the slot you have right now can access user directory but not reading hidden file, is this because of some security concern?

> So we leave the user to manually connect them after the snap is installed and give them full control of his machine. 

Could you give me a bit more information on this part? How do I connect them after I install the snap?

elopio:
Hello @XinyuHou! Sorry for the late reply, I was on holidays.

And you are right. We don't allow snaps access hidden files because of security. I don't want the synergy snap to access my firefox profile, for example. And I don't want any snap to access my synergy configuration.

Here is more information about interfaces, and how to connect and disconnect them: https://snapcraft.io/docs/core/interfaces

XinyuHou:
@elopio 
Thanks for the info. I will look into that.

PR #6002, opened 2017-04-20
Server omits client connections in ｄestructors

SvyatoslavRazmyslov:
We have found and fixed a bug using PVS-Studio tool. [PVS-Studio](https://www.viva64.com/en/pvs-studio/) is a static code analyzer for C, C++ and C#.

We suggests having a look at the emails, sent from @pvs-studio.com.

Analyzer warning: [V625](https://www.viva64.com/en/w/V625/) Consider inspecting the 'for' operator. Initial and final values of the iterator are the same. Server.cpp 253


nlyan:
Thanks a lot!

PR #6048, opened 2017-05-20
Fix icon path on Linux

ccl2of4:
The current icon path on Linux is preventing themes from providing their own icons for synergy.

.desktop files should generally avoid providing absolute icon paths when possible and instead simply adhere to [these conventions](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#directory_layout).

Let me know if you need any more information, if you'd like me to change target branch, etc.

Thanks

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -406,7 +406,7 @@ endif()
 #
 if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
     configure_files (${CMAKE_CURRENT_SOURCE_DIR}/dist/rpm ${CMAKE_BINARY_DIR}/rpm)
-    install(FILES res/synergy.svg DESTINATION share/icons)
+    install(FILES res/synergy.svg DESTINATION share/icons/hicolor/scalable/apps)
     if("${VERSION_MAJOR}" STREQUAL "2") 
         install(FILES res/synergy2.desktop DESTINATION share/applications)
     else()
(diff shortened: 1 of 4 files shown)


ccl2of4:
Actually I have realized now that the icon provided by synergy is in .ico format. I believe due to this, it will not be found by the icon lookup conventions I have mentioned, and so the absolute icon path is necessary for the time being.

Unless we can get an SVG, PNG, or XPM icon for Linux this pull request should not be merged.

yupi2:
The [v1.9 branch](https://github.com/symless/synergy/tree/v1.9/res) has an SVG but still uses an absolute path for the icon.

ccl2of4:
I have updated base/target branch to `v1.9`.

Also I'm now thinking the current installation directory for the icon itself is not correct. It looks like it's being placed in `/usr/share/icons`, but I believe the convention is to put third party icons into the directory for the "hicolor" theme.

nlyan:
Does the installation path for our svg icon need changing here ->  https://github.com/symless/synergy/blob/v1.9/CMakeLists.txt#L413

ccl2of4:
@nlyan I think the file on that branch has been committed to since you made your comment (the line number you linked doesn't make sense), but if you mean to say this line needs changing then I would agree:

````
install(FILES res/synergy.svg DESTINATION share/icons)
````

I can update this pull request with a fix for that as well

nlyan:
@ccl2of4 Please do, and I'll merge asap

ccl2of4:
Updated.

I also pushed a commit to update the icon path in `dist/rpm/synergy.spec.in`. Let me know if this is not something you want.

nlyan:
@ccl2of4 this didn't seem to work on CentOS 7. The desktop environment cannot find our icon

Issue #3083, opened 2014-10-13
Modifier Keys stop working on a VM on Mac client


nbolton:
**Imported issue:**
- Author: Ed Farnbauch
- Date: 2012-04-09 15:03:12
- Legacy ID: 3153
- Version: 1.4.7

**Steps to reproduce:**  
1. setup vm in vmware fusion  
2. start synergy  
3. Attempt to use modifier keys on any window in VM  

Expected: Modifier keys work (ie. shift produces capital letters)  
Actual: Produces just the unmodified key (T is t, or <ctrl-T> is t)  

**Versions (Synergy, OS):**  

Server:  (Linux (has tested on Fedora, Mint, and RHEL 6)  
Client 1: 1.47 (Mac OS X 10.7 (all versions))  
(though this has been an issue for several versions)  

**Workaround:**  

I can use caps lock to get a capital letter. but not for any of the other modifier keys

the-wes:
A user sent in this information:

The Command key does get detected properly. Pressing Shift along with a character does not result in seeing a capital letter. Also, pressing Ctrl-Option does not generate the same keystroke through Synergy that it does with the physical keyboard. The Ctrl-Option combination is used in Parallels to release focus on the VM so the keyboard and mouse control revert back to the OS.

I ran showkeys in the terminal on Fedora in the VM and found the following:

Physical keyboard:

Press and release shift: 
keycode 42 press 
keycode 42 release 
Press and release control: 
keycode 29 press 
keycode 29 release 
Press and release option: 
keycode 56 press 
keycode 56 release 
Press shift, then Z, release Z, release shift: 
keycode 42 press 
keycode 44 press 
keycode 44 release 
keycode 42 release

The last combination also displays an uppercase Z on the terminal window.

lordarcy:
I was the user that sent in the details the-wed mentioned on May 16. More details of the environment:

Server = Mac OS 10.10.5 Yosemite
Client = Mac OS 10.11.4 El Capitan
Using Parallels Desktop 11.2.0.32581
VM = Fedora 23

kryojenik:
Server = Mac OS 10.9.5
Client = Mac OS 10.11.4 El Capitan
Using Microsoft Remote Desktop 8.0.31

Modifier keys in Microsoft Remote Desktop are ignored

TryTryAgain:
Confirmed by me for a while as well, very painful.
Software: Microsoft Remote Desktop 8.0.32 (Build 27142)
[...]

PR #6064, opened 2017-06-10
#3083 Detect currently active window and send modifier keys differently if the active window is the Parallels one

sdcarpenter:
I experimented with a bunch of different ways to send modifier keys to my Windows 10 Parallels VM and the only way I could make it work was by directly sending the shift key code. Unfortunately, this breaks Synergy when interacting with OSX. So this patch detects the active window, and if it is Parallels, send the modifier key directly, otherwise, continue to use the modifier flags. This patch has a couple of limitations:

1) Modifier keys don't work properly in Parallels config windows (not a deal breaker for my use case, as I rarely need a modifier key when configuring a VM, or I can switch to the real keyboard if I do).
2) Only solves this bug for Parallels (although it should be easy to extend this method for other pieces of software that are doing special key handling)
3) Caps lock doesn't work in Parallels (although I don't believe it was working before, but haven't confirmed)

I've been running this patch for a week now in my current dev setup (Windows 10 Synergy server, macOS Sierra client, Windows 10 VM) and it has been very stable for me. If you send me a pointer to how to run the tests, I'd be happy to make sure this patch doesn't break them, and perhaps can come up with a way to test this particular functionality should you decide you want to accept this PR.


eugene-yao-zocdoc:
@sdcarpenter does your fix work if the parallels is running in Coherence mode? I really would love this fix! Thank you!

sdcarpenter:
@eugene-yao-zocdoc I just tested, and it does appear to work in Coherence mode as well.

bricerebsamen:
I really wish this can be merged soon and fix the issue. This issue is extremely annoying and apparently is frustrating a large number of users. I am new to the mac/parallels setup but this is totally killing my workflow on my new mac.

bgehrs:
Is there any chance these fixes will allow the shift key to work in the Microsoft Remote Desktop client running on Mac?  I am using the Synergy client on MacOS Mohave 10.14.2.  When running the Microsoft Remote Desktop client 10.2.4 (downloaded from the Mac App Store) on Mohave the shift key does not work when connecting into a Windows Server.  Thanks!

bgehrs:
Is there any chance these fixes will allow the shift key to work in the Microsoft Remote Desktop client running on Mac?  I am using the Synergy client on MacOS Mohave 10.14.2.  When running the Microsoft Remote Desktop client 10.2.4 (downloaded from the Mac App Store) on Mohave the shift key does not work when connecting into a Windows Server.  Thanks!

Jnewbon:

mariajeli (approved):

SerhiiGadzhilov:
This PR can’t be merged because this functionality was rewritten.
Also this problem should be fixed in the latest version.
There were the following PRs for this functionality:
1. #7089 
2. #7094 
3. #7135

PR #6077, opened 2017-06-16
Stop retry timer properly which unregisters nothing

smejkar:
ServerApp::stopRetryTimer does not remove the handler associated with the retry timer.


nlyan:
Can you tell me a bit more about the impact of this patch and why you sought to address it?

smejkar:
While investigating another timer related issue I have encountered that ServerApp::stopRetryTimer does not unregister the event handler for the retry timer. Note stopRetryTimer attempts to unregister an event handler for target NULL by calling m_events->removeHandler(Event::kTimer, NULL). This actually unregisters nothing.

This should be addressed as the event queue refers to the event handler for deleted retry timer.

Issue #6031, opened 2017-05-07
Dump config file to log


nlyan:
Problem: Asking users for details of their configuration alongside their log is time consuming. 

Solution: At DEBUG and above dump the config file in to the log when Synergy starts.

paul-suarez:
JDDoesIT from forums:

When we report issues we are often asked for the system information as well as the installation details. Could we get a dump button in the help menu that puts all the info one could possibly need into our clipboard so we don't have to type it? I'm just being lazy, but that is not always a bad thing...

[Forums 2966 link.](https://symless.com/forums/topic/2966-system-dump/)

PR #6096, opened 2017-07-07
issue6031

cassinodari:
Add to file  /src/gui/src/ServerConfig.cpp  the libraries:
#include  <stdio.h>
#include <stdlib.h>

And the command section below:
#if defined(__linux)
	system("free - g > /tmp/synergyCPU.txt");
	system("lscpu >> /tmp/synergyCPU.txt");
#elif defined(_WIN32)
	system("systeminfo >> c:/Program files");
#endif

This will generate the log with information the cpu.  In the platform Linux, the file are on /tmp/ and on the platform Windows are on the C:\ProgramFiles


XinyuHou:
@cassinodari 
Thank you for the PR.

By looking at #6031, it is for outputting Synergy configuration info into Synergy logging system.

So it's nothing to do with local CPU info.

PR #6128, opened 2017-08-19
NMake generator, remove GUI, VS2017 compile error

nyetwurk:
The cmake nmake generator makes makefiles that puts all files in bin/ not bin/(Config)

This way we can do:

 cmake -G "NMake Makefiles" -DCMAKE_BUILD_TYPE=Release
 nmake
 cd installer
 msbuild /p:Configuration=Release

And nmake will get the right CMAKE_RUNTIME_OUTPUT_DIRECTORY.

msbuild does not allow an empty /p:Configuration


yupi2:
You can check out my [fork](https://github.com/yupi2/synergy) which works with VS2017 if you want.

nyetwurk:
Uh, where did Include.wxi from dist/wix/Include.wxi.in go?

Jnewbon:
@nyetwurk Is this PR still valid for the master branch?

I use VS2019 currently without problems.

nyetwurk:
Yes it affects nmake iirc but I will recheck time permitting.

PR #6129, opened 2017-08-19
Latest Qt installer puts things in different places.

nyetwurk:


XinyuHou:
Which version of Qt do you install?

nyetwurk:
5.9.1

nlyan:
No longer relevant to master as the v1 GUI was removed

Issue #6135, opened 2017-08-27
SSL private key is world readable


bitsandsalsa:
### Operating Systems ###

Server: Ubuntu 16.04.3
Client: Ubuntu 16.04.3

### Synergy Version ###

1.8.8

### Steps to reproduce bug ###

1. `ls -l ~/.synergy/SSL/Synergy.pem`
2. Notice the following file permissions:

> -rw-rw-r-- 1 user user 1111 Jan 1  2017 Synergy.pem

### Other info ###

To reduce private key exposure, the key should only be readable by the user; not the group or other. I would expect the following file permissions:

> -rw------- 1 user user 1111 Jan 1  2017 Synergy.pem

This is what I have for my SSH private keys created with ssh-keygen:

> -rw------- 1 user user 1111 Jan 1  2017 id_rsa

LeadManPL:
Looks like synergy-v1.8.8-stable-25a8cb2-Linux-x86_64.deb  still have this issue:

patben@patbenamd:~/.synergy/SSL$ ll
razem 20
drwxr-xr-x 3 patben patben 4096 paź  5 11:50 ./
drwxr-xr-x 4 patben patben 4096 sie 10 13:30 ../
drwxr-xr-x 2 patben patben 4096 lip 27  2016 Fingerprints/
-rw-r--r-- 1 patben patben 1653 paź  5 11:50 Synergy.pem
-rw-r--r-- 1 patben patben 1653 paź  5 11:49 Synergy.pem.bkp

PR #6142, opened 2017-09-07
Set appropriate umask when writing SSL certificate

epakai:
Fix for #6135 by setting umask before calling openssl.
previous umask is restored after call.

This doesn't account for files already created (synergy currently does nothing if the file already exists.)

I also noticed in testing a newer version of openssl on Debian (1.1.0f) that it was already setting permissions appropriately. I can't find a relevant changelog entry from Debian or OpenSSL so I'm not sure where the change was introduced.

### DIFF
--- a/src/gui/src/SslCertificate.cpp
+++ b/src/gui/src/SslCertificate.cpp
@@ -19,6 +19,11 @@
 
 #include "Fingerprint.h"
 
+#if defined(Q_OS_LINUX)
+#include <sys/types.h>
+#include <sys/stat.h>
+#endif
+
 #include <QProcess>
 #include <QDir>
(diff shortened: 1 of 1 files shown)


epakai:
Setting a proper umask might also apply to Mac OSX systems (my patch only does linux). I'm not familiar with how the files are stored there, and how the headers and umask call might differ.

epakai:
I chased down why the newer version of openssl did permissions right as well:
https://github.com/openssl/openssl/commit/3b061a00e39d2e4ad524ff01cbdc0c53fe8171ee

So I guess you could count on that behavior with openssl (>=1.1.0). I just wanted to confirm it wasn't something Debian specific.

PR #6174, opened 2017-10-14
Reallocate socket buffer when size is not big enough

TTimo:
The most significant memory leak I've noticed so far. synergys on linux with --enable-crypto causing ~1GB/mn of leaked memory before this.


nlyan:
TLS support has been removed from core and reimplemented in v2. Closing

Jnewbon:
Reopening as this is still a problem and the proposed fix preliminary testing works 

#6460

Jnewbon:
Recreated this in #6488 due to rebasing problems

Issue #4325, opened 2015-01-15
IPv6 support


davidgro:
The place I work has a lot of computers. That's an understatement. The internal network is probably larger than many countries' networks. Anyway despite my work (Synergy server) PC being on what appears to be a class B IPv4 block in the internal network (one of many blocks and types of blocks) any time it goes to sleep or hibernates there is a good chance it will have its IPv4  address stolen by something else that needs one and get itself assigned a new one. My synergy client PC then needs to wait a long while for the DNS to update to pick up my server's new IP. 

In the mean time, everything has an IPv6 and it would be really nice to just give that to the client (or even just have it find it from DNS) since I am pretty sure that one never needs to change.  When I tried it apparently treated it as a hostname and failed to connect.

nbolton:
Duplicates #4154

nbolton:
See PR #4499

tjpatter:
Any updates on IPv6 support?

wltjr:
Synergy is the only program I still have to use IPv4 for....
```
ssh -f -N -L 127.0.0.1:24800:host:24800 host
```
vs
```
ssh -f -N -L [::1]:24800:host:24800 host
```

Really need to add IPv6 support sooner than later. Some of us are looking to stop using IPv4 and synergy is causing it to remain. Thanks!

brianjmurrell:
> Duplicates #4154

That is actually a PR, not an issue.  If that PR actually work{s,ed} and got landed, it would "Fixes #4325", not really duplicate it.

But I guess the latest PR for this is now #4499, but sadly it has not landed either.  It's now listing a ton of conflicts.  I wonder how many of those were present 3 years ago when the PR was pushed.

@nbolton Given that there are patches for this and they just need porting, and given that IPv6 is *now* and not *tomorrow* for [some people](https://en.wikipedia.org/wiki/IPv4_address_exhaustion#Exhaustion_dates_and_impact) is there not even a junior developer who could bang this out in a day or two?

wltjr:
I am not sure about those conflicts. I am using the patches in that PR #4499 , they apply and work fine with synergy-1.8.8. I will comment on the PR.

brianjmurrell:
@wltjr Or push a new PR and let's get this landed already.  :smile:

wltjr:
[...]

PR #4154, opened 2014-11-01
IPv6 support for Windows and Linux

WilliamKyle:
Add ipv6 support, tested in Windows 7 and debian 7. I don't have a MAC OS, it is not tested.

PS: 
synergys run as ipv4 mode as default, use -a [::] to turn ipv6 mode.
I don't know how to modify synergy(GUI), so it is now only working start from terminal with args.

PR #4499, opened 2015-04-02
IPv6 support for Windows and Linux

WilliamKyle:
Sorry for my bad operation here:
https://github.com/synergy/synergy/pull/4154

Tested in Windows 7 and debian 7. I don't have a MAC OS, so it is not tested.

PS: 
synergys run as ipv4 mode as default, use -a [::] to turn ipv6 mode.
I don't know how to modify synergy(GUI), so it is now only working start from terminal with args.

PR #6178, opened 2017-10-20
IPv6 support for Windows and Linux

wltjr:
NOT MY WORK! Updated patches from the following:

From adabfe11d45e9e8fd5bfc445dc02ea80497c308f Mon Sep 17 00:00:00 2001
From: WilliamKyle <williamkyle@126.com>
Date: Thu, 2 Apr 2015 23:45:56 +0800
Subject: [PATCH 1/2] IPv6 support for Windows and Linux

From a4c7ec47ad2447e8bcbb2fa99aac370c0b50d429 Mon Sep 17 00:00:00 2001
From: Fabian Krueger <krueger.fabian@gmail.com>
Date: Sat, 29 Oct 2016 20:21:54 +0200
Subject: [PATCH 2/2] Fix build issue with IpcServer after applying ipv6 patch

PR #4499
https://github.com/symless/synergy-core/pull/4499

Fixes Issue #4325
https://github.com/symless/synergy-core/issues/4325

### DIFF
--- a/src/lib/arch/win32/ArchNetworkWinsock.cpp
+++ b/src/lib/arch/win32/ArchNetworkWinsock.cpp
@@ -26,7 +26,8 @@
 
 static const int s_family[] = {
     PF_UNSPEC,
-    PF_INET
+    PF_INET,
+    PF_INET6,
 };
 static const int s_type[] = {
     SOCK_DGRAM,
@@ -156,7 +157,7 @@ ArchNetworkWinsock::initModule(HMODULE module)
     setfunc(startup, WSAStartup, int(PASCAL FAR*)(WORD, LPWSADATA));
 
     // startup network library
-    WORD version = MAKEWORD(2 /*major*/, 0 /*minor*/);
+    WORD version = MAKEWORD(2 /*major*/, 2 /*minor*/);
     WSADATA data;
     int err = startup(version, &data);
     if (data.wVersion != version) {
@@ -212,6 +213,11 @@ ArchNetworkWinsock::newSocket(EAddressFamily family, ESocketType type)
     }
     try {
         setBlockingOnSocket(fd, false);
+        BOOL flag = 0;
+        int size     = sizeof(flag);
+        if (setsockopt_winsock(fd, IPPROTO_IPV6, IPV6_V6ONLY, &flag, size) == SOCKET_ERROR) {
+            throwError(getsockerror_winsock());
+        }
     }
     catch (...) {
         close_winsock(fd);
@@ -294,7 +300,7 @@ ArchNetworkWinsock::bindSocket(ArchSocket s, ArchNetAddress addr)
     assert(s    != NULL);
     assert(addr != NULL);
 
-    if (bind_winsock(s->m_socket, &addr->m_addr, addr->m_len) == SOCKET_ERROR) {
+    if (bind_winsock(s->m_socket, TYPED_ADDR(struct sockaddr, addr), addr->m_len) == SOCKET_ERROR) {
         throwError(getsockerror_winsock());
     }
 }
@@ -317,10 +323,10 @@ ArchNetworkWinsock::acceptSocket(ArchSocket s, ArchNetAddress* const addr)
 
     // create new socket and temporary address
     ArchSocketImpl* socket = new ArchSocketImpl;
-    ArchNetAddress tmp = ArchNetAddressImpl::alloc(sizeof(struct sockaddr));
+    ArchNetAddress tmp = ArchNetAddressImpl::alloc(sizeof(struct sockaddr_in6));
 
     // accept on socket
-    SOCKET fd = accept_winsock(s->m_socket, &tmp->m_addr, &tmp->m_len);
+    SOCKET fd = accept_winsock(s->m_socket, TYPED_ADDR(struct sockaddr, tmp), &tmp->m_len);
     if (fd == INVALID_SOCKET) {
         int err = getsockerror_winsock();
         delete socket;
@@ -368,7 +374,7 @@ ArchNetworkWinsock::connectSocket(ArchSocket s, ArchNetAddress addr)
     assert(s    != NULL);
     assert(addr != NULL);
 
-    if (connect_winsock(s->m_socket, &addr->m_addr,
+    if (connect_winsock(s->m_socket, TYPED_ADDR(struct sockaddr, addr),
                             addr->m_len) == SOCKET_ERROR) {
         if (getsockerror_winsock() == WSAEISCONN) {
             return true;
@@ -685,6 +691,15 @@ ArchNetworkWinsock::newAnyAddr(EAddressFamily family)
         break;
     }
 
+    case kINET6: {
+        addr = ArchNetAddressImpl::alloc(sizeof(struct sockaddr_in6));
+        struct sockaddr_in6* ipAddr = TYPED_ADDR(struct sockaddr_in6, addr);
+        ipAddr->sin6_family         = AF_INET6;
+        ipAddr->sin6_port           = 0;
+        memcpy(&ipAddr->sin6_addr, &in6addr_any, sizeof(in6addr_any));
+        break;
+    }
+
     default:
         assert(0 && "invalid family");
     }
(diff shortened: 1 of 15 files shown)


src/lib/net/TCPSocketFactory.h
```diff
@@ -31,9 +32,9 @@ class TCPSocketFactory : public ISocketFactory {
 
     // ISocketFactory overrides
     virtual IDataSocket*
-                        create(bool secure) const;
+                        create(bool secure), IArchNetwork::EAddressFamily family = IArchNetwork::kINET) const;
```
yupi2: need to remove a `)` after `secure` here

src/lib/net/TCPSocketFactory.h
```diff
@@ -31,9 +32,9 @@ class TCPSocketFactory : public ISocketFactory {
...
     virtual IListenSocket*
-                        createListen(bool secure) const;
+                        createListen(bool secure), IArchNetwork::EAddressFamily family = IArchNetwork::kINET) const;
```
yupi2: need to remove a `)` after `secure` here

wltjr:
@yupi2 thanks! Total fail there... Corrected and force pushed. Please let me know if I missed anything else. I clearly did not try the patches myself :P

src/lib/arch/win32/ArchNetworkWinsock.cpp
```diff
@@ -212,6 +213,11 @@ ArchNetworkWinsock::newSocket(EAddressFamily family, ESocketType type)
     }
     try {
         setBlockingOnSocket(fd, false);
+        OOL flag = 0;
```
yupi2: `OOL` to `BOOL`

wltjr: Thanks again, corrected and re-force pushed :)

wltjr:
@XinyuHou  @nbolton What is going on guys? This is the 2nd PR now for IPv6 being ignored. I updated the files, it was ready to merge. Now there has been another release and the patches must be updated again...

Is IPv6 support something synergy cares to support like ever? Seriously there is no reason to keep avoiding these PRs for years. Almost 3 years now from the date of the initial patches in April 2015. I have been using this with IPv6 for over a year now patched. I really see no reason to keep omitting this, short of it not being wanted. Which seems to be the case.

Otherwise can we get a reason why it is not being merged? I can update the files again. Which will make the 3rd PR or set of patches for IPv6 support in synergy. IPv6 cannot be avoided for ever. Many have been using it for years.

Thanks!

mrsixw:
@wltjr Thanks for the PR. We are not deliberately ignoring this by any stretch, but we are a small team and our priorities do vary over time. Part of my remit over the coming month will be to try and establish a more regular cadence for reviewing synergy-core PRs.

wltjr:
@yupi2 if you could review as you did before. Pretty sure I did not miss anything but could use another set of eyes. You caught a few before. Thanks!

@mrsixw I rebased and fixed the modifications so its once again ready to merge. Please review and merge ASAP. Thank you!

yupi2:
@wltjr It seems like a lot of recheck / compare so I'm not going to this time.

wltjr:
@yupi2 no worries just seemed like you were the only one to review last time. It should be good to go, as I rebased, and only a few files had conflicts. That was mostly where they dropped secure/tls stuff. Much more minor than before. I remade [another that applies against 2.0.0](https://github.com/Obsidian-StudiosInc/synergy-core/commit/e57e0b40953153b6100321f90085c31ff5c22a8c.patch), less windows stuff, so I can patch that version. Which I am running 2.0.0 with that patch right now, no problems. The patches here are against HEAD and do not apply to 2.0.0. The original set works with 1.8.8, but not to many commits beyond that tag.

That's 3 sets of IPv6 patches.... But the one here should hopefully get merged ASAP. I hope I do not have to do this for a 3rd/4th time...

nlyan:
@mrsixw IPv6 support in core is not really relevant for us now, since all networking in v2.0 is outside of core. We're only using TCP between core and v2 for legacy reasons.

wltjr:
@nlyan people want to use Synergy over IPv6. Where the code goes, be it in core or other is moot. Synergy needs to have support for IPv6, some years ago. Do you all think IPv6 is not a thing? Not in use? IPv4 is here forever?

IPv6 should be relevant somewhere. I care less if its in core or other. Just needs to exist. I am sure others feel the same. PR and bugs have only been open since like 2015...
https://github.com/symless/synergy-core/issues/4325
https://github.com/symless/synergy-core/pull/4499

Really not sure what the hang ups or issue really is. Seems like a stream of excuses or something. Merge the code, move it else where later if your changing the Networking aspect. Though given what synergy is, I am not clear how you are moving networking out of core. Given networking is a core feature of synergy. How else are two computers supposed to exchange information on keyboard and mouse if not via network?

Not to be rude, but it really sounds like you haven't a clue what your talking about. If the network code was moved, the patch would not apply. I still see networking files here in core
https://github.com/symless/synergy-core/tree/master/src/lib/net

Same for 2.0 branch
https://github.com/symless/synergy-core/tree/v2.0.0-stable/src/lib/net

Anyway its moot. I have the patches I need and can continue to modify things for my needs. Clearly synergy developers care less about IPv6. Just be honest. You all either do not want to add IPv6. Lack the knowledge or ability, or something. Thus far just seems like excuses with no merit or factual basis.

If I wasn't updating the patches/PRs and making noise. This would be a non issues going no where even less than it is now. So lets be clear. Synergy does not care about IPv6 support. Thus ignoring the initial PR for years, my 2nd and now 3rd pr, and now just excuse after excuse. For code that can merge per CI, etc.

nlyan:
We're not simply ignoring the issue, but there has been very low demand to support IPv6 from customers, and we just haven't reviewed this code for that reason.

Synergy v2 doesn't use the networking code in synergy-core to connect over the LAN, so supporting IPv6 in core has no upside for us. Merging this to *core* at this point would just risk further instability.

Adding IPv6 support to the proprietary components of v2 will be fairly trivial, we just have to take the time to make a few tweaks and turn it on. Again, we've just had much higher priorities since v2.0.0

I might suggest you try and get this merged in to one of the popular open source v1 forks such as https://github.com/debauchee/barrier/ or https://github.com/yupi2/synergy as I have no desire to see your hard work go to waste.

wltjr:
@nlyan thanks for the explanation! Not so much my work, it was another's I updated twice. Not seeing much demand is the best explanation. I would assume that maybe because there are less people using multiple systems these days in our new mobile world... Who sits and uses a keyboard and mouse? Old fogies...

If you have proprietary components, you all may want to code the IPv6 support yourself to retain copyright. Unless you have some contributor sign off as to such. Its moot to me, but since the original work is not mine. I cannot really sign off any anything along those lines. Maybe a none issue, but came to mind when you said proprietary. FOSS contributions can get tricky there in such situations.

Thanks for the suggestion on forks! I was not aware of them. I am not so concerned about the work being unappreciated as just moving the ball forward. I like to get stuff done ASAP, so if I can help to expedite etc. That was my interest, greasing the wheels. I do not have much personally invested. Though do not want to endless redo PRs either.

Thanks again for further clarification!

nlyan:
Just to clarify, our proprietary codebase is totally separate and doesn't contain any "Synergy core" code.

wltjr:
@nlyan will this project continue to have a FOSS version or just move to proprietary? Based on information thus far, seems like aspects of synergy-core will not be furthered and maybe removed, like the network code. Sorry just trying to clarify, I think I am confused now. It sounded like synergy-core was used with the proprietary one. Just different code bases, one building/running on the other. Thus no point in adding IPv6 to that, if it can be added else where.

Is that correct? Is the proprietary one still using synergy-core? Not sharing code, using synergy-core as a layer per se.

I did not realized there was a proprietary version. I never looked into synergy. I just used the package as provided by distro.

nlyan:
The proprietary (as of v2.0) product known as Synergy, produced by us at Symless, uses "Synergy Core" (everything in this repository) to support its core feature set. A synergy-core(.exe) binary is shipped with v2.0, and this constitutes an unaltered build from master here.

I can't speak for the business wrt to whether future features will be open sourced or not, but so far only features that have been rewritten/replaced in our new codebase have been removed from master here to signal that we don't intend on supporting them any longer. Any additions to core will be added to the this repo as required to comply with the GPL.

wltjr:
@nylan ok that is what I was thinking. Based on Symless needs, if Symless needs IPv6.It would go into the Symless proprietary Synergy products. Thus not adding this code to the underlying synergy-core. That makes sense in that regard. If you all are implementing other features on top of the core.

I understand not being able to say about future features going into synergy-core vs Symless Synergy products. Also makes sense as why some code would be removed from core, and other code like IPv6 support not added to core. I was more asking the other way, which has been answered indirectly. If IPv6 was added, where would it go? IPv6 support would not be added to core but go into Symless Synergy products. I assume further networking code maybe removed from core if not used in Symless Synergy products.

I guess there is not really any road forward for this PR then if that is the case. May need to comment on the bug as to that effect. Which would make this type of PR better suited to a fork as you mentioned.

nlyan:
IPv6 support would definitely go in to our proprietary code because that's where the new network layer lives.

wltjr:
Yeah so I guess we can close this PR then, as it will be invalid. Unless you all implement it in way that leaves the existing networking stuff. But then you would have code in core you are not using, and not sure about maintaining, etc. Likely should comment on general IPv6 bug that it will be going into Synergy products, not the open source core, as the Network layer was moved. Which I assume the rest of the networking in core that remains will be removed someday.

nlyan:
Some kind of networking in core will always be required, whether that be TCP sockets or an IPC, but at the moment that is fine on IPv4 because it only connects on loopback. 

Leaving PR for a bit longer to give forks a chance

wltjr:
Ok no worries, at least I understand the situation. I am fine with the PR remaining open for forks, or if till decided otherwise for synergy-core. Thank you very much for your time in clarification!

nbolton:
We could merge this in, though it has no advantage to Symless or it's customers (for the reasons that @nlyan highlighted), it may actually have some sort of benefit to the open source community at large (i.e. people that use only synergy-core via the command line that they build themselves).

@mrsixw How much impact (on time and risk) would it be on our dev resources to review this PR and potentially merge it in to benefit non-customers?

mrsixw:
@nbolton We'll take some time in our next planning cycle to review this and see where we are. Once reviewed, we'll see about merging.

nbolton:
Fixes #4154

srl295:
This PR breaks mDNS (`.local`) at least on mac: server listens on IPv4, but client attempts to connect on IPv6 : #6964

PR #6182, opened 2017-10-28
Fix building against libressl

jwestfall69:
a73b65431b755aec1f007d11feccd017cc04af6c added a compile time condition
to use SSL_get_client_ciphers() on newer versions of ssl.  libressl
doesn't have this function and sets its OPENSSL_VERSION_NUMBER to
0x20000000L, which is causing compilation to fail.


nlyan:
TLS support removed from core. Closing as not relevant

PR #6197, opened 2017-11-20
fix non-git build

sassmann:
When trying to build from the 2.0.0 tarball the following error occurs.

CMake Warning at cmake/Version.cmake:27 (message):
  Synergy version wasn't set.  Set to 2.0.0
Call Stack (most recent call first):
  CMakeLists.txt:35 (include)

fatal: Not a git repository (or any parent up to mount point /home)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not
set).
CMake Error at cmake/Version.cmake:53 (string):
  string sub-command LENGTH requires two arguments.
Call Stack (most recent call first):
  CMakeLists.txt:35 (include)

CMake Error at cmake/Version.cmake:55 (message):
  SYNERGY_REVISION ('') should be a short commit hash
Call Stack (most recent call first):
  CMakeLists.txt:35 (include)

-- Configuring incomplete, errors occurred!

Signed-off-by: Stefan Assmann <sassmann@kpanic.de>


aaannz:
I don't think the patch is correct. By moving the code chunk forward SYNERGY_REVISION is always set, rendering later `if (NOT DEFINED SYNERGY_REVISION)` check useless.
I fixed the build issue by using environmental variable SYNERGY_VERSION_STAGE=release. No change in cmake files needed.

nlyan:
@aaannz is indeed correct. Closing.

Issue #6037, opened 2017-05-11
PVS Studio reported defects


nlyan:
### Operating Systems ###

Windows

### Synergy Version ###

1.9.0-rc3

### Steps to reproduce bug ###

Run [PVS Studio](https://www.viva64.com/en/pvs-studio/) over Synergy to find many high priority issues in the codebase

### Other info ###

PVS Studio 6.15.2140
Visual Studio 2015 Update 3

As of writing there are **38** high priority issues in the 'General' and '64-bit' categories, and up to 51 if the Optimization category is included.

PR #6214, opened 2017-12-07
PVS-Studio small fix

pavel-pimenov:
V801 Decreased performance. It is better to redefine the second function argument as a reference. Consider replacing 'const .. b' with 'const .. &b'. keystate.h 164
V813 Decreased performance. The 'filename' argument should probably be rendered as a constant reference. string.cpp 177
V813 Decreased performance. The 'buffer' argument should probably be rendered as a constant reference. mswindowswatchdog.cpp 571
V813 Decreased performance. The 'filename' argument should probably be rendered as a constant reference. string.cpp 177
V813 Decreased performance. The 'buffer' argument should probably be rendered as a constant reference. mswindowswatchdog.cpp 571
V815 Decreased performance. Consider replacing the expression 'm_data[index] = ""' with 'm_data[index].clear()'. clipboard.cpp 46
V815 Decreased performance. Consider replacing the expression 'x = ""' with 'x.clear()'. keymap.cpp 1248
V812 Decreased performance. Ineffective use of the 'count' function. It can possibly be replaced by the call to the 'find' function. keystate.cpp 690
V812 Decreased performance. Ineffective use of the 'count' function. It can possibly be replaced by the call to the 'find' function. keymap.cpp 1002
V808 'srcArgString' object of 'basic_string' type was created but was not utilized. config.cpp 963
V808 'dstArgString' object of 'basic_string' type was created but was not utilized. config.cpp 963
V808 'neighbor' object of 'basic_string' type was created but was not utilized. config.cpp 1794
V614 Potentially uninitialized variable 'server' used. main.cpp 48
V614 Potentially uninitialized variable 'client' used. main.cpp 52

### DIFF
--- a/src/cmd/core/main.cpp
+++ b/src/cmd/core/main.cpp
@@ -28,7 +28,7 @@ int
 main(int argc, char** argv) 
 {
     // TODO(andrew): use existing arg parse code
-    bool server, client;
+    bool server = false, client = false;
     if (argc > 1) {
         server = std::string(argv[1]) == "--server";
         client = std::string(argv[1]) == "--client";
--- a/src/lib/base/String.cpp
+++ b/src/lib/base/String.cpp
@@ -174,7 +174,7 @@ findReplaceAll(
 }
 
 String
-removeFileExt(String filename)
+removeFileExt(const String& filename)
 {
     size_t dot = filename.find_last_of('.');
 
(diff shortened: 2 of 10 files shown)


nbolton:
Potentially fixed in #6037 (please let me know if not).

PR #6245, opened 2018-01-23
synergy.conf.example: clarify comments

waldyrious:
instead of saying apparently absolute statements
like "Foo is to the right of Bar",
use relative phrasing "for Bar, Foo is to the right".
This makes it clearer that the configuration file
does not describe a globally consistent spatial arrangement
but rather a set of areas that can be linked in arbitrary ways.

### DIFF
--- a/doc/synergy.conf.example
+++ b/doc/synergy.conf.example
@@ -4,28 +4,29 @@
 # line.  comments may appear anywhere the syntax permits.
 
 section: screens
-	# three hosts named:  moe, larry, and curly
+	# three hosts, named "moe", "larry", and "curly"
 	moe:
 	larry:
 	curly:
 end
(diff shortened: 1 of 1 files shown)


waldyrious:
I'm also a little confused at the existence of three example configuration files in the repo. This one seems to be a mix between the [basic](https://github.com/symless/synergy-core/blob/master/doc/synergy.conf.example-basic) and the [advanced](https://github.com/symless/synergy-core/blob/master/doc/synergy.conf.example-advanced) examples, but its naming doesn't indicate that. Also, the advanced example seems to be incomplete/inconsistent, and partially copied from this one. Perhaps it should it be removed and this one renamed to be the advanced example, for the time being?

PR #6261, opened 2018-02-24
Add support for latin 's' and 't' with comma below

rombert:
These codes were not supported at all, presumably due to being introduced
only in Unicode 3.0.


rombert:
Thanks for merging @nlyan !

PR #6268, opened 2018-03-09
Couple of minor fixes.

erikd:


src/lib/core/App.cpp
```diff
@@ -175,7 +175,7 @@ App::initApp(int argc, const char** argv)
     // this is a simple way to allow the core process to talk to X. this avoids
     // the "WARNING: primary screen unavailable: unable to open screen" error.
     // a better way would be to use xauth cookie and dbus to get access to X.
-    if (static_cast<int>((!(argsBase().m_runAsUid) == 0 != -1))) {
+    if (argsBase().m_runAsUid != 0) {
```
erikd: Not 100% sure this is correct, but the previous code was needlessly confusing.

nlyan: Git blame reveals the previous code was actually rewritten by clang-tidy. It looks like before that the logic was dodgy, clang-tidy just revealed what it was actually doing.

See: https://github.com/symless/synergy-core/blame/5fbada942471ccf2dde01ccfc5577f008c15bd81/src/lib/core/App.cpp#L182

nlyan: Suggest changing to : 

```if (argsBase().m_runAsUid >= 0) {```

erikd: Updated as suggested. Thanks @nlyan

PR #6298, opened 2018-05-07
Removed invalid #include causing compile error

AshtonHarding:
After removing this, I was able to compile it on Windows 10.


Jnewbon:
It seems this change is already done. On v2-dev (2398824b) and v1-dev (f07b765a)

Closing.

PR #6323, opened 2018-06-24
Add missing headers to build against Qt 5.11.0

jpwhiting:
Some Qt headers are not included in QtGui anymore. So add them where
needed to get synergy 1.10 to build on windows with Qt 5.11.0

### DIFF
--- a/src/gui/src/ActionDialog.cpp
+++ b/src/gui/src/ActionDialog.cpp
@@ -26,6 +26,8 @@
 #include <QtCore>
 #include <QtGui>
 
+#include <QButtonGroup>
+
 ActionDialog::ActionDialog(QWidget* parent, ServerConfig& config, Hotkey& hotkey, Action& action) :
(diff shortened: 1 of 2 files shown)


nbolton:
Hey, is this the effective live?

`#include <QHeaderView>`

Difficult to tell because of the code style changes. Perhaps you could split this into 2 commits? 1 for the fix, and another commit for the code style?

jpwhiting:
My bad, one moment, I'll redo the patch without the style changes. Qt Creator did those when I saved the file.

By the way, in order to get v1.10 to build on windows I had to set SYNERGY_ENTERPRISE in my environment so it wouldn't try to look for dnssd.lib when building. What's the purpose of that variable? Is it expected that non enterprise builds require dnssd.lib ? If so where should dnssd.lib come from?

jpwhiting:
Ok, that's better. Sorry for the noise in the earlier patch.

nbolton:
Aha! Perfect, didn't realise it was in 2 files 👍

PR #6408, opened 2018-08-29
Fixed possible data index bug for Unicode read

cuviper:


cuviper:
TBH, I'm not even sure if this code is used, but I think the corrections are straightforward.

Jnewbon:
Closed as this PR would pull in V2 changes to the v1 branch

Recreated #6617

Issue #6409, opened 2018-09-04
Server freezes when clipboard contains an image


nneul:
### Operating Systems ###

Ubunty 16 client and Ubuntu 18 server, both x86_64

### Synergy Version ###

synergys 1.10.0, protocol version 1.6

### Steps to Reproduce ###

Screenshot or similar to copy an image to the X clipboard on the server. It appears to be size based, smaller copies do not appear to trigger the same lockup.

### Log URLs ###

No indications of failure on the client side, only on server. The last messages in the log are it attempting to retrieve text/html and then image/bmp. 

[2018-09-04T14:59:18] DEBUG1: request succeeded after 0.000001s
[2018-09-04T14:59:18] DEBUG:   available targets: TARGETS (450), MULTIPLE (451), image/tiff (857), image/jpeg (660), image/x-MS-bmp (848)
[2018-09-04T14:59:18] DEBUG1: request selection=CLIPBOARD (448), target=text/html (574), window=5000004
[2018-09-04T14:59:18] DEBUG2: can't read property 454 on window 0x05000004
[2018-09-04T14:59:18] DEBUG1: request failed after 0.000003s
[2018-09-04T14:59:18] DEBUG1: can't get data for selection target text/html (574)
[2018-09-04T14:59:18] DEBUG1:   no data for target text/html (574)
[2018-09-04T14:59:18] DEBUG1: request selection=CLIPBOARD (448), target=image/bmp (575), window=5000004
[2018-09-04T14:59:18] DEBUG2: read property 454 on window 0x05000004: bytes=4
[2018-09-04T14:59:25] DEBUG2: reading secure socket

Now that I know how to prevent it, it's easier to deal with.

nneul:
Does not appear to change behavior if I change the size limit on clipboard sharing.

nneul:
And no change with latest 1.10.1 build.

nneul:
Little bit of brute force debug - it's getting stuck oscillating between 1343 and 1348 debug lines, but only when I move the mouse - it's not a busy spin:

src/lib/platform/XWindowsClipboard.cpp:

```1341         if (noWait || XPending(display) > 0) {
1342             while (!m_done && !m_failed && (noWait || XPending(display) > 0)) {
1343 LOG((CLOG_DEBUG1 "dbg %s:%d", __FILE__, __LINE__));
1344                 XNextEvent(display, &xevent);
1345                 if (!processEvent(display, &xevent)) {
1346                     // not processed so save it
[...]

PR #6410, opened 2018-09-04
Add clipboard read events timeout on Linux

nneul:


nneul:
Resolves #6409

nneul:
Or at least masks the hang by timing out. Probably some better fix that actually addresses what isn't working.

Jnewbon:

nneul:
@Jnewbon This is a trivial fix to a clearly defined problem, is there some reason it's still not merged?

Jnewbon:
I have reproduced this bug and I can confirm that this PR fixes it. so I have merged it.

PR #6425, opened 2018-10-04
Fixed about dialog on Linux dark theme

jpwhiting:
afterpatch
beforepatch

If breeze dark desktop theme is used on linux the about dialog
icon "Synergy" text is not legible. The height of the dialog is
also too short cutting off the last line of about text.
Remove "Synergy" from about.png and put label with 20pt text in
the theme foreground color instead.
Set height of about dialog on linux to 350 to make last line visible.

Screenshots showing before and after this patch on kde with breeze dark theme attached.


Jnewbon:

jpwhiting:
Makes sense, ok.

Jnewbon:

PR #6433, opened 2018-10-13
Fixed multi-layout keyboard bugs on macOS client

inlife:
when using multiple layout on a macos client machine, the switching to other than default layout causes input bugs, where some of the keys are still on the default layout and some are on switched one

this fix makes sure that layout will be properly switched

### DIFF
--- a/src/lib/platform/OSXKeyState.cpp
+++ b/src/lib/platform/OSXKeyState.cpp
@@ -398,17 +398,15 @@ OSXKeyState::pollActiveModifiers() const
 SInt32
 OSXKeyState::pollActiveGroup() const
 {
+    bool layoutValid = true;
     TISInputSourceRef keyboardLayout = TISCopyCurrentKeyboardLayoutInputSource();
-    CFDataRef id = (CFDataRef)TISGetInputSourceProperty(
-                        keyboardLayout, kTISPropertyInputSourceID);
     
-    GroupMap::const_iterator i = m_groupMap.find(id);
-    if (i != m_groupMap.end()) {
-        return i->second;
+    if (layoutValid) {
+        GroupMap::const_iterator i = m_groupMap.find(keyboardLayout);
+        if (i != m_groupMap.end()) {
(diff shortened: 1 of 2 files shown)


Jnewbon:
@inlife Does this relate to a specific issue. 

Also are you able to provide steps to reproduce the bug. 

The code looks good and clean, I just want to make sure it fixes the bug intended.

inlife:
@Jnewbon not sure tbh. I had personally experienced that issue and then fixed it.
Had no problems since. Still actively using it, up to this day.

src/lib/platform/OSXKeyState.cpp
```diff
@@ -398,17 +398,15 @@ OSXKeyState::pollActiveModifiers() const
...
 {
+    bool layoutValid = true;
     TISInputSourceRef keyboardLayout = TISCopyCurrentKeyboardLayoutInputSource();
-    CFDataRef id = (CFDataRef)TISGetInputSourceProperty(
-                        keyboardLayout, kTISPropertyInputSourceID);
     
-    GroupMap::const_iterator i = m_groupMap.find(id);
-    if (i != m_groupMap.end()) {
-        return i->second;
+    if (layoutValid) {
```
Jnewbon: Is this `if` really necessary? it will always evaluate to true due to the variable being set just above it

inlife: I think that the `if` was there at the time of my fork being created. And seems like you are referring to an outdated change, as github noted above.

Jnewbon: You'r right, not sure how i clicked the wrong link :)

Jnewbon:
It all looks fine, as this is targeting the v2 branch, I'm going to close this PR and open a new one with v1 as the base branch. 

If I switch this PR to V1 then it will pull in a lot of v2 changes that we don't want on the v1 branch.

I have cherry-picked both commits to preserve author details and the new PR is #6627

PR #6435, opened 2018-10-24
Fixed missing SSL_get_client_ciphers with LibreSSL

vifino:
This patch is for the 1.X versions. 2.X does not have this problem.
Weirdly enough, LibreSSL has a high enough version, yet does not feature SSL_get_client_ciphers.


Jnewbon:

Issue #6470, opened 2019-04-04
CI solution with on-demand containers


Jnewbon:
Problem: We are using Jenkins with static VM instances, So the VM's are always online wasting resources. Also, macOS builder is a physical machine and is always online.

Solution: Use a CI as-a-service which offers on-demand instances preferable using Docker images. Cirrus CI could be a good solution, however, their Mac builder doesn't support Docker.

Note: This issue does not include deployment (i.e. no snapshot upload). This will be a separate issue.

fkorotkov:
How do you use Docker on Mac right now?

Jnewbon:
> How do you use Docker on Mac right now?

Currently, we don't, we use a physical machine in the office connected to Jenkins, for the Mac builds. I have a build working on Mac on Cirrus CI. It would have been nice to have all the builds using Docker for ease of maintaining them.

PR #6486, opened 2019-04-24
Added CI Solution for automated builds

Jnewbon:
#6470

Added Cirrus CI Builds for
- Windows
- MacOS
- Ubuntu 16.04
- Ubuntu 18.04

Linux builds are Via docker containers
MacOS uses Anka VMs via Cirrus CI
Windows uses Google cloud images via Cirus CI (Image managed by Symless)

### DIFF
--- a/src/gui/src/ActionDialog.cpp
+++ b/src/gui/src/ActionDialog.cpp
@@ -25,6 +25,7 @@
 
 #include <QtCore>
 #include <QtGui>
+#include <qbuttongroup.h>
 
 ActionDialog::ActionDialog(QWidget* parent, ServerConfig& config, Hotkey& hotkey, Action& action) :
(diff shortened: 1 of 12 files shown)


src/gui/src/ActionDialog.cpp
```diff
@@ -25,6 +25,7 @@
 
 #include <QtCore>
 #include <QtGui>
+#include <qbuttongroup.h>
```
darkdragon-001: Use `QButtonGroup` instead.

src/gui/src/ScreenSetupView.cpp
```diff
@@ -22,6 +22,7 @@
 
 #include <QtCore>
 #include <QtGui>
+#include <qheaderview.h>
```
darkdragon-001: Use `QHeaderView` instead.

Issue #6375, opened 2018-07-29
Support for Qt 5.11 framework on Windows


nbolton:
Problem: The code is based on an older version of Qt, where things like `qt5_use_modules` are deprecated.

Solution: Update the code so it compiles with Qt 5.11.

PR: #6358 and #6359.

nbolton:
On second thoughts, probably not good timing right now to upgrade to Qt 5.11.

nbolton:
Fixed by #6489

Issue #6366, opened 2018-07-22
Incomplete type build error when using Qt 5.11


targodan:
### Operating Systems ###

Arch Linux with 4.17.5 Kernel

### Synergy Version ###

1.10.0

### Command Line ###

Build as per instructions with qt version 5.11. Will output errors with incomplete type QButtonGroup and more.

### Steps to reproduce bug ###

See above

### Other info ###

There are include files missing. I got the build working. Will try to draft a pull request by Monday evening.

PS: Sorry about the initially empty issue, I accidentally hit enter.

nbolton:
Duplicates #6375

hopeseekr:
Hey, I ran into this trying to install synergy on Arch Linux. I fixed it for you.

nbolton:
@hopeseekr How does your fix compare to PRs #6358 and #6359 by @jpwhiting?

jpwhiting:
It's the same, it was merged into master at one point, then reverted later for some reason...

Jnewbon:
The changes have been merged into `v1-dev` See #6489. We are currently doing the internal review for a release that will pull all the changes on the `v1-dev` branch onto the `master` branch.

nbolton:
Also, https://github.com/symless/synergy-core/issues/6475 is probably relevant.

PR #6496, opened 2019-05-04
(#6366, #6375) Now compiles with Qt v5.11.

hopeseekr:
Please merge this so that Arch Linux users can use the product.

Thank you.

---
This change is [Reviewable](https://reviewable.io/reviews/symless/synergy-core/6496)


Jnewbon:
This fix is already in the pipeline on `v1-dev`

Duplicate of #6489

PR #6500, opened 2019-05-12
Use macdeployqt to install Qt frameworks inside bundle

jpwhiting:
I found make install copies synergy into the bundle/Synergy.app but fails to install the Qt frameworks. Adding this call to macdeployqt when running make install fixes it for macos. A similar solution may be wanted for windows also.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -11,6 +11,13 @@ file (GLOB LEGACY_GUI_UI_FILES src/*.ui)
 file (GLOB LEGACY_ACTIVATION_FILES src/*Activation* src/*License*)
 file (GLOB LEGACY_ZEROCONF_FILES src/Zeroconf*)
 
+# Retrieve the absolute path to qmake and then use that path to find
+# the binaries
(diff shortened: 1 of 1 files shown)


Jnewbon:

Jnewbon:

Jnewbon:
I'm just looking into this again, would this affect the codesiging for deployment
https://github.com/symless/synergy-core/blob/2de8296b879512d26b8c8f3a4c407fe469ccdf97/CI/MacOS/mac-build.yml#L51-L61

That is run for a release candidate?

jpwhiting:
I'm not sure what you are asking. macdeployqt can be used to sign the .app yes. as you are there in the ci script. My change just makes manual builds copy the Qt dependencies into place so the .app will run if building synergy locally on macos.

Jnewbon:
I understand, i just don't want it breaking the signed deployment scripts is all. I spent way too long getting them up and running. 
I should have merged this first really would has save me trouble with the windeployqt part. :)

What ill do is merge this into a branch and run a couple tests, if all goes well i merge that into master. if not ill see if i can get it to work with deployment then merge

PR #6505, opened 2019-05-18
Trap WM_SYSCOMMAND SC_MONITORPOWER to detect monitor off even if no s…

nyetwurk:
…creen

saver running.

HACK: we never get SC_MONITORPOWER lParam = -1, so just assume monitor turned
back on if we get any event on the primary screen.

### DIFF
--- a/src/lib/platform/MSWindowsScreen.cpp
+++ b/src/lib/platform/MSWindowsScreen.cpp
@@ -115,6 +115,7 @@ MSWindowsScreen::MSWindowsScreen(
     m_screensaver(NULL),
     m_screensaverNotify(false),
     m_screensaverActive(false),
+    m_screensaverPowerOff(false),
     m_window(NULL),
     m_nextClipboardWindow(NULL),
     m_ownClipboard(false),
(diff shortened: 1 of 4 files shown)


src/lib/platform/MSWindowsScreen.cpp
```diff
@@ -980,12 +981,24 @@ MSWindowsScreen::onPreDispatch(HWND hwnd,
...
+
     case SYNERGY_MSG_DEBUG:
         LOG((CLOG_DEBUG1 "hook: 0x%08x 0x%08x", wParam, lParam));
         return true;
     }
 
     if (m_isPrimary) {
+        // FIXME: if anything happens, assume monitor powered up
+        if (m_screensaverPowerOff)
+            m_screensaverPowerOff = false;
```
nyetwurk: Argh, this has a race condition. It only works if debugging is enabled.

SerhiiGadzhilov:
This PR can’t be merged because of conflicts.
Also this functionality is temporary removed and needs to rewritten.

nbolton:
I think perhaps we should have helped to solve the conflicts here. 

@nyetwurk Do you think perhaps this is still something we can use?

nyetwurk:
Yes, I believe so but I no longer use app so I cannot really help debug/test it.

Also make sure the note about the race condition is resolved.

nbolton:
> Yes, I believe so but I no longer use app so I cannot really help debug/test it.
> 
> Also make sure the note about the race condition is resolved.

@sithlord48 Do you want to have a go at solving that race condition?

nbolton (commented):
I tried solving the merge conflicts but it's going to need a fair investment of time as the code has changed significantly.

nbolton:
> I tried solving the merge conflicts but it's going to need a fair investment of time as the code has changed significantly.

Unfortunately, I don't think we have the bandwidth to fix this PR at the moment :disappointed:

PR #6553, opened 2019-09-04
Barrier  on client

EAlthuis:
I installed Barrier Maxiberta on both the server and client, worked OK on server but not on client.  The symbol flashed a few times then stayed on but no contact was made with the server.
It worked at installation time, but not the next day.
I copied the server conf file to the client  to no avail. 
I need a fix.... I have nothing to attach.
The server is Ubuntu 19.04, the client is Ubuntu 18.04.3LTS


nbolton:
Hi there, it looks like you meant to ask for tech support instead of submitting a pull request. Please raise a support ticket: https://members.symless.com/contact/customer-service

PR #6563, opened 2019-10-02
Always send clipboard data when changing screen

Waester:
I have observed that the clipboard sharing between the server and clients become out of sync when a client disconnect from the server and during that time perform a copy-action.

If the server's clipboard has remained unchanged, while the client reconnect, the data it hold versus the client are different, the server wont push an update because the server's logic is that "I have already sent the data, so lets not do it again".

I don't know if this is a design decision, but the way I see it is that the server's clipboard is the primary clipboard and any client which is connected afterward should have it's clipboard.

The downside to this change is that the clipboard data is always sent when changing screen, causing higher network usage.

### DIFF
--- a/src/lib/client/Client.cpp
+++ b/src/lib/client/Client.cpp
@@ -276,7 +276,6 @@ Client::setClipboard(ClipboardID id, const IClipboard* clipboard)
 {
      m_screen->setClipboard(id, clipboard);
     m_ownClipboard[id]  = false;
-    m_sentClipboard[id] = false;
 }
 
 void
@@ -284,7 +283,6 @@ Client::grabClipboard(ClipboardID id)
 {
     m_screen->grabClipboard(id);
     m_ownClipboard[id]  = false;
-    m_sentClipboard[id] = false;
 }
 
 void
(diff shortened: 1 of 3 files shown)


Jnewbon:
The change sounds like an idea, however, you have already illuded to the higher network impact, if the connection is rather slow (e.g. wifi) this might adversely affect the performance if always send clipboard is active.

It may be worth having this as an option in the settings so users can opt-in to having the clipboard always sent to clients. 

The design choice is minimal network impact as possible. I think if we had the setting "Always send clipboard" that was defaulted to off then we would have no problem getting this PR in.

Waester:
It seems that Synergy already does some fugly hack to "always send" on the server side, albeit silently.

https://github.com/symless/synergy-core/blob/ca35737ab9882334a6c30fd3c154739727e352b9/src/lib/server/Server.cpp#L525-L534

This essentially bypasses the function onClipboardChanged where the checks are programmed, so my changes most likely will only affect the client side on this PR.

I think I will close this PR and review the clipboard sending mechanic and maybe propose a better change in late date.

PR #6565, opened 2019-10-04
Fix clipboard out-of-sync

Waester:
These changes fixes an issue where, sometime, the client and server's clipboards become out-of-sync, resulting in empty clipboards or clipboards not updating with new content.

Might be related to:
https://github.com/symless/synergy-core/issues/5729
https://github.com/symless/synergy-core/issues/5776

### DIFF
--- a/src/lib/server/Server.cpp
+++ b/src/lib/server/Server.cpp
@@ -501,12 +501,19 @@ Server::switchScreen(BaseClientProxy* dst,
 
 		// update the primary client's clipboards if we're leaving the
 		// primary screen.
-		if (m_active == m_primaryClient && m_enableClipboard) {
+		if (m_enableClipboard) {
 			for (ClipboardID id = 0; id < kClipboardEnd; ++id) {
 				ClipboardInfo& clipboard = m_clipboards[id];
-				if (clipboard.m_clipboardOwner == getName(m_primaryClient)) {
-					onClipboardChanged(m_primaryClient,
-						id, clipboard.m_clipboardSeqNum);
+				if (m_active == m_primaryClient && clipboard.m_clipboardOwner == getName(m_primaryClient)) {
+					m_active->getClipboard(id, &clipboard.m_clipboard);
+				}
+				String data = clipboard.m_clipboard.marshall();
+				if (data != clipboard.m_clipboardData) {
+					for (ClientList::const_iterator index = m_clients.begin(); index != m_clients.end(); ++index) {
+						BaseClientProxy* client = index->second;
+						client->setClipboardDirty(id, client != m_active);
+					}
+					clipboard.m_clipboardData = data;
 				}
 			}
 		}


Jnewbon:
Hi @Waester 

Could you link this PR to an issue number? if there isn't an issue for this PR you can create one and link it. I ask because we track issues rather than PR's and having an issue attached will link this to exactly what it fixed allowing release notes to maintain traceability back to specific bugs.

After that would you mind give a quick explanation of how you fixed the issue? 

Thanks
Jamie

Waester:
Hi @Jnewbon 

So the function switchScreen() calls onClipboardChanged() to retrieve the clipboard from the system and stores it in a variable. But onClipboardChanged() was not designed for how switchScreen() is meant to work. To fix it I removed onClipboardChanged() from switchScreen() and wrote this:

https://github.com/Waester/synergy-core/blob/efa6d42fd4b6a6852cc2947fdffc28e31cf05b60/src/lib/server/Server.cpp#L508-L512

These codes are actually used inside onClipboardChanged() too. But due to how onClipboardChanged() is designed these following codes might be skipped due to some circumstances.

https://github.com/Waester/synergy-core/blob/efa6d42fd4b6a6852cc2947fdffc28e31cf05b60/src/lib/server/Server.cpp#L1589-L1594

What happens is that without these lines the clients are not informed that they should accept new clipboard data from the server.

What causes this is the following code:

https://github.com/Waester/synergy-core/blob/efa6d42fd4b6a6852cc2947fdffc28e31cf05b60/src/lib/server/Server.cpp#L1580-L1583

Imagine the scenario where the server has new clipboard data to send and he informs his clients that he is sending new clipboard data and the clients recieves this information and will start accepting the new clipboard data. But then one client decides to disconnect from the server and proceeds to perform an untracked copy action. During this entire time the server has not performed any new copy action, so the clipboard data it has is still the same as before the client disconnected.
So when the client reconnects, the server will send the clipboard data again when the user is changing screen. But this time this will fail due to that the client has not been informed that new clipboard data is supposed to be retreived, due to being blocked by the above condition.

So what I did was simply copy those necessary codes to switchScreen().
Why I didn't rewrite onClipboardChanged() is due that it is being used by handleClipboardChanged() and onClipboardChanged() is working as intended when used by handleClipboardChanged().

onClipboardChanged() has two conditions and by bypassing onClipboardChanged() those condition won't be checked, but this is untrue, one of the conditions are actually checked in later part of the code in switchScreen(). Namely here:

https://github.com/Waester/synergy-core/blob/efa6d42fd4b6a6852cc2947fdffc28e31cf05b60/src/lib/server/Server.cpp#L535-L537

The rest of the codes I submitted is to avoid calling the clipboard sending code when moving from client to server screen.

https://github.com/Waester/synergy-core/blob/efa6d42fd4b6a6852cc2947fdffc28e31cf05b60/src/lib/server/Server.cpp#L530

This is because the rest of the code is meant for "server to client" communication. But there are cases where "server to itself" communication can occur unless that condition is there. "Client to server" communcation is handled by handleClipboardChanged().

PS:
I will try to find tickets related to this issue, otherwise I will create my own. This issue has been noted by me and my colleagues due to how often we undock our laptop from the docking station and losing connection to the stationary PC (Synergy server). We have two computers.

Jnewbon:
@Waester 

That's a much more in-depth answer than I was expecting, it's much appreciated, thanks.

Might need to give me a bit of time to understand the current code, the improvements and how you have done it but ill add it to the next synergy-core sprint.

Waester:
@Jnewbon 

Yeah, totally understandable.

In the mean time I have updated the branch with more changes. The first ones I had included was mainly "server-to-client" fixes. The one I am including now are "client-to-server" related.

It is nothing special with the changes I am proposing, I am just removing some, which I think, is redundant condition checks.

The variable m_dataClipboard already covers the majority of cases when the client should NOT send clipboard data. While m_sentClipboard seems to be creating cases of blocking the client from sending, legitimate, clipboard data to the server.

Jnewbon:
Updated the target branch to v1-dev as that is where we deliver everything to, for testing and QA before release.

Waester:
I am closing this because I am not really satisfied in the end with the PR. While the changes did fix one issue, it was more like a band-aid for a larger issue.

nbolton:
@Waester I wonder if it would still be merging anyway? Is the band aid still relevant? We could have it as a feature flag so you can turn it on and off to see if it improves things.

nbolton:
Reopening in case of reply.

PR #6574, opened 2019-10-16
Server restart hotkey

### LINKED ISSUE #6518: Key combination that will force a server restart
### Problem ###

There are situations where a server and client may have problems, like the cursor stuck in a location, etc. where the user is not easily able to restart synergy and reconnect quickly.

### Solution ###

Having the ability to press a key combination that will force the server to terminate all current connections and allow all clients to reconnect. All clients will automatically attempt to reconnect once the connection is lost to the server.

Jnewbon:
Closes #6518

### DIFF
--- a/src/lib/server/InputFilter.cpp
+++ b/src/lib/server/InputFilter.cpp
@@ -323,6 +323,40 @@ InputFilter::LockCursorToScreenAction::perform(const Event& event)
                                 Event::kDeliverImmediately));
 }
 
+
+InputFilter::RestartServer::RestartServer(
+        IEventQueue* events, Mode mode) :
+        m_mode(mode),
+        m_events(events)
+{
+    // do nothing
+}
+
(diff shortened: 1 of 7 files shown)


src/lib/server/InputFilter.cpp
```diff
@@ -323,6 +323,40 @@ InputFilter::LockCursorToScreenAction::perform(const Event& event)
...
+    static const char* s_mode[] = { "restart" };
+
+    return synergy::string::sprintf("restartServer(%s)", s_mode[m_mode]);
+}
+
+void
+InputFilter::RestartServer::perform(const Event& event)
+{
+    //HACK Super hack we should gracefully exit
+   exit(0);
```
nbolton: @Jnewbon Hopefully this won't cause issues with the hook DLL (but we need to remove that anyway).

Issue #6588, opened 2019-10-31
Version number header with update check


nbolton:
### Problem ###

The version check request doesn't include a current version header, meaning that the version request server log is missing version numbers.

### Solution ###

Add a header to the version check call.

Jnewbon:
A release candidate is available that incorporates a fix for this issue

[Download v1.11.0-rc2](https://snapshots.symless.com/public/1.11.0-rc2?github#6588)

If you find any bugs in the release candidate related to this issue please comment here.

Issue #6575, opened 2019-10-17
Service start error on Windows


nbolton:
### Operating Systems ###

Windows 10

### Synergy Version ###

1.10.3

### Steps to Reproduce ###

1. Install
2. See service error

> Service "Synergy" (synergy) failed to start. Verify that you have sufficient privileges to start system services.

### Extra Info ###

Installing the [latest VC redist](https://support.microsoft.com/en-ph/help/2977003/the-latest-supported-visual-c-downloads) seems to solve the issue. Perhaps the installer isn't including the latest redist? See forum posts for more detail.

https://members.symless.com/forums/topic/6321-service-start-error-on-windows/
https://members.symless.com/forums/topic/6316-upgrade-to-1103-has-broken-my-installation/

Jnewbon:
A release candidate is available that incorporates a fix for this issue

[Download v1.11.0-rc2](https://snapshots.symless.com/public/1.11.0-rc2?github#6575)

If you find any bugs in the release candidate related to this issue please comment here.

Issue #6569, opened 2019-10-09
Unable to run on macOS 10.15 Catalina


Jnewbon:
### Operating Systems ###

macOS 10.15 Catalina

### Synergy Version ###

<= 1.10.3

### Steps to Reproduce ###

1. Double click Synergy icon
1. See error below

![image.png](https://images.zenhubusercontent.com/5c9901dc5f846800016ee60b/bb9fcf9e-d04d-4c26-853d-88c82e128b15)

Workaround:
1. Open application folder
1. Right-click Synergy
1. Click open

Jnewbon:
Requires Apple to notarize the application, the process is long and complex and requires authorization from a team owner before I can continue.

Passing the task to @nbolton for authorization

Jnewbon:
Gained permissions, 
The app is now notarized. Changes should be reflected automatically on internet-connected machines according to apple documentation. The download also has the notarization ticket stapled to it for offline checking and download the binary update accordingly. 

Next step is to update automated build machine

Jnewbon:
A release candidate is available that incorporates a fix for this issue

[Download v1.11.0-rc2](https://snapshots.symless.com/public/1.11.0-rc2?github#6569)

If you find any bugs in the release candidate related to this issue please comment here.

Issue #6566, opened 2019-10-08
Debug message output mistakenly as info message


Jnewbon:
### Operating Systems ###

All

### Synergy Version ###

1.10.3

### Steps to Reproduce ###

1. Run synergy with log level set to info
1. Move to client
1. Pressed some shortcuts like Shift + f 
1. See log entries

```
[2019-10-02T17:07:13] INFO: found key in group 0
[2019-10-02T17:07:14] INFO: found key in group 0
```

Jnewbon:
A release candidate is available that incorporates a fix for this issue

[Download v1.11.0-rc2](https://snapshots.symless.com/public/1.11.0-rc2?github#6566)

If you find any bugs in the release candidate related to this issue please comment here.

PR #6590, opened 2019-11-01
v1.11.0-stable

Jnewbon:
### Features:

- #6518 Key combination that will force a server restart

### Enhancements:

- #6588 Version number header with update check
- #4957 Server fails to start up due to synwinhk.dll is in use

### Bug fixes:

- #5294 Cursor lockout at Windows server login screen
- #6569 Unable to run on macOS 10.15 Catalina
- #6561 Debian buster no longer supports CA key length of 1024
- #6556 Function missing error from OpenSSL/TLS
- #6575 Service start error on Windows
- #6566 Debug message output mistakenly as info message
- #5959 User interface failed to load local fingerprint


Jnewbon:
@Alvarord Could you elaborate? which part doesn't work for you?

Jnewbon:
Ah @Alvarord I just noticed you commented against a specific commit,

So I guessing your running on Debian and encountering the key size error. The settings I changed only affect newly generated certificates, so you will need to delete the current certificate and let synergy regenerate it with those settings. 

After deleting the certificate located in `~/.synergy/SSL/Synergy.pem` toggle TLS off then back on and synergy will generate a new cert at the larger key length.

This was done so people not affected by this bug will not have their certificate suddenly changed on them. as it would trigger a change in fingerprint warning message.

Alvarord:
Validated the RC2 deleting the file.

Thanks a lot.

Issue #6593, opened 2019-11-06
Non-ASCII characters replaced with ? on client


adamryczkowski:
Server: Ubuntu 18.04
Client: Ubuntu 18.04

Synergy version `1.10.3-stable, protocol version 1.6` downloaded from current GitHub releases and compiled using libraries available in the Ubuntu repository.

### Steps to Reproduce ###

Copy to clipboard text in a text editor on the server and try to paste it to gedit text editor opened on client.

Expected: Text should copy as-is
Actual: Non-ASCII characters got replaced with `?` on the client

------

I believe this is a serious bug, limiting core functionality for many (if not all) users. 

The issue was discussed many times already  (#316, #981, #1511, #1786, #4824, #4973, #6092) but never resolved fully. I hope this bug report sheds some light on the situation.

Thank you

adamryczkowski:
Important update:
Copying to/from some specific applications across screens actualy does work, in particular I was able to test it on *VSCode* (`sudo snap install code`) and in the *Chromium browser*.

Unfortunately, the majority of applications I tried does not receive the pasted Unicode string correctly; in particular: *gedit*, *Libre Office Writer*, *terminator* terminal emulator and *Firefox*.

I believe that you miss to forward some "*this text is in Unicode*" tag, which confuses the majority of applications. But you do copy the actual text right, and perhaps that is why the issue slipped through your QA. 

HTH

nbolton:
Fixed by #6617

NicolasGoeddel:
Just tried the fix in version 1.11.1 between a Windows 10 and Ubuntu 19.10. And the issue persists.
And it behaves exactly like @adamryczkowski mentioned it. I never used Chromium or VSCode together with Synergy. But I always use gedit, the terminal emulator, LibreOffice and Firefox to do my work.
With Chromium and VSCode copy/paste works like a charm. So half of the solution is there. 

Please reopen.

NicolasGoeddel:
I just found out that the difference seems to be GTK2 vs. GTK3. With GTK2 the clipboard works flawlessly. But with GTK3 applications all non-ASCII characters will become a `?`.

xenomachina:
I'm curious to know why this issue is closed.

[...]

PR #6617, opened 2020-01-11
Fixed possible data index bug for Unicode read

Jnewbon:
Original PR #6408 

> TBH, I'm not even sure if this code is used, but I think the corrections are straightforward.

This was a good catch and should defiantly be merged

I have linked this to #6593 as it may affect it


nbolton:
Fixes #6593

PR #6627, opened 2020-01-28
Fixed multi-layout keyboard bugs on macOS client

Jnewbon:
Original PR #6433 by @inlife
> when using multiple layout on a macos client machine, the switching to other than default layout causes input bugs, where some of the keys are still on the default layout and some are on switched one
>
>this fix makes sure that layout will be properly switched

Cherry-pick also removed some whitespace from the files.


Jnewbon:

PR #6635, opened 2020-02-25
updating man pages to reflect current state of --help

### LINKED ISSUE #6601: Documentation (man page) for synergyc missing --enable-crypto option
Even if it's a paid only feature, the feature is still there isn't it? I'm currently using 1.10.2 from the fedora repos with my subscription key. The option is available, but not documented in the man page for synergyc.

Let me know if I'm barking up the wrong tree here, but it should be there even if there is an accompanying note mentioning that it's a paid feature.

67hz:
Fixes #6601 
* generated new man pages for both client and server applications


67hz:
This addresses issue #6601 

Both client and server man pages have been updated to reflect the respective `--help` menus following the existing convention of `help2man`.

We may want to consider adding `help2man` as a `Make` task for future man page generation.

Jnewbon:
I defiantly agree with adding this as a task to make, with the version number being put in there, we would have to manually update it each time otherwise.

Ill merge this in and add a new issue #6713 for adding the man page generation to the make files

PR #6644, opened 2020-03-17
cmake: check for dns_sd header file

### LINKED ISSUE #6640: missing build dependency in cmake
cmake doesn't detect that this library is missing from the system:
```
In file included from /home/willi/src/deb/synergy-core/build/src/gui/synergy_autogen/UVLADIE3JM/moc_ZeroconfBrowser.cpp:9,
                 from /home/willi/src/deb/synergy-core/build/src/gui/synergy_autogen/mocs_compilation.cpp:30:
/home/willi/src/deb/synergy-core/src/gui/src/ZeroconfBrowser.h:25:10: fatal error: dns_sd.h: No such file or directory
   25 | #include <dns_sd.h>
      |          ^~~~~~~~~~
compilation terminated.
make[2]: *** [src/gui/CMakeFiles/synergy.dir/build.make:124: src/gui/CMakeFiles/synergy.dir/synergy_autogen/mocs_compilation.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:958: src/gui/CMakeFiles/synergy.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```
fixed by 
```
apt-get install libavahi-compat-libdnssd-dev
```
Please add a check to the cmake file that will abort the configure run before compiling.

sim590:
Fixes #6640 
This is a fix for #6640. I'm only fixing for `dns_sd.h`. I thought may be to fix for openssl but I have not played with it yet.

~~`EDIT: This is WIP. I'm realizing that I need to handle the case when the file is not present manually.~~


sim590:
A weird thing happening with the current setup: if a header file is not found, one cannot simply install the header file and run `cmake` again. Instead, it's necessary to remove the cmake *cache* files before running `cmake` again. I don't know if this is mandatory. But that's the currents setup.

sim590:
Question: `libavahi-compat-libdnssd-dev` doesn contain a package config file. I have notice that you don't seem to use package config with CMake's `find_package`. Is there any reason not to? For instance, `Xlib` could be handled with package config but it is not. From the CMake man page for `find_package` documentation:

> The REQUIRED option stops processing with an error message if the package cannot be found.

This seems like the wanted behaviour, no?

Jnewbon:
This issue also states the openssl header isn't checked for but that can be done in another PR

PR #6653, opened 2020-03-24
Fixed problems with translations

### LINKED ISSUE #6622: Language files not updated during build
### Operating Systems ###

All

### Synergy Version ###

1.11.0

### Steps to Reproduce ###

1. Change a translation
2. Build synergy
3. See translation is not updated

### Extra

During the delivery of #6621 it was discovered that language files have to be manually updated, and their compiled outputs are stored in the repo. They should be built as part of the build process.

Jnewbon:
Fixes #6622


Jnewbon:
Had to update pipelines due to the removal of MacOS sierra, also Ubuntu/Debian cause problems.

PR #6659, opened 2020-04-01
Added system config selection for all users configuration

### LINKED ISSUE #6538: System-wide config for all system users
### Problem ###

The config doesn't persist across users when you configure for one user, and then switch to another. The config is not shared between users; each user has to have their own config. In other words, once a config has been created, when a new user logs into the same computer, the config is essentially lost.

### Solution ###

Add an option in the Settings dialog to save the config for all users. The tricky part is making that work on all platforms without affecting the default single user config behavior. Command-line arguments should still override any settings files, and user config should override the global config.

Jnewbon:
Fixes #6538


src/gui/src/ConfigWriter.cpp
```diff
@@ -0,0 +1,218 @@
+//
```
nbolton, line 1: Missing standard heading comment.

src/gui/src/ConfigWriter.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include "ConfigBase.h"
+
+namespace GUI {
+    namespace Config {
+        //Assignment of static variable
+        ConfigWriter *ConfigWriter::s_pConfiguration = nullptr;
+
+
+        ConfigWriter *ConfigWriter::make() {
+            // Only one ConfigWriter can exist at any one time (Singolton)
```
nbolton, line 21: Typo: singleton

src/gui/src/main.cpp
```diff
@@ -88,9 +88,10 @@ int main(int argc, char* argv[])
 #ifndef Q_OS_WIN
     QApplication::setQuitOnLastWindowClosed(false);
 #endif
+    //S
```
nbolton, line 91: Rogue comment?

src/gui/src/ConfigWriter.cpp
Jnewbon: Added header with correct dates

src/gui/src/ConfigWriter.cpp
Jnewbon: Corrected

src/gui/src/main.cpp
Jnewbon: That's where the S went... :)

PR #6664, opened 2020-04-17
Limited the log in the GUI to 10,000 lines to prevent RAM hogging

### LINKED ISSUE #6663: Limit GUI log Window to prevent it using up too much RAM over time
### Problem ###
I'm running the laptop for 30 days in a month without a restart.
All the time is synergy running and waiting for connection or works as expected.
But log has around xxx thousands of lines. It is eating 1GB memory RAM now.
So restart of synergy is a solution but then I have to open the laptop, use keyboard and mouse there.

### Solution ###

Add "clean buffer" option to context menu when user press down right mouse button in the log area.

### More Info ###

At the moment there are two options: "Copy", "Select All".
It will be easy to add something what cleans the buffer list/vector/array whatever.

Jnewbon:
Fixes #6663

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -34,7 +34,7 @@ jobs:
       ubuntu18:
         image: symless/ubuntu-sudo:18.04
       ubuntu19:
-        image: symless/ubuntu-sudo:19.04
+        image: symless/ubuntu-sudo:19.10
 
   container: $[ variables['image'] ]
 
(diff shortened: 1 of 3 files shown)


Jnewbon:
Interesting, Ubuntu 19 build test fails as it cant download packages, with 404.

Jnewbon:
The reason U19 build failed is it using the 19.04 image which has bust been EOL'd so the repositories no longer have packages on them.

Issue #6674, opened 2020-04-28
Debian 10 X11 build for CI


nbolton:
### Problem ###

To create a build for Debian 10 X11 (#6689) we need to add it to the CI first.

### Solution ###

Add Debian 10 X11 to the CI.

PR #6685, opened 2020-05-08
Cursor freezes intermittently on Linux

### LINKED ISSUE #6487: Mouse freezes or key repeats on Linux client
### Operating Systems ###

Server: Windows 10 v1809
Client: Ubuntu 18.04

### Synergy Version ###

1.10.1

### Steps to reproduce bug ###

1. Open a Qt-based app on the client (Dolphin, Konsole, Kate)
2. Click-and-drag a file icon or tab
3. Watch as mouse freezes for 5-10 seconds

### Other info ###

* Seems to have started in 1.10 (or did not notice in Ubuntu 16.04 at least)
* Running KDE Plasma 5.12.7 and Qt 5.9.5
* GTK apps (Nautilus, etc) do not seem affected
* Occurs in both KDE and Gnome
* Client and server are connected via wired LAN (GigE)
* Disable Drag and Drop does not change behavior
* Disable shared clipboard does not change behavior
* Still an issue in Synergy 2
* No messages appear in either client or server log, even at DEBUG level
* Does not prevent me from using Synergy entirely, but it's added a lot of friction...
* Bug does not happen when the mouse is plugged into the client machine with Synergy still running.

Jnewbon:
Fixes #6487 

A fix in the xlib package's poll function fixed a bug where even if there where no events in the queue instead of waiting for the time out delay before returning, the function would instead return immediately. Synergy handled this by increasing the number of times the the loop would poll by a factor of 1000. 

So the 15ms delay was instead 15 seconds. An event normally broke the loop well before the 15 seconds was up but this cause noticeable freezes in the cursor as the thread was essentially locked up in that loop.

This PR removes the 1000x multiplayer reducing the loop down and solves the freezing issue.

### DIFF
--- a/src/lib/platform/XWindowsEventQueueBuffer.cpp
+++ b/src/lib/platform/XWindowsEventQueueBuffer.cpp
@@ -162,31 +162,31 @@ XWindowsEventQueueBuffer::waitForEvent(double dtimeout)
     // we want to give the cpu a chance s owe up this to 25ms
 #define TIMEOUT_DELAY 25
 
-    while (((dtimeout < 0.0) || (remaining > 0)) && QLength(m_display)==0 && retval==0){
+//    while (((dtimeout < 0.0) || (remaining > 0)) && QLength(m_display)==0 && retval==0){
 #if HAVE_POLL
-    retval = poll(pfds, 2, TIMEOUT_DELAY); //16ms = 60hz, but we make it > to play nicely with the cpu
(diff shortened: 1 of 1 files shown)


Jnewbon:
Ubuntu 16 build is failing, this may be fixed by changes brought in by #6674 that changes how the pipeline CI is built

Jnewbon:
As the Ubuntu 16 issue is build machine related, I'm going to approve this for merging.

Issue #6709, opened 2020-07-02
Ipc tests failing to build


Jnewbon:
### Operating Systems ###

Windows only

### Synergy Version ###

Since 83d15886

### Steps to Reproduce ###

1. Revert commit e70b1669
1. Build on windows
1. See error

### Log URLs ###

```
[ 96%] Building CXX object src/test/unittests/CMakeFiles/unittests.dir/ipc/IpcLogOutputterTests.cpp.obj
IpcLogOutputterTests.cpp
ext\googletest\googlemock\include\gmock/gmock-matchers.h(745): error C2338: Can't implicitly convert from <base> to <derived>
ext\googletest\googlemock\include\gmock/gmock-matchers.h(733): note: while compiling class template member function 'bool testing::internal::MatcherCastImpl<T,M>::Impl::MatchAndExplain(T,testing::MatchResultListener *) const'
        with
        [
            T=const IpcMessage &,
            M=testing::Matcher<const IpcLogLineMessage &>
        ]
ext\googletest\googlemock\include\gmock/gmock-matchers.h(723): note: see reference to class template instantiation 'testing::internal::MatcherCastImpl<T,M>::Impl' being compiled
        with
        [
            T=const IpcMessage &,
            M=testing::Matcher<const IpcLogLineMessage &>
        ]
ext\googletest\googlemock\include\gmock/gmock-matchers.h(722): note: while compiling class template member function 'testing::Matcher<A1> testing::internal::MatcherCastImpl<T,M>::Cast(const testing::Matcher<const IpcLogLineMessage &> &)'
        with
        [
            A1=const IpcMessage &,
            T=const IpcMessage &,
            M=testing::Matcher<const IpcLogLineMessage &>
        ]
ext\googletest\googlemock\include\gmock/gmock-matchers.h(784): note: see reference to function template instantiation 'testing::Matcher<A1> testing::internal::MatcherCastImpl<T,M>::Cast(const testing::Matcher<const IpcLogLineMessage &> &)' being compiled
        with
        [
            A1=const IpcMessage &,
            T=const IpcMessage &,
            M=testing::Matcher<const IpcLogLineMessage &>
        ]
ext\googletest\googlemock\include\gmock/gmock-matchers.h(784): note: see reference to class template instantiation 'testing::internal::MatcherCastImpl<T,M>' being compiled
        with
        [
            T=const IpcMessage &,
            M=testing::Matcher<const IpcLogLineMessage &>
        ]
[...]

PR #6710, opened 2020-07-02
Updated Gtest and Gmock to 1.8.1

Jnewbon:
Removed the old version of gtest and gmock
Submodule'd to google/googletest and set the version to 1.8.1

Ipc windows test fail to build so they have been disabled. See issue #6709


Jnewbon:
CI will fail due to submodules not being cloned, this will be fixed in #6682

PR #6738, opened 2020-07-27
Update git submodules during build

### LINKED ISSUE #6739: Add submodules cloning to the cmake
### Problem ###
Currently the user need to manually clone sub modules which can be confusing to new users especially with the current state of the documentation

### Solution ###

Add the cloning of sub modules to the cmake files

67hz:
Resolves #6739

Making submodule init/update transparent to the end user.

This change prevents CMake builds from failing if submodules have not been
installed, e.g., Google Test.

This relates to #6710.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,8 +56,10 @@ else()
     set (SYNERGY_ADD_HEADERS TRUE)
 endif()
 
+
+
 set (libs)
-include_directories (BEFORE SYSTEM ./ext/googletest/googletest/include)
+include_directories (BEFORE SYSTEM ${PROJECT_SOURCE_DIR}/ext/googletest/googletest/include)
 
 if (UNIX)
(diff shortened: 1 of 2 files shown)


67hz:
@Jnewbon  There is no ticket for this CMake update, however, it will prevent broken builds due to git submodules not being manually init'd. This brings the submodules as a proper dependency of the build since it currently requires them for success.

CI fails on the PR attempt due to ChangeLog not being updated... must be part of the new dev flow?

Jnewbon:
Only thing left is the ChangeLog update, once that's is done I'll be able to merge this PR in

67hz:
> Only thing left is the ChangeLog update, once that's is done I'll be able to merge this PR in

I am not familiar with the Symless changelog protocol. I pushed an update to the log following what I believe to be the convention. If it needs adjusting, let me know, and I'll update.

PR #6756, opened 2020-08-26
Automatically create issue in jira when submitted to Github

Jnewbon:
Will, require 3 secrets in the repo to be made
JIRA_BASE_URL
JIRA_USER_EMAIL
JIRA_API_TOKEN

PR #6759, opened 2020-09-01
Added job to add issue to jira automatically

Jnewbon:
Replaces #6756 due to sonar gate error


Jnewbon:
No changes to Source, or effect to the way source is built, This is an internal project management PR only so the changelog need not be updated.

PR #6762, opened 2020-09-08
SYNERGY-92 Fixed deamon leaving zombie process hanging about feeding on memory

### LINKED ISSUE #6567: Synergy spawns hundreds of zombie processes
### Operating Systems ###

Windows 10

### Synergy Version ###

<= 1.10.3

### Steps to Reproduce ###

1. Let synergy run for a couple hours
1. Run up RamMap tool
1. See hundreds of syntool.exe and synergy(c|s) processes

### Extra Info ###

image

Jnewbon:
Fixes #6567 
Captured all child process info as they are created, and closed handles when process are killed, Also added a close everything function to clear all all child process information when the deamon resets

Other info
Installed shotgun in deamon for zombie control

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,7 @@ v1.12.x-snapshot
 ===========
 Bug fixes:
 - #6753 Fixed a number of vulnerabilities detected by static analysis
+- #6567 Fixed Synergy spawning hundreds of zombie processes
 
 Enhancements:
 - #6750 Integrate SonarCloud for static analysis and test coverage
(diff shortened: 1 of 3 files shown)


Jnewbon:
Testing is in progress, so this will not be merged till i have confirmation for a few different sources that this resolves the issue for them

Jnewbon:
It has been pointed out that the handle created on this line is not closed

https://github.com/symless/synergy-core/blob/6def12bfbd0221e12707069f02d831f8ede016fb/src/lib/platform/MSWindowsWatchdog.cpp#L575

It should be closed either inside the shutdownProcess function or immediately after it is called

PR #6765, opened 2020-09-10
Fixed  account name with spaces causing TLS path errors

### LINKED ISSUE #6758: Account name with space on windows causes synergy to error when starting
### Operating Systems ###
Windows

### Synergy Version ###
Synergy 1.12.0-stable

### Steps to Reproduce ###
1. Have a user account with a space in the name eg "Jamie Newbon"
1. Set the TLS cert path in settings 
1. Attempt to start synergy

Expected: Synergy to start with the TLS path
Actual: Error starting synergy

### Extra info
The command line argument setting the TLS path doesn't have " around them thus spaces messes with the arguments sent to the child process

Jnewbon:
Resolves #6758

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,7 @@ v1.12.x-snapshot
 ===========
 Bug fixes:
 - #6753 Fixed a number of vulnerabilities detected by static analysis
+- #6758 Account name with space on windows causes synergy to error when starting
 
 Enhancements:
 - #6750 Integrate SonarCloud for static analysis and test coverage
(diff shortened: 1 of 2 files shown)


kmanwar89:
@SerhiiGadzhilov , thanks for linking my issue to this existing one. What is the fix?

SerhiiGadzhilov:
@kmanwar89 
The fix is to add quotes around TLS certificate path.
As workaround you can move the TLS certificate to a path not containing a space.
The fix will be available in the new version which is coming soon.

kmanwar89:
> 
> 
> @kmanwar89
> The fix is to add quotes around TLS certificate path.
> As workaround you can move the TLS certificate to a path not containing a space.
> The fix will be available in the new version which is coming soon.

Thanks!

PR #6792, opened 2020-09-25
SYNERGY-110 license key expiry

SerhiiGadzhilov:
License key expiry has been implemented.
I used the same algorithm as for trial version.
The new word "temp" can be used instead of "trial" in the same column as trial.
 Also purchase link has been corrected.

### DIFF
--- a/src/lib/shared/SerialKeyType.cpp
+++ b/src/lib/shared/SerialKeyType.cpp
@@ -1,0 +1,40 @@
+#include "SerialKeyType.h"
+
+const std::string SerialKeyType::TRIAL = "trial";
+const std::string SerialKeyType::TEMPORARY = "temp";
+
+SerialKeyType::SerialKeyType() :
+    m_isTrial(false),
+    m_isTemporary(false)
+{
+
+}
+
+void SerialKeyType::setKeyType(const std::string& Type)
+{
+    m_isTrial = false;
+    m_isTemporary = false;
+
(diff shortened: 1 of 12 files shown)


src/lib/shared/SerialKeyType.cpp
```diff
@@ -0,0 +1,60 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "SerialKeyType.h"
+
+const std::string SerialKeyType::TRIAL = "trial";
+const std::string SerialKeyType::TEMPORARY = "temp";
```
Andrea-Symless, line 21: Maybe we could rename "temp" to "subscription"

Jnewbon (approved):
I have reviewed and run and can confirm the expiry works

Only problem i can see with it is the program requires a restart for a currently active key to updated as expired. I'm happy with the code, @Andrea-Symless would need to approve the functionality is okay

Andrea-Symless:
> I have reviewed and run and can confirm the expiry works
> 
> Only problem i can see with it is the program requires a restart for a currently active key to updated as expired. I'm happy with the code, @Andrea-Symless would need to approve the functionality is okay

I'll approve this, and have created a new issue for the improvement you spotted: 
https://github.com/symless/synergy-core/issues/6798

Andrea-Symless (approved):

PR #6800, opened 2020-10-02
SYNERGY-324 Update behaviour when unregistered

SerhiiGadzhilov:


src/lib/shared/SerialKey.h
```diff
@@ -37,6 +37,7 @@ class SerialKey {
     bool                isExpired(time_t currentTime) const;
     bool                isTrial() const;
     bool                isTemporary() const;
+	bool                isValid() const;
     time_t                daysLeft(time_t currentTime) const;
     std::string            email() const;
```
Jnewbon, line 42: I think there might be a tab vs space problem. Could you set your IDE to use spaces(4) instead of tabs, otherwise it gets really nasty when viewing diffs in github specially when they become mixed like this

Jnewbon:
Minor style change, the code functionally looks good though

Jnewbon (approved):

PR #6803, opened 2020-10-06
SYNERGY-282 Update Synergy icons

SerhiiGadzhilov:


Jnewbon (changes requested):
The Icons in the 64x64 directory are only 22x22 big, Simon should be making the correct sized images for you, can you use those please as on mac the lower res image look very pixelated

Jnewbon (approved):

PR #6804, opened 2020-10-07
SYNERGY-322 Synergy loses edition when creating a System scope config

SerhiiGadzhilov:


Jnewbon (changes requested):
So i just ran this branch on my computer, and switching to System scope from user scope change me to unlicensed on Linux. I think the key/edition is being changed elsewhere in the code on scope change

Jnewbon (approved):
Cleaning out my system setting resolved the issue. And is working as expected

Jnewbon (approved):

PR #6811, opened 2020-10-13
SYNERGY-323 "No configuration available error" appears because the system fails on windows.

SerhiiGadzhilov:


Jnewbon:
The master branch now has CI fixes in could you update the branch when able to fix failing builds

andreacndn (approved):

PR #6821, opened 2020-10-30
SYNERGY-221 Blocker bugs

SerhiiGadzhilov:


src/lib/synergy/ProtocolUtil.cpp
```diff
@@ -17,6 +17,7 @@
  */
 
 #include <array>
+#include <iterator>
```
Jnewbon, line 20: Doesn't look like this is being used?

src/lib/synergy/ProtocolUtil.cpp
```diff
@@ -29,6 +30,57 @@
 // ProtocolUtil
 //
 
+namespace  {
+
+void
+writeInt(UInt32 Value, UInt32 Length, std::vector<UInt8>& Buffer)
```
Jnewbon, line 36: I like it, removing duplication is always good.

Jnewbon (commented):
Shame we couldn't get higher coverage, but i can see trying to get more would be difficult, and ROI would diminish too much.

Just the header comment and I'm good with this

src/lib/synergy/ProtocolUtil.cpp
SerhiiGadzhilov: It's necessary for std::back_inserter.
Windows compilation fails without this header.

src/lib/synergy/ProtocolUtil.cpp
```diff
@@ -29,6 +30,57 @@
...
+            break;
+        case 4:
+            Buffer.push_back(static_cast<UInt8>((Value >> 24) & 0xff));
+            Buffer.push_back(static_cast<UInt8>((Value >> 16) & 0xff));
+            Buffer.push_back(static_cast<UInt8>((Value >>  8) & 0xff));
+            Buffer.push_back(static_cast<UInt8>( Value        & 0xff));
+            break;
+        case 2:
+            Buffer.push_back(static_cast<UInt8>((Value >> 8) & 0xff));
+            Buffer.push_back(static_cast<UInt8>( Value       & 0xff));
```
Jnewbon, line 51: Looking im my IDE a Clang tidy error happens here. 
> Use of a signed integer operand with a binary bitwise operator
Wondering if we should update the numbers and hex to specifically be unsigned for example 
```suggestion
            Buffer.push_back(static_cast<UInt8>(Value & 0xffu));
            break;
        case 4:
            Buffer.push_back(static_cast<UInt8>((Value >> 24u) & 0xffu));
            Buffer.push_back(static_cast<UInt8>((Value >> 16u) & 0xffu));
            Buffer.push_back(static_cast<UInt8>((Value >>  8u) & 0xffu));
            Buffer.push_back(static_cast<UInt8>( Value        & 0xffu));
            break;
        case 2:
            Buffer.push_back(static_cast<UInt8>((Value >> 8u) & 0xffu));
            Buffer.push_back(static_cast<UInt8>( Value       & 0xffu));
```
As described here https://stackoverflow.com/a/50399245/11680140

Jnewbon (approved):

PR #6826, opened 2020-11-03
SonarCloud Critical bugs

ignacionr:
Fixes structural oddities marked "Critical" by Sonar Cloud (I don't think any of them were actually bugs, but they could become bugs in the future). My take was in most cases, to mark the offending constructors as ```=delete``` so to prevent the compiler from generating any.
In two occasions, such refactoring did show the use of such implementations. Only in one case, it was a somewhat dangerous construct (but which could have generated issues only at program shutdown).


ignacionr:
@Jnewbon please check out ``` src/lib/synergy/ServerApp.cpp ``` where I took the liberty of adding a shared smart pointer. In the long run this should probably be better expressed by moving a unique pointer, but didn't want to go for a bigger change surface at this stage.

Jnewbon (commented):
The comment I mentioned I might have isn't a problem, I was thinking of another object that gets passed around a lot, the shared pointer you made, is only accessed rather then passed about. so it seems fine to me.

Only thing i would watch for is committing the .vscode files. i have always been of the opinion that IDE specific files shouldn't but committed as they are mostly user preferences. So maybe we should add it to the .gitignore.

ignacionr:
> Only thing i would watch for is committing the .vscode files. i have always been of the opinion that IDE specific files shouldn't but committed as they are mostly user preferences. So maybe we should add it to the .gitignore.

Yes; I found that configuration for things such as the Google Test adapter isn't trivial, and that other items such as vocabulary (for spell checkers) are useful to be shared. On the other hand, this product has survived more than an IDE, and it makes sense to keep those separate. Will remove and ignore .vstudio configuration, maybe come up with a GIST or something we can share in case it helps.

Jnewbon (approved):

PR #6828, opened 2020-11-03
SYNERGY-362 Eliminate using std::chrono 

SerhiiGadzhilov:
Unfortunately we are not able to use std::chrono due to compilation errors on: centos7.6, ubuntu1604, debian9


ignacionr (approved):
Thank you for using the unified initialisation syntax! You can see that the code reads quite differently with it. We should still keep tabs on the point we may want to start abandoning time_t (hopefully before 2038); I think your choice of std::chrono was sound.

PR #6829, opened 2020-11-04
fix for older compilers

ignacionr:
It seems that compilers such as the one shipped with CentOS 7.6 can implement ```auto``` rather strangely; avoiding the issue by not using auto on this new test.


ignacionr:
The original problem can be reproduced (before applying this PR) by building the following [Dockerfile.zip](https://github.com/symless/synergy-core/files/5488837/Dockerfile.zip)

ignacionr:
More info: this is the proposal (accepted for C++17) that the compiler is not abiding to: [N3922](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n3922.html) .

SerhiiGadzhilov (approved):

PR #6835, opened 2020-11-09
SYNEGRY-549 Create TLS certificate during upgrade

SerhiiGadzhilov:


andreacndn:
Awesome, could you please leave a message with comments on this in the issues?
I guess we’ll still need to fix the UI big as suggested?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

SerhiiGadzhilov:
The coverage is 0 because all changes were done in gui code which doesn't have infrastructure for testing now.

src/gui/src/AppConfig.cpp
```diff
@@ -414,8 +415,10 @@ ElevateMode AppConfig::elevateMode()
 }
 
 void AppConfig::setCryptoEnabled(bool newValue) {
+    if (m_CryptoEnabled != newValue && newValue){
+        generateCertificate();
```
ignacionr, line 419: I don't pretend that you should come up with an architectural change at this point. But I think you must be as unhappy as I am to see the code that works out business rules, so tightly mixed with the GUI and ultimately dependent on Qt. At this point, let us just make sure we have an item in the backlog to "make business rules independent from a Qt modeled GUI."

src/gui/src/SettingsDialog.cpp
```diff
@@ -66,19 +66,10 @@ void SettingsDialog::accept()
...
     appConfig().setMinimizeToTray(m_pCheckBoxMinimizeToTray->isChecked());
     appConfig().setTLSCertPath(m_pLineEditCertificatePath->text());
-
-    bool keyLengthChanged = appConfig().getTLSKeyLength() != m_pComboBoxKeyLength->currentText();
     appConfig().setTLSKeyLength(m_pComboBoxKeyLength->currentText());
 
     //We only need to test the System scoped Radio as they are connected
     appConfig().setLoadFromSystemScope(m_pRadioSystemScope->isChecked());
-
-    if(m_pCheckBoxEnableCrypto->isChecked()) {
```
ignacionr, line 76: Shouldn't we just spawn a configuration change (App Config) and that in turn run this kind of checks? I am concerned that we may end up re-creating the certificate twice in any case. Also applies the same concern as before (GUI code running the domain rules logic from within).

src/gui/src/SettingsDialog.cpp
```diff
@@ -265,15 +253,6 @@ void SettingsDialog::on_m_pPushButtonBrowseCert_clicked() {
     updateRegenButton();
 }
 
-void SettingsDialog::regenerateSSLCert() {
```
ignacionr, line 268: I would be kind of less unhappy to have the SSL generation code at the AppConfig object, even if it's not ideal (still tied to Qt and ultimately to GUI) it seems more abstracted than directly on the dialog. But there could be a reason why you moved it here, kindly share?

ignacionr (changes requested):
The code works. I am a bit worried about its organisation. We know GUI isn't Synergy 1's strongest suit, so all domain logic needs to be ready to depart on short notice. If there's a good reason to let the dialog take care of SSL generation, please share.
But again, the code does the generation, it does the override when needed, so the most important part, is great work.

src/gui/src/AppConfig.cpp
SerhiiGadzhilov: Yes, I'm not happy as you but I'm not sure that we need such wide task as "make business rules independent from a Qt modeled GUI."
Tasks like this make code and architecture better but they don't provide value for the end users.

I think about such tasks as about revolution which can break something and can provide a lot af benefits in long term.
I like evolution way where we move to good design through small steps and design is constructed according to current users needs.

I think it's a topiс for discussion and the truth is somewhere in between.

src/gui/src/SettingsDialog.cpp
SerhiiGadzhilov: Actually this is what I did.
I removed checks from SettingsDialog and added them into AppConfig::setCryptoEnabled.
In SettingsDialog we just call AppConfig::setCryptoEnabled and this method does checks and generate certificate if it's needed.

src/gui/src/SettingsDialog.cpp
SerhiiGadzhilov: Actually this is exactly what I did.
I removed TLS certificate code generation from SettingsDialog and added a new method "AppConfig::generateCertificate".
Now SettingsDialog just calls appropriate AppConfig methods as result of different GUI events.

src/gui/src/SettingsDialog.cpp
ignacionr: I will take a note to myself: make sure you read chages on the correct side! :D :1st_place_medal: Thank you.

src/gui/src/SettingsDialog.cpp
ignacionr: :1st_place_medal: thank you; this is the best possible approach right now

src/gui/src/AppConfig.cpp
ignacionr: Agreed that it's an interesting topic for a discussion that doesn't belong in this PR. Only let's have this conversation at some point. I'm sure you too feel the need for the unit tests that are amiss here, and they would be much easier to come up with, should there be more separation between the UI and everything else. Again, more using this PR as food for thoughts than anything else.

By the way, now that I see it in the right way (that is what you actually did) I see that you're using the excuse of solving things like this, to add quality in general. :+1:

ignacionr (approved):

PR #6837, opened 2020-11-10
SYNERGY-148 Memory leaks on macOS

SerhiiGadzhilov:


SerhiiGadzhilov:
The following memory leaks have been fixed:
1. MainWindow was changed to use QTemporaryFile as local variable.
2. Every object which acquired using Core Foundation function which contains the word "Create" or "Copy" should be released using CFRelease.

I tried to create tests but unfortunately I failed with this. 
This functionality interacts with system and current implementation doesn't allow to test it.

ignacionr (approved):
Beautiful job. Not only do I agree on the usage of smart pointers with a release parameter, but love the style corrections too (e.g. casting pointers down to boolean instead of comparing to constant).

PR #6844, opened 2020-11-19
SYNERGY-551 update macOS installer.

SerhiiGadzhilov:


SerhiiGadzhilov:
SonarCloud Quality Gate fails because of the new file "installator_settings.py"
This file contains only settings for "dmgbuild" and should be excluded from coverage calculation.
Looks like I don't have permissions to do it myself.

Additionally I fixed GitHub action for Windows build.

Jnewbon (approved):

PR #6872, opened 2020-12-10
Check for dns_sd.h only when building GUI

mgorny:
The <dns_sd.h> header is only used in the synergy GUI component.  Update
the CMake check to be done only if SYNERGY_BUILD_LEGACY_GUI is enabled.
This fixes build failures when GUI is not being built and <dns_sd.h>
is not installed.

While at it, fix indentation around the added check to match the rest
of the file and avoid being misleading.


andreacndn:
Hi @mgorny , thank you for your contribution!

SerhiiGadzhilov (approved):

PR #6882, opened 2020-12-17
SYNERGY-652 Avoid using brew for macOS build

SerhiiGadzhilov:


ignacionr (approved):
Looking good! Python3 comes installed?

SerhiiGadzhilov:
@ignacionr 
Yes. There is Python 3.9.1. Here is a link with installed software: https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md

Also there is warning message about python2. 
This version will not be supported after 01.01.2021 hence I used python3.

PR #6889, opened 2020-12-28
Ubuntu Tray Icon at Start Up

ignacionr:
Refactored Tray Icon, changed verification to ensure new connection to DBus at check.


src/gui/src/MainWindow.cpp
```diff
@@ -228,7 +224,20 @@ MainWindow::~MainWindow()
 
 void MainWindow::open()
 {
-    createTrayIcon();
+    QAction *trayMenu[] = {
```
SerhiiGadzhilov, line 227: Do we really need C-style array?
Isn't it better to use std::vector?

src/gui/src/TrayIcon.h
```diff
@@ -0,0 +1,45 @@
...
+{
+    Q_OBJECT
+public:
+    using TConnector = std::function<void(QObject *, const char *)>;
+
+    template<typename TActionIterator>
+    void create(TActionIterator *begin, TActionIterator *end, TConnector connector) 
+    {
+        m_connector = connector;
+        m_pTrayIconMenu = std::make_unique<QMenu>();
```
SerhiiGadzhilov, line 19: As I can see before these changes QMenu had parent MainWindow: m_pTrayIconMenu = new QMenu(this);
Is it ok that QMenu is created without parent?

src/gui/src/TrayIcon.h
```diff
@@ -0,0 +1,45 @@
...
+#include <memory>
+
+class TrayIcon : public QObject
+{
+    Q_OBJECT
+public:
+    using TConnector = std::function<void(QObject *, const char *)>;
+
+    template<typename TActionIterator>
+    void create(TActionIterator *begin, TActionIterator *end, TConnector connector)
```
SerhiiGadzhilov, line 16: Maybe it's better to use const std::vector& instead of TActionIterator *begin, TActionIterator *end

src/gui/src/TrayIcon.cpp
```diff
@@ -0,0 +1,19 @@
+#include "TrayIcon.h"
+#include <iostream>
```
SerhiiGadzhilov, line 2: Looks like <iostream> is not needed here?

src/gui/src/TrayIcon.h
ignacionr: Sorry I don't agree. The point of using abstract iterators is to allow for any shape of container (including base arrays, std::vector, std::list, etc).

src/gui/src/TrayIcon.h
ignacionr: Yes it is okay. The difference is lifespan. When you use the parent parameter, the menu will be destroyed by Qt when the parent fades away. My design rather uses unique_ptr to take care of that in an easier fashion, and will eventually let the code be increasingly less dependent on Qt.

src/gui/src/MainWindow.cpp
ignacionr: An alternative could be std::array (std::vector would be needed for dynamically allocations, which is not needed).
But since this code is better read and understood directly, I prefer the most basic representation.

src/gui/src/TrayIcon.cpp
ignacionr: Correct! I will remove.

src/gui/src/MainWindow.cpp
ignacionr: I will take your suggestion of using an STL container, but opt for std::array because there is no need for dynamic allocations.

src/gui/src/TrayIcon.h
ignacionr: Since I'm taking the first suggestion to avoid basic arrays, then basic arrays are out of the picture, then a container could be a better argument. This should also suit your request.

SerhiiGadzhilov (changes requested):
Could you please merge ChangeLog from master and update it.
This task should be inside section "v1.13.1-snapshot"

ignacionr:
> Could you please merge ChangeLog from master and update it.
> This task should be inside section "v1.13.1-snapshot"

This was done. Please review.

SerhiiGadzhilov (approved):

SerhiiGadzhilov:
Centos7.6 build failed with the following error: 
```
/__w/1/s/src/gui/src/TrayIcon.h:19:27: error: 'make_unique' is not a member of 'std'
         m_pTrayIconMenu = std::make_unique<QMenu>();
```
Could you please, check the issue.

ignacionr:
> @ignacionr Centos7.6 build failed with the following error:
> 
> ```
> /__w/1/s/src/gui/src/TrayIcon.h:19:27: error: 'make_unique' is not a member of 'std'
>          m_pTrayIconMenu = std::make_unique<QMenu>();
> ```
> 
> Could you please, check the issue.

@SerhiiGadzhilov the issue seems to be that the installed compiler isn't even supporting C++14. Is it possible that the image has a very outdated compiler installed?

andreacndn:
> > @ignacionr Centos7.6 build failed with the following error:
> > ```
> > /__w/1/s/src/gui/src/TrayIcon.h:19:27: error: 'make_unique' is not a member of 'std'
> >          m_pTrayIconMenu = std::make_unique<QMenu>();
> > ```
> > 
> > 
> > Could you please, check the issue.
> 
> @SerhiiGadzhilov the issue seems to be that the installed compiler isn't even supporting C++14. Is it possible that the image has a very outdated compiler installed?

Ignacio, could you please tomorrow update the compiler on this image?

ignacionr:
@andreacndn replied over Slack

Issue #6862, opened 2020-12-07
nothing gets typed on client if server has different keyboard layout


bellerofonte:
Hi 
I'm using Synergy v1.12.0-stable on my Mac (server) and Ubuntu (client). 
On Mac I use 2 keyboard layouts: en-US and ru-RU, on Ubuntu - en-US only. 
So when I type smth on Mac using ru-RU layout and then switch to Ubuntu and try to type smth where - nothing is typed. 
I have to move my mouse to Mac, switch layout there and move mouse to the client again. Too inconvenient.

Is it possible to send to client always en-US symbols/keycodes, even if server has another layout? 
I mean server[ru-RU] -> client[en-US]; server[en-US] -> client[en-US]

Or maybe there is a way to remap cyrillic keycodes/keynames to latin ones in config file?

bellerofonte:
recently made PR #6890

ignacionr:
@bellerofonte congratulations on your first PR! Also, your opening an issue and going on and creating a PR for it, is the kind of attitude that makes Open Source carry great value. As explained on the PR comments, I don't believe we can merge as is, but as a general idea it can certainly be considered. Thank you and please keep in touch! This issue will remain open for the time being.

PR #6890, opened 2020-12-28
Fixed Cyrillic layout issue

bellerofonte:
this PR is for #6862 

I've investigated the issue and came to following results (_macOS runs Synergy server, Linux runs synergy client_):

1. If macOS's keyboard layout is EN(US), **any** layout on Linux works pretty well
2. If macOS's keyboard layout is RU(RU) and Linux **hasn't** RU layout added - **nothing** gets typed
3. If macOS's keyboard layout is RU(RU) and Linux **hasn't** RU layout added - Cyrillic characters get **always** typed **whatever** layout is chosen on Linux.

So I decide to remap every incoming `KeyID` on client side using hardcoded table. I.e. I tried to translate any incoming `Cyrillic unicode key id` into `Basic Latin unicode key id` using may Macbook's _[physical]_ keyboard layout.
This still works not ideal, but **much** more comfortable to use.

### DIFF
--- a/src/lib/synergy/KeyState.h
+++ b/src/lib/synergy/KeyState.h
@@ -196,7 +196,19 @@ private:
     // active modifiers collection callback
     static void            addActiveModifierCB(KeyID id, SInt32 group,
                             synergy::KeyMap::KeyItem& keyItem, void* vcontext);
-
+    
+    // region 'Edited by bellerofonte'
+    using KeyRemap = std::map<KeyID, KeyID>;
+    
+    void remapKeyID(KeyID &id);
+    bool remapKeyIDCyrillic(KeyID &id);
+    /*
+    bool remapKeyIDArabic(KeyID &id);
+    bool remapKeyIDEtc(KeyID &id);
+    */
+    
+    // end region
+    
 private:
     // must be declared before m_keyMap. used when this class owns the key map.
     synergy::KeyMap*            m_keyMapPtr;
--- a/src/lib/synergy/KeyState.cpp
+++ b/src/lib/synergy/KeyState.cpp
@@ -532,6 +532,98 @@ KeyState::addActiveModifierCB(KeyID, SInt32 group,
     }    
 }
 
+// region 'Edited by bellerofonte'
+void 
+KeyState::remapKeyID(KeyID &id)
+{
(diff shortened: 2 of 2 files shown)


ignacionr:
Здравствуйте @bellerofonte  !

First let me say thank you for actively sharing your take on Synergy, this encourages us a great deal.
Secondly, let me empathise with you as a daily user of both EN and RU mappings.

Finally, even though we will surely take into account this approach, I guess you understand it and it may not even need said but the PR cannot be merged as is, and would need a lot of work to get to a point where it can reach our vast user base (re-mapping would need to be an opt-in or auto-assessed element, it would need to cope not only with other languages but even within Russian, it would require the standard, the PC, and the АБВ layouts).

Please remain in contact, and do consider we are hiring.

bellerofonte:
Hi @ignacionr!

Thank you for your response. 
> I guess you understand it and it may not even need said but the PR cannot be merged as is
Sure, I just wanted you to note this approach. I'm closing this PR, but I will continue to improve this branch for my personal environment. You are welcome to use my work as soon as you need it.

nbolton:
Hi @bellerofonte, just noticed this PR was closed without the work being used. I'm reopening this in the hope that we can perhaps use your contribution.

sithlord48:
This should be rechecked to see if its still a vaild issue now that we should have proper utf16 pairs.

src/lib/synergy/KeyState.h
```diff
@@ -196,7 +196,19 @@ class KeyState : public IKeyState {
...
+    
+    // region 'Edited by bellerofonte'
+    using KeyRemap = std::map<KeyID, KeyID>;
+    
+    void remapKeyID(KeyID &id);
+    bool remapKeyIDCyrillic(KeyID &id);
+    /*
+    bool remapKeyIDArabic(KeyID &id);
+    bool remapKeyIDEtc(KeyID &id);
+    */
```
nbolton, line 208: Commented out code needs removing.

nbolton (changes requested):
Are you able to fix these merge conflicts?

nbolton:
> Are you able to fix these merge conflicts?

Unfortunately, seems like another PR we don't have bandwidth to fix :disappointed:

PR #6900, opened 2021-01-06
Remaining SonarCloud Bugs

ignacionr:
Synergy-591

This PR addresses all 25 bugs pointed out by SonarCloud. Most of them are related to the use of for constructs (styles that are prone to errors); one does unveil a potential memory leak on BSD.

All sections addressed have never before been visited by tests, thus the big cost and complexity that the tests related to this PR will have -but there's also its value right there.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,7 +13,7 @@ Bug fixes:
 - #6878 Update googletest library to latest version
 - #6882 Build macOS using qt 5.15.2 and avoid using brew for build
 - #6884 Fix logo for Linux and dark logo for macOS
-- #6886 Upadte Welcome screen
+- #6886 Update Welcome screen
 
 Enhancements:
 - #6867 Create only 1 artefact for macOS
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 v1.13.1-snapshot
 ===========
-===========
+Bug fixes:
+- #6900 Remaining SonarCloud reported bug items
 
 v1.13.0-stable
 ===========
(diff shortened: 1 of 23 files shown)


SerhiiGadzhilov (approved):

SerhiiGadzhilov:
@ignacionr  Looks like something was broken under this PR!
I just installed fresh build (https://snapshots.symless.com/v1-core-standard/1.13.1/snapshot/b4-5c2311a5/) and it hangs at the start. Application start takes 3 minutes on my macOS Big Sure 11.1.
There is no such problem with the previous build: https://snapshots.symless.com/v1-core-standard/1.13.1/snapshot/b3-c06cdda1/

SerhiiGadzhilov:
@ignacionr 
Please, check also the following test which is failed on Windows:
```
[ RUN      ] ServerConfigTests.serverconfig_will_deem_equal_configs_with_same_cell_names
D:\a\synergy-core\synergy-core\src\test\unittests\server\ConfigTests.cpp(79): error: Value of: a == b
  Actual: false
Expected: true
D:\a\synergy-core\synergy-core\src\test\unittests\server\ConfigTests.cpp(80): error: Value of: b == a
  Actual: false
Expected: true
[  FAILED  ] ServerConfigTests.serverconfig_will_deem_equal_configs_with_same_cell_names (0 ms)
```

ignacionr:
> @ignacionr There are infinite loops in the following functions:
> 
> 1. ArgParser::parseServerArgs
> 2. ArgParser::parseClientArgs
> 
> Could you please, correct them.

No, not really. These are not infinite loops. The code is unfortunately following a very poor style, but the loops are not infinite.
The functions called by them, are guaranteed to move the index, which is passed by reference.

PR #6905, opened 2021-01-11
PR changes 6889 and resolution for CentOS 7.6

ignacionr:
Removing C++14 constructs for compatibility on CentOS7.6.

Jira issue Synergy-734.


andreacndn:
> Removing C++14 constructs for compatibility on CentOS7.6.
I very much disagree on this. We should fix the CentOS image instead of blocking the update to more modern C++

PR #6906, opened 2021-01-12
Modifying the Azure Pipeline so that CentOS 7.6 gets additional required commands

ignacionr:
For building on CentOS 7.6, we need to enable the development tools newer version, each time we issue a command.


CI/Linux/linux-build.yml
```diff
@@ -14,6 +14,19 @@ steps:
...
+      cd build-release
+      yum install -y centos-release-scl-rh
+      yum install -y devtoolset-9-gcc-c++
+      scl enable devtoolset-9 'cmake3 -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX:PATH=$(pwd)/rpm/BUILDROOT/usr ..;'
+      . ./version
+      scl enable devtoolset-9 'make -j'
+    condition: eq(variables['packager'],'centos76-rpm')
+    displayName: 'Build Release rpm'
+    env:
+      GIT_COMMIT: $(Build.SourceVersion)
```
SerhiiGadzhilov, line 28: As I can see the “devtoolset” was added only for the “Standard” version build. 
Maybe the “Enterprise” build should be updated as well?
Isn't it better to include the "devtoolset" into docker image for CentOS 7.6?

SerhiiGadzhilov (changes requested):

CI/Linux/linux-build.yml
andreacndn: Agreed that the requirement was to update the CentOS docker image

andreacndn (changes requested):
Agreed that the requirement was to update the CentOS docker image

SerhiiGadzhilov:
Fixed under #6937

PR #6907, opened 2021-01-13
Save configuration on apply

ignacionr:
As discussed.


SerhiiGadzhilov (commented):
I'm not sure if I understand correctly how it should work but I think current implementation works incorrectly. 
Please, review the following scenarios:
##############
**Scenario 1:**
_**Step 1**_
Open settings dialog.
Don’t make any changes and click “OK”
Quit synergy.
_**Expected:**_
The system doesn't show saving settings dialog.
_**Actual:**_
The system shows saving settings dialog. 
Why? I didn’t make any changes!
###############
**Scenario 2:**
_**Step 1**_
Open settings dialog.
Change any setting and click “OK”
Quit synergy.
**_Expected:_**
The system doesn't show saving settings dialog.
(I expect that saving settings dialog is not shown because the system should save them when the user clicks “OK”. Maybe it’s ok because it’s old behavior but in this case something wrong with the next step)
_**Actual:**_
The system shows settings saving dialog.

_**Step 2**_
Click cancel.
Re-open synergy settings
_**Expected:**_
My modifications from step1 are not saved because I clicked “Cancel”.
_**Actual:**_
My modifications from step 1 are saved. Why does it ask me if I want to save changes and then ignore my choice?

SerhiiGadzhilov:
CodeQL checking fails with the following error:
_Preparing to unpack .../24-msodbcsql17_17.7.1.1-1_amd64.deb ...
ERROR: The EULA was not accepted. Installation aborted._

This error doesn't relate to current task and should be fixed under separate task because I don't know quick solution.

andreacndn (approved):

PR #6914, opened 2021-01-18
Duplicating text spaces after dot only for Mac

ignacionr:


src/gui/src/main.cpp
```diff
@@ -97,10 +97,9 @@ int main(int argc, char* argv[])
     QObject::connect(dynamic_cast<QObject*>(&app), SIGNAL(aboutToQuit()),
                 &mainWindow, SLOT(saveSettings()));
 
-    SetupWizard setupWizard(mainWindow, true);
-
     if (appConfig.wizardShouldRun())
     {
+        SetupWizard setupWizard(mainWindow, true);
         setupWizard.show();
```
SerhiiGadzhilov, line 103: setupWizard is destructed before app.exec() as result we will never see Setup Wizard window.

SerhiiGadzhilov (changes requested):

src/gui/src/main.cpp
ignacionr: @SerhiiGadzhilov I truly don't understand what you mean. The setupWizard will only be shown within the if condition where it is now also created. There is no other condition that will show it, and it is thus not needed. Please recheck.

ignacionr: Great catch! The new approach is equivalent to the original, but improves avoiding construction if unneeded.

SerhiiGadzhilov (approved):

PR #6915, opened 2021-01-18
Synergy 591 remaining sonarcloud bugs in synergy core

ignacionr:


SerhiiGadzhilov (changes requested):
There are infinite loops in the following functions:
1. ArgParser::parseServerArgs
2. ArgParser::parseClientArgs

We have the following “continues” instructions inside the loops:
```
if (parsePlatformArg(args, argc, argv, i)) {
           continue;
       }
       else if (parseGenericArgs(argc, argv, i)) {
           continue;
       }
       else if (parseDeprecatedArgs(argc, argv, i)) {
           continue;
       }
```
Even if these functions increment “i” inside the logic was changed because old code did increment as the last instruction inside “for”. Also, these functions not always increment “i”. For example option “--help” or “--version”:
```
else if (isArg(i, argc, argv, "-h", "--help")) {
       if (m_app) {
           m_app->help();
       }
       argsBase().m_shouldExit = true;
   }
```
 For the test just run “synergyc --help” from the console and you will see infinite help.

_**My opinion how it should be fixed:**_

**_Function: ArgParser::parseServerArgs_**
I don’t see any reason to change `for (int i = 1; i < argc; ++i)` to `while` hence I think these changes should be reverted.

_**Function: ArgParser::parseClientArgs**_
There is only one reason to change and it’s the following code outside the loop:
```
if (i == argc) {
       LOG((CLOG_PRINT "%s: a server address or name is required" BYE,
           args.m_pname, args.m_pname));
       return false;
   }
```
I think that this code should be removed because server address is not always required (for example for options --version and --help).
In this case we can use `for (int i = 1; i < argc; ++i)` as for ArgParser::parseServerArgs

ignacionr:
> **_Function: ArgParser::parseServerArgs_**
> I don’t see any reason to change `for (int i = 1; i < argc; ++i)` to `while` hence I think these changes should be reverted.

My friend, you should read SonarCloud's report; the reason is very interesting and valid. Let me explain.

For statements are nothing short of beautiful. The greatest part of them, is the fact that they feature the three most important parts of any loop, on a single instruction: initialisation, invariant, increment.

The problem identified by SonarCloud, is that in this case the "increment" was offset in other parts of the code. So the "clean" analysis of the three parts of a for statement, becomes deceiving.

This is the reason for changing it to a while (a while statement does not claim/imply that the increment or initialisation are accounted for in a particular place).

ignacionr:
> Even if these functions increment “i” inside the logic was changed because old code did increment as the last instruction inside “for”. Also, these functions not always increment “i”. For example option “--help” or “--version”

The increment that originally was on the for() was moved to line https://github.com/symless/synergy-core/blob/master/src/lib/synergy/ArgParser.cpp#L117 so the code is formally equivalent. There was no infinite loop, as there still isn't one.

ignacionr:
> Even if these functions increment “i” inside the logic was changed because old code did increment as the last instruction inside “for”. Also, these functions not always increment “i”. For example option “--help” or “--version”:

I understand this now. It is a valid concern; the semantics of the original continue, changed in the translation from for to while. I will modify and create corresponding tests.

ignacionr:
@SerhiiGadzhilov please review this one with priority. Thank you.

SerhiiGadzhilov (approved):

PR #6920, opened 2021-01-21
Change OpenSSL paths

aquacash5:
In the Azure pipelines CI for windows. We relied on the openssl.exe dep in the ext directory. I moved it and now I'm fixing the CI script.


ignacionr (approved):
Let us test by merging.

PR #6921, opened 2021-01-21
Handling pre-main window status notifications

ignacionr:
SYNERGY-767


src/gui/src/TrayIcon.h
```diff
@@ -30,16 +39,22 @@ class TrayIcon : public QObject
         m_pTrayIcon = std::make_unique<QSystemTrayIcon>();
         m_pTrayIcon->setContextMenu(m_pTrayIconMenu.get());
         m_pTrayIcon->setToolTip("Synergy");
+        m_set = [this](const QIcon& icon) { m_pTrayIcon->setIcon(icon); };
 
         tryCreate();
+        if (m_init) {
+            m_init();
+            m_init = std::function<void()>();
```
ignacionr, line 47: This is for releasing the icon, trapped by the lambda class.

ignacionr:
The general idea is to keep it branchless (i.e. no *if* for deciding whether to save the icon, or execute the change).

ignacionr:
My first approach simply replaced the original method *set* with a public field, but SonarCloud considered it a code smell (after all, a functor is still data, so I took the suggestion since it's also very likely that the compiler will inline the function and thus create the same machine code).

SerhiiGadzhilov (approved):

PR #6925, opened 2021-01-22
Barrier.exe can be opened only once

Almafozelek:
Hello! 

This is also a problem in linux. Windows version deals with it, but on linux things get laggy, when it opened multiple times

Képkivágás (2)


andreacndn:
Hi @Almafozelek I guess there is some confusion, this is the open source repository for Synergy (the original product) and not for Barrier (the fork of an old version of Synergy).

PR #6928, opened 2021-01-27
SYNERGY-722 Remove information about drag and drop from Log

SerhiiGadzhilov:


SerhiiGadzhilov:
We don't need to update the ChangeLog because this is additional change for SYNERGY-722 and the ChangeLog was updated under previous PR.

andreacndn (approved):

PR #6951, opened 2021-02-18
SYNERGY-833 Add extra enterprise validation

aquacash5:
Adds an extra layer of validation on whether to build the enterprise version of the application. This makes creating workflows easier compared to having custom conditional logic in multiple places of the workflow in order to set or not set the environment variable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,7 @@ endif()
 option (BUILD_TESTS "Override building of tests" ON)
 option (ENABLE_COVERAGE "Build with coverage")
 
-if (DEFINED ENV{SYNERGY_ENTERPRISE} AND NOT "$ENV{SYNERGY_ENTERPRISE}" STREQUAL "")
+if ($ENV{SYNERGY_ENTERPRISE})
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ endif()
 option (BUILD_TESTS "Override building of tests" ON)
 option (ENABLE_COVERAGE "Build with coverage")
 
-if (DEFINED ENV{SYNERGY_ENTERPRISE})
+if (DEFINED ENV{SYNERGY_ENTERPRISE} AND NOT "$ENV{SYNERGY_ENTERPRISE}" STREQUAL "")
```
SerhiiGadzhilov, line 32: Looks like this condition can be rewritten as `if ($ENV{SYNERGY_ENTERPRISE})` with the same result.

SerhiiGadzhilov (changes requested):

CMakeLists.txt
aquacash5: This looks way simpler.

SerhiiGadzhilov (approved):

PR #6971, opened 2021-03-30
SYNERGY 512 SonarCloud vulnerabilities in Synergy-Core

abatyiev:


src/lib/platform/XWindowsClipboard.cpp
```diff
@@ -1083,68 +1079,75 @@ XWindowsClipboard::sendReply(Reply* reply)
...
+                const String hex_digits = "0123456789abcdef";
+                String tmp;
+                tmp.reserve(data.length() * 3);
+                std::for_each(data.begin(), data.end(), [hex_digits, &tmp](const unsigned char& c)
+                    {
+                        tmp += hex_digits[c >> 16];
+                        tmp += hex_digits[c & 15];
+                        tmp += ' ';
+                    });
+                data = tmp;
```
SerhiiGadzhilov, line 1132: This code looks ok but looks like it can be easily moved to separate function with one parameter to reduce the complexity.

src/lib/platform/XWindowsClipboard.cpp
```diff
@@ -1083,68 +1079,75 @@ XWindowsClipboard::sendReply(Reply* reply)
...
             }
-            if (props != NULL) {
-                XFree(props);
+            char* type = XGetAtomName(m_display, target);
+            LOG((CLOG_DEBUG2 "  %s (%s): %s", name, type, data.c_str()));
+            if (type != nullptr) {
+                XFree(type);
```
SerhiiGadzhilov, line 1137: This code looks ok but it's better to use smart pointers like std::unique_ptr in such cases to comply with RAII.

SerhiiGadzhilov (changes requested):

src/gui/src/QUtility.cpp
```diff
@@ -97,17 +98,14 @@ QString getOSInformation()
...
-        QString output = cp.run();
 
+    QFile paramFile("/etc/os-release");
+    if(paramFile.open(QFile::ReadOnly | QFile::Text)) {
+        const QString fileStrings = QTextStream(&paramFile).readAll();
         QRegExp resultRegex(".*PRETTY_NAME=\"([^\"]+)\".*");
-        if (resultRegex.exactMatch(output)) {
+        if (resultRegex.exactMatch(fileStrings)) {
             result = resultRegex.cap(1);
         }
```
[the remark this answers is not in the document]
abatyiev: You are right, but I'm not sure about all functional of synergy application, so the main idea of all my first changes - do not change logic, just fixes. In future, with all settled up tests(including different OS and connections) and with experience of usage, I'll be more brave in deleting or changing logic. For now, after your comment I'm free to delete this code block, thanks.

src/lib/net/TCPListenSocket.cpp
```diff
@@ -57,6 +58,7 @@ TCPListenSocket::~TCPListenSocket()
     }
     catch (...) {
         // ignore
+        LOG((CLOG_INFO "error while closing TCP socket"));
```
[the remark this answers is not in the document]
abatyiev: Fixed

src/lib/platform/XWindowsClipboard.cpp
abatyiev: Beside of complexity this code also has logical errors(converting to hex going wrong). 1128 line contains byte offset of uchar but with value 16 instead of 4. I think, that even log changes can lead to problems, when you are not well familiar with projects, that's why I prefer to do my first changes with minimum of intervention. But if you insist - I'll move this part to new function and, also, can fix char conversion.

src/lib/platform/XWindowsClipboard.cpp
abatyiev: Same with previous comments, based on your experience and advice this code can be rewritten

src/lib/platform/XWindowsClipboard.cpp
SerhiiGadzhilov: @abatyiev Could you please, create a new task in Jira for the problem with hex conversion. 
We will change this code under the new task because it requires thorough investigation.
I will merge this code as is.

SerhiiGadzhilov (approved):

PR #6972, opened 2021-04-05
SYNERGY-511 SonarCloud security hotspots in Synergy-core

abatyiev:
* Test commit to check sonar reaction

### DIFF
--- a/src/gui/src/main.cpp
+++ b/src/gui/src/main.cpp
@@ -56,7 +56,7 @@ int main(int argc, char* argv[])
     ::setenv ("QT_BEARER_POLL_TIMEOUT", "-1", 1);
 #endif
     QCoreApplication::setOrganizationName("Synergy");
-    QCoreApplication::setOrganizationDomain("http://symless.com/");
+    QCoreApplication::setOrganizationDomain("https://symless.com/");
     QCoreApplication::setApplicationName("Synergy");
 
     QSynergyApplication app(argc, argv);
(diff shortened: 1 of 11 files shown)


src/gui/src/main.cpp
```diff
@@ -56,7 +56,7 @@ int main(int argc, char* argv[])
     ::setenv ("QT_BEARER_POLL_TIMEOUT", "-1", 1);
 #endif
     QCoreApplication::setOrganizationName("Synergy");
-    QCoreApplication::setOrganizationDomain("http://symless.com/");
+    QCoreApplication::setOrganizationDomain("https://symless.com/");
```
SerhiiGadzhilov, line 59: We can't just change "http" to "https" because this value is used as settings file name (https://doc.qt.io/qt-5/qcoreapplication.html#organizationDomain-prop) as result we will loose settings.
This value is not used as URL to send requests hence I would suggest just mark it as not a problem on SonarCloud and leave "http".
As alternative solution we can create a new task and investigate how to fix it.
@abatyiev In any case this change should be rolled back.

@andreacndn Should we create a new task for this or just mark it as not a problem?

SerhiiGadzhilov (changes requested):

src/gui/src/main.cpp
andreacndn: Agreed. I have marked not a problem on SonarCloud

abatyiev: After this change application start normally, and connection was successful, so I decided that this changes not impact. Old value returned, thanks for correction!

SerhiiGadzhilov (approved):

PR #6978, opened 2021-04-15
SYNERGY-799 Update Synergy UI. Settings window

SerhiiGadzhilov:


src/gui/src/MainWindow.cpp
```diff
@@ -1041,42 +1038,6 @@ void MainWindow::setVisible(bool visible)
 #endif
 }
 
-QString MainWindow::getIPAddresses()
```
abatyiev, line 1044: Is this part would be replaced with better alternatives? This code was useful.

SerhiiGadzhilov: No.
This is autoconfig part which is not used anymore and turned off using option SYNERGY_AUTOCONFIG.
We will remove all autoconfig code under separate task.

abatyiev (approved):

PR #6983, opened 2021-04-21
Synergy 885 mac client listens on ipv4 only but attempts to connect on ipv6

abatyiev:


src/lib/net/NetworkAddress.cpp
```diff
@@ -62,46 +63,51 @@ NetworkAddress::NetworkAddress(const String& hostname, int port) :
...
-        // disambiguate.
-        if ((!doubleColon || dotNotation) && !colonNotation) {
-            // parse port from hostname
-            char* end;
-            const char* chostname = m_hostname.c_str();
-            long suffixPort = strtol(chostname + i + 1, &end, 10);
-            if (end == chostname + i + 1 || *end != '\0') {
-                throw XSocketAddress(XSocketAddress::kBadPort,
-                                            m_hostname, m_port);
+        m_hostname = m_hostname.substr(0, hostIt - 1);
```
SerhiiGadzhilov, line 79: 1. If the hostname contains only port (“:24800”) the system doesn’t erase it from the hostname as the result we have the following error “synergys: unknown error for: :24800:24800”.  

2. As I can see this function doesn't have external dependencies hence it could be easily covered with unit tests to fix "SonarCloud Quality Gate failed."

SerhiiGadzhilov (changes requested):

src/lib/net/NetworkAddress.cpp
abatyiev: Thanks! All issues fixed!

SerhiiGadzhilov (approved):

PR #6989, opened 2021-04-28
SYNERGY-907 Client tray language switcher reverts to EN

abatyiev:
*Dummy commit to update test coverage statistic


src/gui/src/AppConfig.cpp
```diff
@@ -239,6 +239,7 @@ void AppConfig::loadSettings()
     m_UseInternalConfig         = loadSetting(kUseInternalConfig, false).toBool();
     m_ClientGroupChecked        = loadSetting(kGroupClientCheck, false).toBool();
     m_ServerHostname            = loadSetting(kServerHostname).toString();
+    m_LanguageSync              = loadSetting(kLanguageSync, true).toBool();
```
SerhiiGadzhilov, line 242: According to the task description the new option should be disabled by default.
According to the code this option is enabled by default.

src/gui/src/AppConfig.cpp
```diff
@@ -299,6 +300,7 @@ void AppConfig::saveSettings()
         setSetting(kUseInternalConfig, m_UseInternalConfig);
         setSetting(kGroupClientCheck, m_ClientGroupChecked);
         setSetting(kServerHostname, m_ServerHostname);
+        setSetting(kLanguageSync, m_LanguageSync);
```
SerhiiGadzhilov, line 303: Could you please, add the new option into the array AppConfig::m_SynergySettingsName.
You added “kLanguageSync” into the “enum Setting”. 
This constant is used to get value from the AppConfig::m_SynergySettingsName.
AppConfig::m_SynergySettingsName was not updated as a result the program reads outside of the array and the setting doesn’t work.

src/gui/src/MainWindow.cpp
```diff
@@ -748,6 +748,12 @@ bool MainWindow::clientArgs(QStringList& args, QString& app)
         args << "--log" << appConfig().logFilenameCmd();
     }
 
+    if (appConfig().getLanguageSync())
+    {
+        args << "--sync-language";
+    }
```
SerhiiGadzhilov, line 754: As I understand this option is only for synergy client.
I think it should be hidden for servers.
It could lead to misunderstanding and complains from users if users set this option on the server machine and it will not work.

SerhiiGadzhilov (changes requested):

src/gui/src/AppConfig.cpp
abatyiev: Fixed! Thanks!

src/gui/src/AppConfig.cpp
abatyiev: Fixed! Thanks!

src/gui/src/MainWindow.cpp
abatyiev: This code is already in function `MainWindow::clientArgs`. So will be added only for client start. Also parsing of this flag is in `ArgParser::parseClientArgs`. So we are ready for this flag only when parsing clients args.

SerhiiGadzhilov: Yes, that's why I think we should hide this flag on GUI for servers. 
This flag doesn't work for servers and should be hidden for users.
For example user can set this flag on server side and expect that all clients will work according to this flag but it's not true.

andreacndn: Agreed

abatyiev: Done! Now This setting is hidden when server mode is active.

SerhiiGadzhilov (approved):

PR #7008, opened 2021-05-19
SYNERGY-885 mac client listens on ipv4 only but attempts to connect on ipv6

abatyiev:


andreacndn:
SYNERGY-885

SerhiiGadzhilov (changes requested):
According to the task we should implement the following functionality:
1. In case more than one IP address is returned for the same name, log the information to the logs
2. Try to connect to the IPs following the order that has been returned by the name resolver. Connect to the first one that allows a successful connection.

This PR contains only the first part.
Could you please, implement the second part.

PR #7028, opened 2021-06-03
Barrier for Android

alanoliver2020:
Hello your project is excellent, but it would be better if it had the android version, a big hug thanks!


SerhiiGadzhilov:
Looks like this PR was created by mistake

PR #7038, opened 2021-06-10
Synergy 907 client tray language switcher reverts to en

abatyiev:


SerhiiGadzhilov (changes requested):
Could you please, resolve conflicts and sonar issues.

abatyiev:
For now Sonar warnings are, mainly, about files tinyxml2.cpp and tinyxml2.h. This file is from open source, well-known and tested library [tinyxml2](https://github.com/leethomason/tinyxml2), I think it is better to do not modify these files and add them to Sonar exclude list.

SerhiiGadzhilov (changes requested):
Could you please, add checkbox into the "Settings" window.

src/lib/client/Client.cpp
```diff
@@ -746,10 +767,22 @@ Client::handleHello(const Event&, void*)
...
+
     LOG((CLOG_DEBUG1 "say hello version %d.%d", helloBackMajor, helloBackMinor));
     ProtocolUtil::writef(m_stream, kMsgHelloBack,
                             helloBackMajor,
-                            helloBackMinor, & m_name);
+                            helloBackMinor, & m_name, &allKeyboardLayoutsStr);
+
+    if(!missedLanguages.empty()) {
+        AppUtil::instance().showMessageBox("Language synchronization error",
+                                           String("This languages are required for client proper work: ") + missedLanguages);
```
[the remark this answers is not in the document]
abatyiev: Thanks! Fixed

src/lib/server/ClientProxyUnknown.h
```diff
@@ -20,6 +20,7 @@
 
 #include "base/Event.h"
 #include "base/EventTypes.h"
+#include "base/String.h"
```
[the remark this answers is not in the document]
abatyiev: Fixed

src/lib/server/Server.cpp
```diff
@@ -100,7 +100,7 @@ Server::Server(
 	m_waitDragInfoThread(true),
 	m_args(args)
 {
-	// must have a primary client and it must have a canonical name
+    // must have a primary client and it must have a canonical name
```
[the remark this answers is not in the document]
abatyiev: Fixed

ChangeLog
```diff
@@ -9,6 +9,7 @@ Bug fixes:
 - #7029 | #7033 Wrong encoding for text copied between linux and windows
 - #7015 Fix Windows service not starting up after sleep
 - #7036 Fix tray icon not changing theme on Big Sur
+- #7038 Fix client and server different keyboard layout errors
```
SerhiiGadzhilov, line 12: Could you please, resolve conflict and  move this task under section v1.14.2

src/lib/synergy/unix/X11LayoutsParser.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+        appendVectorUniq(*toCopy, iso639_2Codes);
+    }
+}
+
+std::vector<String>
+X11LayoutsParser::getX11LanguageList()
+{
+    std::vector<String> result;
+    std::ifstream file("/etc/default/keyboard");
```
SerhiiGadzhilov, line 157: Is it possible to use the X11 API instead of parsing “/etc/default/keyboard”?

SerhiiGadzhilov (changes requested):
Could you please, resolve conflict

src/lib/synergy/unix/X11LayoutsParser.cpp
abatyiev: Unfortunately, there is no useful X11 api to get all installed languages, parsing of the "/etc/default/keyboard" file is the onliest way to get full language list

ChangeLog
abatyiev: Fixed

src/lib/client/Client.cpp
```diff
@@ -721,13 +722,33 @@ void
...
+        auto serverLayout = keyboardLayoutList.substr(i, 2);
+        if (std::find(localLayouts.begin(), localLayouts.end(), serverLayout) == localLayouts.end()) {
+            missedLanguages += serverLayout;
+            missedLanguages += ' ';
+        }
+        else {
+            supportedLanguages += serverLayout;
+            supportedLanguages += ' ';
+        }
+    }
```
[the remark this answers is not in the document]
abatyiev: This blocks of code for different Synergy parts. One of them is for client, other - for server. For now, mainly, they are same, but it is incorrect to combine them because they works in different time and and for different logic

SerhiiGadzhilov (changes requested):
Looks like tinyxml has too many issues.
Could you please, try to use https://github.com/zeux/pugixml instead of tinyxml.
Also include it as an external module as it’s done for WinToast and googletest.
In this case this module will be updated with all latest fixes from the official repo.

abatyiev:
TinyXML replaced with pugixml, added as a git submodule. Thanks

src/lib/server/ClientProxyUnknown.cpp
```diff
@@ -245,6 +254,23 @@ ClientProxyUnknown::handleData(const Event&, void*)
...
+            std::copy(supported.begin(), supported.end(), std::ostream_iterator<String>(result, ", "));
+            LOG((CLOG_DEBUG "Supported client languages: %s",  result.str().c_str()));
+        }
+
+        if(!missed.empty()) {
+            std::ostringstream result;
+            std::copy(missed.begin(), missed.end(), std::ostream_iterator<String>(result, ", "));
+            AppUtil::instance().showNotification("Language synchronization error",
+                                                 "These languages are required for server proper work: " + result.str());
+        }
```
SerhiiGadzhilov, line 272: The synergy server always shows notification about language synchronization despite the option language sync.

As I understand from the last comment in the task we should show notifications only on the client side.
These notifications should be shown in two cases:
- when the client connects to the server
- ​​every time the focus is on synergy client and the user is typing (server is sending keystrokes to the client)

As I understand, only the first point was implemented.
@andreacndn Could you please, confirm that my understanding is correct.

SerhiiGadzhilov (changes requested):
Looks like synchronization doesn’t work.
The following scenario doesn’t work as for me:

Initial conditions:
Server - Windows
Server Language: RU, EN
Client - MacOS
Client Languages: RU, EN, UK
The option to sync languages is set on the client.

Step 1
Select language EN on the server
Select language RU on the client
Connect server and client
Move cursor to the client and type text
Expected:
Language on the client is switched to server (EN)
The text is typed in English
Actual:
Language on the client was not switched to server (EN)
The text is typed in Russian

Is it possible to increase test coverage for the new code?

src/gui/src/MainWindow.cpp
```diff
@@ -726,6 +726,10 @@ void MainWindow::startSynergy()
         args << "--prevent-sleep";
     }
 
+    if (appConfig().getLanguageSync()) {
+        args << "--sync-language";
+    }
+
```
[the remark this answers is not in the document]
abatyiev: Unfortunately, because of synergy network protocol limitation, only notifications, at start of session, are implemented. Test coverage is in progress, soon will be ready!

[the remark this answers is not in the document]
abatyiev: "--sync-language" moved to clients config, thanks!

src/lib/synergy/ArgParser.cpp
```diff
@@ -111,6 +111,9 @@ ArgParser::parseClientArgs(lib::synergy::ClientArgs& args, int argc, const char*
             // define scroll
             args.m_yscroll = atoi(argv[++i]);
         }
+        else if (isArg(i, argc, argv, nullptr, "--sync-language", 1)) {
```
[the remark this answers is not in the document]
abatyiev: Fixed! Thanks!

SerhiiGadzhilov (changes requested):
As I can see the following scenario is still doesn’t work:

Initial conditions:
Server - Windows
Server Language: RU, EN
Client - MacOS
Client Languages: RU, EN, UK
The option to sync languages is set on the client.

Step 1
Select language EN on the server
Select language RU on the client
Connect server and client
Move cursor to the client and type text

Expected:
Language on the client is switched to server (EN)
The text is typed in English

Actual (Problem):
Language on the client was not switched to server (EN)
The text is typed in Russian

SerhiiGadzhilov (changes requested):
As I can see the scenario described above is still doesn't work.
The first issue is that  `kMsgDKeyDown`  has incorrect format: `"DKDN%2i%2i%2i%s"`
As result language is not send to the client.

After I fixed the format and the client receives the language the scenario still doesn't work.

SerhiiGadzhilov (changes requested):
Language switching doesn’t work if Windows is a server.

Scenario:
Server - Windows; Language EN
Client - Linux/Mac; Language EN
Language sync is on.

Step 1
Connect server and client
Change server language to RU
Go client and type text
Expected: The text is typed in RU
Actual: The text is typed in EN

Note: During switching from server to client the server switches to client language (EN). 
When moving back to server language is switched back (RU).

src/lib/platform/OSXKeyState.cpp
```diff
@@ -866,7 +880,21 @@ void
 OSXKeyState::setGroup(SInt32 group)
 {
     TISInputSourceRef keyboardLayout = (TISInputSourceRef)CFArrayGetValueAtIndex(m_groups.get(), group);
-    TISSetInputMethodKeyboardLayoutOverride(keyboardLayout);
+    if(!keyboardLayout) {
+        LOG((CLOG_WARN "Nedeed keyboard layout is null"));
+        return;
+    }
+    auto canBeSetted = (CFBooleanRef)TISGetInputSourceProperty(TISCopyCurrentKeyboardInputSource(), kTISPropertyInputSourceIsEnableCapable);
+    if(!canBeSetted || canBeSetted == FALSE) {
```
SerhiiGadzhilov, line 888: Isn't it's the same as if(!canBeSetted)

src/lib/platform/OSXKeyState.cpp
```diff
@@ -601,13 +602,26 @@ OSXKeyState::fakeKey(const Keystroke& keystroke)
...
+                    setGroup(getEffectiveGroup(pollActiveGroup(), group));
+                }
+
+                //A minimal delay is needed after a group change because the
+                //keyboard key event often happens immediately after.
+                //Language (TIS) and event (CG) systems are not in the mutual
+                //event queue and without a delay the subsequent key press
+                //event could be applied before the keyboard layout would
+                //actually be changed.
+                ARCH->sleep(.01);
```
SerhiiGadzhilov, line 621: Isn’t it better to move `ARCH->sleep(.01);` inside `setGroup`?

src/lib/platform/MSWindowsDesks.cpp
```diff
@@ -608,14 +614,12 @@ MSWindowsDesks::deskLeave(Desk* desk, HKL keyLayout)
                     GetWindowThreadProcessId(desk->m_window, NULL);
                 DWORD thatThread =
                     GetWindowThreadProcessId(desk->m_foregroundWindow, NULL);
-                AttachThreadInput(thatThread, thisThread, TRUE);
-                SetForegroundWindow(desk->m_window);
-                AttachThreadInput(thatThread, thisThread, FALSE);
```
SerhiiGadzhilov, line 613: I'm really worry about this change because we can break something.
According to Microsoft documentation SetForegroundWindow shouldn't work for background processes.
(https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setforegroundwindow)

In any case this change is incorrect because if we remove `AttachThreadInput` we don't need anymore `GetWindowThreadProcessId` inside this condition. Also if we remove `ActivateKeyboardLayout` we don't need `keyLayout` parameter. In this case we should remove this parameter and fix code which calls this function.

From my perspective we should restore this code and find another solution or discuss possible impact if we don't fix the problem and maybe leave it as is for now.

src/lib/platform/MSWindowsDesks.cpp
```diff
@@ -145,7 +145,13 @@ MSWindowsDesks::MSWindowsDesks(
...
     m_cursor    = createBlankCursor();
     m_deskClass = createDeskWindowClass(m_isPrimary);
-    m_keyLayout = GetKeyboardLayout(GetCurrentThreadId());
+    GUITHREADINFO gti = { sizeof(GUITHREADINFO) };
+    if(GetGUIThreadInfo(0, &gti) && gti.hwndActive) {
+        m_keyLayout = GetKeyboardLayout(GetWindowThreadProcessId(gti.hwndActive, NULL));
+    }
+    else {
+        LOG((CLOG_WARN "Failed to determine correct keyboard layout"));
+    }
```
SerhiiGadzhilov, line 154: As I can see we have the same code in `MSWindowsDesks::MSWindowsDesks, MSWindowsScreen::leave` and `AppUtilWindows::getCurrentLanguageCode.` Maybe we can move it into separate function and use this function in all these places?

SerhiiGadzhilov (changes requested):

src/lib/platform/OSXKeyState.cpp
SerhiiGadzhilov: Fixed

src/lib/platform/OSXKeyState.cpp
SerhiiGadzhilov: Fixed

src/lib/platform/MSWindowsDesks.cpp
SerhiiGadzhilov: Fixed

src/lib/platform/MSWindowsDesks.cpp
SerhiiGadzhilov: Fixed

SerhiiGadzhilov (approved):

PR #7046, opened 2021-07-01
SYNERGY-1032 - Add MacOS version check and disabl UserNotification library if needed

igor-sikachyna:


azure-pipelines.yml
```diff
@@ -80,13 +80,13 @@ jobs:
         catalina-std:
           image: macOS-10.15
           platform: x86-64
-          version: 10.13
+          version: 10.15
```
SerhiiGadzhilov, line 83: This variable is used in CI/mac_build.yml in the following way:
-DCMAKE_OSX_DEPLOYMENT_TARGET=$(version)
According to synergy web-site we support the following versions:
- macOS BigSur (11) Intel and Apple M1
- macOS Catalina (10.15)
- macOS Mojave (10.14)
- macOS High Sierra 10.13

We have separate builds for macOS 10.13 and for M1.
We have only one build for 10.14, 10.15, 11(Intel).
If CMAKE_OSX_DEPLOYMENT_TARGET is set to 10.15 this build will not work on 10.14.

I think “version: 10.15” should be changed to “version: 10.14” to support macOS10.14.

SerhiiGadzhilov (changes requested):

azure-pipelines.yml
igor-sikachyna: Agree, updated to by `10.14` instead, does not impact other changes since UserNotification library was added in `10.14`

SerhiiGadzhilov (approved):

PR #7047, opened 2021-07-01
SYNERGY-856 - Force prevent sleep

igor-sikachyna:


src/gui/src/MainWindow.cpp
```diff
@@ -718,6 +721,9 @@ void MainWindow::startSynergy()
     if (m_AppConfig->getPreventSleep())
     {
         args << "--prevent-sleep";
+#ifdef Q_OS_LINUX
+        setSleepPreventionState(true);
+#endif
```
SerhiiGadzhilov, line 726: As I can see for mac and windows this functionality is implemented through the command line option “--prevent-sleep” but for Linux it’s implemented inside MainWindow.
As a result we have inconsistency which doesn’t allow us to use this functionality through the command line. 
For example it’s a real problem for Synergy 3 which has another GUI and passes options through the command line. Such logic shouldn’t be implemented inside GUI code.

src/gui/src/MainWindow.cpp
```diff
@@ -1504,3 +1514,54 @@ void MainWindow::on_m_pButtonConnect_clicked()
...
+            QDBusInterface screenSaverInterface( services[i], paths[i],services[i], bus);
+
+            if (!screenSaverInterface.isValid())
+                continue;
+
+            QDBusReply<uint> reply;
+
+            if(state)
+            {
+                reply = screenSaverInterface.call("Inhibit", "nzuri-video Downloader", "REASON");
```
SerhiiGadzhilov, line 1548: What is "nzuri-video Downloader"?

SerhiiGadzhilov (changes requested):

src/lib/platform/XWindowsScreen.cpp
```diff
@@ -245,6 +247,12 @@ XWindowsScreen::enable()
 		// warp the mouse to the cursor center
 		fakeMouseMove(m_xCenter, m_yCenter);
 	}
+
+	// disable sleep if the flag is set
+	if (App::instance().argsBase().m_preventSleep &&
+		!ArchSystemUnix::disableSleep()) {
+		LOG((CLOG_INFO "Failed to prevent system from going to sleep"));
```
SerhiiGadzhilov, line 254: Prevent sleep doesn't work on my Ubuntu 20.04.
I always receive the following message: "Failed to prevent system from going to sleep".
Does it work on your Linux system?
Is it possible to add a reason of the failure to the traces?

SerhiiGadzhilov (changes requested):

src/gui/src/MainWindow.cpp
```diff
@@ -38,6 +38,10 @@
 #include "Zeroconf.h"
 #include <QPushButton>
 
+#if defined(Q_OS_LINUX)
+#include <QtDBus/QtDBus>
+#endif
+
```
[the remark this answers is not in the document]
igor-sikachyna: Agree, removed

src/lib/platform/XWindowsScreen.cpp
igor-sikachyna: Added more verbose output to the errors occurring during sleep prevention. Also tested on my Ubuntu 20.04 VM and it seems to be working fine.

src/lib/platform/XWindowsScreen.cpp
```diff
@@ -225,6 +227,9 @@ XWindowsScreen::~XWindowsScreen()
 	s_screen = NULL;
 }
 
+bool disableSleep();
+bool enableSleep();
+
```
SerhiiGadzhilov, line 232: Could you please, remove these unused prototypes.

SerhiiGadzhilov (changes requested):

SerhiiGadzhilov (approved):

PR #7056, opened 2021-07-08
SYNERGY-1065 Mac notification on foreground

igor-sikachyna:
This is an example code, should solve the problem but it does not


SerhiiGadzhilov:
Decided not to fix this. 
It all works if the notifications are configured as “Alert” instead of “Banner”

PR #7063, opened 2021-07-30
SYNERGY-1089-Synergy1. There is no notification if user moves pointer…

abatyiev:
… between 2 clients

*Add messagebox when app is foreground and we need to show notification


SerhiiGadzhilov:
Decided not to merge this. 
It all works if the notifications are configured as “Alert” instead of “Banner”

PR #7067, opened 2021-08-04
Reduce file size of PNGs used by GUI and setup

K0-RR:
Lossless compression.


vitalPixel:
Hi @RDKRACZ ,

thanks for your contribution.
Unfortunately we will not be able to accept this change right now as it would have negative impacts on our current workflow.

Hopefully we'll find a way to fit this in our workflow in the near future.

nbolton:
> negative impacts on our current workflow.

Strange. Not sure I understand what this means. I'll reopen this and investigate.

sithlord48:
This should be closed since we can't apply this to our new images. @nbolton

PR #7074, opened 2021-08-19
Synergy 1161 when clicking on save preferences on configure server window, automatically apply all the new settings

abatyiev:


SerhiiGadzhilov (changes requested):
Is it possible to fix code smells?

abatyiev:
> Is it possible to fix code smells?

Fixed!Thanks

SerhiiGadzhilov (approved):

PR #7084, opened 2021-09-02
SYNERGY-1043

daunchung:
testing github action to upload files to s3


.github/workflows/aws-upload-mac-10.13.yml
```diff
@@ -0,0 +1,51 @@
...
+
+    strategy:
+      matrix:
+        enterprise: ['1', '']
+    env:
+      GIT_COMMIT: ${{ github.sha }}
+      SYNERGY_ENTERPRISE: ${{ matrix.enterprise }}
+      Qt5_DIR: /usr/local/opt/qt/5.15.2/clang_64
+      OpenSSL_DIR: /usr/local/ssl
+      CODESIGN_ID: "Developer ID Application: Symless Ltd (4HX897Y6GJ)"
```
SerhiiGadzhilov, line 19: Looks like this workflow doesn't sign binaries hence we don't need this variable.

ChangeLog
```diff
@@ -11,6 +11,9 @@ Enhancements:
 - #7068 Add Synergy restart when settings changed
 - #7072 Ability to run synergy as a pre-login agent
 - #7074 Add Synergy restart when server settings changed
+
+Github Actions:
+- #1043 Adding upload to s3 feature on ubuntu
```
SerhiiGadzhilov, line 16: Looks like this feature was added for all not only for Ubuntu?

SerhiiGadzhilov (changes requested):

.github/workflows/aws-upload-mac-10.13.yml
```diff
@@ -0,0 +1,49 @@
...
+  release:
+    types: [created]
+
+jobs:
+  build-on-mac-10_13:
+    runs-on: [self-hosted, macOS, X64, 10.13]
+
+    strategy:
+      matrix:
+        enterprise: ['1', '']
```
SerhiiGadzhilov, line 13: As I can see the variable `matrix.enterprise` is not used and should be removed.
Previously, it was used to define `SYNERGY_ENTERPRISE: ${{ matrix.enterprise }} `to make two different builds.
It's not used anymore and we will have two runs for the same build.

SerhiiGadzhilov (changes requested):

SerhiiGadzhilov (approved):

PR #7085, opened 2021-09-14
Synergy 810 synergy icon is removed from system tray if the user clicks red cross in main application window

abatyiev:


SerhiiGadzhilov (changes requested):
This PR implements the following functionality:
If the option “Minimize to tray” is on the system will leave synergy in the tray.
If the option “Minimize to tray” is off the system will close synergy.

According to the attached video this is not what is expected.

As I understand, on MacOS and Linux we can’t close the application except using File->Quit.

@andreacndn Do I understand correctly that we should implement the same behavior on Windows? Do we really need MacOS and Linux user experience on Windows?

SerhiiGadzhilov:
This PR changes close button behaviour.
This is not what is expected to be implemented under this task.

PR #7098, opened 2021-09-29
Add M1 workflow

aquacash5:
- Add workflows for apple's M1 platform
- Fix issue with core builds


.github/workflows/aws-upload-windows.yml
```diff
@@ -1,101 +1,76 @@
...
-        
-
-    - name: Install Qt
-      if: steps.cache-qt.outputs.cache-hit != 'true'
-      run: |
-        pip install aqtinstall
-        python -m aqt install --outputdir $env:QT_BASE_DIR --base $env:QT_BASE_URL $env:QT_VERSION windows desktop ${{ steps.qt-setup.outputs.QT_ARCH }}
+      - uses: actions/checkout@v2
 
+      - name: Set up QT variables
```
SerhiiGadzhilov, line 25: As I can this task doesn't setup QT variables anymore.
Maybe it's better to move this code into `"Install OpenSSL"`
Also as I can see we have only "x64" build hence we can remove the following condition: `if ('${{ matrix.arch }}' -eq 'x64')`

.github/workflows/aws-upload-mac.yml
```diff
@@ -0,0 +1,58 @@
...
+      matrix:
+        include:
+          - os: "macos-10.15"
+            target: "10.14"
+            shell: "bash"
+            destination: macOS_x64
+          - os: [self-hosted, macOS, arm64]
+            target: ""
+            shell: "/usr/bin/arch -arch arm64e /bin/bash --noprofile --norc -eo pipefail {0}"
+            destination: macOS_arm64
```
SerhiiGadzhilov, line 25: As I can see .github/workflows/aws-upload-mac-10.13.yml was removed and .github/workflows/aws-upload-mac.yml doesn't have macOS10.13.
We don't need it anymore?

src/cmd/CMakeLists.txt
```diff
@@ -14,11 +14,7 @@
...
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+add_subdirectory(synergyd)
 add_subdirectory(synergyc)
 add_subdirectory(synergys)
 add_subdirectory(syntool)
-
-if (SYNERGY_BUILD_LEGACY_SERVICE)
-    add_subdirectory(synergyd)
-endif (SYNERGY_BUILD_LEGACY_SERVICE)
```
SerhiiGadzhilov, line 23: As I can see we still have the option `SYNERGY_BUILD_LEGACY_SERVICE` inside `CMakeLists.txt` but after this change it does nothing. Looks like we need to remove the option or revert changes in `src/cmd/CMakeLists.txt`.

SerhiiGadzhilov (changes requested):

.github/workflows/aws-upload-mac.yml
aquacash5: For the core builds, I don't think we need macOS 10.13 builds.

src/cmd/CMakeLists.txt
aquacash5: You are correct. I will make that change

.github/workflows/aws-upload-windows.yml
aquacash5: Until we need a 32-bit version, I will set the action to only make a 64-bit version.

SerhiiGadzhilov (approved):

SerhiiGadzhilov:
@aquacash5 Could you please, check what is wrong with actions after merging this PR
5A9AD548-CADD-4D8A-9642-1A4CA4A8411E
:

PR #7099, opened 2021-09-30
SYNERGY-854 Compiling with MSVC 2014 causes error C2280

abatyiev:
*Fix compilation error


SerhiiGadzhilov (changes requested):
Could you please, fix the code smell if it's possible.

SerhiiGadzhilov (approved):

PR #7104, opened 2021-10-11
Synergy 1267 endless notifications during connection

SerhiiGadzhilov:


nbolton (approved):
I'm guessing it needs to be `CLOG_INFO` so the UI can always see it.

PR #7111, opened 2021-10-21
SYNERGY-1287 display language notifications only on client side

SerhiiGadzhilov:
Under this PR the following changes were implemented:
1. Removed language notification on the server side.
2. Enable notification on the client side even if sync is off.
3. Notification message was changed according to the task SYNERGY-1287
4. Changed Doxygen job to use ubuntu-latest instead of ubuntu-16 because ubuntu-16 is no available anymore on Azure Pipelines.

### DIFF
--- a/src/lib/synergy/protocol_types.cpp
+++ b/src/lib/synergy/protocol_types.cpp
@@ -19,7 +19,7 @@
 #include "synergy/protocol_types.h"
 
 const char* const               kMsgHello            = "Synergy%2i%2i%s";
-const char* const               kMsgHelloBack        = "Synergy%2i%2i%s%s";
+const char* const               kMsgHelloBack        = "Synergy%2i%2i%s";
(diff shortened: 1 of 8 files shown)


src/lib/synergy/protocol_types.cpp
```diff
@@ -19,7 +19,7 @@
 #include "synergy/protocol_types.h"
 
 const char* const               kMsgHello            = "Synergy%2i%2i%s";
-const char* const               kMsgHelloBack        = "Synergy%2i%2i%s%s";
```
nbolton, line 22: @SerhiiGadzhilov Do you think this enough to warrant changing the protocol version?

SerhiiGadzhilov: No, I just removed the field which was added under the language task because it's not needed anymore.

PR #7135, opened 2022-01-12
SYNERGY-1397 MacOS function keys

SerhiiGadzhilov:


src/lib/platform/OSXKeyState.cpp
```diff
@@ -336,34 +382,34 @@ OSXKeyState::fakeCtrlAltDel()
 bool
 OSXKeyState::fakeMediaKey(KeyID id)
 {
-    return fakeNativeMediaKey(id);;
```
nbolton, line 339: @SerhiiGadzhilov I wondered how long that'd been there... turns out 6 years. https://github.com/symless/synergy-core/commit/9115d5570bcf85e68b8ffd9819359e44f8c4999d

src/lib/platform/OSXKeyState.cpp
```diff
@@ -530,12 +552,9 @@ OSXKeyState::getKeyboardEventFlags() const
 {
     // set the event flags for special keys
     // http://tinyurl.com/pxl742y
-    CGEventFlags modifiers = getDeviceIndependedFlags();
+    CGEventFlags modifiers = getModifierStateAsOSXFlags();
 
-    if (m_capsPressed) {
-        modifiers |= kCGEventFlagMaskAlphaShift;
```
nbolton, line 536: @SerhiiGadzhilov What are the risks of removing this?

src/lib/platform/OSXKeyState.cpp
```diff
@@ -567,7 +586,26 @@ OSXKeyState::setKeyboardModifiers(CGKeyCode virtualKey, bool keyDown)
     }
 }
 
-void 
+// Post a key event to HID manager. It posts an event to HID client, a
```
nbolton, line 589: How about using Doxygen style comments here?
Example: https://github.com/symless/synergy-core/blob/master/src/lib/synergy/IScreen.h

src/lib/platform/OSXKeyState.h
```diff
@@ -99,7 +99,7 @@ class OSXKeyState : public KeyState {
     virtual SInt32        pollActiveGroup() const;
     virtual void        pollPressedKeys(KeyButtonSet& pressedKeys) const;
 
-    CGEventFlags getModifierStateAsOSXFlags();
+    CGEventFlags getModifierStateAsOSXFlags() const;
```
nbolton, line 102: Nice to see const correctness being added 👍

nbolton (approved):

src/lib/platform/OSXKeyState.cpp
```diff
@@ -129,6 +130,51 @@ static const KeyEntry    s_controlKeys[] = {
...
+io_connect_t
+getEventDriver()
+{
+    static io_connect_t sEventDrvrRef = 0;
+
+    if (!sEventDrvrRef) {
+        // Get master device port
+        mach_port_t masterPort = 0;
+        kern_return_t kr = IOMasterPort(bootstrap_port, &masterPort);
+        assert(KERN_SUCCESS == kr);
```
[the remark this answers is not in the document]
SerhiiGadzhilov: @nbolton I restored this code and I don't think that we need exceptions here.
According to our logic if getEventDriver fails we will use CGEventPost API.
I will change it to exceptions if you think that it's necessary.
Should I use exceptions here?

src/lib/platform/OSXKeyState.cpp
SerhiiGadzhilov: There is no risk because I switched to use getModifierStateAsOSXFlags which has this code.
I removed getDeviceIndependedFlags because it has the same code as getModifierStateAsOSXFlags.

src/lib/platform/OSXKeyState.cpp
SerhiiGadzhilov: @nbolton As I can see only comments for public methods are in Doxygen format. 
I used the comment format as for other private methods because postHIDVirtualKey is a private method.
I will remove the comment from the cpp file and leave it only in the header. 
Should I change the comment format for the private method according to Doxygen format?

src/lib/platform/OSXKeyState.cpp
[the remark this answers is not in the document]
SerhiiGadzhilov: I will change this function

src/lib/platform/OSXKeyState.cpp
```diff
@@ -129,6 +130,65 @@ static const KeyEntry    s_controlKeys[] = {
...
+getService(io_iterator_t iter) {
+    io_connect_t service = 0;
+    auto nextIterator = IOIteratorNext(iter);
+
+    if (nextIterator) {
+        IOServiceOpen(nextIterator, mach_task_self(), kIOHIDParamConnectType, &service);
+        IOObjectRelease(nextIterator);
+    }
+
+    return service;
```
nbolton, line 145: @SerhiiGadzhilov Could `service` be `0` here?

src/lib/platform/OSXKeyState.cpp
```diff
@@ -129,6 +130,65 @@ static const KeyEntry    s_controlKeys[] = {
     { kKeyBrightnessDown, s_brightnessDown }
 };
 
+namespace {
+
+io_connect_t
+getService(io_iterator_t iter) {
+    io_connect_t service = 0;
```
nbolton, line 137: Should this be `nullptr` instead of `0`?

src/lib/platform/OSXKeyState.cpp
```diff
@@ -129,6 +130,65 @@ static const KeyEntry    s_controlKeys[] = {
...
+        mach_port_t masterPort = 0;
+        if (!IOMasterPort(bootstrap_port, &masterPort)) {
+            io_iterator_t iter = 0;
+            auto dict = IOServiceMatching(kIOHIDSystemClass);
+
+            if (!IOServiceGetMatchingServices(masterPort, dict, &iter)) {
+                sEventDrvrRef = getService(iter);
+            }
+            else {
+                LOG((CLOG_DEBUG1, "IOService not found"));
```
nbolton, line 164: I'm wondering if this should be warning/error... probably fine as debug, but it depends; will the user notice if this fails? If so, maybe make it warning/error.

src/lib/platform/OSXKeyState.cpp
```diff
@@ -129,6 +130,65 @@ static const KeyEntry    s_controlKeys[] = {
...
+            }
+
+            IOObjectRelease(iter);
+        }
+        else {
+            LOG((CLOG_DEBUG1, "Couldn't obtain IO master port"));
```
nbolton, line 170: Same here (warning/error?)

src/lib/platform/OSXKeyState.cpp
```diff
@@ -596,7 +652,11 @@ OSXKeyState::fakeKey(const Keystroke& keystroke)
                     button, virtualKey, keyDown ? "down" : "up", client));
 
             setKeyboardModifiers(virtualKey, keyDown);
-            postKeyboardKey(virtualKey, keyDown);
+            if (postHIDVirtualKey(virtualKey, keyDown) != KERN_SUCCESS) {
+                LOG((CLOG_DEBUG1, "Fail to post HID event"));
```
nbolton, line 656: Warning/error?

nbolton (approved):
Probably fine as is, but maybe consider using warning/error. Though, on second thoughts, error is probably best for when you intend to exit the program, so perhaps warning would work best (if you wanted to change it).

src/lib/platform/OSXKeyState.cpp
SerhiiGadzhilov: If IOServiceOpen filled the "service" it will not be 0.
If it was not filled the function returns 0 and it's expected.

src/lib/platform/OSXKeyState.cpp
SerhiiGadzhilov: From my perspective it shouldn't be warning/error.
This is technical traces which don't provide any useful information for the user.
The system uses CGEventPost if these calls are failed.

src/lib/platform/OSXKeyState.cpp
SerhiiGadzhilov: Fixed

src/lib/platform/OSXKeyState.cpp
SerhiiGadzhilov: Fixed

src/lib/platform/OSXKeyState.cpp
nbolton: Ah, 0 is expected 👍

src/lib/platform/OSXKeyState.cpp
nbolton: Ah, I should not have assumed ptr.

nbolton (approved):

PR #7141, opened 2022-01-26
SYNERGY-1454 Scroll direction refactoring

SerhiiGadzhilov:


src/lib/platform/MSWindowsScreen.h
```diff
@@ -44,7 +45,9 @@ class MSWindowsScreen : public PlatformScreen {
         bool isPrimary,
         bool noHooks,
         bool stopOnDeskSwitch,
-        IEventQueue* events);
+        IEventQueue* events,
+        bool enableLangSync = false,
+        lib::synergy::ClientScrollDirection scrollDirection = lib::synergy::ClientScrollDirection::SERVER);
     virtual ~MSWindowsScreen();
```
nbolton, line 51: @SerhiiGadzhilov Just an idea: I wonder if we should pass a setting struct/class? This would allow us to change settings on the fly without re-creating the object (if that's a good idea, I'm not sure). Also it would reduce the number of arguments in the ctor. What do you think? Happy to go with what we have now as I think it works fine too.

src/lib/platform/OSXScreen.mm
```diff
@@ -786,9 +788,7 @@
 	else {
 		// FIXME -- prevent system from entering power save mode
 
-		if (m_autoShowHideCursor) {
-			hideCursor();
-		}
```
nbolton, line 791: @SerhiiGadzhilov I'm a bit nervous about this change, in show/hide cursor on macOS is a bit weird. I think as long as it's tested thoroughly, it'd be ok. Is it actually related to scrolling though?

nbolton (approved):

src/lib/platform/MSWindowsScreen.h
SerhiiGadzhilov: All “XScreen” classes are used on server and client side.
ClientApp::createScreen()
ServerApp::createScreen()

Server arguments are stored in lib::synergy::ServerArgs
Client arguments are stored in lib::synergy::ClientArgs

If we want to pass a structure instead of separate parameters we need to create a new structure with all parameters, fill it and pass it into ctor or pass parent class lib::synergy::ArgsBase.

If we pass lib::synergy::ArgsBase we need to convert the type inside “XScreen” class to appropriate type which is not good and safe solution.

From my perspective the best solution is to create two separate classes “ServerScreen” and “ClientScreen”. “ServerScreen” should contain only server logic and “ClientScreen” should contain only client logic. Common logic should be inside the parent class.
In this case we can pass appropriate parameters to appropriate classes and totally separate client and server logic.

This approach requires much more time/changes and affects much more functionality.
I’m not sure that the current task is suitable for such refactoring.

src/lib/platform/OSXScreen.mm
SerhiiGadzhilov: m_autoShowHideCursor was initialised in the ctor according to the third parameter with default value autoShowHideCursor=true. There was no code which used the third parameter of the ctor hence this code was not used and m_autoShowHideCursor was always true.

PR #7149, opened 2022-02-26
Address issues with modifiers and dead keys

pedronavf:
After getting a dead key from `ToUnicodeEx`, add an additional `VK_SPACE` to the keyboard state so we reset the dead key flag and subsequent calls with modifiers, like shift, return the right result (-1) instead of 1.

This happened because without reseting the dead key status the new one was attempted to be composed with the old one, which failed and the end result was a single unicode codepoint not marked as a deadkey.

This opens the door to potentially use the returned unicode from the second call as the key character instead of maintaining the `getDeadKey` function.

This should address, hopefully, most reported issues with Shift-dead-key not working correctly (like https://github.com/symless/synergy-core/issues/5185). This patch was tested on a Linux server and a Windows 10 client. The Windows code doesn't use a flag introduced in Windows 10 version 1607 (which also fixes the issue) hoping to support more clients.


mhalano:
Any news about the approval of this PR?

PR #7163, opened 2022-05-03
SYNERGY1-1478 Fix compilation issues

SerhiiGadzhilov:


aquacash5 (approved):

src/lib/platform/XWindowsKeyState.cpp
```diff
@@ -273,7 +273,9 @@ bool XWindowsKeyState::setCurrentLanguageWithDBus(SInt32 group) const
 
     if(!reply.isValid()) {
         auto qerror = reply.error();
-        LOG((CLOG_WARN "Keyboard layout fail %s : %s", qerror.name().toStdString(), qerror.message().toStdString()));
```
nbolton, line 276: Oh wow, how did that ever work before?

SerhiiGadzhilov: Yes but this code is executed in very rare cases. 
I wasn't able to reach this code without code modification.

PR #7164, opened 2022-05-05
SYNERGY1-1497 Fix memory leaks

SerhiiGadzhilov:
This PR fixed the following memory leaks:
1. In function getKeyboardLayoutList call for CFDictionaryCreat and TISCreateInputSourceList
2. A new function freeSSL to call free SSL memory from destructor and function close

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -8,7 +8,7 @@ Bug fixes:
 - #7144 Fix double lines when pasting text from Linux to Windows
 - #7149 Address issues with modifiers and dead keys
 - #7163 Fix compilation issues for FreeBSD
-- #7164 Fix memory leaks
+- #7164 Fix memory leaks in language sync and TLS functionality
 
 Github Actions:
 - #7148 Fix unstable build for windows core
--- a/ChangeLog
+++ b/ChangeLog
@@ -8,6 +8,7 @@ Bug fixes:
 - #7144 Fix double lines when pasting text from Linux to Windows
 - #7149 Address issues with modifiers and dead keys
 - #7163 Fix compilation issues for FreeBSD
+- #7164 Fix memory leaks
 
 Github Actions:
 - #7148 Fix unstable build for windows core
--- a/src/lib/net/SecureSocket.cpp
+++ b/src/lib/net/SecureSocket.cpp
@@ -77,31 +77,13 @@ SecureSocket::SecureSocket(IEventQueue* events,
 
 SecureSocket::~SecureSocket()
 {
-    isFatal(true);
-    // take socket from multiplexer ASAP otherwise the race condition
-    // could cause events to get called on a dead object. TCPSocket
-    // will do this, too, but the double-call is harmless
-    setJob(NULL);
-    if (m_ssl->m_ssl != NULL) {
-        SSL_shutdown(m_ssl->m_ssl);
-
-        SSL_free(m_ssl->m_ssl);
-        m_ssl->m_ssl = NULL;
-    }
-    if (m_ssl->m_context != NULL) {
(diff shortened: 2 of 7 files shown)


src/lib/net/SecureSocket.cpp
```diff
@@ -77,31 +77,13 @@ SecureSocket::SecureSocket(IEventQueue* events,
 
 SecureSocket::~SecureSocket()
 {
-    isFatal(true);
```
nbolton, line 80: @SerhiiGadzhilov This is an odd function. It returns m_fatal (and the result is not used here). Also, I can't find a decl of `isFatal` that takes an arg. Is it the case that function does nothing, or did I miss something?

src/lib/platform/OSXAutoTypes.h
```diff
@@ -0,0 +1,28 @@
+/*
+ * synergy -- mouse and keyboard sharing utility
+ * Copyright (C) 2016 Symless Ltd.
```
nbolton, line 3: @SerhiiGadzhilov I believe this should be the current year.

ChangeLog
```diff
@@ -8,6 +8,7 @@ Bug fixes:
 - #7144 Fix double lines when pasting text from Linux to Windows
 - #7149 Address issues with modifiers and dead keys
 - #7163 Fix compilation issues for FreeBSD
+- #7164 Fix memory leaks
```
nbolton, line 11: @SerhiiGadzhilov Perhaps this could be more specific; something like, "Fix memory leaks in secure socket and key groups"

src/lib/net/SecureSocket.cpp
```diff
@@ -77,31 +77,13 @@ SecureSocket::SecureSocket(IEventQueue* events,
...
 }
 
 void
 SecureSocket::close()
 {
-    isFatal(true);
-
-    SSL_shutdown(m_ssl->m_ssl);
-
+    freeSSL();
```
nbolton, line 86: Good catch.

src/lib/platform/OSXKeyState.h
```diff
@@ -122,7 +118,7 @@ class OSXKeyState : public KeyState {
                             SInt32 group, const IOSXKeyResource& r) const;
 
     // Get the available keyboard groups
-    bool                getGroups(GroupList&) const;
```
nbolton, line 125: @SerhiiGadzhilov I see that `GroupList` is also used on Windows; I wonder if a memory leak exists there too (I didn't really look at the code but thought I'd raise the question). 
https://github.com/symless/synergy-core/blob/e9e21aad8ff14c253785f90b043212c31360f04a/src/lib/platform/MSWindowsKeyState.h#L172

nbolton (changes requested):

src/lib/net/SecureSocket.cpp
SerhiiGadzhilov: There are two functions in the file: synergy-core/src/lib/net/SecureSocket.h
```
bool                isFatal() const { return m_fatal; }
void                isFatal(bool b) { m_fatal = b; }
```

The first one is a getter and the second one is a setter.
In function “close” and destructor we use setter to set this flag.
Getter is used in other functions:
```
SecureSocket::secureAccept
SecureSocket::secureConnect
SecureSocket::checkResult
```

nbolton: Hmm, yeah I was looking for the setter but missed it. It's more the function name that threw me; `isFatal` seems a bit of an odd name, as it suggests something went wrong. I suppose we can leave it as is, but something to think about maybe.

nbolton (approved):

PR #7170, opened 2022-05-13
SYNERGY1-1497 memory leaks in synergyc on mac os

SerhiiGadzhilov:
The following memory leaks were fixed:
1. In `isScreenSaverEngine` free memory allocated by `CopyProcessName`
2. In `nameToAddr` free memory allocated by `getaddrinfo`
3. In ​​`SecureSocket::verifyCertFingerprint` free memory allocated by `SSL_get_peer_certificate`

Sonar shows some code duplication which we had previously.
It requires some refactoring which will increase testing scope hence it should be done under a separate task.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,7 +9,7 @@ Bug fixes:
 - #7144 Fix double lines when pasting text from Linux to Windows
 - #7149 Address issues with modifiers and dead keys
 - #7163 Fix compilation issues for FreeBSD
-- #7164 | #7170 Fix memory leaks in language sync and TLS functionality
+- #7172 Memory leaks in language sync and TLS functionality
 
 Github Actions:
 - #7148 Fix unstable build for windows core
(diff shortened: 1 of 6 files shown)


ChangeLog
```diff
@@ -9,7 +9,7 @@ Bug fixes:
 - #7144 Fix double lines when pasting text from Linux to Windows
 - #7149 Address issues with modifiers and dead keys
 - #7163 Fix compilation issues for FreeBSD
-- #7164 Fix memory leaks in language sync and TLS functionality
+- #7164 | #7170 Fix memory leaks in language sync and TLS functionality
```
nbolton, line 12: Let's merge these issues into one rather than mention both. If they're two different issues, then let's put on two separate lines.

nbolton (changes requested):

ChangeLog
SerhiiGadzhilov: This is the same issue and we have one task for this SYNERGY1-1497.
These are PRs numbers because we have two PRs which fix the same task.

nbolton: Please use the line:
#7172 Memory leaks in language sync and TLS functionality

SerhiiGadzhilov: I have updated the ChangeLog

nbolton (approved):

PR #7175, opened 2022-05-19
SYNERGY1-1497 Memory leaks in copy/paste and drag and drop functionality

SerhiiGadzhilov:


src/lib/client/Client.cpp
```diff
@@ -875,14 +874,13 @@ Client::isReceivedFileSizeValid()
 void
 Client::sendFileToServer(const char* filename)
 {
-    if (m_sendFileThread != NULL) {
+    if (m_sendFileThread) {
```
nbolton, line 877: I think the convention in the code is to use `!= nullptr` but the result is the same. Change if you like.

nbolton (approved):

PR #7217, opened 2022-08-31
Added active screen filename output to enable app integrations with things like key mappers

jglanz:
Added active screen filename output to enable app integrations with things like key mappers

- defaults to `~/.synergy-active-screen`
- Preferences panel option with browse button
- Added command line option `--active-screen-filename` to `synergys` too
- And added support to read from conf file in `options` section named `activeScreenFilename`

I personally use it with `kinto` on top of `xkeysnail` and it is MAGIC

If it's ok to merge then I'll happily clean it up (super busy at work, so it may take a minute), but regardless, hope this helps others.

J


CMakeLists.txt
```diff
@@ -44,8 +44,8 @@ else()
     option (SYNERGY_BUSINESS "Build Business" OFF)
 endif()
 
-set (CMAKE_CXX_STANDARD 14)
-set (CMAKE_CXX_EXTENSIONS OFF)
+set (CMAKE_CXX_STANDARD 17)
+set (CMAKE_CXX_EXTENSIONS ON)
```
SerhiiGadzhilov, line 48: We can’t simply switch to C++17 because some OSs that we have to support don’t support this standard.

src/cmd/synergys/CMakeLists.txt
```diff
@@ -14,46 +14,49 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+find_package(Qt5 COMPONENTS Core Widgets Network LinguistTools)
```
SerhiiGadzhilov, line 17: synergys shouldn't have QT dependency

src/cmd/synergys/synergys.cpp
```diff
@@ -16,6 +16,9 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <QCoreApplication>
+#include <QTimer>
+#include <QThread>
```
SerhiiGadzhilov, line 21: synergys shouldn't have QT dependency

src/lib/server/CMakeLists.txt
```diff
@@ -1,30 +1,34 @@
...
 # but WITHOUT ANY WARRANTY; without even the implied warranty of
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU General Public License for more details.
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+find_package(Qt5 COMPONENTS Core Widgets Network)
+find_package(httplib REQUIRED)
+set(CMAKE_AUTOMOC ON)
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
```
SerhiiGadzhilov, line 19: This project shouldn't have QT dependency

SerhiiGadzhilov (changes requested):
Thank you for the PR but it’s not ok to merge due to the following reasons:
1. This PR introduces QT dependencies for synergys.
2. This PR introduces C++17 dependencies which can’t be used due to some OSs which we still support.
3. Unresolved conflicts
4. Configuration files like “.gdbinit” and “.lldbinit” should be excluded from the PR.

nbolton:
Closing since no reply, please re-open when changes are made.

nbolton:
Actually, I think this could be useful, so I'll pick this up.

nbolton:
Unfortunately, as this is very far behind master now nad we don't have the bandwidth to rebase master to this PR, I'm forced to close the PR. If there is any interest in porting this code to the latest master, I would welcome a new PR.

PR #7221, opened 2022-09-16
Fix the free order

HenryHu:
Currently, XFree(kbdDescr) is called before  XkbFreeNames(kbdDescr, XkbSymbolsNameMask, true). As you can see, it should be called after that, since XkbFreeNames still uses kbdDescr.


SerhiiGadzhilov:
Thank you for your contribution but we already have a PR for this issue: https://github.com/symless/synergy-core/pull/7208.

PR #7236, opened 2022-11-06
Fix out-of-bounds access in XWindowsScreen::updateButtons

DuBistKomisch:
This code triggers an assertion in gcc's std_vector.h on my system due to a logical button 0 causing a write to m_buttons[-1], and a SIGABRT.

I've tried to make this code do what it seems to be intended to do, but m_buttons isn't actually read from anywhere anyway.

### DIFF
--- a/src/lib/platform/XWindowsScreen.cpp
+++ b/src/lib/platform/XWindowsScreen.cpp
@@ -2029,15 +2029,15 @@ XWindowsScreen::updateButtons()
 	}
 
 	// allocate button array
-	m_buttons.resize(maxButton);
+	m_buttons.resize(maxButton + 1);
 
 	// fill in button array values.  m_buttons[i] is the physical
-	// button number for logical button i+1.
-	for (UInt32 i = 0; i < numButtons; ++i) {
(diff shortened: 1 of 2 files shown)


DuBistKomisch:
Ah, I was a bit careless and didn't notice its usage in `mapButtonToX`, as I originally made this change against barrier which doesn't have it. I agree, I'll update when I have time.

nbolton:
Closing since not updated for a while, please re-open when changes are made. Thanks.

nbolton:
On second thoughts, I think the close was a bit premature there, reopening. I'll take a look at this and fix anything that is pending.

nbolton:
I tried to rebase master and update this PR but got an error:
```
>     git push DuBistKomisch HEAD:synergy-fix-linux-out-of-bounds --force
ERROR: Permission to DuBistKomisch/barrier.git denied to nbolton.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

@DuBistKomisch @sithlord48 Closing for now, but I wonder if it's worth reopening as a new PR?

PR #7241, opened 2022-12-15
Upgrade the default hashing algorithm

aquacash5:
[SYNERGY1-1634](https://symless.atlassian.net/browse/SYNERGY1-1634)

Upgrade the default hashing algorithm for message digests from md5 to sha256.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,7 @@
 
 Bug fixes:
 - #7208 Fix crash on FreeBSD when pressing special keys on a client.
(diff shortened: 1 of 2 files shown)


SerhiiGadzhilov (approved):

QuAzI:
Why not blake3?

SerhiiGadzhilov:
> 
According to the following ticket openssl doesn't support blake3.
https://github.com/openssl/openssl/issues/11613

PR #7242, opened 2023-01-05
Fix tls certificate generation for flatpak

SerhiiGadzhilov:
SYNERGY1-1592
This PR fixes problem with tls generation for flatpak package.


aquacash5 (approved):
And that just works!? :rofl:

PR #7274, opened 2023-07-29
Build unified core or separate server and client

aquacash5:
When building Synergy, we should only need to build a single `synergy-core` binary or separate `synergys` and `synergyc` binaries.


nbolton:
I was thinking the same.

PR #7283, opened 2023-09-22
Update workflows

aquacash5:
[S1-1669](https://symless.atlassian.net/browse/S1-1669)

Update workflows and fix broken workflows.

[S1-1669]: https://symless.atlassian.net/browse/S1-1669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/job-build-flatpak.yaml
```diff
@@ -8,26 +8,29 @@ jobs:
...
             remote_folder: v1-core-enterprise
           - name: synergy-business
             remote_folder: v1-core-business
+
     env:
       GIT_COMMIT: ${{ github.sha }}
-      SYNERGY_ENTERPRISE: ${{ matrix.enterprise }}
-      SYNERGY_BUSINESS: ${{ matrix.business }}
+      SYNERGY_ENTERPRISE: ${{ matrix.runtime.enterprise }}
+      SYNERGY_BUSINESS: ${{ matrix.runtime.business }}
```
SerhiiGadzhilov, line 29: As I can see we don't have `enterprise` and `business` in the `runtime`.
Is it correct?

SerhiiGadzhilov (dismissed):

.github/workflows/job-build-flatpak.yaml
aquacash5: These were changed from `include` to `runtime`. `include` is a way for you to add one off custom kinds of strategies. I have changed it to `runtime`, so now they are bundled together as a matrix of runtime "objects". https://github.com/symless/synergy-core/blob/3d4114e0a9633dd5378c4e3561161a8c21453aa9/.github/workflows/job-build-flatpak.yaml#L16-L24

PR #7288, opened 2023-09-26
Export a C++ compatible library from the Rust license tool

nbolton:
S1-1527


nbolton:
This is now quite old and has diverged from master, so I'll start a new branch and try to cherry pick in a separate PR.

PR #7295, opened 2023-09-26
Combine workflows into one CI workflow (original)

nbolton:
https://symless.atlassian.net/browse/S1-1687


nbolton:
Fixed by https://github.com/symless/synergy-core/pull/7352

PR #7297, opened 2023-09-26
Host mode connection issue

nbolton:
S1-1639


SerhiiGadzhilov:
It was test branch to investigate a problem with host mode connection

PR #7298, opened 2023-09-26
Initiate connection from server

nbolton:


SerhiiGadzhilov:
It was implemented under and merged under https://github.com/symless/synergy-core/pull/7222

PR #7361, opened 2024-06-28
Re-implement packaging for GitHub workflows (Linux)

nbolton:
https://symless.atlassian.net/browse/S1-1693
https://symless.atlassian.net/browse/S1-1694

To test on Ubuntu and Fedora:
1. `./scripts/install_deps.py`
2. `cmake -B build --preset=linux-release`
3. `cmake --build build -j8`
4. `./scripts/package.py`


aquacash5:
I don't think you should have to run `install_deps.py` to run `package.py`. I suggest adding a cli option that would create the venv if included. (ex `./scripts/package.py --make-venv`)

aquacash5 (changes requested):
Fedora doesn't produce a package after running `scripts/package.py`

aquacash5 (approved):
Nevermind there was some other issue on my vm

nbolton:
> I don't think you should have to run `install_deps.py` to run `package.py`. I suggest adding a cli option that would create the venv if included. (ex `./scripts/package.py --make-venv`)

I did consider this, but then realised that it's an edge case, so adding complexity seemed wasteful.

PR #7364, opened 2024-07-02
Format all source with Clang and introduce lint workflow

nbolton:
https://symless.atlassian.net/browse/S1-1695


nbolton:
Force merged as SonarCloud is detecting problems in re-formatted files.

PR #7382, opened 2024-07-10
Re-run `macdeployqt6` to copy missing Qt 6 dependencies

nbolton:
https://symless.atlassian.net/browse/S1-1714

To test:
1. Same as https://github.com/symless/synergy-core/pull/7379
2. ... but actually test the binary: `./build/bundle/Synergy.app/Contents/MacOS/synergy`

### DIFF
--- a/scripts/lib/mac.py
+++ b/scripts/lib/mac.py
@@ -1,5 +1,5 @@
 import dmgbuild  # type: ignore
-import os, time, json
+import os, time, json, shutil
 import lib.cmd_utils as cmd_utils
 import lib.env as env
 from lib.certificate import Certificate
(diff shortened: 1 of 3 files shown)


scripts/lib/mac.py
```diff
@@ -101,18 +101,16 @@ def package_env_vars():
...
+        shutil.rmtree(app_path)
+
     print("Building bundle...")
+
     # cmake build install target should run macdeployqt
     cmd_utils.run("cmake --build build --target install", shell=True, print_cmd=True)
 
-    bundle_bin_path = "Contents/MacOS/synergy"
-    cmd_utils.run(
-        'install_name_tool -add_rpath "@executable_path/../Frameworks" '
```
nbolton, line 110: I removed the call to `install_name_tool` because it wasn't actually doing anything.

aquacash5 (approved):
Installed DMG and GUI seems to work fine.

PR #7417, opened 2024-07-29
Save hostname on `connect` and verify in `secureConnect`

nbolton:
https://symless.atlassian.net/browse/S1-1740


nbolton:
Will re-visit later.

PR #7445, opened 2024-08-22
Update `config.yaml` to support Linux Mint build target

import-that:
Linux Mint is a Debian downstream, and it should be supported as such.

I ran a build in a Linux Mint 21.3 x86_64 vm that completed without errors, and running `./build/bin/unittests` garnered 100% test passage.


nbolton (approved):

nbolton:
Tested and working on Linux Mint 22. Thanks for the PR! :+1:

PR #7454, opened 2024-08-30
Rip out stupid ‘no Wayland’ screen

nbolton:
https://symless.atlassian.net/browse/S1-1820


nbolton:
Closing. I was wrong; when the 1.15 server runs on Wayland, it almost works but is buggy as the mouse behaves strangely when transitioning between server and client.

PR #7463, opened 2024-09-02
Scripts to delete garbage releases and tags

nbolton:
https://symless.atlassian.net/browse/S1-1801


nbolton:
Not merging as this these scripts have served their purpose.

PR #7470, opened 2024-09-03
Port EI fixes to use `pipe2` instead of `pipe` and `fcntl`

nbolton:
> just noticed that doing a pipe2(pipefd, O_NONBLOCK) might be smarter here

@whot Looks good, tested on Wayland and X, client and server with libei main and libportal main :+1: 

---

https://symless.atlassian.net/browse/S1-1812


nbolton:
@whot Looks good, tested on Wayland and X, client and server with libei main and libportal main 👍

Issue #7518, opened 2024-09-17
Holding a key and exiting screen of Linux Wayland KDE client gets key stuck


enzious:
### What type of Synergy user are you?

Customer

### Discussion

- [X] I want to report a bug (I don't want to start a discussion)

### Product name

Synergy 1

### Synergy version number

1.16.0 beta

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [X] Linux or Unix
- [ ] Other

### OS versions or Linux distros

- (Windows/Linux) (server)
- Gentoo Linux KDE Plasma 6 (client)

### Synergy configuration

- (windows/linux) (server)
- Gentoo Linux KDE Plasma 6 (client)

### What steps will reproduce the problem?

1. Start client and server
2. Move mouse to client screen
3. Hold key
4. Leave client screen

### Log output

```shell
[2024-09-16T13:37:51] ERROR: ei: 🪲  Bug: ei_device_keyboard_key: device is not emulating
```

enzious:
Not convinced it's just a KDE bug. Looking into the code a bit.

enzious:
- Updating to libei 1.3.0 did not resolve the issue.
- Commenting out the calls to `ei_device_stop_emulating(...)` mitigated the issue but resulted in 'share' dialog on re-entry.
- Commenting out code causing 'device is not emulating' log did not resolve the issue.

enzious:
I was right, here is what fixed it for me: https://github.com/enzious/deskflow/tree/fix/stuck-keys-on-leave

sithlord48:
Tested your fix in input-leap it works perfect please make a pr here for this !

enzious:
> Tested your fix in input-leap it works perfect please make a pr here for this !

That change may break edge cases. I'll submit a PR later that handles the issue better.

sithlord48:
I will update to be in sync with you changes. For both here and there.

nbolton:
Thanks for the bug report and PRs.

PR #7520, opened 2024-09-18
fix: stuck keys on leave

enzious:
Blocked by: https://github.com/deskflow/deskflow/pull/7522

---

On exiting a screen, deskflow attempts to unset keys that are down while leaving the screen. In some cases, such as with Wayland (EiScreen), the keys are reset but after the device emulation has already ended by `Screen::leave()`. This leaves the keys in the pressed state since the reset will fail after emulation stops, and may leave the keys stuck pressed until the environment implementing libei is restarted (such as with Kwin).

This pull request splits `Screen::leave()` into `Screen::canLeave()` and `Screen::leave()`, allowing for the keys to be reset in between.

Basically:
Check for screen leave allowed -> Reset keys -> Actually leave

Resolves: #7518

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -345,7 +345,9 @@ void EiScreen::enter() {
 #endif
 }
 
-bool EiScreen::leave() {
+bool EiScreen::canLeave() { return true; }
+
+void EiScreen::leave() {
   if (!is_primary_) {
     if (ei_pointer_) {
       ei_device_stop_emulating(ei_pointer_);
@@ -359,7 +361,6 @@ bool EiScreen::leave() {
   }
 
   is_on_screen_ = false;
-  return true;
 }
 
 bool EiScreen::setClipboard(ClipboardID id, const IClipboard *clipboard) {
--- a/src/lib/platform/XWindowsScreen.cpp
+++ b/src/lib/platform/XWindowsScreen.cpp
@@ -326,7 +326,15 @@ void XWindowsScreen::enter() {
   m_isOnScreen = true;
 }
 
-bool XWindowsScreen::leave() {
+bool XWindowsScreen::canLeave() {
+  // raise and show the window, required to grab mouse and keyboard
+  XMapRaised(m_display, m_window);
+
+  // see if grabbing the mouse and keyboard, if primary, is possible
+  return !(m_isPrimary && !grabMouseAndKeyboard());
+}
+
+void XWindowsScreen::leave() {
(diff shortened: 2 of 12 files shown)


enzious:
Side note - after rebasing on master, I'm getting the following error while attempting to build for Windows:

image

src/lib/platform/XWindowsScreen.cpp
```diff
@@ -326,7 +326,16 @@ void XWindowsScreen::enter() {
   m_isOnScreen = true;
 }
 
-bool XWindowsScreen::leave() {
+bool XWindowsScreen::canLeave() {
+  // see if grabbing the mouse and keyboard, if primary, is possible
+  if (m_isPrimary && !grabMouseAndKeyboard()) {
```
sithlord48, line 331: This should be this same logic.
`return !(m_isPrimary && !grabMouseAndKeybaord());`

sithlord48 (commented):
The initial commit can be dropped as its just changed completely by the second one. 
Otherwise looks pretty good. Testing on inputleap

nbolton:
Thanks for the PR, we're currently renaming the project and I'm in the process of restoring CI functionality.
Updates: https://github.com/deskflow/deskflow/discussions/7517
Once CI is working, we can get your PR merged in.

src/lib/platform/XWindowsScreen.cpp
```diff
@@ -326,7 +326,12 @@ void XWindowsScreen::enter() {
   m_isOnScreen = true;
 }
 
-bool XWindowsScreen::leave() {
+bool XWindowsScreen::canLeave() {
```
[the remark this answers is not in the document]
enzious: Interesting, I'll have to test this after work.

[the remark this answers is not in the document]
enzious: Actually, fix was pretty simple. Pushed it.

src/lib/platform/EiScreen.cpp
```diff
@@ -345,7 +345,9 @@ void EiScreen::enter() {
 #endif
 }
 
-bool EiScreen::leave() {
+bool EiScreen::canLeave() { return true; }
```
sithlord48, line 348: Future improvement: This should be our default from the base class

nbolton: Perhaps it's worth making an issue for this.

enzious: Should I just resolve it here in this PR?

sithlord48 (approved):

Issue #7523, opened 2024-09-18
Compatibility with Synergy broken


enzious:
### Project

Deskflow

### Deskflow version number

1.17.0

### Operating systems (OS)

- [X] Windows
- [X] macOS
- [X] Linux
- [X] BSD-derived
- [ ] Other

### OS versions/distros

- Windows
- Linux
- Mac OSX

### Deskflow configuration

- N/A

### What steps will reproduce the problem?

1. Start Deskflow Server/Client
2. Attempt to connect with Synergy Server/Client

### Additional information

It appears that the following name change to Deskflow breaks compatibility with Synergy:
https://github.com/deskflow/deskflow/blob/9dc6377bd921f64109581746ddef3f50cb54d25a/src/lib/deskflow/protocol_types.cpp#L21

Is this desired or should I open a PR to revert this particular change?

If it is desirable to maintain compatibility, then leaving it as Synergy should be fine since the 'protocol' is still basically Synergy.

PR #7524, opened 2024-09-18
fix: protocol compatibility with Synergy

sithlord48:
Fixes: #7523 
revert protocol hello to `Synergy`

 NOTE: For compatibility with Barrier / InputLeap we need to use `Barrier` 
TODO: Check for and Send the correct one based on what we are connecting to

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,11 @@
 # 1.17.0
 
+Bug fixes:
+
+- #7524 Restore protocol compatibility with Synergy
+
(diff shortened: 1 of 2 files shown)


nbolton (dismissed):

nbolton (approved):

nbolton:
> NOTE: For compatibility with Barrier / InputLeap we need to use `Barrier`

It would be great if this was a setting in the GUI.

Issue #7527, opened 2024-09-18
Add helpful comment to PR when lint fails


nbolton:
### Project

Deskflow

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] BSD-derived
- [X] Other

### OS versions/distros

n/a

### Deskflow configuration

n/a

### What problem are you trying to solve?

It's time consuming to manually apply formatting for each PR.

### What solution do you propose?

~~Modify GitHub action to auto format, commit append and rebase.~~
Add a comment to the PR instructing the author how to fix the problem.

nbolton:
@sithlord48 RFC: Lint comment idea...

[Mascot icon] Meow'dib: The sleeper must awaken, and fix these lint errors...

PR #7530, opened 2024-09-18
ci: add helpful comment to PR when lint fails

nbolton:
Fixes: https://github.com/deskflow/deskflow/issues/7527

TODO: Instead of commit ammdend and rebase (which GitHub doesn't really want you to do because of infinite loop bugs), we will add a comment to the PR instructing the author how to fix the problem.

### DIFF
--- a/.github/workflows/lint-source-code.yml
+++ b/.github/workflows/lint-source-code.yml
@@ -19,6 +19,9 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.ref || github.ref }}
+          fetch-depth: 2
 
       - name: Setup Python venv
         uses: ./.github/actions/init-python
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-cmake_minimum_required(VERSION 3.5)
+  cmake_minimum_required(VERSION 3.5)
```
sithlord48, line 17: Please remove this unneeded change

sithlord48 (changes requested):

CMakeLists.txt
nbolton: This is to trigger the linter. I'll let you know when the PR is out of draft for you to review.

sithlord48: Looks like its not pushing back to the repo

nbolton: Oh I've been having fun with that one! 

It does, but I keep pushing up the lint error to re-trigger it.

See #7529 and #7528 for my previous mistakes; turns out if you --amend to fetch depth 1, it pushes changes to all files. I figured out the solution is to add:
```
        with:
          ref: ${{ github.event.pull_request.head.ref || github.ref }}
          fetch-depth: 2
```
to
```
      - name: Checkout
        uses: actions/checkout@v4
```

sithlord48: why not depth of 0 ?

nbolton: I don't think that depth is needed; `0` brings in all the tags which I don't need here. Depth of `2` works well.

The problem I have now is that GitHub workflows can't be triggered by GitHub (e.g. if GitHub does `git push`), so once the linter makes it's fixes, the workflow doesn't re-run and gets stuck... not quite sure where to go from here. I'll sleep on it.

image

sithlord48:
Moved to discussion https://github.com/deskflow/deskflow/discussions/7536

PR #7543, opened 2024-09-20
ci: give write permission for PR comments

nbolton:
When the comment action is triggered by a PR originating from an external fork (i.e. a contributor) the error, `Resource not accessible by integration` shows because the worklow doesn't automatically have write permission to the PR.

Example: https://github.com/deskflow/deskflow/actions/runs/10964481892?pr=7511

We need to explicitly grant this permission with:
```
permissions:
  pull-requests: write
```

{no-changelog-lint}


sithlord48:
Not sure how i feel about this i think id rather just have the user run the format command and skip the whole apply this diff part. . Instead have the bot say it needs linting run this script and / or copy it into your git hooks

sithlord48 (approved):

Issue #7545, opened 2024-09-20
`error: macro "__DATE__" might prevent reproducible builds`


nbolton:
### Project

Deskflow

### Deskflow version number

1.17.0

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [X] Linux
- [ ] BSD-derived
- [ ] Other

### OS versions/distros

Debian

### Deskflow configuration

n/a

### What steps will reproduce the problem?

1. Build on Debian stable

### Log output

```shell
In file included from /build/deskflow-1.16.1~beta+r1/src/lib/synergy/App.cpp:30:
/build/deskflow-1.16.1~beta+r1/src/./lib/common/copyright.h:32:28: error: macro "__DATE__" might prevent reproducible builds [-Werror=date-time]
   32 |   const std::string date = __DATE__;
      |                            ^~~~~~~~
cc1plus: all warnings being treated as errors
make[3]: *** [src/lib/synergy/CMakeFiles/synlib.dir/build.make:79: src/lib/synergy/CMakeFiles/synlib.dir/App.cpp.o] Error 1
```

### Additional information

Reported by WrathOfAchilles on IRC

nbolton:
The solution could be either:
1. Make `-Werror` optional (so package maintainers don't need to patch it out)
2. Add a `#pramga` ignore warning on the offending line of code (if possible)

sithlord48:
This feature should go.. we can not make a reproducible build if the contents of the file change based only on when it was built.

PR #7547, opened 2024-09-20
build: set `BUILD_DATE` instead of using `__DATE__`

nbolton:
Fixes: #7545

We are using `__DATE__` macro which does not respect `SOURCE_DATE_EPOCH` making it difficult for downstream packagers to create reproducible builds.

Using `string(TIMESTAMP BUILD_DATE "%Y-%m-%d" UTC)` in CMake respects [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/). ~~And for older versions of CMake (Debian Trixie for instance uses CMake 3.30.0) we have the less desirable way.~~

> only respects SOURCE_DATE_EPOCH since version 3.8.0 ([April 2017](https://cmake.org/pipermail/cmake-developers/2017-April/029946.html)). Note that the final argument UTC is required or the timestamp may vary between timezones.

Edit: The min version for CMake is set to 3.8 for this reason.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,13 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-cmake_minimum_required(VERSION 3.5)
+# Why CMake 3.8?
+# This allows package maintainers to create reproducible builds:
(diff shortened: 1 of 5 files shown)


sithlord48 (changes requested):
This can be done better with `configure_file` . Not sure how i feel about having to encode the build date into the application.

sithlord48 (approved):

PR #7551, opened 2024-09-21
feat: use new Deskflow graphics

nbolton:
The Synergy graphics are still everywhere.

Replace with the new Deskflow logo.


res/deskflow.svg
sithlord48, line 1: can we get the svg larger then 20x20

sithlord48 (changes requested):

res/deskflow.svg
nbolton: Sure, what size do you want it?

nbolton: I went with 400x400, I think that's big enough.

sithlord48 (approved):

Issue #7558, opened 2024-09-21
CI: `Error: Resource not accessible by integration`


nbolton:
### Project

Deskflow

### Deskflow version number

1.17.0

### Build type

Local developer build (built it yourself)

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] BSD-derived
- [X] Other

### OS versions/distros

n/a

### Deskflow configuration

n/a

### What steps will reproduce the problem?

1. Create a PR from an external contrib fork
1. Cause Clang or CMake format error
2. Wait for lint workflow to fail

Expected: Comment with patch
Actual:
```
Run marocchino/sticky-pull-request-comment@v2
Error: Resource not accessible by integration
```

### Additional information

According to the `marocchino/sticky-pull-request-comment` [docs](https://github.com/marocchino/sticky-pull-request-comment), adding this to the job should solve the issue:
```
permissions:
  pull-requests: write
```

But that doesn't seem to be the case.

From the [GitHub docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token):
> You can use the permissions key to add and remove read permissions for forked repositories, but typically you can't grant write access. The exception to this behavior is where an admin user has selected the Send write tokens to workflows from pull requests option in the GitHub Actions settings. For more information, see "[Managing GitHub Actions settings for a repository](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#enabling-workflows-for-private-repository-forks)."

I cannot see the 'Send write tokens to workflows from pull requests' option anywhere.

nbolton:
A little frustrating; GitHub doesn't allow PRs from forks write access on PRs (so they can't create comments). Apparently, using `pull_request_target` solves this, but I have not found this to be the case; the workflow never starts.

nbolton:
I think not; tentatively closing. My memory is foggy here, but I suspect that I might have fixed this a couple of months ago by introducing [`ci-comment.yml`](https://github.com/deskflow/deskflow/blob/master/.github/workflows/ci-comment.yml) which runs from `master` using `workflow_run`. I have a vague recollection of following the article I mentioned before:
[...]

PR #7560, opened 2024-09-21
ci: use pull_request_target for lint workflows

nbolton:


nbolton:
Closing, doesn't seem to solve the problem (https://github.com/deskflow/deskflow/issues/7558)

Issue #7540, opened 2024-09-20
`vcpkg` does not use cached `openssl` build


nbolton:
### Project

Deskflow

### Deskflow version number

1.17.0

### Operating systems (OS)

- [x] Windows
- [ ] macOS
- [ ] Linux
- [ ] BSD-derived
- [ ] Other

### OS versions/distros

n/a

### Deskflow configuration

n/a

### What steps will reproduce the problem?

1. Trigger CI 
2. Wait for Windows CI to run

Expected: Cached `openssl` build is used if the version number hasn't changed; the build should normally take around 5mins.
Actual: It's rebuilt every time even when cache is restored, the build always takes 15 mins.

### Log output

```shell
Running: vcpkg install
...
Fetching registry information from https://github.com/microsoft/vcpkg (HEAD)...
Detecting compiler hash for triplet x64-windows...
Compiler found: C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.41.34120/bin/Hostx64/x64/cl.exe
The following packages will be rebuilt:
    openssl[core,tools]:x64-windows@3.3.1#1 -- C:\Users\runneradmin\AppData\Local\vcpkg\registries\git-trees\038ce0eca1d7e3f285a8de86c75bb8b89255994d
```

### Additional information

This was working before we renamed the project to Deskflow. The cache seems to be uploaded and restored; it is the expected size. Maybe we need to build openssl as a separate package and pull it in that way rather than using `vcpkg`? We were previously using the "Shining Light" OpenSSL distribution (via `choco`), but their website URLs are unreliable and occasionally get changed without warning (so then you have to wait for all the package managers to update their URLs, meanwhile CI is failing).

nbolton:
@sithlord48 Any thoughts on this problem? I thought your `vcpkg` and CI expertise might provide some insight.  I can't see why changing the repo name has buggered up `vcpkg`'s interpretation of the restored cache. It worked before when the repo was `symless/synergy`; `vcpkg` saw the cache and said something like "hash looks good, using built openssl" whereas now it's saying "hmm, nope, don't like this... rebuilding openssl" which adds an annoying 10 mins to the Windows CI build every time.

PR #7566, opened 2024-09-22
ci: remove `restore-keys` for `vcpkg` and use local bootstrap

nbolton:
Fixes: #7540

Using the GitHub system-installed `vcpkg` seems to cause issues with caching. Instead, we will ignore the system-installed `vcpkg` and use our own local bootstrapped one; this seems to allow us greater control over caching. Using `restore-keys` seems to have no benefit, so this was removed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,10 +68,9 @@ jobs:
         uses: actions/cache@v4
         with:
           path: |
+            vcpkg
             vcpkg_installed
-            ${{ env.LOCALAPPDATA }}/vcpkg
           key: vcpkg-${{ runner.os }}-${{ hashFiles('vcpkg.json', 'vcpkg-configuration.json') }}
-          restore-keys: vcpkg-${{ runner.os }}
 
       # Should only restore the .venv directory from cache.
       - name: Init Python venv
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -71,7 +71,6 @@ jobs:
             vcpkg_installed
             ${{ env.LOCALAPPDATA }}/vcpkg
           key: vcpkg-${{ runner.os }}-${{ hashFiles('vcpkg.json', 'vcpkg-configuration.json') }}
-          restore-keys: vcpkg-${{ runner.os }}
```
nbolton, line 74: @sithlord48 Hmm, removing this line doesn't seem to solve the issue.
```
restore-keys: vcpkg-${{ runner.os }}
```

It's still rebuilding: https://github.com/deskflow/deskflow/actions/runs/10981219510/job/30488232214?pr=7566

nbolton: I'll try another approach.

nbolton: Using local `vcpkg` instead of system seems to fix the cache issue, so I'll run with that.

I will keep the change to remove `restore-keys` since I believe it has no effect.

sithlord48 (approved):

PR #7567, opened 2024-09-22
ci: use `github-script` instead of `download-artifact`

nbolton:
Blocked by: #7569 

Follow-on from: #7563

Fixes: https://github.com/deskflow/deskflow/issues/7558


sithlord48 (approved):

nbolton:
Actually, this is probably not the right way to go. I merged a more generic workflow, `ci-comment.yml`, which can be used for anything (not just lint results).

Issue #7568, opened 2024-09-22
Use GitHub step/job summary instead of PR comments


nbolton:
### Project

Deskflow

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] BSD-derived
- [X] Other

### OS versions/distros

n/a

### Deskflow configuration

n/a

### What problem are you trying to solve?

The unit/integ test PR comments have 3 disadvantages:
- They don't work for external fork PRs (due to permissions)
- They are quite noisy
- There are parallel bugs

![image](https://github.com/user-attachments/assets/a5833455-664e-4323-a6c4-aba9256d1c76)

### What solution do you propose?

Use the GitHub job/step summary instead so that the summary is contained on the workflow job summary.

PR #7573, opened 2024-09-22
ci: create SonarScanner send file workflow

nbolton:
Relates to: https://github.com/deskflow/deskflow/issues/7568 and https://github.com/deskflow/deskflow/issues/7558

:warning: As this is a `workflow_run`, it must be landed in master to test.


nbolton:
Abandoning: On second thoughts, bad idea, since `sonarscanner` needs the source of the PR checked out into the same dir in order for it to analyze the build.

Issue #7443, opened 2024-08-20
openssl config file not found: res/openssl/synergy.conf


JustSxm:
### Discussion

- [X] Yes, and I still want to report a bug

### What type of Synergy user are you?

Community user

### Product name

Synergy 1

### Synergy version number

1.15.0 r3

### Operating systems (OS)

- [X] Windows
- [ ] macOS
- [ ] Linux or Unix
- [ ] Other

### OS versions or Linux distros

Windows 11

### Synergy configuration

Windows 11 Server.

### What steps will reproduce the problem?

1. Start synergy.exe after building

### Log output

```shell
Sorry, a fatal error has occurred and the application must now exit.
Please report a bug and copy/paste the following error:
v1.15.0
openssl config file not found:  "res/openssl/synergy.conf"
```

### Additional information

the "res" folder does not seem to exist in the build/bin directory. Perhaps missing information?

nbolton:
Thanks for reporting this. I'll take a look and see if I can reproduce.

nbolton:
Hmm, ok reproduced it. 

So, the issue here is that if `OpenSSL\synergy.conf` doesn't exist (which is expected in production when the app is installed), the app then looks in the current working dir for the path `res/openssl/synergy.conf`. Therefore, if you run it like so:
```
synergy-core> build/bin/synergy
```

...you won't get the error.

If you `cd` into the dir (therefore changing the current working dir) and run it, then you get the error:
```
synergy-core> cd .\build\bin
synergy-core\build\bin> .\synergy.exe
```

I'll have a think about what the best solution should be.

elegos:
Hello,

I've got the same issue on 1.16.1-beta-r1. I wanted to compile on a machine and run on another, and wanted to copy the bin folder only. Maybe as a workaround we could copy `res\openssl\synergy.conf` in the build folder after the compiling process?

nbolton:
> Maybe as a workaround we could copy res\openssl\synergy.conf in the build folder after the compiling process?

Sure, that's one solution. Alternatively, we could add an env var to specify the OpenSSL `.conf` path or allow this to be configured in the GUI. Would either of those ideas? 

It might be a good idea to avoid adding extra files to the bin dir unless we absolutely have to.

elegos:
Once packaged, the application won't (shouldn't?) include the sources, and thus even if we specify the path, we might not have the configuration file at all.

nbolton:
[...]

PR #7579, opened 2024-09-23
Relocate commercial code downstream

nbolton:
Blocked by: #7595

The final step to rename the project to Deskflow (#7517) is to move all commercial code downstream. 

Indication of commercial code removed:
image

As well as removing any commercial code (serial key dialog, etc). This involves making it simple for downstream project forks to re-skin the app, which includes making Deskflow usable as a CMake subproject.

**Important:** Making the app name _so_ variable is perhaps not ideal as it adds a degree of complexity, though there is an advantage of not repeating the app name all over the place. Our long term goal is to lib-ize the guts so that the binaries can be tailored to the downstream forks (rather than the upstream binaries being skinnable). This PR is a temporary solution to allow downstream forks to apply their own makeup to binaries. Discussion: #7590

{no-sonar}


sithlord48 (commented):
You should always use `CMAKE_SOURCE_DIR` , `CMAKE_BINARY_DIR` or `CMAKE_CURRENT_BINARY_DIR` or `CMAKE_CURRENT_BINARY_DIR` to find these paths. for what reason do you feel like you need to change them?

res/dist/flatpak/deskflow.yml
```diff
@@ -1,4 +1,4 @@
-app-id: com.symless.Deskflow
+app-id: org.deskflow.Deskflow
```
sithlord48, line 1: This is incorrect, the id should be `io.github.deskflow.Deskflow`.

nbolton: Please help me to understand why.

res/dist/flatpak/deskflow.yml
```diff
@@ -1,4 +1,4 @@
...
 runtime: org.kde.Platform
 runtime-version: "5.15-21.08"
```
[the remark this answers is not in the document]
nbolton: Thanks, I was planning on fixing Flatpack in a separate PR. I'm going to leave this for now.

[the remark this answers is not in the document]
nbolton: https://github.com/deskflow/deskflow/issues/7593

res/dist/flatpak/deskflow.yml
nbolton: Almost ready to submit PR for review, so let me know about this if you have time. If not we can discuss in another PR.

sithlord48:
Converted to a draft until your ready here

CMakePresets.json
```diff
@@ -41,7 +41,7 @@
       "generator": "Unix Makefiles"
     },
     {
-      "name": "macos",
```
sithlord48, line 44: Why change this from macos ? I would think that would be pretty common

CMakePresets.json
```diff
@@ -41,7 +41,7 @@
...
+      "name": "mac",
       "hidden": true,
       "inherits": "base",
       "generator": "Unix Makefiles"
```
sithlord48, line 47: On mac os you may want to use the [XCode](https://cmake.org/cmake/help/latest/generator/Xcode.html)  generator

Doxyfile
```diff
@@ -51,7 +51,7 @@ PROJECT_BRIEF          =
 # pixels and the maximum width should not exceed 200 pixels. Doxygen will copy
 # the logo to the output directory.
 
-PROJECT_LOGO           = res/deskflow.svg
+PROJECT_LOGO           = res/app.svg
```
sithlord48, line 54: I am not sure i am a fan of the general binary names.

cmake/Build.cmake
```diff
@@ -14,8 +14,22 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 macro(configure_build)
+
+  set(CMAKE_CXX_STANDARD 20)
+  set(CMAKE_CXX_EXTENSIONS OFF)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
+  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/bin")
```
sithlord48, line 21: This is generally not needed when your using cmake it will take care of the various parts finding one anther for you. This is also not making deployment any easier or harder.

cmake/Build.cmake
```diff
@@ -14,8 +14,22 @@
...
+  set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/lib")
+
+  if(APPLE)
+    message(STATUS "Configuring for Apple")
+    set(CMAKE_OSX_DEPLOYMENT_TARGET "12.0")
```
sithlord48, line 26: Qt6 needs at least   `11`

cmake/Build.cmake
```diff
@@ -37,6 +51,11 @@ macro(set_build_date)
   add_definitions(-DBUILD_DATE="${BUILD_DATE}")
 endmacro()
 
+macro(configure_file_shared)
+  configure_file(src/lib/gui/gui_config.h.in
+                 ${PROJECT_BINARY_DIR}/config/gui_config.h)
```
sithlord48, line 56: Perhaps prefix the source with `${PROJECT_SOURCE_DIR}` ? .

cmake/Build.cmake
```diff
@@ -58,28 +77,39 @@ endmacro()
...
       VERBATIM
       COMMENT "Copying files to bin dir")
 
     add_dependencies(
       run_post_build
-      deskflow
-      deskflowc
-      deskflows
-      deskflowd)
+      ${GUI_BINARY_NAME}
```
sithlord48, line 109: Not sure how i feel about this.

cmake/Definitions.cmake
```diff
@@ -16,46 +16,20 @@
...
 
   set(INTEG_TESTS_BIN integtests)
   set(UNIT_TESTS_BIN unittests)
 
-  if("${VERSION_URL}" STREQUAL "")
-    set(VERSION_URL "https://api.deskflow.org/version")
-  endif()
-  add_definitions(-DDESKFLOW_VERSION_URL="${VERSION_URL}")
-
   if(NOT "$ENV{GIT_SHA}" STREQUAL "")
```
sithlord48, line 26: Why is this an Environmental Variable ? 

What if i didn't get the code via git ?

cmake/Definitions.cmake
```diff
@@ -16,46 +16,20 @@
...
     # Shorten the Git SHA to 8 chars for readability
     string(SUBSTRING "$ENV{GIT_SHA}" 0 8 GIT_SHA_SHORT)
     message(STATUS "Short Git SHA: ${GIT_SHA_SHORT}")
     add_definitions(-DGIT_SHA_SHORT="${GIT_SHA_SHORT}")
```
sithlord48, line 30: consider using configure file to make a version header

sithlord48 (changes requested):
I got a few things then i stopped reviewing the rest. Most of the things i think we can fix in additional prs (at least the bit i did look at) but i am not sure about the naming changes using cmake vars being helpful or just more vars i need to set .

sithlord48:
Can you split the fix for #7443 ? (sorry)

CMakePresets.json
nbolton: I'm easy either way. When I originally wrote that code, `macos` felt weird but I ran with it and I've been meaning to try it with just "mac" to see if it's more readable. I figured that `mac` looked a bit better, but it was purely subjective. I can change it back if you like.

nbolton:
> Can you split the fix for #7443 ? (sorry)

I'll take a look :slightly_smiling_face:

Doxyfile
nbolton: Can you think of any disadvantages of this image name? Thinking about it from a contributors perspective.

cmake/Build.cmake
nbolton: Hmm, I can't remember why I did this but I seem to remember it being important for some reason. Happy to look at taking this out in a different PR?

cmake/Build.cmake
nbolton: I believe we have to set it to 12 because of C++20, but I can't remember exactly. Again, this isn't really related to this PR, just moving code around. Ok to look at it in a separate PR?

cmake/Build.cmake
nbolton: Good catch. Will take a look.

cmake/Build.cmake
nbolton: Me too, but I can't think of a better way to fix the problem now. This feels like part of a much bigger problem that we need to solve separately.

nbolton:
> I got a few things then i stopped reviewing the rest. Most of the things i think we can fix in additional prs (at least the bit i did look at) but i am not sure about the naming changes using cmake vars being helpful or just more vars i need to set . 

Sounds good. I think we can iterate on this in other PRs. Nothing is set in stone and getting this done is probably more important than getting it perfect.

.github/workflows/ci.yml
```diff
@@ -19,20 +19,43 @@ on:
...
 env:
-  GIT_SHA: ${{ github.sha }}
-  DESKFLOW_PRODUCT_NAME: ${{ vars.DESKFLOW_PRODUCT_NAME }}
-  DESKFLOW_PACKAGE_PREFIX: ${{ vars.DESKFLOW_PACKAGE_PREFIX || 'deskflow' }}
-  PACKAGE_BUILD: ${{ !github.event.pull_request.draft }}
-  PACKAGE_UPLOAD: ${{ !github.event.pull_request.draft }}
+  GIT_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
+  PACKAGE_PREFIX: "deskflow"
+  PACKAGE_PATH: ./dist
+  PACKAGE_RETENTION_DAYS: 3
```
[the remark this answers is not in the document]
nbolton: I like to follow the DRY (don't repeat yourself) principle. So if there's something repeated several times, I prefer to set it as a constant. Env vars aren't awesome for consts, but I don't believe there is another way in GitHub workflows.

cmake/Definitions.cmake
sithlord48: i will likey want to use what i do else where for versions so i may do this in time as well.

cmake/Build.cmake
sithlord48: Ah yes idk about c++20 's version im ok with using 12 in that case.

.github/workflows/ci.yml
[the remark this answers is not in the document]
sithlord48: Sorry what i meant was that there is almost no reason to set the retention time.

CMakePresets.json
sithlord48: i feel like macos or some variant of it MacOS macOS what have you is the norm. so i would just have kept `macos`

CMakePresets.json
sithlord48: ill have to see what the default is. I know we need to then open that in xcode it it will produce a project and build files that xcode (the native) toolchain will consume . you should then be able to use cmake --build to build it

Doxyfile
sithlord48: well its the project logo so i think if i was forking and making a new one of my first things would be to replace this and using the project name.svg seams to make sense to me .. same idea w/ the binaries

cmake/Definitions.cmake
sithlord48: just seams weird . can fix it later its not hurting anything

Doxyfile
nbolton: I was thinking about it this way: For downstream forks, it'd be easier if they didn't have to rename files, that way less would get out of sync. IMO, using the project name for any file is a bit vague and lazy. It seems to be the kind of thing a junior developer tends to do. How about we call it `logo.svg` instead? I'm a big fan of naming files after their purpose.

nbolton:
@sithlord48 Quick thought, maybe the commit prefix is `chore!:` and not `ci:` (perhaps not `feat!:`)

nbolton:
> Can you split the fix for #7443 ? (sorry)

Done 👍 #7595

sithlord48 (commented):
Land it if you want I think the generic names are not needed but do you.

nbolton:
> Land it if you want I think the generic names are not needed but do you.

We can change them back if they don't have a positive effect.

Issue #7513, opened 2024-09-13
Unable to install Qt 6.7 with deps script on Windows


dk-mushiyoke:
### What type of Synergy user are you?

Community user

### Discussion

- [ ] I want to report a bug (I don't want to start a discussion)

### Product name

Synergy 1

### Synergy version number

1.16.0-beta

### Operating systems (OS)

- [X] Windows
- [ ] macOS
- [ ] Linux or Unix
- [ ] Other

### OS versions or Linux distros

Windows 10

### Synergy configuration

host: opensuse slowroll (have to duplicate config.yaml section for tumbleweed but works just fine otherwise)
client: windows 10

### What steps will reproduce the problem?

1. Run `python scripts/install_deps.py` from the Build Quick Start section for windows in readme file

### Log output

```shell
C:\Users\dkong\Tools\synergy>python scripts/install_deps.py
Using virtual environment for: C:\Users\dkong\Tools\synergy\scripts\install_deps.py
Updating pip...
Running: C:\Users\dkong\Tools\synergy\.venv\Scripts\python.exe -m pip install --upgrade pip
Requirement already satisfied: pip in c:\users\dkong\tools\synergy\.venv\lib\site-packages (24.2)
Installing required modules...
Running: C:\Users\dkong\Tools\synergy\.venv\Scripts\python.exe -m pip install -e scripts
Obtaining file:///C:/Users/dkong/Tools/synergy/scripts
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Preparing editable metadata (pyproject.toml) ... done
Requirement already satisfied: cmake-format in c:\users\dkong\tools\synergy\.venv\lib\site-packages (from scripts==0.0.1) (0.6.13)
Requirement already satisfied: clang-format in c:\users\dkong\tools\synergy\.venv\lib\site-packages (from scripts==0.0.1) (18.1.8)
Requirement already satisfied: python-dotenv in c:\users\dkong\tools\synergy\.venv\lib\site-packages (from scripts==0.0.1) (1.0.1)
Requirement already satisfied: pyyaml in c:\users\dkong\tools\synergy\.venv\lib\site-packages (from scripts==0.0.1) (6.0.2)
Requirement already satisfied: colorama in c:\users\dkong\tools\synergy\.venv\lib\site-packages (from scripts==0.0.1) (0.4.6)
[...]

PR #7585, opened 2024-09-24
ci: lock qt version to 6.7.2 on win/mac

nbolton:
Fixes: #7513

For some reason `aqt` is trying to download `6.7.3` even though it doesn't exist in [`official_releases/qt/6.7`](https://qt.mirror.constant.com/official_releases/qt/6.7/) on the Qt mirror.

Note: We have discussed using `vcpkg` instead of `aqt` on Windows, but this PR fixes the problem for now so we should land it before looking at using `vcpkg`. See: [Discussion on `vcpkg` vs `aqt`](https://github.com/deskflow/deskflow/discussions/7586)

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -12,7 +12,7 @@ config:
       # already installed on the Windows runner.
       command: if not defined CI (winget install ninja-build.ninja cmake)
       qt:
-        version: 6.7
+        version: 6.7.2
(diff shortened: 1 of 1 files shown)


sithlord48 (approved):
Will do what it says

PR #7592, opened 2024-09-26
ci: Disable SonarCloud PR comment

### LINKED ISSUE #7583: Move SonarCloud status to job summary (if possible)
### Project

Deskflow

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] BSD-derived
- [X] Other

### OS versions/distros

n/a

### Deskflow configuration

n/a

Every time I push a change to my PR, SonarCloud alerts me with a message.

image

Disable the PR comment alert, and instead query the API from the SonarCloud workflow, and share the analysis status as a job summary. If, and only if there are problems, then update the mono-summary-comment on the PR indicating that there is a SonarCloud analysis problem to deal with.

nbolton:
Fixes #7583


nbolton:
On second thoghts, this can be done in the SonarCloud admin GUI, so I'll do it there instead.

PR #7597, opened 2024-10-01
build: Bump min `libei` and `libportal` versions

nbolton:
Rationale: `libei` 1.3 and `libportal` 0.8 are required for non-X support on both server and client, therefore setting these as minimum versions makes sense.

### DIFF
--- a/cmake/Libraries.cmake
+++ b/cmake/Libraries.cmake
@@ -13,8 +13,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-set(LIBEI_MIN_VERSION 1.2.1)
-set(LIBPORTAL_MIN_VERSION 0.6)
+set(LIBEI_MIN_VERSION 1.3)
+set(LIBPORTAL_MIN_VERSION 0.8)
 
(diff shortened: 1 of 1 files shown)


sithlord48 (approved):

cmake/Libraries.cmake
```diff
@@ -13,8 +13,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-set(LIBEI_MIN_VERSION 1.2.1)
```
nbolton, line 16: @whot @sithlord48 Hmm, I just noticed that Ubuntu 24 has 1.2.1. Maybe we should drop this back down to 1.2.1?

sithlord48:
Nope  24.10 will be the first we can build in w/o a ppa

nbolton:
> Nope 24.10 will be the first we can build in w/o a ppa

Not sure what you mean.

> Confirmed. Works fine on Ubuntu 24.04 with `libei` 1.2.1 as both client and server.

Works for me on Ubuntu 24.04.

sithlord48:
24.10 will be the first version where all our dependencies are in distro .

PR #7601, opened 2024-10-02
Bump version to 1.18.0

nbolton:


nbolton:
@sithlord48 How about we land this for now, and then make sure the next thing we work on is moving the version logic into CMake? My concern is that we don't want master to be confused with 1.17.0, which is out in the wild now and people are referencing that version number in bug reports.

sithlord48 (approved):

PR #7608, opened 2024-10-02
Use unified binary when `BUILD_UNIFIED` set

nbolton:
When the `BUILD_UNIFIED` option is set, the individual server/client binaries don't exist (only `$CORE_BINARY_NAME` does).

```
CMake Error at cmake/Build.cmake:61 (add_dependencies):
  The dependency target "synergyc" of target "run_post_build" does not exist.
Call Stack (most recent call first):
  CMakeLists.txt:33 (post_config_all)

CMake Error at cmake/Build.cmake:61 (add_dependencies):
  The dependency target "synergys" of target "run_post_build" does not exist.
Call Stack (most recent call first):
  CMakeLists.txt:33 (post_config_all)
```

This PR uses `CORE_BINARY_NAME` when `BUILD_UNIFIED` is set.


cmake/Build.cmake
```diff
@@ -104,12 +104,17 @@ macro(post_config_all)
...
 
-    add_dependencies(
-      run_post_build
-      ${GUI_BINARY_NAME}
-      ${CLIENT_BINARY_NAME}
-      ${SERVER_BINARY_NAME}
-      ${DAEMON_BINARY_NAME})
+    if(BUILD_UNIFIED)
+      add_dependencies(run_post_build ${GUI_BINARY_NAME} ${CORE_BINARY_NAME}
+                       ${DAEMON_BINARY_NAME})
```
nbolton, line 109: This is a big awkward, it's because `max_pargs_hwrap` is set to `4` in `cmake-format.yaml`, so it only wraps when there are 4 args.

PR #7615, opened 2024-10-04
Update `config.yaml` to support Linux KDE Neon build target

dmitrii-galantsev:


dmitrii-galantsev:
on hold since I ran into issues compiling for arch for steam deck.. So i don't have a machine to connect to! :cry:

nbolton:
We are planning to move the setup script to a .sh file instead of .py so it's easier to run. Python is a bit painful for some people to setup. I'm thinking a good approach might be to fall back to the ID-LIKE property, eg since all Debian-like distros the apt command and package names seem to be the same.

PR #7622, opened 2024-10-06
Use modern Qt connections in GUI

sithlord48:
Requires: #7624 Landed first To Pass Lint
Update the Qt Connections to use the "Qt5" Style.

### DIFF
--- a/src/gui/src/AddClientDialog.cpp
+++ b/src/gui/src/AddClientDialog.cpp
@@ -39,25 +39,33 @@ AddClientDialog::AddClientDialog(const QString &clientName, QWidget *parent)
   m_pButtonLeft->setIcon(icon);
   m_pButtonLeft->setIconSize(IconSize);
   gridLayout->addWidget(m_pButtonLeft, 2, 0, 1, 1, Qt::AlignCenter);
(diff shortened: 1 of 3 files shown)


nbolton:
Another great PR. Also been meaning to do this so I'm happy to see modern connections.

nbolton:
> gui: use modern connections

I think if we're going to stick to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) then perhaps we shouldn't deviate from the standard. 

So, this should probably be `chore:` instead of `gui:`.

nbolton (changes requested):
> gui: use modern connections

I think if we're going to stick to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) then perhaps we shouldn't deviate from the convention. 

So, this should probably be `chore:` instead of `gui:`

nbolton:
> Requires: https://github.com/deskflow/deskflow/pull/7624 Landed first To Pass Lint

I wonder if maybe we should treat the code style change as separate and unrelated, and merge this PR inline with the current established standards.

nbolton (approved):

PR #7624, opened 2024-10-07
Adjust Clang-format linter rules

sithlord48:
Blocked by: https://github.com/deskflow/deskflow/discussions/7626

 - Adjust the clang format
   - 120 Column Limit
   - BlockIndent for align after opening bracket
   - No Single line functions
   - Spelled out break rules


nbolton:
ColumnLimit 0?

nbolton:
How's that PR title? A bit more comprehensive I think. Quite a big change so it'll take me time to review.

.clang-format
```diff
@@ -1,28 +1,105 @@
-# Important: Use the same version of clang-format as our linter, which is:
+---
+# SPDX-FileCopyrightText: 2024 Chris Rizzitello <sithlord48@gmail.com>
```
nbolton, line 2: @sithlord48 Did you mean to use your @kde.org email?

src/gui/src/Action.cpp
```diff
@@ -21,133 +21,128 @@
...
-          if (i != screens.size() - 1)
-            text += ":";
+    : m_KeySequence()
+    , m_Type(keystroke)
+    , m_TypeScreenNames()
+    , m_SwitchScreenName()
+    , m_SwitchDirection(switchLeft)
+    , m_LockCursorMode(lockCursorToggle)
+    , m_ActiveOnRelease(false)
+    , m_HasScreens(false)
```
nbolton, line 38: Personally, I'm not a huge fan of this style, but I'll roll with it since it's the KDE style.

src/gui/src/ActionDialog.cpp
```diff
@@ -27,88 +27,87 @@
...
+    setupUi(this);
+
+    // work around Qt Designer's lack of a QButtonGroup; we need it to get
+    // at the button id of the checked radio button
+    QRadioButton *const typeButtons[] = {m_pRadioPress,
+                                         m_pRadioRelease,
+                                         m_pRadioPressAndRelease,
+                                         m_pRadioSwitchToScreen,
+                                         m_pRadioSwitchInDirection,
+                                         m_pRadioLockCursorToScreen,
```
nbolton, line 47: This level of indentation makes me quite sad, but if you feel strongly about this I'm happy to go with it.

nbolton:
:warning: Word of Caution: Changing the code style to accommodate a contributor's personal preference could have broader implications:

- Precedent for Contributor Bias: Adjusting style for one contributor may set a precedent, increasing friction when others request changes to accommodate their preferences.

- Impact on Current Contributors: Those already accustomed to the LLVM style might find the switch frustrating, affecting productivity and possibly reducing engagement.

src/gui/src/ActionDialog.cpp
sithlord48: Yeah im not a fan also but if we gotta lint. I find this generally more readable.

.clang-format
sithlord48: Nope, I get way to much email there already

whot:
fwiw, I'm happy enough reviewing the odd patch related to the code I introduced and the odd higher-level comment but I cannot keep up with every pull requests. Please only tag me on issues/MRs that really matter, otherwise they'll just get lost in the noise.

nbolton:
> Maybe [just] ColumnLimit 0?

Hmm, no, that does something weird as [others are reporting](https://stackoverflow.com/questions/47683910/can-you-set-clang-formats-line-length), e.g.:

```diff
diff --git a/src/gui/src/Action.h b/src/gui/src/Action.h
index 63c1ff1d..2ddb27d7 100644
--- a/src/gui/src/Action.h
+++ b/src/gui/src/Action.h
@@ -45,8 +45,13 @@ public:
     mouseUp,
     mousebutton,
   };
-  enum SwitchDirection { switchLeft, switchRight, switchUp, switchDown };
-  enum LockCursorMode { lockCursorToggle, lockCursonOn, lockCursorOff };
+  enum SwitchDirection { switchLeft,
+                         switchRight,
```

gabeklavans:
I would be hesitant to overwrite the line history for the whole codebase like this.

What are thoughts on just setting the style guide and only having new changes be format-checked on PRs? Either by git diff or by files changed in the PR? It might be too convoluted, and I've only tried something like this once (it worked *okay* but was definitely a bit annoying to have two styles at once whenever I touched an older file).

sithlord48:
> I would be hesitant to overwrite the line history for the whole codebase like this.
> 
> What are thoughts on just setting the style guide and only having new changes be format-checked on PRs? Either by git diff or by files changed in the PR? It might be too convoluted, and I've only tried something like that once (it worked _okay_ but was definitely a bit annoying to have two styles at once whenever I touched an older file).

I Completely for this, but the linter is gonna lint. so if we don't do it for this PR (i.e land by override our lint rules). The next PR submitted will again fail linter because the linter is gonna lint and flag all the changes from the style change.

nbolton:
Quick thought, I'll revisit this later to share my full perspective... even if we delay the linting, the history rewrite problem is still present. It also has the compounding problem of making all future commits even harder to grok due to them also having format changes.

sithlord48:
> Quick thought, I'll revisit this later to share my full perspective... even if we delay the linting, the history rewrite problem is still present. It also has the compounding problem of making all future commits even harder to grok due to them also having format changes.

well then we just have to fix the linting if we change the lint rules every time . So lets take this time to get it right so we do not have to change it again

nbolton:
@sithlord48 Just a thought, since you're struggling with the code style, how about turning on auto Clang format on save in Kate? This is a common approach and means you don't need to worry about code style (clang-format does that for you).

image

sithlord48:
> @sithlord48 Just a thought, how about turning on auto Clang format on save in Kate? This is a common approach and means you don't need to worry about code style (clang-format does that for you).
> 
> image

I still want readable code the linter settings as have way to short of line and keep forcing lines be full length etc.. Honestly just get rid of it

nbolton:
I replied on the main discussion: https://github.com/deskflow/deskflow/discussions/7626#discussioncomment-10948105

src/cmd/deskflowc/MSWindowsClientTaskBarReceiver.cpp
```diff
@@ -34,16 +34,20 @@
 //
 
 const UINT MSWindowsClientTaskBarReceiver::s_stateToIconID[kMaxState] = {
```
sithlord48, line 36: This is like alot better

nbolton:
Resolved a conflict in `src/lib/gui/dialogs/AboutDialog.cpp`

nbolton:
The GitHub diff tool crashes because of the number of changes, so I resorted to good old fashioned.

```diff
-void MainWindow::updateScreenName() {
-  m_pLabelComputerName->setText(
-      QString("This computer's name: %1 "
-              R"((<a href="#" style="color: %2">change</a>))")
-          .arg(m_AppConfig.screenName())
-          .arg(kColorSecondary));
+void MainWindow::updateScreenName()
+{
+  m_pLabelComputerName->setText(QString("This computer's name: %1 "
+                                        R"((<a href="#" style="color: %2">change</a>))")
+                                    .arg(m_AppConfig.screenName())
+                                    .arg(kColorSecondary));
   m_ServerConfig.updateServerName();
 }
```

I wonder if we could improve this so that the indentation isn't so harsh?

I much preferred the way this was before.

sithlord48:
Could be the result of 
`AlwaysBreak` vs `BlockIndent` ? 

None of the indent should be changed now its llvm based

sithlord48:
> The GitHub diff tool crashes because of the number of changes, so I resorted to good old fashioned.
> 
> ```diff
> -void MainWindow::updateScreenName() {
> -  m_pLabelComputerName->setText(
> -      QString("This computer's name: %1 "
> -              R"((<a href="#" style="color: %2">change</a>))")
> -          .arg(m_AppConfig.screenName())
> -          .arg(kColorSecondary));
> +void MainWindow::updateScreenName()
> +{
> +  m_pLabelComputerName->setText(QString("This computer's name: %1 "
> +                                        R"((<a href="#" style="color: %2">change</a>))")
> +                                    .arg(m_AppConfig.screenName())
> +                                    .arg(kColorSecondary));
>    m_ServerConfig.updateServerName();
>  }
> ```

Either way that is one hard to read line..

nbolton:
> Either way that is one hard to read line..

Yeah true. Maybe the solution is outside the linter.

sithlord48:
> Yeah true. Maybe the solution is outside the linter.

yeah the one is odd cause they use the return to add a raw string 

better code would be  (105 width)

```
QString string = QStringLiteral("This Computer's name: %1" <a href=\"#\" style="color: %2\">change</a>")
m_pLabelComputerName->setText(string.arg.arg(m_AppConfig.screenName, kColorSecondary));
```

nbolton:
> QString string = QStringLiteral("This Computer's name: %1 <a href=\"#\" style="color: %2\">change</a>\"");
> m_pLabelComputerName->setText(string.arg.arg(m_AppConfig.screenName, kColorSecondary));

Yeah, the solution is to refactor code for readability rather than trying to bend the linter to the code.

sithlord48:
> > QString string = QStringLiteral("This Computer's name: %1 <a href="#" style="color: %2">change"");
> > m_pLabelComputerName->setText(string.arg.arg(m_AppConfig.screenName, kColorSecondary));
> 
> Yeah, the solution is to refactor code for readability rather than trying to bend the linter to the code.

And that is why you have to trust the programmer  and not the linter

nbolton (approved):
Approving inline with my earlier comment: https://github.com/deskflow/deskflow/discussions/7626#discussioncomment-10975671

PR #7639, opened 2024-10-11
chore: rm unused string_utils file

sithlord48:
Remove unused file `src/lib/utils/string_utils.h`


nbolton (approved):
Ah yeah that was related to licensing code that was moved downstream.

PR #7640, opened 2024-10-11
build: findSSL via cmake

sithlord48:
- Use Find_package to get openssl 1.1.1 + 
 - Link with proper names
 - Rm Windows specific macro

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,10 +149,12 @@ jobs:
 
       - name: Install dependencies
         env:
-          VCPKG_ROOT: "" # Unset deliberately to suppress 'already installed' warning.
+          VCPKG_ROOT: ${{github.workspace}}/vcpkg
         run: python ./scripts/install_deps.py
 
       - name: Configure
+        env:
+          VCPKG_ROOT: ${{github.workspace}}/vcpkg
         run: cmake -B build --preset=windows-release
 
       - name: Build
--- a/cmake/Libraries.cmake
+++ b/cmake/Libraries.cmake
@@ -485,26 +485,8 @@ macro(configure_windows_libs)
   configure_file(${PROJECT_SOURCE_DIR}/res/win/version.rc.in
                  ${PROJECT_BINARY_DIR}/src/version.rc @ONLY)
 
-  configure_windows_openssl()
+  configure_openssl()
 
-endmacro()
-
-macro(configure_windows_openssl)
-  set(OPENSSL_ROOT_DIR ${PROJECT_SOURCE_DIR}/vcpkg_installed/x64-windows)
-  set(OPENSSL_EXE_DIR ${OPENSSL_ROOT_DIR}/tools/openssl)
-
-  if(EXISTS ${OPENSSL_EXE_DIR})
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -149,10 +149,12 @@ jobs:
 
       - name: Install dependencies
         env:
-          VCPKG_ROOT: "" # Unset deliberately to suppress 'already installed' warning.
+          VCPKG_ROOT: ${{github.workspace}}/vcpkg
```
nbolton, line 152: Perhaps we could set this once in the root of `steps`, less code to maintain.

cmake/Libraries.cmake
```diff
@@ -538,8 +520,15 @@ macro(configure_openssl)
     set(OPENSSL_USE_STATIC_LIBS TRUE)
   endif()
 
-  find_package(OpenSSL REQUIRED)
-  include_directories(${OPENSSL_INCLUDE_DIR})
+  find_package(OpenSSL 1.1.1 REQUIRED COMPONENTS SSL Crypto)
```
nbolton, line 523: I assume that if OpenSSL v3 is available, it will be used.

src/lib/net/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ add_library(net STATIC ${sources})
 
 target_link_libraries(
   net
-  PUBLIC ${OPENSSL_LIBRARIES}
+  PUBLIC OpenSSL::SSL OpenSSL::Crypto
```
nbolton, line 28: I had used `${OPENSSL_LIBRARIES}` (defined by the OpenSSL `find_package` module) since historically these are prone to change. Using a var could reduce maintenance cost, especially if defined by the CMake module. What is the rationale for using the explicit library names?

nbolton (approved):
Only some minor comments, happy to land it as is.

.github/workflows/ci.yml
nbolton: Also, I wonder if `${{ env.LOCALAPPDATA }}/vcpkg` on line 130 is redundant now? I had to set that so that vcpkg was actually cached; without this it kept re-running the OpenSSL build but I guess setting `VCPKG_ROOT` fixes that properly?

cmake/Libraries.cmake
sithlord48: Correct

src/lib/net/CMakeLists.txt
sithlord48: Using the `LIBRARIES`  is the old way. 

 The targets `OpenSSL::SSL` and `OpenSSL:Crypto` are what we want to link so we should put them explicitly.

.github/workflows/ci.yml
sithlord48: Well we will not know until this lands and the cache gets saved again.

src/lib/net/CMakeLists.txt
nbolton: Thanks, yes, I see that this is the modern CMake way. Still not sure why it's better but I'll go with it. I asked a [question on SO](https://stackoverflow.com/questions/79081079/why-is-using-openssl-libraries-not-modern-cmake) so I can complete my understanding.

PR #7645, opened 2024-10-12
CI: Requires Lint to pass before running other jobs. Allow CI to be skipped for non code changes

sithlord48:
- Adjust CI so that all jobs need linting to pass to run . 
 - CI will not run if the PR only contains changes to these files:
   - any md files  (README etc..)
   -`.gitignore`
   -`.gitattributes` 
   - github `ISSUES_TEMPLATES`
   -  `.env.example`
   -  the `cspell.json` file

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,21 +82,24 @@ jobs:
     uses: ./.github/workflows/lint-clang.yml
 
   analyse-valgrind:
+    needs: [lint-cmake, lint-clang]
     if: ${{ github.event_name == 'pull_request' }}
     uses: ./.github/workflows/valgrind-analysis.yml
 
   analyse-codeql:
+    needs: [lint-cmake, lint-clang]
     if: ${{ github.event_name == 'pull_request' }}
     uses: ./.github/workflows/codeql-analysis.yml
 
   analyse-sonarcloud:
-    needs: pr-comment-flags
+    needs: [lint-cmake, lint-clang, pr-comment-flags]
     if: ${{ github.event_name == 'pull_request' && needs.pr-comment-flags.outputs.no-sonar != 'true' }}
     uses: ./.github/workflows/sonarcloud-analysis.yml
     secrets:
       SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
 
   windows:
+    needs: [lint-cmake, lint-clang]
     name: ${{ matrix.target.name }}
     runs-on: ${{ matrix.target.runs-on }}
     container: ${{ matrix.target.container }}
@@ -179,6 +182,7 @@ jobs:
           path: ${{ env.PACKAGE_PATH }}
 
   macos:
+    needs: [lint-cmake, lint-clang]
     name: ${{ matrix.target.name }}
     runs-on: ${{ matrix.target.os }}
     timeout-minutes: ${{ matrix.target.timeout }}
@@ -269,7 +273,7 @@ jobs:
           jq-filter: .distro |= map(select(.["runs-on"] | contains("arm64") | not))
 
   linux:
-    needs: linux-matrix
+    needs: [linux-matrix, lint-cmake, lint-clang]
     name: linux-${{ matrix.distro.name }}
     runs-on: ${{ matrix.distro.runs-on }}
     container: ${{ matrix.distro.container }}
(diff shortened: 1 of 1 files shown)


nbolton (approved):
Alright, let's give this a try.

.github/workflows/ci.yml
```diff
@@ -15,6 +15,13 @@ on:
       - reopened
       - synchronize
       - ready_for_review
+    paths-ignore:
+      - '**/*.md'
+      - '.env-example'
+      - '.gitignore'
+      - 'Brewfile'
+      - 'Chocolatey.config'
```
nbolton, line 23: Not sure about this. CI uses these so I think triggering CI would be wise...

.github/workflows/codeql-analysis.yml
```diff
@@ -5,6 +5,27 @@ on:
...
+      - '!.github/workflow/codeql-analysys.yml'
+      - 'doc/**'
+      - 'res/**'
+      - '!res/config.h.in'
+      - 'scripts/**'
+      - '!scripts/install_deps.py'
+      - '.env-example'
+      - '.gitignore'
+      - 'Brewfile'
+      - 'Chocolatey.config'
```
nbolton, line 21: Same comment as above.

.github/workflows/codeql-analysis.yml
```diff
@@ -5,6 +5,27 @@ on:
   workflow_call:
   push:
     branches: [master]
+    paths-ignore:
+      - '.clang-format'
```
nbolton, line 9: Ah, I'm certain this should trigger CI change.

nbolton (changes requested):

.github/workflows/ci.yml
sithlord48: I was wondering that these (and lot of other stuff are doing

.github/workflows/codeql-analysis.yml
sithlord48: yes but, it should not trigger a codeql change when a pr is landed into master

nbolton:
> Minor CI Changes 

Some of these proposed changes are actually quite major. I think that we should avoid using subjective PR titles; we should describe the change rather than summarizing in a subjective/ambiguous way.

.github/workflows/sonarcloud-analysis.yml
```diff
@@ -8,6 +8,25 @@ on:
         required: true
   push:
     branches: [master]
+    paths-ignore:
+      - '.clang-format'
```
nbolton, line 12: Same as above, quite a major change.

.github/workflows/codeql-analysis.yml
sithlord48: unless codeQL is building on windows on mac these will not affect this jobs run

.github/workflows/sonarcloud-analysis.yml
```diff
@@ -8,6 +8,25 @@ on:
...
+      - '**/*.md'
+      - '.github/**'
+      - '!.github/workflow/sonarcloud-analysys.yml'
+      - 'doc/**'
+      - 'res/**'
+      - '!res/config.h.in'
+      - 'scripts/**'
+      - '!scripts/install_deps.py'
```
nbolton, line 20: All script changes should be tested in CI. I really don't think it's a good idea to ignore these changes.

.github/workflows/sonarcloud-analysis.yml
```diff
@@ -8,6 +8,25 @@ on:
         required: true
   push:
     branches: [master]
+    paths-ignore:
+      - '.clang-format'
+      - '**/*.md'
+      - '.github/**'
```
nbolton, line 14: Not sure I quite understand why this would be a good thing.

.github/workflows/sonarcloud-analysis.yml
sithlord48: This is only code analsys when landed into master

.github/workflows/sonarcloud-analysis.yml
sithlord48: we dont' need to run this job if only those items changed they are not scanned at all

.github/workflows/sonarcloud-analysis.yml
sithlord48: no its not this file is not even scanned. code format changes are not going to change those results so no need to run the job and waste ci time for this file change

nbolton:
Converted to draft to reduce notification noise. Let me know if you need it to be open for some reason.

.github/workflows/sonarcloud-analysis.yml
nbolton: Ah, yeah just noticed the workflow name, never mind.

nbolton:
> TL;DR: I would really appreciate a more descriptive PR title.

@sithlord48 In case you missed it ^

nbolton (approved):

nbolton (approved):

PR #7651, opened 2024-10-12
fix: Set correct QGuiApplication::desktopFileName

nicolasfella:
The implicity default doesn't match the actual name of the desktop file.

This fixes e.g. the application icon on Wayland


sithlord48:
Thanks Nico!! One more item off my todo list

sithlord48:
This is not working for me I suspect its because I don't have it installed and we are for some reason using an intermediate name for the desktop file before installing.

sithlord48 (approved):

nbolton:
> This fixes e.g. the application icon on Wayland

Oh, wonderful!

PR #7653, opened 2024-10-12
build: Don't force CMAKE_INSTALL_PREFIX

nicolasfella:
This should only ever be set from the outside, not the project itself

Distribution build systems will set it to the right value

This breaks installing it to a custom if desired

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -298,7 +298,7 @@ jobs:
           DEBIAN_FRONTEND: noninteractive
 
       - name: Configure
-        run: cmake -B build --preset=linux-release ${{ matrix.distro.extra-cmake-args }}
+        run: cmake -B build --preset=linux-release ${{ matrix.distro.extra-cmake-args }} -DCMAKE_INSTALL_PREFIX=/usr
 
       - name: Build
         run: cmake --build build -j8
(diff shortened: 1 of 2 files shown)


sithlord48 (changes requested):
The change itself is great ! 
Can you also add change to the ci.yml file so the our CI linux builds use the /usr install prefix

nicolasfella:
> The change itself is great ! Can you also add a change to `.github/workflows/ci.yml` line 312. to set `-DCMAKE_INSTALL_PREFIX=/usr`.

I'm unsure where exactly it needs changing. The packaging code in 312 seems to run after the cmake build?

sithlord48:
> > The change itself is great ! Can you also add a change to `.github/workflows/ci.yml` line 312. to set `-DCMAKE_INSTALL_PREFIX=/usr`.
> 
> I'm unsure where exactly it needs changing. The packaging code in 312 seems to run after the cmake build?

Please make the CI for the linux jobs set the CMAKE_INSTALL_PREFIX=/usr

sithlord48 (approved):

nbolton:
@sithlord48 A bit late, but... I can't remember why I changed `CMAKE_INSTALL_PREFIX` in the first place. Did anyone look at the history? Also how about moving this to the presets?

PR #7660, opened 2024-10-13
Revert "ci: Allow ci to skip running if only specific files are changed"

sithlord48:
This reverts commit 14dfcb672d4225d30d1b29c6cf6e80ff5e6a43e3.
If  #7662 can not correct the start issue

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,13 +15,6 @@ on:
       - reopened
       - synchronize
       - ready_for_review
-    paths-ignore:
-      - '**/*.md'
-      - '.github/ISSUE_TEMPLATE/**'
-      - '.env-example'
-      - '.gitignore'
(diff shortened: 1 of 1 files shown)


sithlord48:
We can't land our changes without CI passing and there is no consideration for skipped CI :(

nbolton (approved):

nbolton:
> We can't land our changes without CI passing and there is no consideration for skipped CI :(

Ah of course. Didn't think of that. I bet we can pass CI if only certain files are changed through a bash step or something. Might be a bit less clean though. Worth doing do you think?

sithlord48:
Yeah ill just drop that that check for now and ill add it back when we find the right combo or rules / overrides so we can land those w/o CI

sithlord48:
Fixed by #7662

PR #7662, opened 2024-10-13
ci: fix how doc only jobs are scanned

sithlord48:
- Allow for the Ci jobs `CI Passed` to run as long as its dependent jobs are have not ` failed` and were not `cancelled`
 - Allow for the `Test results` job to be skipped if the windows , linux and macos jobs were also skipped


nbolton (approved):
Let's give it a shot.

PR #7669, opened 2024-10-13
ci: attempt releases

### LINKED ISSUE #7555: Auto continuous/rolling/canary release for each PR land
### Project

Deskflow

### Operating systems (OS)

- [X] Windows
- [X] macOS
- [X] Linux
- [ ] BSD-derived
- [ ] Other

### OS versions/distros

n/a

### Deskflow configuration

n/a

Testers shouldn't have to wait for a release to be manually published in order to test the latest changes.

Each time a PR lands, publish/update a release called continuous/rolling/canary which always shows at the top of the releases as the latest release. This way, the most available package is always the latest bleeding-edge code.

sithlord48:
fixes #7555 
- Attempt to do continuous and normal releases

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -197,7 +197,7 @@ jobs:
       - name: Upload
         uses: actions/upload-artifact@v4
         with:
-          name: ${{ env.PACKAGE_PREFIX }}-${{ matrix.target.name }}
+          name: package-${{ env.PACKAGE_PREFIX }}-${{ matrix.target.name }}
           path: ${{ env.PACKAGE_PATH }}
 
   macos:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -200,7 +200,7 @@ jobs:
       - name: Upload
         uses: actions/upload-artifact@v4
         with:
-          name: ${{ env.PACKAGE_PREFIX }}-${{ matrix.target.name }}
+          name: windows-${{ env.PACKAGE_PREFIX }}-${{ matrix.target.name }}
```
nbolton, line 203: Heads up, this would result in `windows-deskflow-windows-2022-x64`

.github/workflows/ci.yml
```diff
@@ -380,3 +380,33 @@ jobs:
...
           run: ${{ env.DESKFLOW_BUILD_CMD }}
+
+  release_assets:
+    name: Release
+    needs: [windows, linux, macos]
+    runs-on: ubuntu-latest
+    steps:
+    - name: Download Files
+      uses: actions/download-artifact@v4
+    - name: Deploy Continuous
```
nbolton, line 391: Spaces above `-name` entries would make it a touch more redable.

.github/workflows/ci.yml
nbolton: How about prefixing each one with `package-` then you only need to filter on that in 399? Just an idea.

PR #7673, opened 2024-10-14
docs: Update Readme

sithlord48:
Update the readme ONLY [Preview](https://github.com/deskflow/deskflow/blob/readme_update/README.md)
Split other changes into #7791

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,27 @@
 ![Deskflow](https://github.com/user-attachments/assets/f005b958-24df-4f4a-9bfd-4f834dae59d6)
 
-> [!TIP]
-> Join us! Real-time discussion on Matrix: [`#deskflow:matrix.org`](https://matrix.to/#/#deskflow:matrix.org)
+> [!NOTE]
+> 🚀 **Deskflow** is now the upstream project for Synergy.
 >
-> Alternatively, we have [other ways](https://github.com/deskflow/deskflow/wiki/Chat-with-us) to communicate.
+> The Deskflow project was established to cultivate community-driven development where everyone can collaborate.
+>
+> Synergy sponsors the Deskflow project by contributing code and providing financial support while maintaining its customer-oriented code downstream.
+>
+> More info: [Relationship with Synergy](https://github.com/deskflow/deskflow/wiki/Relationship-with-Synergy)
 
-[![Sponsored by: Synergy](https://raw.githubusercontent.com/deskflow/deskflow-artwork/b2c72a3e60a42dee793bd47efc275b5ee0bdaa5f/misc/synergy-sponsor.svg)](https://github.com/deskflow/deskflow/wiki/Relationship-with-Synergy)
+**Deskflow** is a free and open source keyboard and mouse sharing app.
+Use the keyboard, mouse, or trackpad of one computer to control nearby computers,
+and work seamlessly between them.
+It's like a software KVM (but without the video).
+TLS encryption is enabled by default. Wayland is supported. Clipboard sharing is supported.
 
-[![GitHub Release](https://img.shields.io/github/v/release/deskflow/deskflow?display_name=release&label=latest%20version)](https://github.com/deskflow/deskflow/releases)
-[![GitHub commit activity](https://img.shields.io/github/commit-activity/m/deskflow/deskflow)](https://github.com/deskflow/deskflow/commits/master/)
-![GitHub top language](https://img.shields.io/github/languages/top/deskflow/deskflow)
-[![GitHub License](https://img.shields.io/github/license/deskflow/deskflow)](LICENSE)
+[![Downloads: Stable Release](https://img.shields.io/github/downloads/deskflow/deskflow/v1.17.0/total?style=for-the-badge&logo=github&label=Download%20Stable)](https://github.com/deskflow/deskflow/releases/tag/v1.17.0)&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[![Downloads: Continuous Build](https://img.shields.io/github/downloads/deskflow/deskflow/continuous/total?style=for-the-badge&logo=github&label=Download%20Continuous)](https://github.com/deskflow/deskflow/releases)
+
+To use Deskflow you can use one of our [packages](https://github.com/deskflow/deskflow/releases), install `deskflow` (if available in your package repository), or [build it](#build-quick-start) yourself from source.
+
+[![GitHub commit activity](https://img.shields.io/github/commit-activity/m/deskflow/deskflow?logo=github)](https://github.com/deskflow/deskflow/commits/master/)
+[![GitHub top language](https://img.shields.io/github/languages/top/deskflow/deskflow?logo=github)](https://github.com/deskflow/deskflow/commits/master/)
+[![GitHub License](https://img.shields.io/github/license/deskflow/deskflow?logo=github)](LICENSE)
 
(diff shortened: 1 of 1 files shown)


nbolton:
> * mv BUILD.MD -> doc/build.md
> * mv CONTRIBUTING.md -> doc/contributing.md

Not a fan of all caps rage, eh? :smile:

doc/build.md
```diff
@@ -1,4 +1,32 @@
-# Build Deskflow
+<!--
+    SPDX-FileCopyrightText: 2024 Symless Ltd.
```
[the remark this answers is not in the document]
nbolton: This is the way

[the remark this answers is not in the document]
nbolton: This is the way

doc/server-configs.md
nbolton, line 1: @sithlord48 I renamed the file to `doc/server-configs.md` and added more context around how server config files work. I also made the example computer names consistent (moe, larry, and curly) and fixed the ASCII art (which was completely wrong in the originals).

nbolton:
@sithlord48 Ready to land this? If not, what's left?

doc/server-configs.md
nbolton, line 1: @sithlord48 I renamed the file to `doc/server-configs.md` and added more context around how server config files work. I also made the example computer names consistent (moe, larry, and curly) and fixed the ASCII art (which was completely wrong in the originals).

doc/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+  set(DOXYGEN_STRIP_FROM_PATH ${CMAKE_SOURCE_DIR})
+  set(DOXYGEN_USE_MDFILE_AS_MAINPAGE mainpage.md)
+  set(DOXYGEN_QUIET YES)
+
+  doxygen_add_docs(docs ${CMAKE_SOURCE_DIR}/doc ${CMAKE_SOURCE_DIR}/src
+                   COMMENT "Generating Documentation" ALL)
+
+  #This is only to show files used by the documentation
+  #within the project tree
+  target_sources(docs PRIVATE build.md contributing.md configuration.md
```
sithlord48, line 18: can we get these one per line please

doc/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+                              mainpage.md)
+
+  install(
+    DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/html"
+    DESTINATION ${CMAKE_INSTALL_DOCDIR}
+    COMPONENT deskflow_docs)
```
sithlord48, line 24: closing ) on its own line in line with install

doc/CMakeUserPresets.json
```diff
@@ -0,0 +1,87 @@
...
+    },
+    {
+      "name": "windows",
+      "inherits": "base",
+      "hidden": true,
+      "generator": "Ninja",
+      "cacheVariables": {
+        "CMAKE_C_COMPILER": "cl.exe",
+        "CMAKE_CXX_COMPILER": "cl.exe",
+        "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake"
```
sithlord48, line 21: fwiw this breaks my local build on windows where i have the cache outside the build dir.. 
windows uses on cli may need to set this path for their machine specificity

doc/config.toml.example
```diff
@@ -0,0 +1,15 @@
+# Example file for `--config-toml` arg.
```
sithlord48, line 1: this should be in the `configuration.md` page as a code block  like the other config examples.

doc/server-configs.md
sithlord48: did you not update this link ?

doc/mainpage.md
```diff
@@ -0,0 +1,9 @@
+OUR Documents main page
```
sithlord48, line 1: This main page should be similar to our README.md ?

I tend to think of the `readme` as the intro page to our documents at least.

doc/server-configs.md
nbolton: Which link?

doc/CMakeLists.txt
nbolton: > If you like, but I didn't touch this... I think this was from when we had `cmake-format`?

doc/CMakeUserPresets.json
nbolton: Shame JSON doesn't support comments :|

doc/config.toml.example
nbolton: Ah. So, rename `doc/server-configs.md` back to `doc/configuration.md` and have it for all types of config?

doc/mainpage.md
nbolton: IDK... you do your thing!

.gitignore
```diff
@@ -26,12 +26,14 @@ deskflow-config.toml
...
 
-#CMake ignores
-CMakeLists.txt.user
-CMakeCache.txt
-CMakeUserPresets.json
-CMakeFiles/*
+# CMake files
+/CMakeLists.txt.user
+/CMakeCache.txt
+/CMakeUserPresets.json
```
sithlord48, line 35: why the / , i have never used a preceding / in other repos before.

nbolton:
Added task:
> Add protocol compatibility for Barrier/IL to readme

doc/config.toml.example
sithlord48: i would rather have it all in one spot  its all under the configuration right . figure its easier to find if its all in one document

.gitignore
nbolton: It matches only files in the root. I set it like that so we can add `doc/CMakeUserPresets.json` and the rest was for consistency and clarity, but if you're against the `/` we can remove it and either force add `doc/CMakeUserPresets.json` or call that file `doc/CMakeUserPresets.json.example`.

nbolton:
@sithlord48 You mentioned yesterday you're thinking about dropping this PR. How about dropping some of the stuff such as the hard-coded build instructions? I think there's still a lot of good stuff in here. 

I also think it's worth keeping files like `build.md` and `contribute.md` but simply as links to the relevant wiki pages, since these are common files that people look for in the source tree.

sithlord48:
> @sithlord48 You mentioned yesterday you're thinking about dropping this PR. How about dropping some of the stuff such as the hard-coded build instructions? I think there's still a lot of good stuff in here, such as the config examples. Those could go in the wiki though, I suppose!
> 
> To confirm: you did say we should only have the build guide on the wiki and not in source, right?
> 
> Also, I think it's worth keeping files like `build.md` and `contribute.md` but simply as links to the relevant wiki pages, since these are common files that people look for in the source tree.

I don't want to drop this pr we should ship some documentation 
HOWEVER, 
  I think that the documentation we have started is not the documentation we want to ship with our deskflow installation.
We need to move the development parts of this in to relevant  Wiki pages and  work more to make these documents more user facing

I will work to this modified to reflect this change

README.md
```diff
@@ -67,82 +67,6 @@ Here are a few ways to join in with the project and get involved:
 > Check the [Build Guide](https://github.com/deskflow/deskflow/wiki/Build-Guide)
```
nbolton, line 67: Let's remove the `[!TIP]` and quote (`>`), then to the new page:
```
For instructions on building Deskflow, use the wiki page: [Building](https://github.com/deskflow/deskflow/wiki/Building)
```

nbolton (approved):

Issue #7674, opened 2024-10-14
CI test results error: `line 31: [: too many arguments`


nbolton:
### Project

Deskflow

### Deskflow version number

1.18.0

### What steps will reproduce the problem?

1. Make a failing test so the failure test summary comment shows on the PR
2. Fix the test so it goes away

Expected: PR comment disappears
Actual: PR comment remains, and workflow fails silently with: `line 31: [: too many arguments`
Example: https://github.com/deskflow/deskflow/actions/runs/11320188396/job/31478800082

The error relates to this:
```
  count=$(echo -n "$table" | grep -o "❌" | wc -l || echo 0)
  if [ $count -gt 0 ]; then
```

### Log output

```shell
Run # Sets the step summary and creates a CI summary file.
  # Sets the step summary and creates a CI summary file.
  
  table="# Test results
  | Job name | Unit tests | Integration tests |
  | --- | --- | --- |
  | linux-almalinux-9-amd64 | ✅ Pass | ✅ Pass |
  | linux-archlinux-amd64 | ✅ Pass | ✅ Pass |
  | linux-debian-12-amd64 | ✅ Pass | ✅ Pass |
  | linux-debian-12-arm64 | ✅ Pass | ✅ Pass |
  | linux-debian-13-amd64 | ✅ Pass | ✅ Pass |
  | linux-fedora-39-amd64 | ✅ Pass | ✅ Pass |
  | linux-fedora-40-amd64 | ✅ Pass | ✅ Pass |
  | linux-fedora-40-arm64 | ✅ Pass | ✅ Pass |
  | linux-manjaro-amd64 | ✅ Pass | ✅ Pass |
  | linux-opensuse-amd64 | ✅ Pass | ✅ Pass |
  | linux-rockylinux-9-amd64 | ✅ Pass | ✅ Pass |
  | linux-ubuntu-22.04-amd64 | ✅ Pass | ✅ Pass |
  | linux-ubuntu-24.04-amd64 | ✅ Pass | ✅ Pass |
  | macos-13-x64 | ✅ Pass | ✅ Pass |
  | macos-14-arm64 | ✅ Pass | ✅ Pass |
  | windows-2022-x64 | ✅ Pass | ✅ Pass |
  "
  if [ -z "$table" ]; then
    echo "No test results found" > $GITHUB_STEP_SUMMARY
    exit 1
  else
    echo "$table" > $GITHUB_STEP_SUMMARY
  fi
  
[...]

PR #7675, opened 2024-10-14
Fixed error count logic for test summary

nbolton:
Fixes: #7674 

In future: We should move the `test-summary` action and related code from `test-actions.yml` into a new actions repository, possibly in a personal repo. It could be useful for other projects and doesn't seem like something that needs to be maintained within Deskflow.

### DIFF
--- a/.github/workflows/test-actions.yml
+++ b/.github/workflows/test-actions.yml
@@ -1,0 +1,41 @@
+name: Test actions
+
+# Intended to only be run manually for testing actions in isolation.
+on:
+  workflow_dispatch:
+
+jobs:
+  test-summary:
+    runs-on: ubuntu-latest
+    name: Test summary (${{ matrix.message.name }})
+
+    strategy:
+      matrix:
+        message:
+          - name: success
+            text: "|Success|✅|✅|"
+            expect: ""
+          - name: failure
+            text: "|Failure|❌|❌|"
+            expect: "❌🔬 Tests failed: 2"
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Generate
+        shell: bash
+        run: |
+          echo "${{ matrix.message.text }}" > test-result.md
+
+      - name: Upload
+        uses: actions/upload-artifact@v4
+        with:
+          path: test-result.md
+          name: test-result-${{ matrix.message.name }}
+
+      - name: Test
+        uses: ./.github/actions/test-summary
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/test-actions.yml
```diff
@@ -0,0 +1,41 @@
+name: Test actions
```
sithlord48, line 1: how do you feel about this being in ci.yml this will remove the sub folder it looks to be under in the action summary.

nbolton: I believe `workflow_dispatch` in a separate workflow is the way forward, rather than it always running in CI. Also, if it was in `ci.yml` then it'd add a bunch of irrelevant test artefacts to the workflow.

The real solution here is to move this to a separate repo and maintain it as a separate action outside of Deskflow... but I want to get it fixed for now and do it properly later.

sithlord48: workflow dispatch or not its still run from our ci.yml script

sithlord48: Its in the CI and we have al these test artifacts on every build.

nbolton: Not sure what you mean. The `test-actions.yml` file is a workflow for testing actions. Incidentally, it only tests the `test-summary/action.yml` composite action (which is run from `ci.yml`). I didn't want to call it `test-test-summary` because that looked silly and I don't want to introduce a new test workflow for each new composite action I want to test.

sithlord48:
doesnt the ci job call this action ?

nbolton:
> doesnt the ci job call this action ?

Yes, `ci.yml` calls `./.github/actions/test-summary` from the `test-results` step. I have found 2 ways of testing the `test-summary` action:
1. Add temporary hacks in `ci.yml` to skip ahead (which isn't fun at all)
2. Use a specific test workflow (much less stressful)

sithlord48:
> > doesnt the ci job call this action ?
> 
> Yes, `ci.yml` calls `./.github/actions/test-summary` from the `test-results` step.
> 
> I have found 2 ways of testing the `test-summary` action:
> 
>     1. Add temporary hacks in `ci.yml` to skip ahead (which isn't fun at all)
> 
>     2. Use a specific test workflow (much less stressful)
> 
> 
> Cannot stress enough how much I want to move this action to it's own repo.

in my other repos code QL is its own job no run by the ci.yml action

sithlord48:
yeah lol

Issue #7657, opened 2024-10-12
Run CodeQL separate from CI workflow


nbolton:
### What problem are you trying to solve?

CodeQL is currently being run as a sub-workflow of CI workflow, which potentially causes it not work properly:

![image](https://github.com/user-attachments/assets/59b5ac8e-4e1c-46e1-8dfe-ebcb892aaa8f)

### What solution do you propose?

We should consider restoring the old behavior where it runs separately from the CI workflow.

PR #7676, opened 2024-10-14
ci: Restore `pull_request` trigger for `codeql-analysis.yml`

nbolton:
Fixes: #7657


.github/workflows/codeql-analysis.yml
```diff
@@ -1,8 +1,19 @@
...
+#
+# We should not trigger this workflow with `workflow_call` as this causes the error:
+#   "1 configuration present on `master` was not found"
+#
+# Sadly, this means we can't roll it into our monolithic CI workflow.
 on:
   workflow_dispatch:
-  workflow_call:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
```
sithlord48, line 14: do we really care about types here ?

nbolton: If it runs on PR drafts, it's kind of wasteful.
> (we only run this on open PRs as it's very slow)

sithlord48: If i use a draft i do like to only updraft if i know its gonna pass and only review nits are left. Either way is fine for me

nbolton: I've never seen the CodeQL workflow fail from an issue isolated to that workflow; in my experience so far, it always passes if the other workflows pass.

sithlord48 (approved):

PR #7678, opened 2024-10-14
Add leading edge philosophy to README.md

nbolton:
I think it's important to help people understand the difference between Deskflow and Input Leap. e.g. We don't support Qt5, but IL seem to want to support Qt5.


sithlord48:
ok with me

nbolton:
> ok with me

Will take this as approval.

Issue #7680, opened 2024-10-14
Usage and description doesn't matched in deskflow-server and deskflow-client help message 


kenhys:
### Project

Deskflow

### Deskflow version number

v1.18.0

### Build type

Local developer build (built it yourself)

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [X] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other

### Wayland on Linux

- [ ] I have reviewed the Wayland known issues and my issue is new
- [X] I am not using Wayland on Linux

### Signing on mac os

- [ ] I have authorized the application to run on my mac
- [ ] I am not using Mac os

### OS versions/distros

- Debian sid

### Deskflow configuration

N/A (default)

### What steps will reproduce the problem?

1. Execute deskflow-server -h
2. Execute deskflow-client -h

In deskflow-server help message:

* Usage and explanation doesn't matched
  * `--no-wayland-ei` exist on Usage, but not explained in description
     * https://github.com/deskflow/deskflow/blob/master/src/lib/deskflow/ServerApp.cpp#L139 
  * `--enable-drag-drop` exist in description, but it is not available on Linux. It should be omitted on Linux. This is caused by [HELP_COMMON_INFO_1](https://github.com/deskflow/deskflow/blob/master/src/lib/deskflow/App.h#L162-L177) macro.
    * https://github.com/deskflow/deskflow/blob/master/src/lib/deskflow/ArgParser.cpp#L238-L242

In deskflow-client help message:

* Usage and explanation doesn't matched
  * `--use-x-window` exist on Usage, but not explained in description
    * https://github.com/deskflow/deskflow/blob/master/src/lib/deskflow/ClientApp.cpp#L130-L132
  * `--enable-drag-drop` exist in description, but it is not available on Linux
    * same as deskflow-server.

### Log output

```shell
deskflow-server help message:


# ./deskflow-server -h
Usage: deskflow-server [--address <address>] [--config <pathname>] [--display <display>] [--no-xinitthreads] [--no-wayland-ei] [--daemon|--no-daemon] [--name <screen-name>] [--restart|--no-restart] [--debug <level>]

Start the Deskflow mouse/keyboard sharing server.

  -a, --address <address>  listen for clients on the given address.
  -c, --config <pathname>  use the named configuration file instead.
  -d, --debug <level>      filter out log messages with priority below level.
[...]

PR #7682, opened 2024-10-14
fix: align --display explanation in help message

kenhys:
Broken "." position was fixed like this.

Before:

```
        --display <display>  when in X mode, connect to the X server
                               at <display>
  .      --no-xinitthreads    do not call XInitThreads()
```

After:

```
        --display <display>  when in X mode, connect to the X server
                               at <display>.
        --no-xinitthreads    do not call XInitThreads()
```


kenhys:
Related: https://github.com/deskflow/deskflow/issues/7680

In this PR, focus on trivial cosmetic change.

nbolton (approved):
Thanks, you beat me to it!

PR #7683, opened 2024-10-14
Remove `max_pargs_hwrap` from `cmake-format.yaml`

nbolton:
Solves horrible formatting problem.

e.g.
```
  string(REGEX MATCH [0-9]+\- PATCH ${GITREV})
```

Was formatted as:
```
string(REGEX
MATCH
[0-9]+
- PATCH
${GITREV})
```

### DIFF
--- a/cmake/Build.cmake
+++ b/cmake/Build.cmake
@@ -93,12 +93,8 @@ macro(post_config_all)
       VERBATIM
       COMMENT "Copying files to bin dir")
 
-    add_dependencies(
-      run_post_build
-      ${GUI_BINARY_NAME}
-      ${CLIENT_BINARY_NAME}
-      ${SERVER_BINARY_NAME}
-      ${DAEMON_BINARY_NAME})
+    add_dependencies(run_post_build ${GUI_BINARY_NAME} ${CLIENT_BINARY_NAME}
+                     ${SERVER_BINARY_NAME} ${DAEMON_BINARY_NAME})
   endif()
 
 endmacro()
--- a/cmake/CodeCoverage.cmake
+++ b/cmake/CodeCoverage.cmake
@@ -164,11 +164,8 @@ set(CMAKE_SHARED_LINKER_FLAGS_COVERAGE
           "Flags used by the shared libraries linker during coverage builds."
           FORCE)
 mark_as_advanced(
-  CMAKE_Fortran_FLAGS_COVERAGE
-  CMAKE_CXX_FLAGS_COVERAGE
-  CMAKE_C_FLAGS_COVERAGE
-  CMAKE_EXE_LINKER_FLAGS_COVERAGE
-  CMAKE_SHARED_LINKER_FLAGS_COVERAGE)
+  CMAKE_Fortran_FLAGS_COVERAGE CMAKE_CXX_FLAGS_COVERAGE CMAKE_C_FLAGS_COVERAGE
+  CMAKE_EXE_LINKER_FLAGS_COVERAGE CMAKE_SHARED_LINKER_FLAGS_COVERAGE)
 
 if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
   message(
@@ -199,19 +196,10 @@ function(setup_target_for_coverage_lcov)
 
   set(options NO_DEMANGLE)
   set(oneValueArgs BASE_DIRECTORY NAME)
-  set(multiValueArgs
-      EXCLUDE
-      EXECUTABLE
-      EXECUTABLE_ARGS
-      DEPENDENCIES
-      LCOV_ARGS
-      GENHTML_ARGS)
-  cmake_parse_arguments(
-    Coverage
-    "${options}"
-    "${oneValueArgs}"
-    "${multiValueArgs}"
-    ${ARGN})
+  set(multiValueArgs EXCLUDE EXECUTABLE EXECUTABLE_ARGS DEPENDENCIES LCOV_ARGS
+                     GENHTML_ARGS)
+  cmake_parse_arguments(Coverage "${options}" "${oneValueArgs}"
+                        "${multiValueArgs}" ${ARGN})
 
   if(NOT LCOV_PATH)
     message(FATAL_ERROR "lcov not found! Aborting...")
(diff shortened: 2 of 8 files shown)


cmake/Build.cmake
```diff
@@ -93,12 +93,8 @@ macro(post_config_all)
       VERBATIM
       COMMENT "Copying files to bin dir")
 
-    add_dependencies(
-      run_post_build
-      ${GUI_BINARY_NAME}
-      ${CLIENT_BINARY_NAME}
-      ${SERVER_BINARY_NAME}
-      ${DAEMON_BINARY_NAME})
+    add_dependencies(run_post_build ${GUI_BINARY_NAME} ${CLIENT_BINARY_NAME}
```
sithlord48, line 96: worse.

cmake/CodeCoverage.cmake
```diff
@@ -233,12 +221,7 @@ function(setup_target_for_coverage_lcov)
   foreach(EXCLUDE ${Coverage_EXCLUDE} ${COVERAGE_EXCLUDES}
                   ${COVERAGE_LCOV_EXCLUDES})
     if(CMAKE_VERSION VERSION_GREATER 3.4)
-      get_filename_component(
-        EXCLUDE
-        ${EXCLUDE}
-        ABSOLUTE
-        BASE_DIR
-        ${BASEDIR})
+      get_filename_component(EXCLUDE ${EXCLUDE} ABSOLUTE BASE_DIR ${BASEDIR})
```
sithlord48, line 224: better

cmake/CodeCoverage.cmake
```diff
@@ -278,10 +261,8 @@ function(setup_target_for_coverage_lcov)
     COMMAND ${GENHTML_PATH} ${GENHTML_EXTRA_ARGS} ${Coverage_GENHTML_ARGS} -o
             ${Coverage_NAME} ${Coverage_NAME}.info
     # Set output files as GENERATED (will be removed on 'make clean')
-    BYPRODUCTS ${Coverage_NAME}.base
-               ${Coverage_NAME}.capture
-               ${Coverage_NAME}.total
-               ${Coverage_NAME}.info
+    BYPRODUCTS ${Coverage_NAME}.base ${Coverage_NAME}.capture
```
sithlord48, line 264: worse
General cmake "style" would be
```
ByProducts 
   A
   B
   C 
   D
```

cmake/CodeCoverage.cmake
```diff
@@ -324,12 +305,8 @@ function(setup_target_for_coverage_gcovr_xml)
   set(options NONE)
   set(oneValueArgs BASE_DIRECTORY NAME)
   set(multiValueArgs EXCLUDE EXECUTABLE EXECUTABLE_ARGS DEPENDENCIES)
-  cmake_parse_arguments(
-    Coverage
-    "${options}"
-    "${oneValueArgs}"
-    "${multiValueArgs}"
-    ${ARGN})
+  cmake_parse_arguments(Coverage "${options}" "${oneValueArgs}"
```
sithlord48, line 308: should be one line really but better

src/gui/CMakeLists.txt
```diff
@@ -23,13 +23,7 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
...
 set(qrc_file ${GUI_QRC_FILE})
 
-file(
-  GLOB
-  sources
-  src/*.cpp
-  src/*.h
-  src/validators/*
-  src/widgets/*)
+file(GLOB sources src/*.cpp src/*.h src/validators/* src/widgets/*)
```
sithlord48, line 26: Eww Globs.. 
but also  general style would be 
```
file (GLOB 
   A 
   B
   B
)
```
or

``` 
file (
   Glob
     A
     B
     C
)
```
are common ways i have seen this formatted in cmake

src/gui/CMakeLists.txt
```diff
@@ -48,13 +42,8 @@ include_directories(${PROJECT_BINARY_DIR}/config)
...
 add_executable(${target} WIN32 ${sources} ${ui_files} ${rc_files} ${qrc_file})
 
-target_link_libraries(
-  ${target}
-  ${DESKFLOW_GUI_HOOK_LIB}
-  gui
-  Qt6::Core
-  Qt6::Widgets
-  Qt6::Network)
+target_link_libraries(${target} ${DESKFLOW_GUI_HOOK_LIB} gui Qt6::Core
```
sithlord48, line 45: Style 
when linking only one type of links this form is common
```
target_link_libraries (target [ PRIVATE | PUBLIC | INTERFACE ]
    liba
    libb 
    Group::A Group::b 
)
```
or this if more setting public , private and or interface  links
```
target_link_libraries (target
    PUBLIC
      liba
      libb 
    PRIVATE
       Libc
       Group::A Group::b 
)
```
or 

```

```

src/test/CMakeLists.txt
```diff
@@ -33,12 +33,8 @@ endmacro()
 
 macro(config_test)
 
-  include_directories(
-    ${test_base_dir}
-    ${src_dir}
-    ${src_dir}/lib
-    ${gui_dir}
-    ${ext_dir})
+  include_directories(${test_base_dir} ${src_dir} ${src_dir}/lib ${gui_dir}
```
sithlord48, line 36: same general style as linking libs

sithlord48 (commented):
Its a mixed bag

nbolton:
> Its a mixed bag

Yeah... IDK, I tried fiddling around with the cmake-format config a while ago to fix this shit but it seems limited.

I initially used this because over the years various inconsistencies developed, such as a mix of tabs and spaces, etc.

It'd be nice if cmake-format wasn't so damn fussy... or at least wrapped things sensibly by default.

nbolton:
Maybe I'll see what [`wrap=False`](https://cmake-format.readthedocs.io/en/latest/format-algorithm.html) is all about.

sithlord48:
if your main thing is tabs vs spaces perhaps a using an [editor-config](https://editorconfig.org/) file would be enough for the project to not need lint for cmake / clang

nbolton:
> if your main thing is tabs vs spaces perhaps a using an [.editorconfig](https://editorconfig.org/) file would be enough for the project to not need lint for cmake / clang

For cmake my main concern was tabs/spaces (there are probably other style inconsistencies you can get me grumbling about though), but for clang my concerns are much more elaborate.

nbolton:
> to clarify i meant it could be replace cmake and or clang lint. If your mainly worried about tabs / spaces in that format

Yep, got that. What I meant was there are things beyond tabs/spaces that I'm worried about in both cmake and clang, but my concerns about clang style are much greater than make.

My point is, I don't think we should lump both linters into the same category.

sithlord48:
> > to clarify i meant it could be replace cmake and or clang lint. If your mainly worried about tabs / spaces in that format
> 
> Yep, got that. What I meant was there are things beyond tabs/spaces that I'm worried about in both cmake and clang, but my concerns about clang style are much greater than make.
> 
> My point is, I don't think we should lump both linters into the same category.

That was not what i was suggesting. sorry if you took it that way

nbolton:
> sorry if you took it that way

The curse of text based communication!

nbolton:
> if your main thing is tabs vs spaces perhaps a using an [.editorconfig](https://editorconfig.org/) file would be enough for the project to not need lint for cmake / clang

Thoughts on something like [editorconfig-checker](https://github.com/editorconfig-checker/editorconfig-checker)?

nbolton:
Now irrelevant as `cmake-format` is being removed.

Issue #5762, opened 2016-12-04
Apply XDG base directory specifications for configuration files


z0rc:
### Operating Systems ###

Server: Windows 10
Client: Kubuntu 16.10

### Synergy Version ###

1.8.6-rc1-d0db743

### Steps to reproduce bug ###

1. Set connect to server with SSL enabled

### Other info ###

* Is there a way to work around it? No
* Does this bug prevent you from using Synergy entirely? No

Synergy creates unnecessary top level dotdir with SSL configuration in it. Please use $XDG_CONFIG_HOME (with fallback to ~/.config) for storing such data. More at https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

adrianinsaval:
@nbolton this is still a problem in deskflow, it should probably default to .local/share/deskflow

sithlord48:
`~/.config` .. would be the place for this .

adrianinsaval:
sounds good, can you reopen the issue since it's not actually fixed yet?

nbolton:
Personally I'm not happy that our config files are spread around a few different places.

This is good:
```
$ ls ~/.config | grep Deskflow
Deskflow
```

This is not:
```
$ ls -R ~/.synergy/
~/.synergy/:
SSL

~/.synergy/SSL:
Fingerprints  Synergy.pem

~/.synergy/SSL/Fingerprints:
Local.txt  TrustedServers.txt
```

sithlord48:
@nbolton pleas make pr for this we should not be looking in `~/.deskflow` at all for sure

nbolton:
> @nbolton please make pr for this we should not be looking in `~/.deskflow` at all for sure

On it: #7685

PR #7685, opened 2024-10-14
Change TLS dir to same as Qt config

nbolton:
OMG... how many places is this dir string defined? Will fix that in a separate PR.

Fixes: #5762

### DIFF
--- a/src/lib/arch/unix/ArchFileUnix.cpp
+++ b/src/lib/arch/unix/ArchFileUnix.cpp
@@ -19,6 +19,7 @@
 #include "arch/unix/ArchFileUnix.h"
 
 #include <cstring>
+#include <filesystem>
 #include <pwd.h>
 #include <stdio.h>
 #include <sys/types.h>
@@ -100,17 +101,21 @@ std::string ArchFileUnix::getPluginDirectory() {
 }
 
 std::string ArchFileUnix::getProfileDirectory() {
-  String dir;
   if (!m_profileDirectory.empty()) {
-    dir = m_profileDirectory;
+    return m_profileDirectory;
   } else {
+    const std::filesystem::path homeDir = getUserDirectory();
 #if WINAPI_XWINDOWS
-    dir = getUserDirectory().append("/.config/" DESKFLOW_APP_NAME);
+    const auto xdgDir = std::getenv("XDG_CONFIG_HOME");
+    if (xdgDir != nullptr) {
+      return std::filesystem::path(xdgDir) / DESKFLOW_APP_NAME;
+    } else {
+      return homeDir / ".config" / DESKFLOW_APP_NAME;
+    }
 #else
-    dir = getUserDirectory().append("/Library/" DESKFLOW_APP_NAME);
+    return homeDir / "Library" / DESKFLOW_APP_NAME;
 #endif
   }
-  return dir;
 }
 
 std::string
--- a/src/lib/arch/unix/ArchFileUnix.cpp
+++ b/src/lib/arch/unix/ArchFileUnix.cpp
@@ -105,7 +105,7 @@ std::string ArchFileUnix::getProfileDirectory() {
     dir = m_profileDirectory;
   } else {
 #if WINAPI_XWINDOWS
-    dir = getUserDirectory().append("/." DESKFLOW_APP_ID);
+    dir = getUserDirectory().append("/.config/" DESKFLOW_APP_NAME);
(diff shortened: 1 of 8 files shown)


sithlord48:
I fixed this in [inputleap](https://github.com/input-leap/input-leap/pull/1892) there was alot of extra defines and if def guards on the name when they even were the same

src/lib/arch/unix/ArchFileUnix.cpp
```diff
@@ -105,7 +105,7 @@ std::string ArchFileUnix::getProfileDirectory() {
     dir = m_profileDirectory;
   } else {
 #if WINAPI_XWINDOWS
-    dir = getUserDirectory().append("/." DESKFLOW_APP_ID);
+    dir = getUserDirectory().append("/.config/" DESKFLOW_APP_NAME);
```
sithlord48, line 108: Use `$XDG_CONFIG_DIR` if defined or `~/.config` if its not

sithlord48 (changes requested):

src/lib/arch/unix/ArchFileUnix.cpp
nbolton: How about `$XDG_CONFIG_DIR` or fall back to `~/.config/Deskflow`?

nbolton: I'll rip it from your IL PR.

nbolton: Done: bcc41ea

sithlord48 (commented):
Looks good did not run it

nbolton:
> Looks good did not run it

Tested on Linux with client and server, also tested with XDG_CONFIG_HOME set and seems to work well.

src/lib/arch/unix/ArchFileUnix.cpp
```diff
@@ -100,17 +101,21 @@ std::string ArchFileUnix::getPluginDirectory() {
...
-    dir = getUserDirectory().append("/." DESKFLOW_APP_ID);
+    const auto xdgDir = std::getenv("XDG_CONFIG_HOME");
+    if (xdgDir != nullptr) {
+      return std::filesystem::path(xdgDir) / DESKFLOW_APP_NAME;
+    } else {
+      return homeDir / ".config" / DESKFLOW_APP_NAME;
+    }
 #else
-    dir = getUserDirectory().append("/Library/" DESKFLOW_APP_NAME);
+    return homeDir / "Library" / DESKFLOW_APP_NAME;
```
sithlord48, line 116: This needs to be double checked iirc its `~/Library/Preferences` for user and /Library/Preferences` for global settings on mac os

nbolton: Ah, I belive the `~/Library/Preferences` dir is just for `.plist` files.

nbolton: Seems that `~/Library/Application Support` is an alternative, but seems like plenty of apps just make a dir in root of `~/Library`, so looks like common practice. Not sure what Apple wants us to do though...

sithlord48, line 116: Shouldn't this be something like 
return QStringLitera("%1/Library/%2).arg(homeDir, DESKFLOW_APP_NAME);

sithlord48 (changes requested):

src/lib/arch/unix/ArchFileUnix.cpp
nbolton: If it was in Qt, sure, but that's not the case.

sithlord48: append says otherwise

sithlord48: well if its not should there some concat or something?

nbolton: I'm confused, you mean [`std::string:append`](https://cplusplus.com/reference/string/string/append/)?

sithlord48 (approved):

src/lib/arch/unix/ArchFileUnix.cpp
nbolton: Implicit string conversion: https://en.cppreference.com/w/cpp/filesystem/path/string

PR #7689, opened 2024-10-14
build: linux packages with proper names using cpack

sithlord48:
- Linux Packages via CPack only with proper name
 
   - [x] Linux - DEBIAN
   - [x] Linux - RPM
   - [x] Linux -Arch Like

- new macro to set linux distro name and rpm / deb package release info

### DIFF
--- a/.github/workflows/ci-linux.json
+++ b/.github/workflows/ci-linux.json
@@ -64,13 +64,13 @@
       "name": "archlinux-amd64",
       "container": "deskflow/deskflow:archlinux-amd64",
       "runs-on": "ubuntu-latest",
-      "package-user": "build"
+      "arch-like": "true"
     },
     {
       "name": "manjaro-amd64",
       "container": "deskflow/deskflow:manjaro-amd64",
       "runs-on": "ubuntu-latest",
-      "package-user": "build"
+      "arch-like": "true"
     }
   ]
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -326,8 +326,22 @@ jobs:
       - name: Configure
         run: cmake -B build --preset=linux-release ${{ matrix.distro.extra-cmake-args }} -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_COMPILE_WARNING_AS_ERROR=ON
 
-      - name: Build
-        run: cmake --build build -j8
+      - name: Build Package
+        if: ${{ matrix.distro.arch-like != 'true'}}
+        run: cmake --build build -j8 --target package
+
+      - name: Build Package (Arch-like)
+        if: ${{ matrix.distro.arch-like == 'true' }}
+        run: |
+          cmake --build build -j8
+          sudo chown -R build build
+          cd build
+          sudo -u build makepkg -s
+          export OSNAME=$(cat /etc/os-release | grep ^ID= | sed 's/ID=//g')
+          export ARCH=$(uname -m)
+          mv *.pkg.* $(ls *.pkg.* | sed "s/$ARCH/$OSNAME-$ARCH/g")
+          cd ..
+
 
       - name: Tests
         uses: ./.github/actions/run-tests
@@ -335,20 +349,11 @@ jobs:
         with:
           job: linux-${{ matrix.distro.name }}
 
-      - name: Get version
-        uses: ./.github/actions/get-version
-
-      - name: Package
-        env:
-          LINUX_EXTRA_PACKAGES: ${{ matrix.distro.extra-packages }}
-          LINUX_PACKAGE_USER: ${{ matrix.distro.package-user }}
-        run: ./scripts/package.py --package-version ${{env.DESKFLOW_VERSION}}
-
       - name: Upload
         uses: actions/upload-artifact@v4
         with:
           name: package-${{ env.PACKAGE_PREFIX }}-${{ matrix.distro.name }}
-          path: ${{ env.PACKAGE_PATH }}
+          path: ${{github.workspace}}/build/deskflow[-_]*.*
 
   # Technically, "unix" is a misnomer, but we use it here to mean "Unix-like BSD-derived".
   unix:
--- a/cmake/Packaging.cmake
+++ b/cmake/Packaging.cmake
@@ -1,18 +1,8 @@
-# Deskflow -- mouse and keyboard sharing utility
-# Copyright (C) 2012-2024 Symless Ltd.
-# Copyright (C) 2009-2012 Nick Bolton
-#
-# This package is free software; you can redistribute it and/or
-# modify it under the terms of the GNU General Public License
-# found in the file LICENSE that should have accompanied this file.
-#
-# This package is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+# SPDX-FileCopyrightText: (C) 2024 Chris Rizzitello <sithlord48@gmail.com>
+# SPDX-FileCopyrightText: (C) 2012 - 2024 Symless Ltd.
+# SPDX-FileCopyrightText: (C) 2009 - 2012 Nick Bolton
+# SPDX-License-Identifier: MIT
+
 
 #
 # If enabled, configure packaging based on OS.
@@ -34,6 +24,7 @@ macro(configure_packaging)
     if(NOT CPACK_PACKAGE_VERSION)
       set(CPACK_PACKAGE_VERSION ${DESKFLOW_VERSION})
     endif()
+
     if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
       configure_windows_packaging()
     elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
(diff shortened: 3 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -327,28 +327,33 @@ jobs:
...
 
       - name: Build
-        run: cmake --build build -j8
+        if: ${{ matrix.distro.archlike != 'true'}}
+        run: cmake --build build -j8 --target package
+
+      - name: Arch Build
+        if: ${{ matrix.distro.archlike == 'true' }}
+        run: |
+          cmake --build build -j8
```
nbolton, line 336: Same as above, I think this should be its own step (a unified build step that doesn't also package). It's duplication at this point. Now there are 2 places where we have to maintain the build args.

.github/workflows/ci.yml
```diff
@@ -327,28 +327,33 @@ jobs:
...
+          sudo chown -R build build
+          cd build
+          sudo -u build makepkg -s
+          export OSNAME=$(cat /etc/os-release | grep ^ID= | sed 's/ID=//g')
+          export ARCH=$(uname -m)
+          mv *.pkg.* $(ls *.pkg.* | sed "s/$ARCH/$OSNAME-$ARCH/g")
+          cd ..
```
nbolton, line 343: I know you're not a big fan of scripts, but I think we could explore using a `.sh` script here. If that's not an option, how about we consider moving this to a composite action called `package-arch`?

Embedding bash scripts directly in workflows can make the workflow harder for new contributors to grok. 

However, I do see the flip side, that having to go to another file can break a developer's mental workflow when they are already in the place that the logic affects.

.github/workflows/ci-linux.json
```diff
@@ -64,13 +64,13 @@
       "name": "archlinux-amd64",
       "container": "deskflow/deskflow:archlinux-amd64",
       "runs-on": "ubuntu-latest",
-      "package-user": "build"
+      "archlike": "true"
```
nbolton, line 67: I think `arch-like` might be a tad more readable.

cmake/Packaging.cmake
```diff
@@ -1,18 +1,8 @@
...
-# This package is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+# SPDX-FileCopyrightText: 2024 Chris Rizzitello <sithlord48@gmail.com>
+# SPDX-FileCopyrightText: 2012 -2024 Symless Ltd.
+# SPDX-FileCopyrightText: 2009-2012 Nick Bolton
```
nbolton, line 3:
```suggestion
# SPDX-FileCopyrightText: 2012 Symless Ltd.
# SPDX-FileCopyrightText: 2009 Nick Bolton
```

cmake/Packaging.cmake
```diff
@@ -42,8 +33,13 @@ macro(configure_packaging)
       configure_linux_packaging()
     elseif(${CMAKE_SYSTEM_NAME} MATCHES "|.*BSD")
       message(STATUS "BSD packaging not yet supported")
+      set(OS_STRING ${CMAKE_SYSTEM_NAME}_${CMAKE_SYSTEM_PROCESSOR})
     endif()
 
+    # cmake-format: off
```
nbolton, line 39: No longer needed :tada:

cmake/Packaging.cmake
```diff
@@ -98,16 +101,90 @@ endmacro()
 #
 # Linux packages
 #
+
+macro(configure_linux_package_name)
+  ##Get Distro name for fall back
```
nbolton, line 106:
```suggestion
  # Get distro name for fall back.
```

cmake/Packaging.cmake
```diff
@@ -98,16 +101,90 @@ endmacro()
...
+
+  execute_process(
+    COMMAND bash "-c"
+            "cat /etc/os-release | grep ^VERSION_ID= | sed 's/VERSION_ID=//g'"
+    OUTPUT_VARIABLE _DISTRO_VERSION_ID
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+  string(REPLACE "\"" "" DISTRO_VERSION_ID "${_DISTRO_VERSION_ID}")
+  message(STATUS "DISTRO ID: ${DISTRO_VERSION_ID}")
+
+  #CHECK IF RUNNING ON GITHUB FOR A RUN ATTEMPT TO USE IT
```
nbolton, line 139:
```suggestion
  #CHECK IF RUNNING ON GITHUB FOR A RUN ATTEMPT TO USE IT
  # Check if it is running on GitHub and start the counter
```

Not sure I understand this comment, perhaps you could reword it? Maybe add some context as to why this is needed.

cmake/Packaging.cmake
```diff
@@ -98,16 +101,90 @@ endmacro()
...
+  set(PACKAGE_BUILD_ATTEMPT $ENV{GITHUB_RUN_ATTEMPT})
+  if("$ENV{GITHUB_RUN_ATTEMPT}" STREQUAL "")
+    set(PACKAGE_BUILD_ATTEMPT 1)
+  endif()
+
+  #Check if its Debian Like
```
nbolton, line 145:
```suggestion
  # Check if Debian-like
```

cmake/Packaging.cmake
```diff
@@ -98,16 +101,90 @@ endmacro()
...
+          "${PACKAGE_BUILD_ATTEMPT}~${DISTRO_CODENAME}")
+    endif()
+    set(CPACK_GENERATOR "DEB")
+    message(STATUS "DEB RELEASE INFO: ${CPACK_DEBIAN_PACKAGE_RELEASE}")
+  endif()
+
+  check_is_rpm()
+
+  if("${DISTRO_NAME}" STREQUAL "arch")
+      set(DISTRO_VERSION_ID "") #this has no meaning no arch for packages
```
nbolton, line 159:
```suggestion
      # Arch-like doesn't use version ID.
      set(DISTRO_VERSION_ID "")
```

cmake/Packaging.cmake
```diff
@@ -98,16 +101,90 @@ endmacro()
...
+  endif()
+  #Determain the code name to be used
```
nbolton, line 161:
```suggestion
  endif()
  
  # Determain the code name to be used.
```

cmake/Packaging.cmake
```diff
@@ -98,16 +101,90 @@ endmacro()
...
+
+endmacro()
+
 macro(configure_linux_packaging)
 
   message(VERBOSE "Configuring Linux packaging")
 
-  set(CPACK_GENERATOR "DEB;RPM")
+  # set fallback name
+  configure_linux_package_name()
```
nbolton, line 180:
```suggestion
  # Set fallback name.
  configure_linux_package_name()
```

Maybe some context around this might be helpful. Why set a fallback name? What will replace it?

cmake/Packaging.cmake
```diff
@@ -179,3 +256,41 @@ macro(configure_files srcDir destDir)
   endforeach(templateFile)
 
 endmacro(configure_files)
+
+macro(check_is_rpm)
+  #Check if its RPM Like
```
nbolton, line 261:
```suggestion
  # Check if RPM-like.
```

cmake/Packaging.cmake
```diff
@@ -179,3 +256,41 @@ macro(configure_files srcDir destDir)
...
+      set(DISTRO_CODENAME "tumbleweed")
+    elseif(("${DISTRO_NAME}" STREQUAL "almalinux"))
+      string(REGEX MATCH [0-9]+ V "${DISTRO_VERSION_ID}")
+      set(RPM_CN "alma${V}")
+    elseif(("${DISTRO_NAME}" STREQUAL "rocky"))
+      string(REGEX MATCH [0-9]+ V "${DISTRO_VERSION_ID}")
+      set(RPM_CN "rocky${V}")
+    endif()
+
+    #Determain the code name to be used
```
nbolton, line 277:
```suggestion
    # Determain the code name to be used.
```

Seems like this kind of logic might be duplicated or similar to elsewhere. Shouldn't it be a macro?

nbolton (changes requested):

.github/workflows/ci.yml
```diff
@@ -334,28 +334,29 @@ jobs:
         run: cmake -B build --preset=linux-release ${{ matrix.distro.extra-cmake-args }} -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_COMPILE_WARNING_AS_ERROR=ON
 
       - name: Build
-        run: cmake --build build -j8
+        run: cmake --build build -j8 --target package
```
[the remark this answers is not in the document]
sithlord48: because the package target will make packages on arch that makes (for some reason ) a tar.gz , self extracting tar and tar.z archive so rather then fight CPack (it must have a generator ) we just don't use the package target on arch build as cpack can not those packages anyway.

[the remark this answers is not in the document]
sithlord48: >Rolling it into one makes it harder to identify if the step failed because of a build problem or package problem.

Its one step very clearly logged if you look at the output NO reason to split these for any job where cpack can package.

.github/workflows/ci.yml
sithlord48: The workflows are expected to have bash  in them these are commands.
Leave the CI naked as possible scripts just make it harder to follow .

cmake/Packaging.cmake
sithlord48: only arch doesn't use that

nbolton:
@sithlord48 Did you see the conversations that GitHub _helpfully_ hid?

image

sithlord48:
> @sithlord48 Did you see the conversations that GitHub _helpfully_ hid?
> 
> image

Yes i did but i chose to not respond to some .

cmake/Packaging.cmake
sithlord48: I plan to do so seperate Cmake sweep to "fix" formating all in one go

cmake/Packaging.cmake
sithlord48: It does not start a counter this var is set by github when the job starts. 
Its how many Attempts the job has had , in packages it the -# before the distro name 
`-1~jammy`  for example

cmake/Packaging.cmake
sithlord48: ranges are expect for years you have worked on them 
valid forms are 

2000 (one year you did changes) 
2000 2005 (a change in those two years)
2000 - 2020 (changes every year for that span)

changes must be non white space change etc..

Also we are missing the (C) that is expected

nbolton (approved):

PR #7690, opened 2024-10-15
ci: clearer and cleaner steps

sithlord48:
Sight reorder to the steps so that 

 All jobs need lint first  
 Simplify conditions 

Needs #7691 to pass arch

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ env:
 jobs:
   pr-comment-flags:
     runs-on: ubuntu-latest
+    needs: [lint-cmake, lint-clang]
     if: ${{ github.event_name == 'pull_request' }}
 
     outputs:
@@ -57,7 +58,7 @@ jobs:
   # Quality gate to allow PR merge, used in the branch protection rules.
   ci-passed:
     runs-on: ubuntu-latest
-    needs: [lint-cmake, lint-clang, windows, macos, linux, unix]
+    needs: [test-results, unix]
 
     steps:
       - run: echo "✅ CI passed" > $GITHUB_STEP_SUMMARY
@@ -106,7 +107,7 @@ jobs:
     uses: ./.github/workflows/valgrind-analysis.yml
 
   analyse-sonarcloud:
-    needs: [lint-cmake, lint-clang, pr-comment-flags]
+    needs: pr-comment-flags
     if: ${{ needs.pr-comment-flags.outputs.no-sonar != 'true' }}
     uses: ./.github/workflows/sonarcloud-analysis.yml
     secrets:
@@ -282,6 +283,7 @@ jobs:
           path: ${{ env.PACKAGE_PATH }}
 
   linux-matrix:
+    needs: [lint-cmake, lint-clang]
     runs-on: ubuntu-latest
 
     outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -57,7 +58,7 @@ jobs:
   # Quality gate to allow PR merge, used in the branch protection rules.
   ci-passed:
     runs-on: ubuntu-latest
-    needs: [lint-cmake, lint-clang, windows, macos, linux, unix]
+    needs: [test-results, unix]
```
nbolton, line 61: I'm not sure this will work as you expect. The `test-results` step will pass when other steps fail:
```
always() && needs.windows.result != 'skipped' && needs.macos.result != 'skipped' && needs.linux.result != 'skipped'
```

If you want to simplify it like this, we'd need another step that has only `needs: [windows, macos, linux]` with no extra conditions.

.github/workflows/ci.yml
```diff
@@ -106,7 +107,7 @@ jobs:
     uses: ./.github/workflows/valgrind-analysis.yml
 
   analyse-sonarcloud:
-    needs: [lint-cmake, lint-clang, pr-comment-flags]
+    needs: pr-comment-flags
```
nbolton, line 110: Seems like it could be a good change. I'm guessing you're doing this because `analyse-sonarcloud` is the same kind of thing as linting, so it'd be better to run it in parallel?

.github/workflows/ci.yml
```diff
@@ -282,6 +283,7 @@ jobs:
           path: ${{ env.PACKAGE_PATH }}
 
   linux-matrix:
+    needs: [lint-cmake, lint-clang]
```
nbolton, line 286: Yeah this was bugging me. Good change.

.github/workflows/ci.yml
```diff
@@ -401,7 +403,7 @@ jobs:
 
   release_assets:
     name: Release
-    needs: [windows, linux, macos]
+    needs: ci-passed
```
nbolton, line 406: I guess a side effect here is that this now depends on the `unix` step which doesn't provide a package. Won't cause any harm though.

.github/workflows/ci.yml
```diff
@@ -401,7 +403,7 @@ jobs:
 
   release_assets:
```
nbolton, line 404: Minor nit I didn't notice before; this should be `kabab-case` not `snake_case`.

```suggestion
  release-assets:
```

nbolton (changes requested):

.github/workflows/ci.yml
nbolton: I stand corrected.
Screenshot from 2024-10-15 09-20-13

Even though the test results job passes, the `ci-passed` job is skipped. I'm baffled.
image

nbolton: I stand corrected. That is baffling.
Screenshot from 2024-10-15 09-24-59
Screenshot from 2024-10-15 09-36-18

.github/workflows/ci.yml
nbolton: Fixed a bit more thoroughly in: https://github.com/deskflow/deskflow/pull/7703

.github/workflows/ci.yml
sithlord48: CI here has all steps passed what is the problem ?

.github/workflows/ci.yml
sithlord48: `pr-comment-flags` requires the lint jobs so this will require them implicitly

.github/workflows/ci.yml
nbolton: My concern was in the failure scenario; I was worried that `ci-passed` is conditional based on `test-results` which always passes (because it runs on failure). But for some reason that doesn't happen, which is why I'm stumped.

Doesn't matter, it's a gap in my knowledge that I can't figure out right now.

nbolton (approved):

Issue #7671, opened 2024-10-13
Use Squish instead of writing Google Test unit tests for the GUI


nbolton:
### Project

Deskflow

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] BSD-derived
- [ ] Other

### OS versions/distros

n/a

### Deskflow configuration

n/a

### What problem are you trying to solve?

Unit testing Qt GUI code is challenging because GUIs are event-driven, making it hard to isolate components and simulate real user interactions. GUI tests can become fragile, frequently breaking with minor layout or design changes, and they often fail to cover visual and usability aspects effectively. Simulating events in Qt is complex and leads to tests that are difficult to maintain, while the resource-intensive nature of GUI tests slows down execution. Instead, functional or integration testing tools like Squish, automated GUI tests, or behavior-driven frameworks offer better alternatives for testing the full user experience and interaction.

### What solution do you propose?

Instead of unit testing Qt GUIs, use functional or integration testing tools like Squish or BDD frameworks to simulate real user interactions and test end-to-end workflows. Automated GUI testing tools, behavior-driven testing, and end-to-end frameworks like TestComplete or Appium provide better coverage for GUI behavior, while snapshot testing and manual testing handle visual aspects effectively. These approaches ensure the entire user interface functions as intended, unlike unit tests which struggle with isolated components and interactions.

nbolton:
@sithlord48 Does this accurately represent your views on unit testing GUIs?

As part of this we can also move the GUI lib files back into GUI and perhaps also fix the weird `/src/gui/src` tree too... maybe.

PR #7691, opened 2024-10-15
Solve QCheckBox 6.7+ warning and exclude GUI from coverage

sithlord48:
- Fix warnings causing arch build to fail for new Qt signal 

 Something is wrong with sonar coverage is too low but all the uncovered code is everything but the code i changed in that file


nbolton:
> Something is wrong with sonar coverage is too low but all the uncovered code is everything but the code i changed in that file

I pushed a commit to your branch to add `src/gui/**` to the coverage exclusions since we'll be moving to Squash/similar (#7671).

nbolton:
> * Fix warnings causing arch build to fail for new Qt signal

I won't yammer on about this too much but for me, this kind of externally-caused issue proves the usefulness of a nightly build; CI failed due to factors outside of our control. This will have been easy for you to solve since you're a Qt expert, but for someone like me I would have wasted time wondering what I broke (had I seen this in one of my PRs).

That said, I will concede that nightly might not have helped in this particular case since the change in Arch may not have occurred between nightly and next push. So, maybe the cron is trying to solve an impossible problem.

nbolton (approved):
I did a cursory check with my older Qt version (Debian Trixie). I'll assume you tested these on Arch Linux.

PR #7697, opened 2024-10-15
Ignore paths for CodeQL Analysis

nbolton:
Copied ignore paths from `ci.yml` now that CodeQL Analysis is running on it's own trigger.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -12,6 +12,13 @@ on:
   workflow_dispatch:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
+    paths-ignore:
+      - '**/*.md'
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -12,6 +12,13 @@ on:
   workflow_dispatch:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
+    paths-ignore:
```
sithlord48, line 15: you may want to use ' not " to prevent possible expansion of escape characters

nbolton: Yup, accidents happen and the YAML error for that scenario is horrible.

sithlord48 (approved):

PR #7703, opened 2024-10-15
Fixed code style in ci.yml and add .editorconfig

nbolton:


.editorconfig
sithlord48, line 1: This will affect all files this can now fight the linter every save.

.github/workflows/ci.yml
```diff
@@ -6,22 +6,22 @@ name: CI
...
       - synchronize
       - ready_for_review
     paths-ignore:
-      - '**/*.md'
-      - '.github/ISSUE_TEMPLATE/**'
-      - '.env-example'
-      - '.gitignore'
-      - '.gitattributes'
-      - 'cspell.json'
+      - "**/*.md"
```
sithlord48, line 19: ' and " are not the same in yml

We want `'` as `"` will try to parse escape characters

sithlord48 (changes requested):

.editorconfig
nbolton: I'll move this to it's own PR: #7706

.github/workflows/ci.yml
```diff
@@ -399,30 +399,30 @@ jobs:
...
-
-    - name: Deploy Release
-      if: contains(github.ref, '/tags/v')
-      uses: crowbarmaster/GH-Automatic-Releases@latest
-      with:
-       repo_token: "${{ secrets.DEPLOY_TOKEN }}"
-       prerelease: false
-       files: |
-         package-*/*
+      - name: download-artifacts
```
sithlord48, line 407: These name are meant to be on labels. Spaces are fine

nbolton: It's not the spaces that I'm worried about; consistency is my main concern here. If you're vehemently against using the kebab case for consistency, I'm happy to change all the other names to title case.

_Why I have such a hard-on for consistency:_ Consistency is important for reducing cognitive load because it creates a predictable environment where contributors don’t need to spend extra mental energy figuring out patterns or exceptions. When code or naming conventions are consistent, contributors can focus on the task at hand rather than constantly adapting to small variations. This familiarity allows for quicker understanding and reduces the risk of errors. Inconsistent patterns, even minor ones, force the brain to switch gears more often, leading to mental fatigue and distraction from actual problem-solving or development work.

sithlord48: This is a label shown on the web page we don't need to have the -

sithlord48: If you want to that is fine i was just pointing out that is was a user string (all the names are) so Spaces are ok to have in those.

sithlord48 (dismissed):

.github/workflows/ci.yml
nbolton: Sorry, I got mixed up. It's only job names that were kabab case. Step names are of course not kebab case. I only had a few hours sleep last night so I'm not thinking straight today.

PR #7706, opened 2024-10-15
docs: Add .editorconfig file

nbolton:
Intent: Attempting to keep indentation consistent across workflow files and CMake.


nbolton:
> maybe we want *.cmake and CMakeLists.txt (cmake files) to have their own config

How's that?

sithlord48 (approved):

PR #7713, opened 2024-10-16
docs: Example file for `--config-toml` arg

nbolton:
For anyone wondering how to use the `--config-toml` arg.


nbolton:
> Great Idea but put it into #7673.

Will do.

nbolton:
Rebased to #7673.

Issue #58, opened 2014-10-12
Extra mouse buttons


nbolton:
**Imported issue:**
- Author: nick.bolton.uk (Google user)
- Date: 2009-05-24 00:00:00
- Legacy ID: 67

It looks like some users are having trouble using additional mouse buttons  
such back and forward, etc. I know from experience that mouse buttons can  
have an array of extra buttons, which some users do make use of.  

From user, Id:  
"http://groups.google.com/group/synergy-":http://groups.google.com/group/synergy-  
plus/browse_thread/thread/015cb5e4da7d7ed4  

Is there a way to get the mouse buttons (back and forward on the side  
of the mouse) to work using synergy?   

I am using a Windows synergy server and a Ubuntu client.   

Also, when I press a keyboard button and keep it pressed, the key is  
not sent several times afterwards. For example, while editing this  
text, if I put my cursor at the end, then press backspace (and keep it  
pressed), only one letter will be erased.

keab42:
I have a Microsoft mouse with the forward and back buttons. I'm using a Mac as the server with the intellipoint drivers to get them working correctly, however Synergy is mapping them oddly to the windows client. Instead of acting as back and foward, they seem to act to snap windows to the side of the screen. What mappings do I need to make to correct this?

nkoroste:
Same issue.
Logitech G500s
Server: Ubuntu 14.04
Client: OS X 10.10.5

noscript:
Doesn't work for me either. Tried Logitech MX Master and Logitech Performance MX.

Host: Debian
Guest: Windows
Synergy 1.7.4

Lihis:
Does not work here either. Back and forward buttons does nothing. Mouse is SteelSeries Rival.

Host: Arch Linux
Guest: Windows 10
Synergy: 1.7.5

However, got the buttons working by adding this to Synergy server configuration in the section: options
```
mousebutton(6) = mousebutton(4)
mousebutton(7) = mousebutton(5)
```

noscript:
> However, got the buttons working by adding this to Synergy server configuration in the section: options

It will not work with config file generated by wizard, it uses a completely a different format.

That's how mapping mouse button 4 to 4 generated:
```
hotkeys\2\keys\1\key=5
hotkeys\2\keys\size=1
hotkeys\2\actions\1\keys\1\key=5
hotkeys\2\actions\1\keys\size=1
hotkeys\2\actions\1\type=2
[...]

Issue #7688, opened 2024-10-14
Bug: MacOS client randomly disconnects


rodalpho:
Maybe related: #7329

### Project

Deskflow

### Deskflow version number

1.17.0+r1 44bd69fd

### Build type

Deskflow package (downloaded from us)

### Operating systems (OS)

- [X] Windows
- [X] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other

### Wayland on Linux

- [ ] I have reviewed the Wayland known issues and my issue is new
- [X] I am not using Wayland on Linux

### Signing on macOS

- [X] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### OS versions/distros

MacOS -- latest as of today, M3
Win11 -- also latest x64

### Deskflow configuration

Win11 server, MacOS client on left

### What steps will reproduce the problem?

I randomly get disconnects with my MacOS client. This is extremely annoying when using it as disconnects move my mouse pointer back to the center of the windows screen. This is not related to a network disconnection even though the logs pretty clearly suggest it is; the mac is connected to wired ethernet and I had a ping running from the windows machine to the MacOS laptop the entire time just to confirm, not a single ping was dropped.

I have the RU keyboard installed (but not used) on windows is because I read somewhere that many trojans would silently exit if they thought you were Russian. I didn't add it to the macOS side but I can if you think that's the issue. 

### Log output

```shell
Log from MacOS side:

[2024-10-14T14:51:24] ERROR: tls error occurred (generic failure)
[2024-10-14T14:51:24] ERROR: openssl error: error:0A000126:SSL routines::unexpected eof while reading
[2024-10-14T14:51:24] NOTE: disconnected from server
[2024-10-14T14:51:25] NOTE: connecting to '<<removed>>': <<removed>>:24800
[2024-10-14T14:51:25] NOTE: server fingerprint: <<removed>>
[2024-10-14T14:51:25] WARNING: You need to install these languages on this computer and restart Deskflow to enable support for multiple languages: ru
[2024-10-14T14:51:25] WARNING: cursor may not be visible
[2024-10-14T14:51:25] INFO: connected to secure socket
[2024-10-14T14:51:25] INFO: server tls certificate info: /CN=Deskflow
[2024-10-14T14:51:25] INFO: network encryption protocol: TLSv1.3
[2024-10-14T14:51:25] INFO: local languages: en
[...]

PR #7718, opened 2024-10-16
Increase NumButtonIDs to 7, to support mice with more buttons

dkniffin:
[Maybe] fixes: #58

### DIFF
--- a/src/lib/deskflow/mouse_types.h
+++ b/src/lib/deskflow/mouse_types.h
@@ -39,4 +39,4 @@ static const ButtonID kMacButtonRight = 2;
 static const ButtonID kMacButtonMiddle = 3;
 //@}
 
-static const UInt8 NumButtonIDs = 5;
+static const UInt8 NumButtonIDs = 7;


dkniffin:
@nbolton I was able to install the build artifacts on server and client and got them connected. It didn't resolve the issue, but now when I press one of the two buttons, I get the following error on the server: 
```
[2024-10-16T17:10:10] ERROR: tls error occurred (generic failure)
[2024-10-16T17:10:10] ERROR: openssl error: error:0A000126:SSL routines::unexpected eof while reading
```
and this error on the client:
```
[2024-10-16T17:10:10] WARNING: desktop process exited with error code: 4
```

any ideas?

nbolton:
> [2024-10-16T17:10:10] ERROR: tls error occurred (generic failure)
> [2024-10-16T17:10:10] ERROR: openssl error: error:0A000126:SSL routines::unexpected eof while reading

That's funny, literally just talking about that in: #7688

I think it might be unrelated to your change though and just a random bug.

> [2024-10-16T17:10:10] WARNING: desktop process exited with error code: 4

Might need a bit more context. Could you turn up the logging to debug?

dkniffin:
I turned it up to debug1, and I see this:
```
[2024-10-16T17:20:48] DEBUG1: recv mouse down id=1
[2024-10-16T17:20:48] DEBUG1: faking mouse button id: 0 press: pressed
[2024-10-16T17:20:49] DEBUG1: recv mouse up id=1
[2024-10-16T17:20:49] DEBUG1: faking mouse button id: 0 press: released
[2024-10-16T17:20:51] DEBUG1: recv mouse down id=7
[2024-10-16T17:20:51] DEBUG1: faking mouse button id: 6 press: pressed
[2024-10-16T17:20:51] WARNING: desktop process exited with error code: 4
```

So I guess it's faking it successfully now, but somehow that's causing a crash.

dkniffin:
Maybe a new item needs to be added in [here](https://github.com/deskflow/deskflow/blob/52f616a15d094bb3ba7ec30a2a04b764f7b2bf22/src/lib/platform/OSXScreen.mm#L448-L454)?

nbolton:
> Maybe a new item needs to be added in [here](https://github.com/deskflow/deskflow/blob/52f616a15d094bb3ba7ec30a2a04b764f7b2bf22/src/lib/platform/OSXScreen.mm#L448-L454)?
> 
> Edit: I'm thinking that because I suspect an out-of-bounds error [here](https://github.com/deskflow/deskflow/blob/52f616a15d094bb3ba7ec30a2a04b764f7b2bf22/src/lib/platform/OSXScreen.mm#L590)

Those lines do look related. I'm sure there's somewhere else that mouse buttons are defined but I'd need to dig.

nbolton:
@sithlord48 Aha, I think our branch protection rules (linear history) can only apply to _our_ branches in _our_ repo, i.e. not PRs from external branches. I just tried the merge button out of curiosity, and it allowed it, which makes sense since `dkniffin:increase-numbuttonids` is not under our control.

dkniffin:
Another question/concern I have about this, is how to know which button IDs are which? Do all mice transmit the same button event IDs? It seems no, since mine uses 4 and 5 for the scroll wheel. Is there a way to get a consistent behavior with that?

nbolton:
> Another question/concern I have about this, is how to know which button IDs are which? Do all mice transmit the same button event IDs? 

Deskflow listens to the OS API, not the mouse, so it's down to the OS.

> It seems no, since mine uses 4 and 5 for the scroll wheel. Is there a way to get a consistent behavior with that?

I'm not sure. I wonder if the OS API docs might reveal something?

dkniffin:
This didn't fix the issue, and I don't have more time to troubleshoot

Issue #7719, opened 2024-10-16
Unable to find `openssl.exe` on Windows in dev env


nbolton:
### Project

Deskflow

### Deskflow version number

v1.17.0.65

### Build type

Local developer build (built it myself)

### Operating systems (OS)

- [X] Windows
- [ ] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [ ] I have reviewed the Wayland known issues and my issue is new
- [X] I am not using Wayland on Linux

### Signing on macOS

- [ ] I have authorized the app to run on my Mac
- [X] I am not using macOS

### OS versions/distros

Windows 10

### What steps will reproduce the problem?

Start GUI

![image](https://github.com/user-attachments/assets/5cc69bb7-82de-47a5-95c8-ad986748238c)

### Log output

```shell
v1.17.0.65
openssl binary not found:  "D:/Projects/deskflow/deskflow/build/vcpkg_installed/x64-windows/bin/openssl.exe"
D:\Projects\deskflow\deskflow\src\lib\gui\tls\TlsCertificate.cpp:75
```

### Additional information

Caused by: #7640

PR #7687, opened 2024-10-14
Protocol interoperability for Input Leap (Barrier) and Synergy

sithlord48:
Fixes: #7526

Blocked by: #7741, #7745, #7744, #7752

Tasks:
- [x] Client dynamically responds as "Synergy" or "Barrier" hello back
- [x] Server can be set statically to "Synergy" or "Barrier" hello
- [x] Remove commit: `hack: Enable TOML++ and CLI11`
- [x] Remove commit:  `hack: workaround pending #7741`
- [x] Set default protocol depending on [vote](https://github.com/deskflow/deskflow/discussions/7742)

**Testing:**
1. Run either an Input Leap server or a Deskflow server, and the client works with either.
2. Run either an Input Leap client or a Deskflow client, and the server works with either.

Set the protocol in the server config:
image

PR #7720, opened 2024-10-16
fix: Use `OPENSSL_EXE_DIR` instead of `OPENSSL_ROOT_DIR` on Windows

nbolton:
Fixes: #7719

### DIFF
--- a/cmake/Libraries.cmake
+++ b/cmake/Libraries.cmake
@@ -525,9 +525,9 @@ macro(configure_openssl)
   if(WIN32) #Used for dev in TLS and WIX
     cmake_path(SET OPENSSL_ROOT_DIR NORMALIZE
                "${CMAKE_BINARY_DIR}/vcpkg_installed/x64-windows")
-    message(STATUS "SET OPENSSL ROOT: ${OPENSSL_ROOT_DIR}")
+    message(VERBOSE "Set OPENSSL_ROOT_DIR: ${OPENSSL_ROOT_DIR}")
     set(OPENSSL_EXE_DIR "${OPENSSL_ROOT_DIR}/tools/openssl")
-    add_definitions(-DOPENSSL_EXE_DIR="${OPENSSL_ROOT_DIR}")
+    add_definitions(-DOPENSSL_EXE_DIR="${OPENSSL_EXE_DIR}")
   endif()
 
 endmacro()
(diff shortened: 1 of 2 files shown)


cmake/Libraries.cmake
```diff
@@ -525,9 +525,9 @@ macro(configure_openssl)
   if(WIN32) #Used for dev in TLS and WIX
     cmake_path(SET OPENSSL_ROOT_DIR NORMALIZE
                "${CMAKE_BINARY_DIR}/vcpkg_installed/x64-windows")
-    message(STATUS "SET OPENSSL ROOT: ${OPENSSL_ROOT_DIR}")
+    message(VERBOSE "Set OPENSSL_ROOT_DIR: ${OPENSSL_ROOT_DIR}")
```
sithlord48, line 528: Why not as a status

I use CAPS so it stands out in the command output.

cmake/Libraries.cmake
```diff
@@ -525,9 +525,9 @@ macro(configure_openssl)
...
     set(OPENSSL_EXE_DIR "${OPENSSL_ROOT_DIR}/tools/openssl")
-    add_definitions(-DOPENSSL_EXE_DIR="${OPENSSL_ROOT_DIR}")
+    add_definitions(-DOPENSSL_EXE_DIR="${OPENSSL_EXE_DIR}")
```
sithlord48, line 530: I can't test this so ill take your word you tested and it works.

cmake/Libraries.cmake
nbolton: Personal preference: Caps annoy the hell out of me. The shoutyness of it degrades the SNR. What one developer is interested in may not be the same as another.

I would say it's verbose because it's related to debugging the OpenSSL path, which isn't something you'd normally need to know about.

cmake/Libraries.cmake
nbolton: Yep, it was blocking me on #7687 and now I'm unblocked.

sithlord48 (approved):

cmake/Libraries.cmake
sithlord48, line 528: In the future split these kinds of non code changes  that don't have anything todo w/ other changes in  the commit.

sithlord48 (approved):

PR #7731, opened 2024-10-17
chore: Update feature request template

sithlord48:
For a feature request we only want to know about the proposed feature

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -6,85 +6,11 @@ body:
     attributes:
       value: Thanks for taking the time to help us improve Deskflow.
 
-  - type: dropdown
-    id: project
-    attributes:
-      label: Project
-      description: Are you using Deskflow or a fork/derivative?
-      options:
-        # Empty option to force selection
-        -
-        - Deskflow
-        - Barrier
-        - Input Leap
-        - Synergy
-      default: 0
-    validations:
-      required: true
-
-  - type: checkboxes
-    id: os
-    attributes:
-      label: Operating systems (OS)
-      description: Which operating systems (OS) are you using?
-      options:
-        - label: Windows
-        - label: macOS
-        - label: Linux
-        - label: BSD-derived
-        - label: Other
-    validations:
-      required: true
-
   - type: textarea
-    id: os-version
+    id: request
     attributes:
-      label: OS versions/distros
+      label: Feature Request
       description: |
-        Please provide the version number of your operating system (OS).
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -6,85 +6,11 @@ body:
...
-      label: What problem are you trying to solve?
-      description: |
-        The most important part of problem solving is understanding the problem.
-        The more detail you provide, the better we can understand your idea.
-        “If I had only one hour to save the world, 
-        I would spend fifty-five minutes defining the problem,
-        and only five minutes finding the solution.”
-        — Albert Einstein
-    validations:
-      required: true
```
nbolton, line 81: I'd like to at least keep the problem/solution format. I have found it yields much more coherent feature requests when people are guided into thinking about the problem they're trying to solve.

I can see the other fields being a bit much though.

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -6,85 +6,11 @@ body:
...
-    attributes:
-      label: OS versions/distros
-      description: |
-        Please provide the version number of your operating system (OS).
-        If you're using Linux, please provide the name of the distribution.
-      placeholder: |
-        - Windows 11
-        - macOS 15
-        - Ubuntu 24.04
-        - FreeBSD 14.0
```
nbolton, line 50: You sure you don't want to keep the OS but make it optional? I'm happy to let this one go if you think it's just noise.

.github/ISSUE_TEMPLATE/feature_request.yml
sithlord48: Feature requests do not always have a "problem"

.github/ISSUE_TEMPLATE/feature_request.yml
sithlord48: Why do you think only ppl with some os will care about the feature>?

.github/ISSUE_TEMPLATE/feature_request.yml
sithlord48: My general thought here more mandatory fields the more friction for a feature request.

nbolton: Ok, let's try it.

.github/ISSUE_TEMPLATE/feature_request.yml
nbolton: You got me. Can't think of a good reason.

nbolton (approved):

PR #7736, opened 2024-10-18
build: use vcpkg for all depends, vcpkg action on ci to cache

sithlord48:


.github/workflows/ci.yml
```diff
@@ -109,7 +109,6 @@ jobs:
     name: ${{ matrix.target.name }}
     runs-on: ${{ matrix.target.runs-on }}
     container: ${{ matrix.target.container }}
-    timeout-minutes: 20
```
nbolton, line 112: Without this, the runner can run indefinitely. I've seen this happen many times... it just hangs there forever. Big waste of energy.

nbolton (changes requested):

cmake/Libraries.cmake
```diff
@@ -523,8 +523,7 @@ macro(configure_openssl)
 
   find_package(OpenSSL 3.0 REQUIRED COMPONENTS SSL Crypto)
   if(WIN32) #Used for dev in TLS and WIX
-    cmake_path(SET OPENSSL_ROOT_DIR NORMALIZE
-               "${CMAKE_BINARY_DIR}/vcpkg_installed/x64-windows")
+    cmake_path(SET OPENSSL_ROOT_DIR NORMALIZE "${OPENSSL_INCLUDE_DIR}/..")
```
nbolton, line 526: How about `OPENSSL_ROOT_DIR`?

nbolton (approved):

PR #7737, opened 2024-10-18
docs: Checkbox for continuous build on bug template

nbolton:
I hate to add one more checkbox, but this is important as we keep asking it and sometimes forget. I also made one more field optional to reduce friction. Still no support for [radio buttons](https://github.com/orgs/community/discussions/4318) yet from GitHub.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -80,6 +80,15 @@ body:
         - label: I have authorized the app to run on my Mac
         - label: I am not using macOS
 
+  - type: checkboxes
+    id: continuous-build
+    attributes:
+      label: Continuous build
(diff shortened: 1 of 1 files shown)


sithlord48 (approved):

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -92,8 +101,6 @@ body:
         - macOS 15
         - Ubuntu 24.04
         - FreeBSD 14.0
-    validations:
-      required: true
```
sithlord48, line 96: why don't we want to require os entry ?

nbolton: > I also made one more field optional to reduce friction.

I was worried that we're overloading the reporter with too many fields, so I did this to relieve the friction.

I'm happy to keep this in if you think it's a good idea though.

sithlord48 (approved):

PR #7756, opened 2024-10-19
Static cast screen column count for scan alert 559

sithlord48:
Resolve: https://github.com/deskflow/deskflow/security/code-scanning/559

### DIFF
--- a/src/gui/src/ScreenSetupModel.cpp
+++ b/src/gui/src/ScreenSetupModel.cpp
@@ -39,7 +39,8 @@ ScreenSetupModel::ScreenSetupModel(ScreenList &screens, int numColumns, int numR
     return;
   }
 
-  if (m_NumColumns * m_NumRows > screens.size()) {
+  const long span = static_cast<long>(m_NumColumns) * m_NumRows;
+  if (span > screens.size()) {
     qFatal("scrren list (%lld) too small for %d columns x %d rows", screens.size(), m_NumColumns, m_NumRows);
   }
 }


src/gui/src/ScreenSetupModel.cpp
```diff
@@ -39,7 +39,8 @@ ScreenSetupModel::ScreenSetupModel(ScreenList &screens, int numColumns, int numR
     return;
   }
 
-  if (m_NumColumns * m_NumRows > screens.size()) {
+  long span = static_cast<long>(m_NumColumns) * m_NumRows;
```
nbolton, line 42:
```suggestion
  const auto span = static_cast<long>(m_NumColumns) * m_NumRows;
```

nbolton: Minor.

nbolton (approved):

nbolton:
Commit message suggestion:
> fix: Static cast screen column count for scan alert #559

sithlord48:
i was  hoping git would attach the proper issue but tagging code scan you need a full link :(

src/gui/src/ScreenSetupModel.cpp
sithlord48: we don't want auto in this case as the issue is that the larger result can overflow if stored in a temp `int32_t` and not forced into the `long` bounds.

nbolton (approved):

PR #7757, opened 2024-10-19
ci: Normalize CI step names, descriptions, etc

nbolton:
:warning: Nickpick alert (minor nits ahead)

Also:
- Surfaced the comment about the Kitware repo so it's more visible
- Remove the hard-coded CMake version in comment which is likely to get out of sync and confuse someone

Names of things in the GitHub workflows and actions should be consistent.

Right now the convention is "Sentence case" but if we want to change to "lower case" then we can, or we can even change to "Title Case"... I don't mind what we use as long as we're consistent.


.github/actions/add-kitware-repo/action.yml
```diff
@@ -1,30 +1,33 @@
-# Add Kitware repo action
```
nbolton, line 1: I think page titles in comment serve a good purpose, but only when there isn't already a `name:` key with the same/similar value (in this case it's redundant).

.github/actions/add-kitware-repo/action.yml
```diff
@@ -1,30 +1,33 @@
...
+
           wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null \
             | gpg --dearmor - \
             > /usr/share/keyrings/kitware-archive-keyring.gpg
-          echo 'deb [signed-by=/usr/share/keyrings/kitware-archive-keyring.gpg] https://apt.kitware.com/ubuntu/ ${{ inputs.distro }} main' \
+
+          echo 'deb [signed-by=/usr/share/keyrings/kitware-archive-keyring.gpg] \
+            https://apt.kitware.com/ubuntu/ ${{ inputs.distro }} main' \
             > /etc/apt/sources.list.d/kitware.list
+
```
nbolton, line 30: Should be no functional changes here; simply making things a tad more readable.

.github/workflows/ci.yml
```diff
@@ -110,8 +110,10 @@ jobs:
     name: ${{ matrix.target.name }}
     runs-on: ${{ matrix.target.runs-on }}
     container: ${{ matrix.target.container }}
+
+    # CI should normally takes less than 10 mins to run,
+    # but if vcpkg has to rebuild the cache (should be rare), it takes about 90 mins.
     timeout-minutes: 120
-    # If vcpkg has to rebuild the cache it takes a while
```
nbolton, line 114: Comments below the thing they refer to can be confusing. Best to have it above and break the thing away from the other things so it's clear what the comment refers to.

.github/workflows/ci.yml
```diff
@@ -320,7 +322,7 @@ jobs:
         with:
           distro: jammy
 
-      - name: Update Cmake
+      - name: Update CMake
```
nbolton, line 325: Heads up, I have a strong preference for spelling names exactly the way the brand owner intends.

.github/actions/add-kitware-repo/action.yml
```diff
@@ -1,30 +1,32 @@
...
 
   steps:
-    - name: Cmake-3.24
+
+    # This mirrors instructions at https://apt.kitware.com
+    - name: CMake
       run: |
           apt update -y -qqq
           apt install ca-certificates gpg wget -y -qqq
-          # Deskflow requires at least CMake 3.24
```
nbolton, line 21: Intentionally removed. My concern is that this version number will get out of sync with the single source of truth and cause confusion.

.github/actions/add-kitware-repo/action.yml
```diff
@@ -1,30 +1,32 @@
...
     required: true
 
 runs:
   using: "composite"
 
   steps:
-    - name: Cmake-3.24
+
+    # This mirrors instructions at https://apt.kitware.com
+    - name: CMake
```
sithlord48, line 18: This step should be call "Add Repo" as it does not Install Cmake at all

.github/actions/add-kitware-repo/action.yml
```diff
@@ -1,30 +1,32 @@
...
       run: |
           apt update -y -qqq
           apt install ca-certificates gpg wget -y -qqq
-          # Deskflow requires at least CMake 3.24
-          # This mirrors instructions at https://apt.kitware.com
+
```
sithlord48, line 22: are you allowed empty new lines even in run block ?

sithlord48 (dismissed):

sithlord48:
Next time lets get all these  Nits Before the code lands. I really don't like fixing commit nits with more commits

.github/actions/add-kitware-repo/action.yml
nbolton: Fun fact, these composite action step names don't show up at all in the GitHub workflow log output (not yet anyway). Super helpful... not.

I'll fix it anyway in case that changes.

sithlord48 (approved):

sithlord48 (approved):

PR #7765, opened 2024-10-19
build: Use Fetch Content for Gtest

sithlord48:
Use Fetch Content to get gtest and gmock


src/test/CMakeLists.txt
```diff
@@ -14,6 +14,17 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+include(FetchContent)
+set(INSTALL_GTEST OFF)
+FetchContent_Declare(
```
nbolton, line 19: No `find_package`? Downstream packagers want us to check system first.

sithlord48: Nope, part of what makes this cool is that by simply adding `FIND_PACKAGE_ARGS` to this `FetchContent_Declare` call it will use find package automagiclly. Then failing that we will download from git

Technically this happens when `FetchContent_MakeAvailable(name)`

nbolton: Ooooh damn, I didn't read the docs thoroughly enough! So cool. Because there was no log output about GTest, it confused me, but yeah I should have read the docs.

> This option is for scenarios where the [FetchContent_MakeAvailable()](https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_makeavailable) command may first try a call to [find_package()](https://cmake.org/cmake/help/latest/command/find_package.html#command:find_package) to satisfy the dependency for <name>. By default, such a call would be simply find_package(<name>), but FIND_PACKAGE_ARGS can be used to provide additional arguments to be appended after the <name>. FIND_PACKAGE_ARGS can also be given with nothing after it, which indicates that [find_package()](https://cmake.org/cmake/help/latest/command/find_package.html#command:find_package) can still be called if [FETCHCONTENT_TRY_FIND_PACKAGE_MODE](https://cmake.org/cmake/help/latest/module/FetchContent.html#variable:FETCHCONTENT_TRY_FIND_PACKAGE_MODE) is set to OPT_IN, or is not set.

nbolton (approved):

PR #7768, opened 2024-10-20
ci: New macOS CI with Qt action

### LINKED ISSUE #7759: Fatal error on macOS due to libtomlplusplus.3.dylib
### Project

Deskflow

### Deskflow version number

1.17.0.86

### Build type

### Operating systems (OS)

- [ ] Windows
- [X] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [ ] I have reviewed the Wayland known issues and my issue is new
- [X] I am not using Wayland on Linux

### Signing on macOS

- [X] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### Continuous build

- [X] I have tried the latest continuous build and the issue persists

### OS versions/distros

- MacOS 15

### Deskflow configuration

_No response_

1. Start deskflow
378061420-45160b08-48ab-42f0-b441-4ee276f2f3ca

### Log output

_No response_

### Additional information

_No response_

sithlord48:
- New mac for mac os 
    -  Use Brew to get `cmake`, `openssl` and `googletest`
    -  Qt Install action to install Qt. 
    -  Allow cmake fetch and embed, `tomlplusplus` and `cli11` fix #7759

- New `env.CMAKE_CONFIGURE` holds our basic cmake command 
    - `cmake -S. -Bbuild -DCMAKE_BUILD_TYPE=Release -DCMAKE_WARNINGS_AS_ERRORS=ON`
    - Use this everywhere we can

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ env:
   GIT_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
   PACKAGE_PREFIX: "deskflow"
   PACKAGE_PATH: ./dist
-  CMAKE_ARGS: "-DCMAKE_COMPILE_WARNING_AS_ERROR=ON"
+  CMAKE_CONFIGURE: "cmake -Bbuild -DCMAKE_BUILD_TYPE=Release -DCMAKE_COMPILE_WARNING_AS_ERROR=ON"
 
 jobs:
   # Always run this job, even if not on PR, since other jobs need it.
@@ -162,10 +162,8 @@ jobs:
 
       - name: Configure
         run: |
-          cmake ${{ steps.vcpkg.outputs.vcpkg-cmake-config }} `
-            -B build -G Ninja `
-            -DCMAKE_BUILD_TYPE=Release `
-            ${{ env.CMAKE_ARGS }}
+          ${{env.CMAKE_CONFIGURE}} -G Ninja `
+            ${{ steps.vcpkg.outputs.vcpkg-cmake-config }}
 
       - name: Build
         run: cmake --build build -j8
@@ -197,10 +195,6 @@ jobs:
     runs-on: ${{ matrix.target.os }}
     timeout-minutes: ${{ matrix.target.timeout }}
 
-    defaults:
-      run:
-        shell: ${{ matrix.target.shell }}
-
     strategy:
       # Normally, we want to fail fast, but in this case we shouldn't since one target may
       # fail due to transient issues unrelated to the build.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -214,9 +214,9 @@ jobs:
             arch: arm64
             shell: "/usr/bin/arch -arch arm64e /bin/bash --noprofile --norc -eo pipefail {0}"
 
-          - name: ${{ vars.CI_MAC_INTEL_NAME || 'macos-13-x64' }}
```
nbolton, line 217: This makes it harder for downstream forks to change the runner name (e.g. if they want to use a different runner). Ok with that?

Doesn't affect Synergy because it uses different CI.

.github/workflows/ci.yml
```diff
@@ -232,17 +232,25 @@ jobs:
...
 
-      - name: Cache deps dir
-        uses: actions/cache@v4
-        with:
-          path: ./deps
-          key: ${{ runner.os }}-deps-${{ hashFiles('config.yaml') }}
-
       - name: Install dependencies
-        run: ./scripts/install_deps.py
+        run: brew install cmake openssl googletest --quiet
```
nbolton, line 236: Why not use the `Brewfile`? That's the standard way.

This way is like using vcpkg without `vcpkg.json`

.github/workflows/ci.yml
```diff
@@ -232,17 +232,25 @@ jobs:
...
+         cache: true
+         cache-key-prefix: ${{matrix.target.os}}-${{env.qt-version}}
+        env:
+         qt-version: 6.7.2
+         qt-install-dir: "/Users/runner"
 
       - name: Configure
-        run: cmake -B build ${{ env.CMAKE_ARGS }}
+        run: |
+          cmake -B build -DCMAKE_BUILD_TYPE=Release `
```
nbolton, line 251: Isn't it release build by default? If not, how about moving `-DCMAKE_BUILD_TYPE=Release` to `CMAKE_ARGS`?

nbolton (approved):
Approved but I'm not 100% convinced dropping `Brewfile` is the right direction.

.github/workflows/ci.yml
sithlord48: downstream will (and should)  have their own ci setup

.github/workflows/ci.yml
sithlord48: 1. Its has extra packages only suitable for local builds
2. I want to be rid of such files.

.github/workflows/ci.yml
sithlord48: No Release is not default. 
I could as long as no other jobs that use those arg need a RelWithDebInfo or Debug build 
i.e none of the "scanning jobs"

sithlord48:
> Approved but I'm not 100% convinced dropping `Brewfile` is the right direction.

I have not removed the BREWFILE (yet). 
I will when I have updated the build wiki .
Since you can have several combo of depends simply saying use this BREWFILE is not gonna work for every one ..

.github/workflows/ci.yml
sithlord48: If the vcpkg file (and config) were not required to make it "automatic" on windows i would also remove those.

nbolton (approved):

PR #7770, opened 2024-10-20
build: remove meson

sithlord48:
fix 7679

Removes the ability for static portals and libei / libportal from meson. 
This will affect the builds for ubuntu 24 where we did inject libportal


sithlord48:
superseeded by #7771

PR #7771, opened 2024-10-20
Unify Jobs into one matrix

sithlord48:
Blocked by: #7781

 - Unify all os builders under one step (other then freebsd) 
 - New install-dependencies action to handle dependencies
 - remove the old matrix and put it into the main ci build


nbolton:
> Error: Process completed with exit code 100.

Looks like Fedora 40 is failing due to `dnf install` returning `100` which AFAIK means "Updates available." (kinda silly)

nbolton:
Heads up, the `Dockerfile` for each distro used by `.github/workflows/build-containers.yml` also references `./scripts/install_deps.py` which is how the containers have the correct packages pre-loaded (and are at most a week out of date).

nbolton:
> sithlord48 requested a review from nbolton 3 hours ago

I'm guessing you intended to switch this to open.

.github/actions/install-dependencies/action.yml
```diff
@@ -0,0 +1,102 @@
...
+  vcpkg-cmake-config:
+    description: "windows vcpkg output for cmaket"
+    value: ${{ steps.vcpkg.outputs.vcpkg-cmake-config }}
+
+runs:
+  using: "composite"
+
+  steps:
+
+    # This mirrors instructions at https://apt.kitware.com
```
nbolton, line 26: Is this comment still relevant?

.github/actions/install-dependencies/action.yml
```diff
@@ -0,0 +1,102 @@
...
+        cache-key-prefix: ${{matrix.target.os}}-${{env.qt-version}}
+
+    # Install Ninja with an action instead of using Chocolatey, as it's more
+    # reliable and faster. The Ninja install action is pretty good as it
+    # downloads directly from the `ninja-build` GitHub project releases.
+    - name: Install Ninja
+      if: ${{ runner.os == 'Windows' }}
+      uses: seanmiddleditch/gha-setup-ninja@master
+
+    - name: Dependencies (vcpkg)
```
nbolton, line 87:
```suggestion
    - name: Build and cache vcpkg
```

.github/workflows/build-containers.yml
nbolton, line 1: We could rename our Debian runner to `debian-trixie-amd64` for accuracy but I'm happy either way.

.github/workflows/ci.yml
```diff
@@ -122,8 +119,56 @@ jobs:
...
+            like: "debian"
+            timeout: 20
+            extra-cmake-config: "-G Ninja -DCMAKE_INSTALL_PREFIX"
+
+          - name: "fedora-40-arm64"
+            runs-on: ubuntu-24.04-4-core-arm64
+            container: deskflow/deskflow:fedora-40-arm64
+            like: "fedora"
+            timeout: 20
+            extra-cmake-config: "-G Ninja -DCMAKE_INSTALL_PREFIX"
```
nbolton, line 150: I thought you wanted to ditch the arm runner.

.github/workflows/ci.yml
```diff
@@ -132,33 +177,44 @@ jobs:
...
-      # reliable and faster. The Ninja install action is pretty good as it
-      # downloads directly from the `ninja-build` GitHub project releases.
-      - name: Install Ninja
-        uses: seanmiddleditch/gha-setup-ninja@master
-
-      - name: Dependencies (vcpkg)
-        id: vcpkg
-        uses: johnwason/vcpkg-action@v6
+      - name: Install dependencies
+        id: deploy
```
nbolton, line 184:
```suggestion
        id: dependencies
```

.github/workflows/ci.yml
```diff
@@ -255,94 +243,9 @@ jobs:
...
-          job: linux-${{ matrix.distro.name }}
-
-      - name: Upload
-        uses: actions/upload-artifact@v4
-        with:
-          name: package-${{ env.PACKAGE_PREFIX }}-${{ matrix.distro.name }}
-          path: ${{github.workspace}}/build/deskflow[-_]*.*
+          path: |
+            ${{github.workspace}}/build/deskflow[-_]*.*
+            ${{github.workspace}}/dist/deskflow[-_]*.*
```
nbolton, line 248: I'm guessing the `dist` dir is eventually going away?

.github/workflows/ci.yml
```diff
@@ -379,6 +282,7 @@ jobs:
 
   release:
     needs: ci-passed
+    if: (github.ref == 'refs/heads/master')
```
nbolton, line 285: Oh, hang on a sec... our un-landed PR builds were being sent to release? Oops.

.github/workflows/sonarcloud-analysis.yml
```diff
@@ -25,18 +25,15 @@ jobs:
...
-      - name: Add Kitware repo
-        uses: ./.github/actions/add-kitware-repo
+      - name: Install dependencies
+        uses: ./.github/actions/install-dependencies
         with:
-          distro: jammy
+          like: "debian"
 
-      - name: Install dependencies
+      - run: apt install curl unzip gcovr  -y
```
nbolton, line 33:
```suggestion
      - name: Install dependencies (extra)
        run: apt install curl unzip gcovr  -y
```

.github/workflows/valgrind-analysis.yml
```diff
@@ -14,15 +14,12 @@ jobs:
...
+        uses: ./.github/actions/install-dependencies
         with:
-          distro: jammy
+          like: "debian"
 
-      - name: Install dependencies
-        run: |
-          ./scripts/install_deps.py
-          apt install valgrind cmake -y
+      - run: apt install valgrind cmake -y
```
nbolton, line 22:
```suggestion
      - name: Install dependencies (extra)
        run: apt install valgrind cmake -y
```

meson.build
nbolton, line 1: Makes me happy.

src/gui/src/MainWindow.cpp
```diff
@@ -543,7 +543,7 @@ void MainWindow::open()
 void MainWindow::onCoreProcessStarting()
 {
 
-#if defined(WINAPI_XWINDOWS) or defined(WINAPI_LIBEI)
+#ifdef Q_OS_LINUX
```
nbolton, line 546: Good change... this is defined on BSD-derived Unix-like too, right? (e.g. FreeBSD)

src/lib/platform/EiScreen.cpp
```diff
@@ -30,10 +30,8 @@
 #include "platform/EiEventQueueBuffer.h"
 #include "platform/EiKeyState.h"
```
nbolton, line 32:
```suggestion
```

nbolton (approved):
Some optional suggestions you may want to implement, but your call. Spotted some nits, but will fix in new PR.

src/gui/src/MainWindow.cpp
sithlord48: Afaik bsd does not have wayland so this check is really still linux only

sithlord48: We do not need this check if the var we check is empty we will say we are not on wayland .

src/lib/platform/EiScreen.cpp
sithlord48: part of #7781

src/gui/src/MainWindow.cpp
sithlord48: On #7781

.github/workflows/ci.yml
sithlord48: Well the other method of using qemu will take about 40 minutes to build each .
I figured we can lookin to this in another pr.

For now i would like to leave them as is

.github/workflows/build-containers.yml
sithlord48: Id rather use the numeric release because to people that don't follow debian we all know 13 > 12 but not all of us know if trixie , bookwork , or buster is the current codename.

.github/workflows/ci.yml
sithlord48: well when cmake makes the packages they are produced in the root of the build dir so that is where we are gonna look for them. 

The `dist` is temp until i have the mac and windows packages being produced property via cpack

.github/workflows/ci.yml
sithlord48: No they are not but the step starting and spending time to dl the artifacts that we dont release. This just makes it so that does not happen

.github/workflows/build-containers.yml
nbolton: Ok sounds good, that's why I used the number in the first place so it sounds like we both have the same point of view on this one. Debian wants us to use "Trixie" because they're not officially calling it 13 yet, but we all know that's what's going to happen.

.github/workflows/ci.yml
nbolton: Cool... to be honest GitHub charges next to nothing for their ARM runners, so it's no big deal to stick with paid runners. Problem is in this PR, that CI will now fail for downstream forks (contributors) when they enable workflows in their fork. IIRC, the workflow will just hang there forever waiting for the ARM runner to come online (which it never will unless they have a paid runner with exactly the same name).

We may want to drop ARM for now in this PR (since we're no longer using the 'clever' JSON filter solution) and reintroduce it in a different PR (to fix #7785) as a new workflow or job that only runs if you have paid runners.

sithlord48: Yes ill remove it for now and we can add it back later.

.github/workflows/ci.yml
```diff
@@ -105,130 +105,108 @@ jobs:
...
+      - name: Setup Python
+        if: ${{runner.os != 'Linux' }}
+        run: python ./scripts/setup_venv.py
 
       - name: Configure
-        run: ${{env.CMAKE_CONFIGURE}} ${{matrix.target.extra-cmake-config}}
+        run: |
+          ${{env.CMAKE_CONFIGURE}} `
+            ${{ matrix.target.extra-cmake-config }} `
+            ${{ steps.install-dependencies.outputs.vcpkg-cmake-config }}
```
nbolton, line 190: Oops, didn't spot this on my last pass.

I don't think the PowerShell backtick linebreak approach works on `sh`/`bash`:
https://github.com/deskflow/deskflow/actions/runs/11462609763/job/31894430876?pr=7771
```
Run cmake -Bbuild -DCMAKE_BUILD_TYPE=Release -DCMAKE_COMPILE_WARNING_AS_ERROR=ON `
  cmake -Bbuild -DCMAKE_BUILD_TYPE=Release -DCMAKE_COMPILE_WARNING_AS_ERROR=ON `
    -G Ninja -DCMAKE_INSTALL_PREFIX `
    
  shell: sh -e {0}
  env:
    GIT_SHA: b9a86a074f903af10db460b44cce68dbe4[2](https://github.com/deskflow/deskflow/actions/runs/11462609763/job/31894430876?pr=7771#step:7:2)c0288
    PACKAGE_PREFIX: deskflow
    PACKAGE_PATH: ./dist
    CMAKE_CONFIGURE: cmake -Bbuild -DCMAKE_BUILD_TYPE=Release -DCMAKE_COMPILE_WARNING_AS_ERROR=ON
/__w/_temp/901[3](https://github.com/deskflow/deskflow/actions/runs/11462609763/job/31894430876?pr=7771#step:7:3)d1f1-1ed3-43c0-8e89-6b897e72bae9.sh: 2: -G: not found
```

Maybe the best approach is to split this into two steps, "Configure (Unix)" and "Configure (Windows)":
```
     # Windows with backtick (`) linebreaks
      - name: Configure (Windows)
        if: ${{ runner.os == 'Windows' }}
        run: |
          ${{env.CMAKE_CONFIGURE}} `
            ${{ matrix.target.extra-cmake-config }} `
            ${{ steps.install-dependencies.outputs.vcpkg-cmake-config }}

     # Non-Windows with backslack (\) linebreaks
      - name: Configure (Unix)
        if: ${{ runner.os != 'Windows' }}
        run: |
          ${{env.CMAKE_CONFIGURE}} \
            ${{ matrix.target.extra-cmake-config }}
```

Or, if Windows will let you run CMake outside of PowerShell (not sure if possible), maybe we could have a single step and use bash backslashes:
```
      - name: Configure
        shell: bash
        run: |
          ${{env.CMAKE_CONFIGURE}} \
            ${{ matrix.target.extra-cmake-config }} \
            ${{ steps.install-dependencies.outputs.vcpkg-cmake-config }}
```

And if that's all too much trouble, just remove the backticks and have it on one line:
```
      - name: Configure
        run: |
          ${{env.CMAKE_CONFIGURE}} ${{ matrix.target.extra-cmake-config }} ${{ steps.install-dependencies.outputs.vcpkg-cmake-config }}
```

IDK though, feels a bit dupey. Fucking PowerShell, eh? Why couldn't they adopt backslashes... oh yeah, it's a dir separator! :roll_eyes:

.github/workflows/ci.yml
```diff
@@ -105,130 +105,108 @@ jobs:
...
+            runs-on: "windows-2022"
+            timeout: 120 #Windows can take while if it has to build vcpkg cache
+            extra-cmake-config: "-G Ninja"
+
           - name: "macos-14-arm64"
+            runs-on: "macos-14"
             timeout: 10
-            os: "macos-14"
             arch: arm64
             extra-cmake-config: "-DCMAKE_OSX_ARCHITECTURES=\"arm64\" -DMACOSX_DEPLOYMENT_TARGET=12"
```
nbolton, line 130: I wonder if it'd make things a bit readable if we called it `cmake-args` instead of `extra-cmake-config`?

nbolton (changes requested):
Just spotted an error in CI.
```
d1f1-1ed3-43c0-8e89-6b897e72bae9.sh: 2: -G: not found
```

nbolton (approved):

PR #7772, opened 2024-10-21
Clazy nits

sithlord48:
- remove `foreach` macro for c++11 style ranged for
- ensure ranged loops do not detach
- Use Q_EMIT in place of emit 
- Do not #include full QModules 
- use multi arg
- add context object for lambdas

### DIFF
--- a/src/gui/src/MainWindow.cpp
+++ b/src/gui/src/MainWindow.cpp
@@ -48,12 +48,11 @@
 #include <QMenuBar>
 #include <QMessageBox>
 #include <QNetworkAccessManager>
+#include <QNetworkInterface>
 #include <QPushButton>
 #include <QRegularExpression>
 #include <QScrollBar>
-#include <QtCore>
-#include <QtGui>
-#include <QtNetwork>
+
 #include <memory>
 
 #if defined(Q_OS_MAC)
--- a/src/gui/src/MainWindow.cpp
+++ b/src/gui/src/MainWindow.cpp
@@ -1012,8 +1012,7 @@ void MainWindow::updateScreenName()
 {
   ui->m_pLabelComputerName->setText(QString("This computer's name: %1 "
                                             R"((<a href="#" style="color: %2">change</a>))")
-                                        .arg(m_AppConfig.screenName())
-                                        .arg(kColorSecondary));
+                                        .arg(m_AppConfig.screenName(), kColorSecondary));
   m_ServerConfig.updateServerName();
 }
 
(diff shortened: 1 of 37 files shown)


src/lib/gui/VersionChecker.cpp
```diff
@@ -86,11 +86,12 @@ int VersionChecker::getStageVersion(QString stage)
   if (stage.isEmpty() || stage == stableName) {
     return stableValue;
   } else if (stage.toLower().startsWith(rcName)) {
-    QRegularExpression re("\\d*", QRegularExpression::CaseInsensitiveOption);
-    if (re.match(stage).hasMatch()) {
+    static QRegularExpression re("\\d*", QRegularExpression::CaseInsensitiveOption);
+    auto match = re.match(stage);
+    if (match.hasMatch()) {
```
nbolton, line 91: I think we should delete `VersionChecker::getStageVersion`. It's a stupid function.

I wrote a unit test against `master` to see if the logic had changed here and I believe that this code is actually broken in `master`.

Come to think of it, the original code is kind of dumb anyway because: 

1. Why would we ever need to compare `1.2.3-beta` and `1.2.3-stable`...? That makes no sense to me.
2. If we were to do use `rc1`, `rc2`, etc (which I don't think we will) then why would we not change the patch version?

```
TEST_F(VersionCheckerTests, compareVersions_majorMinorPatch_isValid)
{
  EXPECT_EQ(compareVersions("0.0.0-rc1", "0.0.0-rc2"), -1);
  EXPECT_EQ(compareVersions("0.0.0-rc2", "0.0.0-rc1"), -1);
  EXPECT_EQ(compareVersions("0.0.0-rc1", "0.0.0-rc1"), 0);
}
```

src/gui/src/MainWindow.cpp
```diff
@@ -199,7 +198,7 @@ void MainWindow::connectSlots()
   connect(&m_CoreProcess, &CoreProcess::error, this, &MainWindow::onCoreProcessError);
 
   connect(
-      &m_CoreProcess, &CoreProcess::logLine, //
+      &m_CoreProcess, &CoreProcess::logLine, this, //
```
nbolton, line 201: Aha, is this to automatically disconnect when the object is destroyed?

nbolton (changes requested):

src/gui/src/MainWindow.cpp
sithlord48: well if auto disconnect would happen it will not with the context of the receiver.

src/lib/gui/VersionChecker.cpp
nbolton: Cool, since it's dead code I suppose it doesn't matter if it is changed.

nbolton (approved):

PR #7779, opened 2024-10-21
ci: use fancy checkout action

sithlord48:


nbolton (approved):
The force is strong with this one

PR #7781, opened 2024-10-21
build: Require libei and libportal on linux

### LINKED ISSUE #7679: Migrate Meson dependency fetching to CMake (FetchContent)

As discussed (#7546) using Meson to get deps was a mistake.

Use CMake `FetchContent` instead and remove our `Meson` config entirely.

sithlord48:
- Require libei 1.3 from system on linux
 - Require libportal 0.8 be on the system for linux 
 - Remove runners without these libs
    - debian 12
    - ubuntu 22
    - ubuntu 24
    - fedora 39
    - manjaro (we only need one arch like to test) 

 - Removing remaining meson parts fix #7679


nbolton (dismissed):
See comments in: #7771

nbolton (approved):

Issue #7785, opened 2024-10-22
Re-implement Linux ARM64 runners with QEMU


nbolton:
### Feature Request

GitHub ARM64 runners are a bit of a problem for the open source community. Because they're a paid runner, we had to make them optional for downstream contributor forks (the reason for implementing [`ci-linux.json`](https://github.com/deskflow/deskflow/blob/2136af68b542fa1ec185a56c03535f03bb587170/.github/workflows/ci-linux.json)) but as this was creating a maintenance burden on our dev community, we decided to drop that approach; it's a clever solution to a problem we shouldn't be solving.

As a community, we decided that we will be building on ARM64 Linux in a separate QEMU workflow (until GitHub makes ARM64 Linux runners free for open source).

nbolton:
@sithlord48 Please remind me what that action was.

sithlord48:
[run-on-arch](https://github.com/uraimo/run-on-arch-action)

PR #7796, opened 2024-10-23
Add run on arch

sithlord48:
resolve #7767, #7785 
 - Add back arch runners using [run-on-arch](https://github.com/uraimo/run-on-arch-action) 
     -  `fedora_latest` (fc40)

Adds in #7909 
 -  Add aarch64 (arm64) build for flatpak
 -  Split this as im not sure we want to build this here as it makes our minimum build time around 25 minutes.


sithlord48:
Very slow .. not sure why I mean very extra slow

sithlord48:
Superseeded by #8101

PR #7801, opened 2024-10-23
docs: Add CMake presets example

nbolton:
Example CMake presets


sithlord48:
I thought we had decided this belonged on the wiki  

In the the [Window's getting started](https://github.com/deskflow/deskflow/wiki/Building#windows) we add a small section after the vcpkg setup with an example for setting up a `CmakeUserPreset.json`  that will tell cmake where vcpkg's toolchain is.

nbolton:
Ah yes.

PR #7813, opened 2024-10-25
docs: Make unstable download clearer

nbolton:
Some users may be downloading the continuous build thinking it's stable. e.g. Continuous doesn't work with Wayland on Ubuntu right now.

Maybe we should warn them that it's unstable, otherwise perhaps it's unfair on them.

Only an idea, happy for this to not be landed. This change it makes the button a bit uglier, and also makes the button a bit bigger which may backfire due to prominence effect, aka big red button effect.

[Preview](https://github.com/deskflow/deskflow/blob/readme-unstable/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,8 @@ and work seamlessly between them.
 It's like a software KVM (but without the video).
 TLS encryption is enabled by default. Wayland is supported. Clipboard sharing is supported.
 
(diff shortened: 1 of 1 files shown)


sithlord48 (changes requested):
The Continuous is not unstable the problem is ubuntu's lack of a proper portal we can use

sithlord48:
Why is this even installing for ppl on ubuntu it should fail install saying i need libportal >0.8. 

I think that is the real problem here.

nbolton:
> Why is this even installing for ppl on ubuntu it should fail install saying i need libportal >0.8.
> 
> I think that is the real problem here.

As a workaround, in 1.17.0 we static linked to libportal, but this hack was later removed.

I'll close this PR.

PR #7821, opened 2024-10-28
docs: readme add mac os xattr note

sithlord48:
add a tip for mac users to run xattr -c on the app post download
[Preview](https://github.com/deskflow/deskflow/blob/readme-update-2/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,10 @@ TLS encryption is enabled by default. Wayland is supported. Clipboard sharing is
 
 To use Deskflow you can use one of our [packages](https://github.com/deskflow/deskflow/releases), install `deskflow` (if available in your package repository), or [build it](#build-quick-start) yourself from source.
 
+> [!TIP]
+> macOS users may need to run `xattr -c /Applications/Deskflow.app` after copying the app to the `Applications` dir.
+> You will also have to allow accessibility access to both the `Deskflow` app and the `deskflow` process.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,6 +19,9 @@ TLS encryption is enabled by default. Wayland is supported. Clipboard sharing is
 
 To use Deskflow you can use one of our [packages](https://github.com/deskflow/deskflow/releases), install `deskflow` (if available in your package repository), or [build it](#build-quick-start) yourself from source.
 
+> [!TIP]
+> Mac os users may need to run `xattr -c /Applications/Deskflow.app` After copying the app too applicatons, You will also have to allow accessibility access to both the deskflow gui and the server / client process.
```
nbolton, line 23:
```suggestion
> macOS users may need to run `xattr -c /Applications/Deskflow.app` after copying the app to `Applications`.
> You will also have to allow accessibility access to both the `Deskflow` app and the `deskflow` process.
```

nbolton (changes requested):
I found that that adding permission to the `deskflow` process (the GUI process) is sufficient. The only entry that shows by default is the `Deskflow` app bundle entry (capital 'D') but you have to manually add the `deskflow` GUI process (lower 'D).

nbolton (approved):

DominikDeak:
cheers for adding the tip

PR #7834, opened 2024-10-30
doc: correct target for documents

sithlord48:
typo doc target is `user-docs`
- Remove `DOCS` option for `BUILD_DOCS` 
Breaks builds with `BUILD_DOCS` on or doxygen and it not specified .


nbolton (approved):

nbolton:
Nice

image

PR #7838, opened 2024-10-31
docs: Improve error messages around Doxygen

nbolton:
Fixes minor nits around Doxygen messages.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -2,9 +2,10 @@
 # SPDX-License-Identifier: MIT
 
 find_package(Doxygen QUIET)
-option(BUILD_DOCS "Build and Install Documents" ${DOXYGEN_FOUND})
+option(BUILD_DOCS "Build and install documents" ${DOXYGEN_FOUND})
 
-if(BUILD_DOCS)
+if (BUILD_DOCS AND DOXYGEN_FOUND)
+
   set(DOXYGEN_EXTRACT_ALL YES)
   set(DOXYGEN_STRIP_FROM_PATH ${CMAKE_SOURCE_DIR})
   set(DOXYGEN_USE_MDFILE_AS_MAINPAGE mainpage.md)
@@ -12,7 +13,7 @@ if(BUILD_DOCS)
 
   # Files used to make our documents
   # User facing documents will not include doxy comments in source code
-  doxygen_add_docs(user-docs ${CMAKE_SOURCE_DIR}/doc COMMENT "Generating User Documentation" ALL)
+  doxygen_add_docs(user-docs ${CMAKE_SOURCE_DIR}/doc COMMENT "Generating user documentation" ALL)
 
   # HACK Only these will show in your IDE
   target_sources(user-docs PRIVATE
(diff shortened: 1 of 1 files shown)


doc/CMakeLists.txt
```diff
@@ -2,17 +2,21 @@
 # SPDX-License-Identifier: MIT
 
 find_package(Doxygen QUIET)
-option(BUILD_DOCS "Build and Install Documents" ${DOXYGEN_FOUND})
+option(BUILD_DOCS "Build and install documents" ${DOXYGEN_FOUND})
 
 if(BUILD_DOCS)
```
sithlord48, line 7: if (BUILD_DOCS AND DOXYGEN_FOUND)

doc/CMakeLists.txt
```diff
@@ -24,6 +28,4 @@ if(BUILD_DOCS)
     DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/html"
     DESTINATION ${CMAKE_INSTALL_DOCDIR}
     COMPONENT deskflow_docs)
-elseif(NOT DOXYGEN_FOUND)
-  message(STATUS:"Unable to build documentation Doxygen not found")
```
sithlord48, line 28: Leave this message 
Change it to 
` message(STATUS "Unable to build documentation Doxygen not found") `

sithlord48 (changes requested):

doc/CMakeLists.txt
nbolton: How about a `WARNING`? I think `STATUS` is too subtle. If someone forces `BUILD_DOCS` on then the message should jump out at them.

sithlord48: Status is fine as this will be show every time you build without doxygen present its not an error or warning its literately just reporting the status of the check for us

doc/CMakeLists.txt
```diff
@@ -24,6 +25,13 @@ if(BUILD_DOCS)
     DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/html"
     DESTINATION ${CMAKE_INSTALL_DOCDIR}
     COMPONENT deskflow_docs)
+
+elseif(BUILD_DOCS AND NOT DOXYGEN_FOUND)
```
sithlord48, line 29: Why ?? just use else 
please remove these empty lines we don't need them to read the else

sithlord48 (changes requested):

doc/CMakeLists.txt
```diff
@@ -24,6 +25,13 @@ if(BUILD_DOCS)
...
+
+  message(WARNING "Doxygen not found, cannot build docs")
```
sithlord48, line 31: Why do we need two messages to say the same thing ?

doc/CMakeLists.txt
nbolton: In the event that someone doesn't read the docs and adds `-DBUILD_DOCS=ON`... I don't think we should punish people for missing something in the docs. Without this, it'd be confusing. You could see a misleading error message.

There are 3 scenarios:
1. They have Doxygen installed
2. They don't have Doxygen installed
3. They think they have Doxygen installed but don't or have a damaged Doxygen install and set `-DBUILD_DOCS=ON`

sithlord48: IMHO its a waste of runtime .. `BUILD_DOCS=ON` is on by default if doxygen is found. 
How about 
message(STATUS "Documents will not be built Doxygen not found or BUILD_DOCS is OFF")

doc/CMakeLists.txt
nbolton: Let's ignore the 3rd scenario. I'm not sure it's worth discussing further.

doc/CMakeLists.txt
nbolton: Let's ignore the 3rd scenario. I'm not sure it's worth discussing further.

sithlord48 (approved):

PR #7845, opened 2024-11-01
Better package names

sithlord48:
This does not change any package's internal version they will still be  `M.m.P.T` internally

This does change how the packages are named: 
 - Use `continuous` for the version portion of package names when not a tagged release
 - Use the Tag for others 

Side effects include new package names 
 - Arch packages no longer have the package revision in our deployment
 - Continuous  = `deskflow-continuous_os.ext`
 - Release = `deskflow-1.17.0_os.ext` (When tagged v1.17.0)

Why ? 
  This will create a permalink for continuous packages making them just easier
  This makes our release tag no longer require having the .0 for tweak in the filename for the package

We are not going to change the names for 1.17.0 release unless everyone is onboard with the change 

Continuous build title changed to Continuous build: real version ex: `Continuous Build: 1.17.0.170` 
this can be used to see if  version is newer then your current continuous version.

## Current continuous package names
i need this for example to make a deskflow@head cask in https://github.com/deskflow/homebrew-tap that can auto update alot easier

- Arch: `deskflow-continuous-arch-x86_64.pkg.tar.zst` 
- Debian: `deskflow-continuous-debian-trixie-x86_64.deb`
- Fedora 40: `deskflow-continuous-fedora-40-x86_64.rpm`
- Fedora 41: `deskflow-continuous-fedora-41-x86_64.rpm`
- mac OS (arm): `deskflow-continuous-macos-arm64.dmg`
- mac OS (64 bit): `deskflow-continuous-macos-x86_64.dmg`
- Suse: `deskflow-continuous-opensuse-tumbleweed-x86_64.rpm`
- Windows (x86_64) `deskflow-continuous-win-x64.msi`
- Windows (arm) `deskflow-continuous-win-arm64.msi`

## Current release package names
example using a v1.17.1 as the tagged version

- Arch: `deskflow-1.17.1-arch-x86_64.pkg.tar.zst` 
- Debian: `deskflow-1.17.1-debian-trixie-x86_64.deb`
- Fedora 40: `deskflow-1.17.1-fedora-40-x86_64.rpm`
- Fedora 41: `deskflow-1.17.1-fedora-41-x86_64.rpm`
- mac OS (arm): `deskflow-1.17.1-macos-arm64.dmg`
- mac OS (64 bit): `deskflow-1.17.1-macos-x86_64.dmg`
- Suse: `deskflow-1.17.1-opensuse-tumbleweed-x86_64.rpm`
- Windows (x86_64) `deskflow-1.17.1-win-x64.msi`
- Windows (arm) `deskflow-1.17.1-win-arm64.msi`

## THERE WILL BE NO RENAMING AFTER THIS POINT LETS GET IT RIGHT

Questions for Windows Packagers
1. `windows` or `win` ? 
1. `win$arch` or `win_$arch`
1. How do you all prefer the arm one be labeled ? 
1. MSI being windows only do we need to include `windows` or just the `arch` ?  (i.e `deskflow-v_x64.msi`)

Questions for Mac users
1. `macOS` or `mac` ? 
1.  `arm64` or `AppleSilicon` ? 
1. `x64` or `Intel` 
1. since the dmg is mac on should we even put mac in the file name ?  (i.e `deskflow-v-_arm64.dmg`

Attn: @nicola02nb @marlop352 @teknowledgist @cdwmhcc @SelfRef


marlop352:
- Not sure the continuous name is good since there would be no way to know if there was an update by the tag or filename(thinking about scripting/automations), there may be ways around that but since packagers shouldn't be using those, as far as I understood, then if deskflow itself can notify the user that there is an update it should be fine.

- The extra ".0" should be irrelevant if I understood you versioning scheme correctly as all tagged versions would end in ".0".

- **Do not** rename or mess around with the artifacts already released for 1.17.0.

Regarding the Windows questions:

1. windows or win ?
    - Either should work fine.
2. win$arch or win_$arch
    - Either should work fine but it's easier to read with the "_".
3. How do you all prefer the arm one be labeled ?
    - arm64
4. MSI being windows only do we need to include windows or just the arch ? (i.e deskflow-v_x64.msi)
    - From a user's(specially one that is less technical) point of view it's easier to find the right file if it has win or windows in the file name

Not a mac user so not gonna give an opinion about the mac questions.

cdwmhcc:
Will 1.19.0 and 1.19.0.3 coexist in the future? If they exist at the same time, then 1.19.0.3 > 1.19.0 or 1.19.0.3 < 1.19.0?

Several common combinations of macOS architectures:

- x86_64/aarch64
- x86_64/arm64
- x64/arm64
- intel/apple silicon

Recommended: x86_64/arm64

A complete example:
deskflow-1.17.0.0_macos_x86_64.dmg
deskflow-1.17.0.0_macos_arm64.dmg

sithlord48:
> Will 1.19.0 and 1.19.0.3 coexist in the future? If they exist at the same time, then 1.19.0.3 > 1.19.0 or 1.19.0.3 < 1.19.0?
> 

They can. Within deskflow and within the packages the full 4 part version is always used.  
If your package manager / format does a version check for your example it would conclude that  `1.19.0.0` < `1.19.0.3 `.


> * Not sure the continuous name is good since there would be no way to know if there was an update by the tag or filename(thinking about scripting/automations), there may be ways around that but since packagers shouldn't be using those, as far as I understood, then if deskflow itself can notify the user that there is an update it should be fine.

The packages use the real 4 part version internally. The continuous name is only used so that `https://github.com/deskflow/deskflow/releases/download/continuous/deskflow-continuous_win64.msi` can be used as a perma link  to the 
last build from from master branch. 

All of the packages we make here are version aware so they should do the correct thing based on the internal versions 

You are correct in that the filename alone will not tell you if its newer then the last continuous you had downloaded. A simple solutions for such script could be to check the commit SHA, or release time vs the version last installed.  The oldest continuous release is our current one. A new one is always an upgrade. 


> * The extra ".0" should be irrelevant if I understood you versioning scheme correctly as all tagged versions would end in ".0".

 Future tagged versions will not have the Tweak Value of `0` in the name any longer  for example v 1.18.0  will have a name of `deskflow-1.18.0_win64.msi` (added example to the pr description)

.github/actions/get-version/action.yml
```diff
@@ -6,5 +6,10 @@ runs:
 
   steps:
     - run: |
-        echo "DESKFLOW_VERSION=$(git describe --long | sed 's/-/./g;s/v//g'  | grep -o '^[0-9]*.[0-9]*.[0-9]*.[0-9]*')">> $GITHUB_ENV
+        echo "DESKFLOW_SEM_VERSION=$(git describe --long | sed 's/-/./g;s/v//g'  | grep -o '^[0-9]*.[0-9]*.[0-9]*.[0-9]*')">> $GITHUB_ENV
```
nbolton, line 9: It seems that `DESKFLOW_SEM_VERSION` could contain a 4-digit version number, which is not SemVer (which can only have 3 dot-delimited digits followed by `-` and/or `+`). I think this may confuse people.

nbolton:
> A complete example:
> deskflow-1.17.0.0_macos_x86_64.dmg
> deskflow-1.17.0.0_macos_arm64.dmg

My preference is to use `kebab-case` (i.e. hyphen for delimiter) rather than mixing `-` and `_` for delimiters. That way, we can reserve underscore for use in things like `x86_64`.

Given that @sithlord48 is set on not using the 4th digit for stable releases, I'd be happy with:

- `deskflow-1.18.0-macos-x86_64.dmg`
- `deskflow-1.18.0-macos-arm64.dmg`
- `deskflow-1.18.0-windows-x64.msi`

sithlord48:
you want - or _ I honestly don't care but i think name-version_OS just makes the OS part sick out a bit more.

.github/actions/get-version/action.yml
sithlord48: It is a SemVer as SemVer is M.m.P with `Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.` we use the label `.#`

nbolton: The key issue here is that while SemVer does allow labels for pre-release or build metadata, it explicitly specifies that these should be added after a hyphen or a plus sign, not as an extra numeric segment.

In SemVer:
- A pre-release version uses a hyphen: 1.0.0-alpha.1.
- Build metadata uses a plus sign: 1.0.0+build.1.

Simply adding an extra number like 1.2.3.4 isn’t a compliant way to indicate a pre-release or build label because it bypasses these standardized separators. This is more than just formatting; many SemVer-compatible tools require this structure to accurately interpret the version’s intent, so using 1.2.3.4 can disrupt tooling compatibility.

If you’re adding build-specific info, a proper SemVer example would be 1.2.3+4 or 1.2.3-4 rather than 1.2.3.4.

Therefore I recommend not using the word "SEM" or "SemVer" in this situation as it will confuse people.

Perhaps developers like @cdwmhcc @SelfRef @fuzzzerd @marlop352 can help out here with their interpretation of SemVer.

cdwmhcc:
> My preference is to use `kebab-case` (i.e. hyphen for delimiter) rather than mixing `-` and `_` for delimiters. That way, we can reserve underscore for use in things like `x86_64`.

My fault, it should be: `deskflow_1.17.0.0_macos_x86_64.dmg`

teknowledgist:
> Questions for Windows Packagers
> 1. windows or win ?
> 2. win$arch or win_$arch
> 3. How do you all prefer the arm one be labeled ?
> 4. MSI being windows only do we need to include windows or just the arch ? (i.e deskflow-v_x64.msi)

1. "win" is fine.
2. "win_$arch" please.
3. "arm64" is fine.
4. Please include "win" even though `.msi` is Windows-only.

If there will never be a 4-level version number, then please don't add the extra "0".  The download URL with two different kinds of versioning is driving me nuts (https://github.com/deskflow/deskflow/releases/download/v1.17.0/deskflow-1.17.0.0_win64.msi).

Please use underscores, `_`,  to separate data bits in the name (e.g. `deskflow_1.18.0_windows_x64.msi` in `$name_$version_$platform_$arch.$extension` format).  Hyphens, `-`, should be reserved for hyphenated names (e.g. `Deskflow-Pro`)

My $.02

sithlord48:
I will use underscore but i will not do 
`deskflow_1.14.0` I will do only `deskflow-1.14.0` as this how most linux distros want the package name.. 
also having deskflow-verison_os makes it really easy to read the version you look for the - and read until the next _ .  need the os look for _ and read until the next _

marlop352:
I would prefer `deskflow_type-version-os-arch`, the type being "pro" or something like that if it's ever needed.
For now it would be: `deskflow-version-os-arch`, having a `_` for the `x86_64` arch

sithlord48:
the architecture part will come from cmake that is x86_64 , AMD64 , ARM64 and afew others depending on the OS and how it reports it  
for windows we are using x64 for `x86_64` all others oses are using `x86_64` in the new scheme. If you windows ppl want the windows one to say x86_64 too that is fine with me.

teknowledgist:
In Windows world, `x64` means 64-bit only.  Use `x86_64` only if it will work on either 32-bit or 64-bit systems.  Ideally, that means it has one installer with both builds, because 32-bit (i.e. `x86` or `win32`) should also work on 64-bit Windows.

marlop352:
For me `x86_64` and `x64` are the same so I don't really care which of the two is used.

scripts/package.py
```diff
@@ -85,16 +85,13 @@ def get_filename_base(version, prefix):
...
-            os_part = "win64"
+            os_part = "x64"
     elif os == "mac":
-        if machine == "amd64" or machine == "x86_64":
-            os_part = "mac_x64"
-        else:
-            os_part = "mac_arm64"
-
+            os_part = "macos"
     # Underscore is used to delimit different parts of the filename (e.g. version, OS, etc).
```
marlop352, line 91: The comments should also be updated to reflect the new scheme.

marlop352 (changes requested):

scripts/package.py
sithlord48: good point i should remove this comment its kinda obvious what the next line does

marlop352 (approved):

nbolton:
> * Arch: `deskflow-continuous-arch-x86_64.pkg.tar.zst`
> * Debian: `deskflow-continuous-debian-trixie-x86_64.deb`
> * Fedora 40: `deskflow-continuous-fedora-40-x86_64.rpm`
> * Fedora 41: `deskflow-continuous-fedora-41-x86_64.rpm`
> * mac OS (arm): `deskflow-continuous-macos-arm64.dmg`
> * mac OS (64 bit): `deskflow-continuous-macos-x86_64.dmg`
> * Suse: `deskflow-continuous-opensuse-tumbleweed-x86_64.rpm`
> * Windows (x86_64) `deskflow-continuous-win-x64.msi`
> * Windows (arm) `deskflow-continuous-win-arm64.msi`

The PR description seems to have gone out of sync with the code.

.github/workflows/ci.yml
```diff
@@ -228,7 +231,7 @@ jobs:
             sudo -u build makepkg -s
             export OSNAME=$(cat /etc/os-release | grep ^ID= | sed 's/ID=//g')
             export ARCH=$(uname -m)
-            mv *.pkg.* $(ls *.pkg.* | sed "s/$ARCH/$OSNAME-$ARCH/g")
+            mv *.pkg.tar.zst deskflow-${{env.DESKFLOW_PACKAGE_VERSION}}_${OSNAME}_${ARCH}.pkg.tar.zst
```
nbolton, line 234: Are the underscores intentional?
https://github.com/deskflow/deskflow/pull/7845#issuecomment-2459866387

sithlord48: no let me update that

nbolton (approved):

PR #7848, opened 2024-11-02
fix: crash on --help arg

### LINKED ISSUE #7847: calling --help crashes server / client
### Discussed in https://github.com/deskflow/deskflow/discussions/7846

<div type='discussions-op-text'>

<sup>Originally posted by **nPHYN1T3** November  1, 2024</sup>
...
deskflow-client --help
[2024-11-01T14:37:44] FATAL: a runtime error occurred: This should be caught in your main function, see examples
....
</div>

sithlord48:
fixes #7847

### DIFF
--- a/src/lib/deskflow/App.cpp
+++ b/src/lib/deskflow/App.cpp
@@ -187,13 +187,18 @@ void App::initApp(int argc, const char **argv)
 {
 
   std::string configFilename;
-  CLI::App cliApp{kAppDescription, kAppName};
+  CLI::App cliApp{kAppDescription};
   cliApp.add_option("--config-toml", configFilename, "Use TOML configuration file");
 
   // Allow legacy args.
   cliApp.allow_extras();
 
-  cliApp.parse(argc, argv);
+  // Having the help argument crashes without try / catch around it
+  try {
+    cliApp.parse(argc, argv);
+  } catch (const CLI::Error &e) {
+    cliApp.exit(e);
+  }
 
   if (!configFilename.empty()) {
     Config config(configFilename, configSection());


src/lib/deskflow/App.cpp
```diff
@@ -193,7 +193,12 @@ void App::initApp(int argc, const char **argv)
   // Allow legacy args.
   cliApp.allow_extras();
 
-  cliApp.parse(argc, argv);
+  // Having the help argument crashes without try / catch around it
+  try {
+    cliApp.parse(argc, argv);
+  } catch (CLI::Error &e) {
+    printf("Error: %s", e.get_name().c_str());
+  }
```
nbolton, line 201: I wonder if we should exit here instead, and fix the underlying throw?

When you use `--help` now, the error shows in front of the help text. We could certainly add a couple of newlines, but that means the error is always going to be there.
```
 ./build/bin/deskflow-server --help
Error: CallForHelpUsage: deskflow-server
```

nbolton (changes requested):

sithlord48:
Its not throwing that is the odd part

src/lib/deskflow/App.cpp
nbolton, line 201:
```suggestion
  // Handle CLI11 error (CLI::Error) according to the docs.
  try {
    cliApp.parse(argc, argv);
  } catch (const CLI::Error &e) {
    cliApp.exit(e);
    return;
  }
```

nbolton (changes requested):
We should also change the args for `CLI:App` to drop the `kAppName` arg, since this is actually meant to be used for the binary name. Unfortunately, I couldn't make a GitHub suggestion here, as this wasn't code that changed in your PR:

```diff
- CLI::App cliApp{kAppDescription, kAppName};
+ CLI::App cliApp{kAppDescription};
```

src/lib/deskflow/App.cpp
nbolton: I added a suggestion further down.

PR #7869, opened 2024-11-07
Prepare 1.17.1

sithlord48:
A Change to the release workflow that we missed having not tagged yet. 
 - Strip `refs/tags/v` from the `DESKFLOW_PACKAGE_VERSION`  that the get-version action reports when run on a `v` tag
 - Release step can run if the ref is 'master ' or a tag   It can not be both

Note: release steps:
   - Bump the version in the main CMakeLists.txt
   - Make PR (like this)  

After landing.
   - Tag the commit with an annotated tag like `git tag -a v1.71.1 -m "v1.17.1"`
   - Push the tag  `git push origin v1.17.1`


nbolton (approved):
LGTM

Issue #7699, opened 2024-10-15
Configuration not found in CodeQL result


nbolton:
### Project

Deskflow

### Deskflow version number

1.17.0

### Build type

Local developer build (built it yourself)

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [ ] Linux (X11)
- [X] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other

### Wayland on Linux

- [ ] I have reviewed the Wayland known issues and my issue is new
- [X] I am not using Wayland on Linux

### Signing on macOS

- [ ] I have authorized the app to run on my Mac
- [X] I am not using macOS

### OS versions/distros

n/a

### Deskflow configuration

n/a

### What steps will reproduce the problem?

Wait for CodeQL to finish:
![image](https://github.com/user-attachments/assets/7a37af47-7a76-4093-8f7a-85fab753f515)

Expect: Passed
Actual: Configuration not found
![image](https://github.com/user-attachments/assets/b5f76d2a-f487-475a-8d46-74026a4f9165)

![image](https://github.com/user-attachments/assets/1e8aae07-62d8-477e-8fda-588840afceb3)
https://github.com/deskflow/deskflow/pull/7692/checks?check_run_id=31537606446

nbolton:
Last attempted solution: #7657

nbolton:
@sithlord48 I think to solve this we may need to delete/rename and then restore the `ci.yml` workflow because there is no way to remove "CI" from the "Setup types" under [Code Scanning](https://github.com/deskflow/deskflow/security/code-scanning/tools/CodeQL/status):
![image](https://github.com/user-attachments/assets/38323fd7-154d-4ab7-857e-8e75dd217032)

The CodeQL result is failing because GitHub thinks there is still a CodeQL workflow in master (called `ci.yml`). The settings are inadequate in that they a) do not let you manually remove a setup type and b) don't seem to detect when a `workflow_call` was removed. I think this stems from the fact that CodeQL was never intended to be used as a `workflow_call` in the first place.

sithlord48:
in #7880  I rename ci.yml => build.yml

nbolton:
@sithlord48 Did it work?

sithlord48:
well you still can't remove the old ci.yml from the security "tools" or at least i can't maybe you can ?

nbolton:
Sadly not. Time for a GitHub support ticket, I think.

[...]

PR #7880, opened 2024-11-10
ci: rename ci.yml to => build.yml

sithlord48:
rename the ci job to build.yml attempt to solve https://github.com/deskflow/deskflow/issues/7699


nbolton (commented):
Since we're doing this purely to rename the file, how about `ci-temp.yml` instead? If it fixes the problem, we can try renaming it back to `ci.yml` (by reverting the commit).

My concern is that it will permanently become `continuous-integration.yml` which looks a bit strange since CI is such a common acronym.

sithlord48:
I don't plan to rename it again

nbolton:
Ok, any ideas other than `continuous-integration.yml`? Maybe something prefixed with `ci-`?

It looks weird as `continuous-integration.yml`, and other CI-related workflows are no longer necessarily grouped with it (e.g. `ci-comment.yml`).

sithlord48:
Continuous integration == ci

nbolton:
> looks a bit strange since CI is such a common acronym.

> Continuous integration == ci

Yes. I know. It's a common acronym.

sithlord48:
So I figured if we were gonna rename it might as well just make it `continuous integration` since that is what is it and it looks nice on the action label

nbolton (approved):

PR #7892, opened 2024-11-13
Gui: Expose Check for updates option

### LINKED ISSUE #7890: Option to disable update check
### Feature Request

Currently it looks like deskflow tries to reach cloudfront to check for updates. When the deskflow is run on a computer which blocks this, it seems like the connection does not work. When it can go out to cloudfront, there is no problem. The Address i see in the firewall is dll7xpq8c5ev0.cloudfront.net
Please add an option to disable the update check.
Thanks!

sithlord48:
Fixes #7890  By exposing a check box in the application settings to enable or disable the check

 - Add the option to enable or disable update checks in the settings dialog
 - Fix the SettingsDialog UI files so it stops generating invalid Alignment values when the form is edited.
 - Use `AppConfig` object directly in the SettingsDialog
 - Fix an issue where the initial setting for auto update is not set.

### DIFF
--- a/src/gui/src/MainWindow.cpp
+++ b/src/gui/src/MainWindow.cpp
@@ -528,6 +528,7 @@ void MainWindow::open()
 
   if (!m_AppConfig.enableUpdateCheck().has_value()) {
     m_AppConfig.setEnableUpdateCheck(messages::showUpdateCheckOption(this));
+    m_ConfigScopes.save();
   }
 
   if (m_AppConfig.enableUpdateCheck().value()) {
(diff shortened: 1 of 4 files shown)


nbolton:
Might have found a bug. 
1. Clear settings (Help -> Clear settings)
2. Click 'No' for update check pop
3. Restart GUI

Expected: My update check choice is remembered
Actual: The check update pop shows again

It seems like the setting from the pop isn't getting saved; `enableUpdateCheck` does not exist in `~/.config/Deskflow/Deskflow.conf`

Workaround: Go to **Edit** -> **Preferences**, and simply click **Save** (which sets `enableUpdateCheck` in `Deskflow.conf`

image

sithlord48:
> Edit: I found a bug. It's in master. Do you want to fix as a separate PR or add a commit to this PR? I think a simple fix for this PR is to save the settings after dismissing the check update pop.
 
Added a commit to fix that here.

nbolton (approved):

nbolton:
> Set to automerge , we are go for review :P

Snap!

Issue #7896, opened 2024-11-14
GUI process does not exit when clearing settings


nbolton:
### Project

Deskflow

### Deskflow version number

1.17.1.26

### Build type

Local developer build (built it myself)

### What steps will reproduce the problem?

1. Clear settings (Help -> Clear settings)
2. Enter computer name
3. File -> Quit

Expected: Settings are saved and process ends
Actual: Process hangs instead of saving and exiting

### Log output

```shell
[2024-11-14T11:17:25] DEBUG: emitting config saving signal
        /home/nick/Projects/deskflow/src/lib/gui/config/ConfigScopes.cpp:74
[2024-11-14T11:17:25] DEBUG: committing app config
        /home/nick/Projects/deskflow/src/lib/gui/config/AppConfig.cpp:184
[2024-11-14T11:17:25] DEBUG: writing config to filesystem
        /home/nick/Projects/deskflow/src/lib/gui/config/ConfigScopes.cpp:78
[2024-11-14T11:17:25] DEBUG: recalling server config
        /home/nick/Projects/deskflow/src/gui/src/ServerConfig.cpp:173
[2024-11-14T11:17:25] DEBUG: server disabled
        /home/nick/Projects/deskflow/src/gui/src/MainWindow.cpp:1018
[2024-11-14T11:17:25] DEBUG: client disabled
        /home/nick/Projects/deskflow/src/gui/src/MainWindow.cpp:1042
[2024-11-14T11:17:25] DEBUG: active settings path: /home/nick/.config/Deskflow/Deskflow.conf
        /home/nick/Projects/deskflow/src/gui/src/MainWindow.cpp:260
[2024-11-14T11:17:26] DEBUG: saving window size and position
        /home/nick/Projects/deskflow/src/gui/src/MainWindow.cpp:135
[2024-11-14T11:17:26] DEBUG: emitting config saving signal
        /home/nick/Projects/deskflow/src/lib/gui/config/ConfigScopes.cpp:74
[2024-11-14T11:17:26] DEBUG: committing app config
        /home/nick/Projects/deskflow/src/lib/gui/config/AppConfig.cpp:184
[2024-11-14T11:17:26] DEBUG: committing server config
        /home/nick/Projects/deskflow/src/gui/src/ServerConfig.cpp:121
[2024-11-14T11:17:26] DEBUG: writing config to filesystem
        /home/nick/Projects/deskflow/src/lib/gui/config/ConfigScopes.cpp:78
[...]

PR #7897, opened 2024-11-14
Restart Qt app instance instead of process

nbolton:
fixes: #7896

Instead of launching a new process, which is a bit hacky, this simply starts a new Qt app instance after the existing one exits. As well as fixing a bug (where the new process hangs on exit), this also makes the process far easier to debug, since there is no new process to attach to.

I suggest we land this before: #7896

**To test:**
1. Clear settings (Help -> Clear settings)
2. Enter computer name
3. File -> Quit

Expect: Instead of the app hanging like before (I had to press ctrl+C to kill the process), it should exit cleanly.

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -164,5 +164,5 @@ _last = "moe:24800"
 # DESKFLOW_GUI_VERBOSE=true
 
 # Reset all settings and delete all data on startup
-# DESKFLOW_RESET_ALL=true
+# DESKFLOW_CLEAR_SETTINGS=true
 ```
--- a/src/gui/src/MainWindow.cpp
+++ b/src/gui/src/MainWindow.cpp
@@ -367,7 +367,12 @@ void MainWindow::on_m_pActionClearSettings_triggered()
 
   m_Quitting = true;
   m_SaveOnExit = false;
-  diagnostic::clearSettings(m_ConfigScopes, true);
+  diagnostic::clearSettings(m_ConfigScopes);
+
+  // Expected to cause the app to restart after exiting.
+  Q_EMIT clearedSettings();
+
+  QApplication::quit();
 }
 
 bool MainWindow::on_m_pActionSave_triggered()
--- a/src/gui/src/MainWindow.h
+++ b/src/gui/src/MainWindow.h
@@ -96,6 +96,7 @@ public:
 signals:
   void created();
   void shown();
+  void clearedSettings();
 
 public slots:
   void onAppAboutToQuit();
--- a/src/gui/src/main.cpp
+++ b/src/gui/src/main.cpp
@@ -119,7 +119,7 @@ int main(int argc, char *argv[])
   // --no-reset
   QStringList arguments = QCoreApplication::arguments();
   const auto noReset = hasArg("--no-reset", arguments);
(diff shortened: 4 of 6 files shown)


doc/configuration.md
```diff
@@ -164,5 +164,5 @@ _last = "moe:24800"
 # DESKFLOW_GUI_VERBOSE=true
 
 # Reset all settings and delete all data on startup
-# DESKFLOW_RESET_ALL=true
+# DESKFLOW_CLEAR_SETTINGS=true
```
sithlord48, line 167: We need a setting to clear the setting? WHY this is a horrible pattern.

src/gui/src/MainWindow.cpp
```diff
@@ -367,7 +367,12 @@ void MainWindow::on_m_pActionClearSettings_triggered()
 
   m_Quitting = true;
   m_SaveOnExit = false;
-  diagnostic::clearSettings(m_ConfigScopes, true);
+  diagnostic::clearSettings(m_ConfigScopes);
+
+  // Expected to cause the app to restart after exiting.
+  Q_EMIT clearedSettings();
+
+  QApplication::quit();
```
sithlord48, line 375: This is the bug fix

sithlord48 (changes requested):
This is alot of code to do a very simple thing. If you wanna exit when settings are cleared then tell the user you cleared the settings then exit. That's it .

src/gui/src/MainWindow.cpp
nbolton: I disagree, this is part of the fix.

sithlord48:
> > If you wanna exit when settings are cleared then tell the user you cleared the settings then exit.
> 
> You're missing the point. Exiting the app, or starting a new process is horrible UX.

Then fix the settings so restart is not needed.. Cause when we do that this code is now useless...

doc/configuration.md
nbolton: I’d like to keep this functionality (it’s already in `master`) as it serves a specific purpose for development and debugging. Developer conveniences like this contribute directly to product quality by streamlining our workflow. This setting allows us to quickly reset all configurations without having to manually delete files or navigate through the UI.

It’s particularly valuable in testing scenarios where we need to reset the application to a 'fresh' state to verify behaviors reliably. Removing it would increase both the time and cognitive load for developers who need to clear configurations repeatedly while testing or troubleshooting. For instance, when developing the wizard.

In short, this setting boosts development efficiency and maintains a smoother workflow without impacting end users. It’s an established part of our process, adding real value with a minimal code footprint. Removing it, in my opinion, would only make things harder for the rest of us, with no tangible upside.

sithlord48: If its in master that is fine we can remove it later.

nbolton:
> Then fix the settings so restart is not needed.. Cause when we do that this code is now useless...

What are you proposing exactly? Somehow reset the state of the main window without restarting it then show the wizard?

sithlord48:
> > Then fix the settings so restart is not needed.. Cause when we do that this code is now useless...
> 
> What are you proposing exactly? Somehow reset the state of the main window without restarting it then show the wizard?

Yes.

PR #7898, opened 2024-11-14
Use k app name more

sithlord48:
Land After #7824  
 -  Consolidate our "Deskflow" strings to use kAppName where possible


sithlord48:
added to #7824

PR #7900, opened 2024-11-17
fix: single typo in README.md

jcrm1:
`brew instal` works as intended, but only because Homebrew has aliases for a few common typos. Changed to `brew install`


nbolton:
Whoops! Thanks for the PR.

PR #7908, opened 2024-11-20
Flatpak nits

sithlord48:
- Adjust the generated name for flatpak artifacts. `deskflow-version-linux-arch.flatpak`
 - Do not cleanup /lib/debug from the generated flatpak
 - Fallback to 'linux' if `DISTRO_NAME` is empty
 - Update the description in the metadata
 - User `master` version of flatpak builder
 - Use `x86_64` for linux job name sufixes

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -306,7 +306,7 @@ jobs:
       - name: Get version
         uses: ./.github/actions/get-version
 
-      - uses: flatpak/flatpak-github-actions/flatpak-builder@v6
+      - uses: flatpak/flatpak-github-actions/flatpak-builder@master
         name: "Build"
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -313,10 +313,10 @@ jobs:
       - name: Get version
         uses: ./.github/actions/get-version
 
-      - uses: flatpak/flatpak-github-actions/flatpak-builder@v6
```
nbolton, line 316: Usually it's advisable to version lock GitHub actions, as master tends to be less stable. Did we need something specific from master that wasn't in latest release?

nbolton (approved):

.github/workflows/continuous-integration.yml
sithlord48: yes the ability to not upload artifacts .. i will pin to a release as soon as they make one

PR #7909, opened 2024-11-21
Arm flatpaks

sithlord48:
based on #7908 . That must land first. 

- Add aarch64 (arm64) build for flatpak
- Split this as im not sure we want to build this here as it makes our minimum build time around 25 minutes.


nbolton:
I'll wait for #7908 to land before reviewing.

PR #7929, opened 2024-11-27
Clean smells

sithlord48:
- Clean up code smells reported by sonar cloud
   - scripts/package.py, remove unused vars in package method
   - IArchTaskBarReciver subclasses Icon pointer should not be a const
   - IArchTaskBarReeiver subclasses should return nullptr for empty icon not NULL
   - IArchTrayBarReceiver subclass should override virtual methods not mark them virtual
   - src/gui/Action initilize items in class when possible
   - Replace most cases of `typedef` with `using`

### DIFF
--- a/src/lib/server/Config.cpp
+++ b/src/lib/server/Config.cpp
@@ -715,7 +715,7 @@ void Config::readSectionOptions(ConfigReadContext &s)
     if (handled) {
       // make sure handled options aren't followed by more values
       if (i < line.size() && (line[i] == ',' || line[i] == ';')) {
-        throw XConfigRead(s, "to many arguments to %s", name.c_str());
+        throw XConfigRead(s, "to many arguments to %s", name);
       }
     } else {
       // make filter rule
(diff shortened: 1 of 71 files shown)


nbolton:
Big kahuna! Will get to this soon. A tad swamped right now.

src/lib/server/Config.cpp
```diff
@@ -715,7 +715,7 @@ void Config::readSectionOptions(ConfigReadContext &s)
     if (handled) {
       // make sure handled options aren't followed by more values
       if (i < line.size() && (line[i] == ',' || line[i] == ';')) {
-        throw XConfigRead(s, "to many arguments to %s", name.c_str());
+        throw XConfigRead(s, "to many arguments to %s", name);
```
nbolton, line 718: I tested this line, and it prints:
```
[2024-12-04T14:04:24] ERROR: cannot read configuration "/home/nick/.config/Deskflow/deskflow-server.conf": read error: line 49: to many arguments to %s
        /home/nick/Projects/deskflow/src/lib/deskflow/ServerApp.cpp:257
```

However, problem is the same on `master`, so I raised a bug: https://github.com/deskflow/deskflow/issues/7942

nbolton (approved):
Great PR

src/lib/server/Config.cpp
sithlord48: Are you saying this fixes those issues ?

nbolton: No, I'm saying that I discovered those issues while testing the changes in this PR, and found that they exist in master and so are neither caused by or fixed by this PR, however they are related to this PR in that the changed code is related.

Issue #7806, opened 2024-10-24
CVEs related to TLS and TCP connections


nbolton:
💵 **Bounty**: $1000 ([details](https://github.com/deskflow/deskflow/issues/7806#issuecomment-2545811616))

### Feature Request

There are a few CVEs that affect Deskflow, as well as one SonarCloud hotspot:

- [ ] CVE-2021-42072
- [ ] CVE-2021-42073
- [x] CVE-2021-42074
- [x] CVE-2021-42075
- [x] CVE-2021-42076
- [ ] [SonarCloud hotspot: cpp:S5527](https://sonarcloud.io/project/issues?open=AZH_1rPbUm1X5WafpxUI&id=deskflow_deskflow) (Enable server hostname verification on this SSL/TLS connection)
- [ ] https://github.com/deskflow/deskflow/security/code-scanning/554

The fixes downstream need to be ported upstream. Except for [SonarCloud hotspot: cpp:S5527](https://sonarcloud.io/project/issues?open=AZH_1rPbUm1X5WafpxUI&id=deskflow_deskflow), which may be fixed by one of the CVE fixes, or may need to be fixed as well. There is also this GitHub detected issue, but that appears to be the same as the SonarCloud issue: https://github.com/deskflow/deskflow/security/code-scanning/554

Related:
- #8010
- [SonarCloud hotspot: cpp:S5527](https://sonarcloud.io/project/issues?open=AZH_1rPbUm1X5WafpxUI&id=deskflow_deskflow)
- https://github.com/debauchee/barrier/commit/229abab99f39f11624e5651f819e7f1f8eddedcc

nbolton:
@mgerstner It looks like you helped with a lot of these, so this may be of interest.

mgerstner:
> @mgerstner It looks like you helped with a lot of these, so this may be of interest.

Ah yes, that was in Barrier. Quite some time ago, I completely forgot about it. Thanks for the hint!

sithlord48:
https://github.com/debauchee/barrier/pulls?q=is%3Apr+is%3Amerged+Cve

sithlord48:
Some of this is started by this pr 
https://github.com/deskflow/deskflow/pull/7931

sithlord48:
No we only have a text file for the fingerprints at this time.. I suspect #7943  its partly related to this.

sithlord48:
If you like you can work off my Pr for those two CVEs

sithlord48:
[...]

PR #7931, opened 2024-11-28
Feat: Peer Id Checking

sithlord48:
Based on #8195 

The final part that fixes: #7806 ( It may solve the sonar hotspot as we have a mode where both clients and server keys are stored so they can say their name is whatever but w/o the key they must be readded again as the key changed) 

Recreates : https://github.com/debauchee/barrier/pull/1346   (and more) 
From that PR
> This PR implements client identity verification. Essentially server gets the same process of accepting and rejecting clients like the clients can accept or reject the server. This is important because even though the client can't move the mouse on the server, it can still receive input and potentially set the clipboard.
> 

This PR fixes the following security vulnerabilities:

  * [CVE-2021-42072](https://github.com/advisories/GHSA-wvmp-5fvm-29ff) server does not verify client identity (certificate fingerprint)
  * [CVE-2021-42073](https://github.com/advisories/GHSA-fm32-wgpg-7rc9) By guessing/listening in on valid client names server clipboard content can be manipulated.
 
- [x] Work with Peer ID Check OFF
- [x]  Works with Peer ID Check ON

In Addition: 
 - Modify the fingerprint dialog a bit so we can look at the local keys when pairing. 

Current local fingerprint dialog header text 
`Local computer's fingerprints`


nbolton:
I removed myself as reviewer while this is draft. Let me know if you actually need me to review.

nbolton:
Oh, its finally ready! Will try to review tomorrow.

nbolton:
I am concerned that this will confuse users. Why would they need to click this button? 
image

It's likely in my experience that they will click it and think, "Hey those don't match" to which we will respond "Of course they don't match, they're not supposed to"... and we'll repeat that response until ultimately we remove the button as we tire of this discussion.

TL;DR: People don't read things and will end up wasting our time with questions that don't make sense.

nbolton:
> With the dialog up there is no way to see your local keys to confirm what is on the screen in the case a new server connecting to a new client both will have the dialog up showing the others keys with no way to view their own. I do not want to add the local keys into dialog as having to see them to compare can happen in cases where you are not needing to see the local ones for example a known client or sever generates new keys.

Which problem are we trying to solve?
1. In case of a MITM, the user will see this dialog abruptly, and because they can't see what their existing client-side fingerprint was (to compare with the server) they may be tricked into thinking that the fingerprint hasn't changed.
2. When the client connects and you see the dialog on the server, you cannot check your local fingerprints to ensure that the fingerprint on the client side is valid.

In both cases, I see the value of the button but I still think the wording will confuse some people.

I also think that the word "this" will cause confusion: Does "this computer" mean the client that is connecting or this current computer that I'm on? To us devs, it's obviously this current computer that I'm on but that's because we know how it works.

sithlord48:
> > With the dialog up there is no way to see your local keys to confirm what is on the screen in the case a new server connecting to a new client both will have the dialog up showing the others keys with no way to view their own. I do not want to add the local keys into dialog as having to see them to compare can happen in cases where you are not needing to see the local ones for example a known client or sever generates new keys.
> 
> Which problem are we trying to solve?
> 
>     1. In case of a MITM, the user will see this dialog abruptly, and because they can't see what their existing client-side fingerprint was (to compare with the server) they may be tricked into thinking that the fingerprint hasn't changed.
> 
>     2. When the client connects and you see the dialog on the server, you cannot check your local fingerprints to ensure that the fingerprint on the client side is valid.
> 
> 
> In both cases, I see the value of the button but I still think the wording will confuse some people.
> 
> I also think that the word "this" will cause confusion: Does "this computer" mean the client that is connecting or this current computer that I'm on? To us devs, it's obviously this current computer that I'm on but that's because we know how it works.

Yes lets word it better

nbolton:
Hmm, problem. Simple setup with 2 computers. I cleared my settings on both computers. On the server, I see no way to see the local server's fingerprint.
image

The only dialog I see is on the client:
TODO

sithlord48:
Setting clearing really needs to restart the Gui , i would consider that a seperate issue

src/lib/gui/config/AppConfig.h
```diff
@@ -325,6 +329,7 @@ class AppConfig : public QObject, public deskflow::gui::IAppConfig
   std::optional<bool> m_EnableUpdateCheck;
   bool m_logExpanded = true;
   bool m_colorfulTrayIcon = false;
+  bool m_RequireClientCert = false;
```
nbolton, line 332: Less secure by default is less secure. Would defaulting the 'require client certs' setting to `true` make sense? I fear with this disabled, 95% of users won't ever turn it on.

```suggestion
  bool m_RequireClientCert = true;
```

src/lib/gui/core/ServerConnection.cpp
```diff
@@ -93,7 +93,8 @@ void ServerConnection::handleNewClient(const QString &clientName)
   Q_EMIT messageShowing();
 
   m_messageShowing = true;
-  const auto result = m_pDeps->showNewClientPrompt(m_pParent, clientName);
+  const auto result =
+      m_pDeps->showNewClientPrompt(m_pParent, clientName, m_appConfig.tlsEnabled() && m_appConfig.requireClientCerts());
```
nbolton, line 97: Does it make sense to call the 3rd bool arg `previouslyAccepted`? 
Wouldn't something like `requireClientCerts` make sense?

```suggestion
  const auto result =
      m_pDeps->showNewClientPrompt(m_pParent, clientName, m_appConfig.tlsEnabled() && m_appConfig.requireClientCerts());
```

src/lib/gui/messages.cpp
```diff
@@ -205,23 +205,30 @@ void showClientConnectError(QWidget *parent, ClientError error, const QString &a
...
+    );
     return Add;
-  } else if (message.clickedButton() == ignore) {
-    return Ignore;
   } else {
-    qFatal("no expected dialog button was clicked");
-    abort();
+    QMessageBox message(parent);
+    const QPushButton *ignore = message.addButton("Ignore", QMessageBox::RejectRole);
+    const QPushButton *add = message.addButton("Add client", QMessageBox::AcceptRole);
```
nbolton, line 221: For consistency.
```suggestion
    const QPushButton *add = message.addButton("New Client", QMessageBox::AcceptRole);
```

src/lib/gui/messages.cpp
```diff
@@ -205,23 +205,30 @@ void showClientConnectError(QWidget *parent, ClientError error, const QString &a
...
-  const QPushButton *ignore = message.addButton("Ignore", QMessageBox::RejectRole);
-  const QPushButton *add = message.addButton("Add client", QMessageBox::AcceptRole);
-  message.setText(QString("A new client called '%1' wants to connect").arg(clientName));
-  message.exec();
-
-  if (message.clickedButton() == add) {
+  if (previouslyAccepted) {
+    QMessageBox::information(
+        parent, QString("New Client"),
+        QString("A new client \"%1\" has been accepted you must add it to your servers screen layout").arg(clientName)
```
nbolton, line 215: Improved grammar.
```suggestion
        QString("A new client called '%1' has been accepted. You'll need to add it to your server's screen layout.").arg(clientName)
```

src/lib/gui/messages.cpp
```diff
@@ -205,23 +205,30 @@ void showClientConnectError(QWidget *parent, ClientError error, const QString &a
   dialog.exec();
 }
 
-NewClientPromptResult showNewClientPrompt(QWidget *parent, const QString &clientName)
+NewClientPromptResult showNewClientPrompt(QWidget *parent, const QString &clientName, bool previouslyAccepted)
```
nbolton, line 208: Why do we not allow users to ignore clients when 'require client certs' is on?

Also the UI wording is now very different between the two logic branches. If we really do need the logic branch, perhaps some de-duping is needed.

src/lib/net/SecureListenSocket.h
```diff
@@ -16,8 +18,14 @@ class IDataSocket;
...
+  SecureListenSocket(
+      IEventQueue *events, SocketMultiplexer *socketMultiplexer, IArchNetwork::EAddressFamily family,
+      SecurityLevel securityLevel = SecurityLevel::PlainText
+  );
 
   // IListenSocket overrides
   virtual IDataSocket *accept();
+
+private:
+  SecurityLevel m_securityLevel;
```
nbolton, line 30: As this is only set in ctor, const makes sense rather than it being mutable.
```suggestion
  const SecurityLevel m_securityLevel;
```

src/lib/net/SecureSocket.cpp
```diff
@@ -335,6 +340,11 @@ bool SecureSocket::loadCertificates(std::string &filename)
   return true;
 }
 
+static int cert_verify_ignore_callback(X509_STORE_CTX *, void *)
```
nbolton, line 343: Free functions are normally defined separately--form the class functions--at the top of the source file.

Also it should be `camelCase`.

```suggestion
static int verifyIgnoreCertCallback(X509_STORE_CTX *, void *)
```

src/lib/net/SecureSocket.cpp
```diff
@@ -452,6 +479,17 @@ int SecureSocket::secureAccept(int socket)
 
 int SecureSocket::secureConnect(int socket)
 {
+
+  std::string certDir =
+      deskflow::string::sprintf("%s/%s/%s", ARCH->getProfileDirectory().c_str(), kSslDir, kCertificateFilename);
+
+  if (!loadCertificates(certDir)) {
+    LOG((CLOG_ERR "could not load client certificates"));
+    // FIXME: this is fatal
```
nbolton, line 488: I wouldn't call this fatal. Normally "fatal" means the application must exit.

src/lib/net/SecurityLevel.h
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * \enum SecurityLevel
+ * \brief This enum is used to set how the client and server will communicate.
+ */
+enum class SecurityLevel
+{
+  PlainText, /** The connection is not encrypted */
+  Encrypted, /** The connection is encrypted */
+  PeerAuth   /** The connection is encrypted and the peer has been authenticated */
```
nbolton, line 17: Do these comments really add much?

nbolton (changes requested):
Works well. Nits aside, I'm concerned about a couple of things.

src/lib/net/SecureSocket.cpp
sithlord48: the comment is from the ported code so i left is as is.

src/lib/net/SecurityLevel.h
sithlord48: Yes these are doxygen comments and will create a table tell you about the enum the generated table will look something like [this](https://sithlord48.github.io/ff7tk/classFF7SaveInfo.html#a26b833c52018990f2bd00a27bed3b958)

src/lib/gui/core/ServerConnection.cpp
sithlord48: Nope when we call this if we have peer auth or encryption then the use has all ready accepted that they want the client to be able to connect.  so `previouslyAccepted` is what that represents.

src/lib/gui/messages.cpp
sithlord48: We don't . When its on we show a slightly different box that just says you have to add this to the screen layout.

src/lib/gui/messages.cpp
sithlord48: With in the context of the that wording is sane
```
      A New client 'foo' want to connect
                         [Add client] [Ignore]
```

Vs. 

```
      A New client 'foo' want to connect
                         [New client] [Ignore]
```
 
If i was gonna change the previous text (as it been in the dialog forever) 
I would use the standard action names
```
      A New client 'foo' want to connect
                                      [Accept] [Reject]
```

src/lib/gui/messages.cpp
sithlord48: > Why do we not allow users to ignore clients when 'require client certs' is on? I.e. there is no 'Ignore' button on this logic branch. Also, the UI wording is now different between the two logic branches. Should it be?

Good question, When peer auth is on the user has already accepted the client connection in the dialog where they view the remote keys. This dialog will show after that one. This is why that version is only an information box to tell the user that they need to add the client before they are taken to the sever config page. I added it cause it was a bit weird to just be thrown into the screen config .

src/lib/gui/core/ServerConnection.cpp
nbolton: I don't see how this makes sense. How can `requireClientCerts` returning true prove that a client has been previously accepted? All it proves is that you checked the 'Require client certificates' checkbox.

src/lib/gui/messages.cpp
nbolton: I can't see how the code in its current state makes sense. In both cases, you should add it to the layout if you want the client to connect. Why would one give you the opportunity to ignore and not the other. And why would the variable be called `previouslyAccepted` when it's related to a new client?

This is all seriously confusing code.

nbolton: Not for future self: After having a quick call, this all makes way more sense now :tada:

nbolton (approved):

nbolton:
@sithlord48 The milestone set for this PR (2.0.0) may be causing some confusion. We should review what issues/PRs are in 2.0.0 milestone.

sithlord48:
This is a goal we need to completed before 2.0.0 can be released.. We don't hold code back for a specific release we just land it when ready and our milestones act  as a guide for us  to know when we can say we can bump to version ...

PR #7934, opened 2024-12-01
Add Requirement for Qt 6.7+

sithlord48:
- Added missing Qt Version to CMake

Without it those without Qt 6.7 can configure successfully but the will have build failures.


sithlord48:
added to #7927

PR #7944, opened 2024-12-04
ReAdd NDEBUG define

sithlord48:
Removing the Define for `NDEBUG` was a mistake 
 
1. Set `NDEBUG` in main CMakeLists.txt when needed
2. Make note to unify the logic to not need this define

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,6 +96,14 @@ set(REQUIRED_LIBEI_VERSION 1.3)
 set(REQUIRED_LIBPORTAL_VERSION 0.8)
 set(REQUIRED_QT_VERSION 6.7.0)
 
+# Control debug item visibility
+# When not set logging is forced to DEBUG and show code locations
+# Also exposes a test menu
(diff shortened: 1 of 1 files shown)


nbolton (approved):
Interesting... When I configure `master` with `-DCMAKE_BUILD_TYPE=Release`, then `NDEBUG` is defined already.

I guess you're not setting `CMAKE_BUILD_TYPE`? This is a valid case so I agree we should default to release... but perhaps we should also set `CMAKE_BUILD_TYPE` too rather than just setting the `NDEBUG` define?

My thinking is: What else does setting `CMAKE_BUILD_TYPE=Release` do other than defining `NDEBUG`?

sithlord48:
I do not want to set a CMAKE_BUIILD_TYPE that is the builders job. But we do want to make sure if you build in RelWithDebInfo (my normal build) or Undefined (how the aur's deskflow-git ) package is made. 

1. Add this back for undefined.
2.  Report to on the aur package for them to add a build type to their config step

PR #7958, opened 2024-12-06
fix: Tray Icon Click to hide / show

sithlord48:
- Fix up the tray icon so the app shows or hides when the icon is Clicked 
 - Remove Show / Hide from the trayMenu
 - Remove Show from the window menu , it is not possible to trigger this its disabled when the window is shown

### DIFF
--- a/src/gui/src/MainWindow.cpp
+++ b/src/gui/src/MainWindow.cpp
@@ -288,14 +288,9 @@ void MainWindow::onAppConfigTlsChanged()
 
 void MainWindow::onTrayIconActivated(QSystemTrayIcon::ActivationReason reason)
 {
-
-  if (reason == QSystemTrayIcon::DoubleClick) {
-    if (isVisible()) {
-      hide();
-    } else {
-      showAndActivate();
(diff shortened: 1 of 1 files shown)


nbolton (approved):
Works on all OS, especially works well on Linux and Windows. 

macOS is a bit fiddly, but this might be down to the OS design language. It seems that tray icons do not normally show the app on left click; there is normally an 'Open app' or 'Preferences' menu item. Right click is often used for alternative actions such as toggle on/off.

Happy to merge but you may want to take another look at the macOS design language for this.

sithlord48:
> macOS is a bit fiddly, but this might be down to the OS design language. It seems that tray icons do not normally show the app on left click; there is normally an 'Open app' or 'Preferences' menu item. Right click is often used for alternative actions such as toggle on/off.
> 
> Happy to merge but you may want to take another look at the macOS design language for this.

Yeah sounds more like mac being different cause mac is different. Lets land this and we can lookin to what kinda mac specific stuff we could do in a later PR

Issue #7789, opened 2024-10-22
Unicode bug: `failed to cd up to profile root dir`


L1Q:
### Project

Deskflow

### Deskflow version number

1.17.0.120

### Build type

Other (please specify)

### Operating systems (OS)

- [X] Windows
- [ ] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [ ] I have reviewed the Wayland known issues and my issue is new
- [ ] I am not using Wayland on Linux

### Signing on macOS

- [ ] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### Continuous build

- [X] I have tried the latest continuous build and the issue persists
- [ ] I am unable to try the latest continuous build

### OS versions/distros

- Windows 10 22H2 build 19045.5011

### Deskflow configuration

This happens on the Windows 10 client, before anything connects to anything, eg. blank or known invalid server ip.

### What steps will reproduce the problem?

- Have user dir: `C:\Users\Пользователь`
- Install the Deskflow windows using the msi isntaller from [latest CI build](https://github.com/deskflow/deskflow/releases/download/continuous/deskflow_1.17.0.120_windows_x64.msi).
- Run Deskflow, literally just press the "Connect" button.
- Get this window:

![deskflow_2024-10-22_16-28-05](https://github.com/user-attachments/assets/1b5de78f-9482-47a8-84c9-b696a97e9dc1)

> v1.17.0.120 (2136af68)
> failed to cd up to profile root dir

### Additional information

Verly likely due to cyrillic home directory name, it looks jumpbled in the preferences
![deskflow_2024-10-22_16-31-49](https://github.com/user-attachments/assets/5fcd7b34-34e4-492e-a39a-0880a2227261)

(checking or unchecking the TLS box makes no difference for this issue)

nbolton:
Oooo, nice username! Looks like bad UTF-8 support on our side. What's the actual path to your user dir?

sithlord48:
we should use utf8 stings for this stuff for sure.

nbolton:
> Пользователь

Thanks, updated your repro steps to include this.

sithlord48:
Duplicate #7995  Has this issue with Deskflow on C:\Deskflow

sithlord48:
Duplicate #7961 does not state the install path.

Temporalin:
[...]

PR #7959, opened 2024-12-07
Fixed UTF-16 surrogate handling

pedronavf:
The data pointer needs to move before decoding the second surrogate, and not after. Because of this the first surrogate was begin decoded again, resulting in an invalid codepoint.

This affected clipboard operations originating on Windows machines, where the text in CF_UNICODE is encoded in UTF-16 and copying characters from a high plane (like emojis) was broken.

### DIFF
--- a/src/lib/base/Unicode.cpp
+++ b/src/lib/base/Unicode.cpp
@@ -459,9 +459,9 @@ String Unicode::doUTF16ToUTF8(const UInt8 *data, UInt32 n, bool *errors)
       setError(errors);
       toUTF8(dst, s_replacement, NULL);
     } else if (c >= 0x0000d800 && c <= 0x0000dbff) {
-      UInt32 c2 = decode16(data, byteSwapped);
       data += 2;
       --n;
+      UInt32 c2 = decode16(data, byteSwapped);
       if (c2 < 0x0000dc00 || c2 > 0x0000dfff) {
         // error -- [d800,dbff] not followed by [dc00,dfff]
         setError(errors);


sithlord48:
could you please change the commit message to use `fix:` instead of `Fixed` this will make sure its picked up by the auto notes generated on release.

pedronavf:
@sithlord48 done!

sithlord48:
Possible solution for : #7961  #7789

sithlord48 (approved):
:+1:

sithlord48:
@pedronavf  i will take over any needed updates for this PR until it lands. So no need to worry about doing rebases.

nbolton:
> This affected clipboard operations originating on Windows machines, where the text in CF_UNICODE is encoded in UTF-16 and copying characters from a high plane (like emojis) was broken.

Do you have steps to repro the bug? On `master` I am able to copy emojis to and from a Windows client.

pedronavf:
> > This affected clipboard operations originating on Windows machines, where the text in CF_UNICODE is encoded in UTF-16 and copying characters from a high plane (like emojis) was broken.
> 
> Do you have steps to repro the bug? On `master` I am able to copy emojis to and from a Windows client.

You need to copy an emoji from a Unicode high plane (like U+1F600), where the use of UTF-16 surrogate pairs is necessary.

Try to copy the emoji from this page: https://graphemica.com/1F600

nbolton:
Tested and working. Thanks for the PR.

Before: Copied 😀 (09 08 00 00) from Windows client to macOS client, resulted in � (FD FF 00 00)
After: Copied 😀 (09 08 00 00) from Windows client to macOS client and result was 😀 (09 08 00 00)

Also works the opposite way round (macOS to Windows) and also when Windows is server.

PR #7960, opened 2024-12-08
Access the Windows clipboard using newer APIs

pedronavf:
The way of monitoring the clipboard on Windows, using clipboard viewer windows, has been kept on modern Windows for backwards compatibility. It involves inserting a window in the clipboard viewer chain and relying on every program passing the message to the next, which could result in a program failing or stopping processing messages to disconnect the rest of the chain from receiving clipboard messages.

This was causing Synergy, and now Deskflow, to stop processing clipboard messages. The solution was to restart the client or copy something to the clipboard from another machine (I used Linux) which would trigger a clipboard re-opening. For me this happened 3 or 4 timnes per day.

With this PR I have been copying and pasting between Windows and Linux for 3 days without a single clipboard hang.


sithlord48:
It's nice to see someone working on some windows platform updates !

Two small nits about this 
  1. you need to run [clang-format](https://github.com/deskflow/deskflow/wiki/Clang%E2%80%90Format-Tips-%26-Tricks) on your commits
  2. your commit messages are missing words like `fix` and `feat` explained in our [hacking guide](https://github.com/deskflow/deskflow/wiki/Hacking-Guide). As an example I would use
    - fix: windows, don't encode empty data to clipboard
    - feat: windows use clipboard listener

pedronavf:
Got it. I read that pages and I misinterpreted "CI enforced"; I took it to mean that clang-format would run as a pre-commit hook, but I still tried to maintain the same code style.

Ah! Thanks for the link to the Hacking Guide. I actually read that page a couple of days ago and found it very cool how different tags were created, but for the life of me I couldn't find it today when I created the PR! I ended up thinking that I probably had seen those instructions on Input Leap's repo 🤦

pedronavf:
Should I create a bug first and then mention it in my commit message?

sithlord48:
> Should I create a bug first and then mention it in my commit message?

No you do not have to make an issue.

sithlord48 (approved):
Looks good on the surface I am unable to test this however. @nbolton  can you please test this

nbolton:
Thanks for the PR, testing this now. Will report my findings in 2-3 days.

nbolton (approved):
LGTM

PR #7977, opened 2024-12-16
fix: `qDebug` doesn't work on Fedora

nbolton:
IDK why, but on Fedora, `QT_LOGGING_RULES` is set in a way that turns off `qDebug()`. We have our own env var, `DESKFLOW_GUI_DEBUG` to turn on debug logging, so I think overriding `QT_LOGGING_RULES` should be ok. 

Open to alternatives.

### DIFF
--- a/src/apps/deskflow-gui/main.cpp
+++ b/src/apps/deskflow-gui/main.cpp
@@ -44,6 +44,10 @@
 #include <cstdlib>
 #endif
 
+#if defined(Q_OS_UNIX) && defined(QT_DEBUG)
+#include <QLoggingCategory>
+#endif
+
 using namespace deskflow::gui;
 
 class QThreadImpl : public QThread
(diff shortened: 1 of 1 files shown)


src/gui/src/main.cpp
```diff
@@ -66,6 +66,10 @@ bool hasArg(const QString &arg, const QStringList &args)
 
 int main(int argc, char *argv[])
 {
+#if defined(Q_OS_UNIX)
+  // Fixes Fedora bug where qDebug() messages aren't printed.
+  qputenv("QT_LOGGING_RULES", "*.debug=true;qt.*=false");
```
nbolton, line 71: Alternative: `QLoggingCategory::setFilterRules(QStringLiteral("driver.usb.debug=true"));`

sithlord48: This rule could be for any OS but i would guard it on it being a debug build

sithlord48:
Updated it please test.

sithlord48 (approved):
This way works for me

sithlord48 (dismissed):
LGTM

nbolton:
> Updated it please test.

Did not work, restored original hack.

src/apps/deskflow-gui/main.cpp
```diff
@@ -66,6 +67,10 @@ bool hasArg(const QString &arg, const QStringList &args)
 
 int main(int argc, char *argv[])
 {
+#if defined(Q_OS_UNIX)
```
sithlord48, line 70: Please only change for debug builds and also include the item only for these builds as well

sithlord48 (changes requested):

sithlord48 (approved):

PR #7982, opened 2024-12-16
Fix(CVE-2021-42074): Handle SSL race conditions and segmentation fault 

varshith257:
Part of #7806


sithlord48 (commented):
Should there be an unlock for the mutex ?

nbolton:
@mgerstner

sithlord48 (approved):
Does not break my current connections and disconnect seams to also work fine
+1

PR #7984, opened 2024-12-17
Fix(CVE-2021-42076): Enforce maximum message length

varshith257:
Part of #7806


sithlord48 (approved):
Looks Good does not break my connections

PR #7986, opened 2024-12-17
Improve the macOS tray UX

sithlord48:
General:
 - Use QList<QAction*> instead of a std::vector<QAction*>
 - Clean up the `showAndActivate` Logic just show always no matter the previous state
 - Clang-Format the `.m` and `.mm` files ( I had to edit one so i did a commit formatting them all) 
 - Added *.m and *.mm to `scipts/lint-clang.py`

 Update the tray for mac OS
 -  Do not show / hide when the icon is pressed
 -  Show "Bring Window to Front"  in the tray menu (ActionRestore)
 - "Bring Window to Front" works when the app is hidden with ⌘ + H

### DIFF
--- a/src/gui/src/MainWindow.cpp
+++ b/src/gui/src/MainWindow.cpp
@@ -240,7 +240,10 @@ void MainWindow::connectSlots()
 
   connect(&m_WindowSaveTimer, &QTimer::timeout, this, &MainWindow::onWindowSaveTimerTimeout);
 
+// Mac os tray will only show a menu
+#ifndef Q_OS_MAC
   connect(&m_TrayIcon, &TrayIcon::activated, this, &MainWindow::onTrayIconActivated);
+#endif
 
   connect(
       &m_ServerConnection, &ServerConnection::configureClient, this, &MainWindow::onServerConnectionConfigureClient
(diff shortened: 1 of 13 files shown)


src/gui/src/MainWindow.cpp
```diff
@@ -226,7 +226,10 @@ void MainWindow::connectSlots()
 
   connect(&m_WindowSaveTimer, &QTimer::timeout, this, &MainWindow::onWindowSaveTimerTimeout);
 
+// Mac os tray will only show a menu
```
nbolton, line 229: Silencio, Bruno!

sithlord48: ... We don't talk about Bruno... :P

src/gui/src/MainWindow.cpp
```diff
@@ -523,9 +526,14 @@ void MainWindow::moveEvent(QMoveEvent *event)
 void MainWindow::open()
 {
 
-  std::vector<QAction *> trayMenu = {ui->m_pActionStartCore, ui->m_pActionStopCore, nullptr, ui->m_pActionQuit};
+  QList<QAction *> trayActions{ui->m_pActionStartCore, ui->m_pActionStopCore, nullptr, ui->m_pActionQuit};
 
-  m_TrayIcon.create(trayMenu);
+#ifdef Q_OS_MAC
+  ui->m_pActionRestore->setText(tr("Bring Window to Front"));
```
nbolton, line 532: Is it possible to only enable this menu option if the window is hidden?

sithlord48: Not in a way that works in all cases mac os does not for example tell the app when the user has used command +H  to hide it

nbolton (approved):

PR #7987, opened 2024-12-18
Rm post config macros

### LINKED ISSUE #7754: Replace hacky `fancy_copy.py` script with CMake copy
### Feature Request

The `fancy_copy.py` script was created to work around problems with Windows and file in use running process lock problems related to working with the daemon and Windows services. We build to a temp dir on Windows so that we can guarantee the build always works, even if processes are running (such as the Windows service). It is not practical to shut down the Windows service every time you want to compile.

There is probably a way of doing this in CMake, so that we can get rid of `fancy_copy.py`. We need to mimic what `fancy_copy.py` does, which is copy the files and ignore each file copy if the file is in use by Windows, but show a warning. There is also a related `daemon.py` script which installs and reinstalls the Windows service daemon, and that uses the same copy logic in ``fancy_copy.py`.

TL;DR: Developing with Windows services (and Windows in general) is a horrible experience and `fancy_copy.py` works around this problem. It would be better if CMake could do this so we don't have to use Python.

Idea: Make this funky copy logic a CMake option, so that by default it just builds normally to the build dir. A developer can optionally turn on the "safe" build/copy feature.

sithlord48:
- Remove fancy copy resolves #7754 
 - Clean up script dir
 - Stop using post_config (i.e fancy copy)
 - No Longer Replace running exes on windows,  Users on windows should install a copy using package for testing or copy a built copy to a safer place to run from. Small inconvenience for the greater good of clean up .
 
@pedronavf  , can you check this out please since you are using windows and let me know if this is worse the before for a dev.


scripts/daemon.py
nbolton, line 1: I guess we can move this to a new scripts repo. It's an important time saver for me on Windows. 

I do understand that it's a big burden on the project to maintain this kind of script, so it makes sense to move it to an external repo.

nbolton (approved):

nbolton (approved):

scripts/daemon.py
sithlord48: sure but this cmake stuff has to go in order to be able to move this to a new repo

PR #7988, opened 2024-12-18
ci: simplify the lint step

sithlord48:
Simplifies the lint step renames it to `lint-check`
 - Modify action/lint-clang to use clang-format directly 
 - Remove now unused action init-python
 - Consolidate the actions lint-clang and lint-error into new  lint-check


nbolton (changes requested):
The preview now looks a bit harder to read and the command is also incorrect for local dev:
image

Example: https://github.com/deskflow/deskflow/actions/runs/12454611838?pr=8001

This used to be a short script, but now that it's a long command, it's a bit chewy.

I think this is more helpful:
```diff
diff --git a/.github/actions/lint-check/action.yml b/.github/actions/lint-check/action.yml
index 76c36085..b1901b31 100644
--- a/.github/actions/lint-check/action.yml
+++ b/.github/actions/lint-check/action.yml
@@ -60,7 +60,7 @@ runs:
         summary=$(cat<<EOF
         ❌ \`${{ inputs.format-tool }}\`: It looks like your changes don't match our code style.
 
-        🛠️ Please either run \`${{ inputs.format-command }}\` or apply this patch with \`git apply\`:
+        🛠️ Please either run \`clang-format -i\` on the file, or apply this patch with \`git apply\`:
 
         [\`${{ steps.changes.outputs.file }}\`](${{ steps.upload.outputs.artifact-url }})
         $code_block diff
```
image

sithlord48:
done

nbolton (approved):
Yeah, that's better. Previously the inputs with defaults had become like constants since the inputs were not being changed, which is confusing.

TL;DR: Bad abstraction is worse than bad duplication. This is a good change.

Issue #7941, opened 2024-12-04
Process name is not printed correctly on invalid config


nbolton:
### Project

Deskflow

### Deskflow version number

1.17.2.24

### Build type

Local developer build (built it myself)

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [ ] Linux (X11)
- [x] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### OS versions/distros

Fedora 41

### What steps will reproduce the problem?

Have an error in the config file, e.g. to `section: options` add `foo = bar`

### Log output

```shell
[2024-12-04T14:00:29] FATAL: rver.conf: failed to load config:
```

### Additional information

Instead of the pname, we get `rver.conf`

Issue #7940, opened 2024-12-04
Default config path is incorrect and contains `%s`


nbolton:
### Project

Deskflow

### Deskflow version number

deskflow-1.17.2.24

### Build type

Local developer build (built it myself)

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [ ] Linux (X11)
- [X] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### OS versions/distros

Fedora 41

### What steps will reproduce the problem?

Run: `deskflow-server --config-toml deskflow-config.toml`
```
[server.args]
no-daemon = true
no-tray = true
debug = "DEBUG"
name = "spock-tux"
address = ":24800"
```

### Log output

```shell
[2024-12-04T13:52:01] DEBUG: opening configuration "/home/nick/%s.conf"
        /home/nick/Projects/deskflow/src/lib/deskflow/ServerApp.cpp:243
[2024-12-04T13:52:01] DEBUG: cannot open configuration "/home/nick/%s.conf"
        /home/nick/Projects/deskflow/src/lib/deskflow/ServerApp.cpp:249
[2024-12-04T13:52:01] DEBUG: opening configuration "/etc/%s.conf"
        /home/nick/Projects/deskflow/src/lib/deskflow/ServerApp.cpp:243
[2024-12-04T13:52:01] DEBUG: cannot open configuration "/etc/%s.conf"
        /home/nick/Projects/deskflow/src/lib/deskflow/ServerApp.cpp:249
[2024-12-04T13:52:14] FATAL: : failed to load config: /etc/%s.conf
        /home/nick/Projects/deskflow/src/lib/deskflow/ServerApp.cpp:234
```

### Additional information

For default config path, looks like the %s isn't replaced.

Workaround: Use `--config` option.

PR #7990, opened 2024-12-19
Fix: #7940 and #7941

sithlord48:
fixes: #7940 
fixes: #7941 

by using std::string 's insert method in place is %s sub for the static string

### DIFF
--- a/src/lib/deskflow/ServerApp.h
+++ b/src/lib/deskflow/ServerApp.h
@@ -22,6 +22,7 @@
 #include "arch/IArchMultithread.h"
 #include "base/EventTypes.h"
 #include "base/String.h"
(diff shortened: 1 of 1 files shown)


nbolton (changes requested):
This seems over complicated. Shouldn't it simply be...
```
inline static const std::string CONFIG_NAME = kAppName + ".sgc";
```

Or am I missing something?

sithlord48:
> This seems over complicated. Shouldn't it simply be...
> 
> ```
> inline static const auto CONFIG_NAME = std::string(kAppName) + ".conf";
> ```
> 
> Or am I missing something?

append is safer then using concatenation.

nbolton (approved):

Issue #7942, opened 2024-12-04
Too many arguments error shows `%s` instead of option name


nbolton:
### Project

Deskflow

### Deskflow version number

deskflow-1.17.2.16

### Build type

Local developer build (built it myself)

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [ ] Linux (X11)
- [X] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### OS versions/distros

Fedora 41

### What steps will reproduce the problem?

Add a extra arg to config option, e.g.
```
section: options
	protocol = barrier,foo
```

### Log output

`[2024-12-04T14:17:55] ERROR: cannot read configuration "/home/nick/.config/Deskflow/deskflow-server.conf": read error: line 49: to many arguments to %s
        /home/nick/Projects/deskflow/src/lib/deskflow/ServerApp.cpp:257`

### Additional information

`%s` is not replaced by string

sithlord48:
When i try todo this the file is cleaned and i don't see the error can you confirm this please.

PR #7991, opened 2024-12-19
fix: #7942, Correctly substitute config name part

sithlord48:
fixes https://github.com/deskflow/deskflow/issues/7942

### DIFF
--- a/src/lib/server/Config.cpp
+++ b/src/lib/server/Config.cpp
@@ -715,7 +715,7 @@ void Config::readSectionOptions(ConfigReadContext &s)
     if (handled) {
       // make sure handled options aren't followed by more values
       if (i < line.size() && (line[i] == ',' || line[i] == ';')) {
-        throw XConfigRead(s, "to many arguments to %s", name);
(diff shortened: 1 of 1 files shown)


src/lib/server/Config.cpp
```diff
@@ -715,7 +715,7 @@ void Config::readSectionOptions(ConfigReadContext &s)
     if (handled) {
       // make sure handled options aren't followed by more values
       if (i < line.size() && (line[i] == ',' || line[i] == ';')) {
-        throw XConfigRead(s, "to many arguments to %s", name);
+        throw XConfigRead(s, std::string("to many arguments to ").append(name));
```
nbolton, line 718:
```suggestion
        throw XConfigRead(s, std::string("too many arguments for:").append(name));
```

nbolton, line 718:
```suggestion
        throw XConfigRead(s, std::string("too many arguments for:").append(name));
```

nbolton (changes requested):
Good opportunity to improve the grammar here

nbolton (approved):

PR #7993, opened 2024-12-19
Single source for apps

sithlord48:
- Move all apps to `src/apps`
 - Single source for all app icons for windows
 - Update deskflowc deskflows mstray icons to use deskflow icon not synergy.


nbolton (changes requested):
I think this is a good time to get rid of the legacy `deskflowc` and `deskflows` names (originally `synergys` and `synergyc` to fit within the Windows 8 char limit). 

We should use `deskflow-server` and `deskflow-client` instead, which matches the binaries.

sithlord48:
> I think this is a good time to get rid of the legacy `deskflowc` and `deskflows` names (originally `synergys` and `synergyc` to fit within the Windows 8 char limit).
> 
> We should use `deskflow-server` and `deskflow-client` instead, which matches the binaries.

also changed deskflowd to deskflow-daemon

nbolton (approved):

PR #7994, opened 2024-12-20
feat: Visiblity toggle for the log on the main window

sithlord48:
Add the ability to toggle the log (its state is saved on exit)

When Deskflow's log is collapsed the widow's size is fixed: 
Screenshot_20241219_214321

If you expand the log: you can resize how you like and this size is remembered if you close the log it goes back to the above state. 
Screenshot_20241219_214454

### DIFF
--- a/src/apps/deskflow-gui/MainWindow.cpp
+++ b/src/apps/deskflow-gui/MainWindow.cpp
@@ -112,6 +112,16 @@ MainWindow::MainWindow(ConfigScopes &configScopes, AppConfig &appConfig)
   m_actionStartCore->setShortcut(QKeySequence(tr("Ctrl+S")));
   m_actionStopCore->setShortcut(QKeySequence(tr("Ctrl+T")));
 
+#ifdef Q_OS_MAC
+  ui->btnToggleLog->setFixedHeight(ui->lblLog->font().pixelSize());
+#endif
+
+  ui->btnToggleLog->setStyleSheet(QStringLiteral("background:rgba(0,0,0,0);"));
+  if (m_AppConfig.logExpanded())
+    ui->btnToggleLog->click();
+
+  toggleLogVisible(m_AppConfig.logExpanded());
+
   createMenuBar();
   setupControls();
   connectSlots();
@@ -127,6 +137,8 @@ MainWindow::MainWindow(ConfigScopes &configScopes, AppConfig &appConfig)
   restoreWindow();
 
   qDebug().noquote() << "active settings path:" << m_ConfigScopes.activeFilePath();
+
+  updateSize();
 }
 
 MainWindow::~MainWindow()
@@ -306,6 +318,8 @@ void MainWindow::connectSlots()
 
   connect(ui->rbModeServer, &QRadioButton::clicked, this, &MainWindow::setModeServer);
   connect(ui->rbModeClient, &QRadioButton::clicked, this, &MainWindow::setModeClient);
+
+  connect(ui->btnToggleLog, &QAbstractButton::toggled, this, &MainWindow::toggleLogVisible);
 }
 
 void MainWindow::onAppAboutToQuit()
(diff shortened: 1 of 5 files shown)


nbolton:
> Screenshot_20241219_214321

Cool. I think it'd look neater if the server and client groupboxes were equally sized (50%+50%). Right now the left box is a bit smaller and it looks unintentional.

nbolton (changes requested):
> Screenshot_20241219_214321

Cool. I think it'd look neater if the server and client groupboxes were equally sized (50%+50%). Right now the left box is a bit smaller and it looks unintentional.

sithlord48:
The boxes are sized based on their contents that is how layouts work , I am not going to do anything to change that .

src/apps/deskflow-gui/MainWindow.cpp
```diff
@@ -465,6 +494,25 @@ void MainWindow::resetCore()
...
 }
 
+void MainWindow::updateSize()
+{
+  if (ui->textLog->isVisible()) {
+    setMaximumHeight(16777215);
+    setMaximumWidth(16777215);
+#ifdef Q_OS_MAC
+    // On mac os the titlebar is part of the height so we need to adjust our Y coord to not "jump" up
+    setGeometry(x(), y() + 28, m_logRect.width(), m_logRect.height());
```
nbolton, line 504: For readability, other developers would find it helpful for the magic number to be a constant. This is optimized out by the compiler so it won't add a perf overhead.

```suggestion
    // On macOS, the titlebar is part of the height so we need to adjust our Y coord to not "jump" up
    const auto kTitlebarOffset = 28;
    
    setGeometry(x(), y() + kTitlebarOffset, m_logRect.width(), m_logRect.height());
```

src/apps/deskflow-gui/MainWindow.cpp
```diff
@@ -315,6 +329,21 @@ void MainWindow::onAppAboutToQuit()
...
+    ui->btnToggleLog->setArrowType(Qt::DownArrow);
+    ui->textLog->setVisible(true);
+    m_AppConfig.setLogExpanded(true);
+  } else {
+    ui->btnToggleLog->setArrowType(Qt::RightArrow);
+    m_logRect = size();
+    ui->textLog->setVisible(false);
+    m_AppConfig.setLogExpanded(false);
+  }
+  QTimer::singleShot(10, this, &MainWindow::updateSize);
```
nbolton, line 344: Why 10?

nbolton (commented):
Up to you if you want to change.

src/apps/deskflow-gui/MainWindow.cpp
sithlord48: we just need a few ms for the events to be processed.  10 is the mininum

src/apps/deskflow-gui/MainWindow.cpp
sithlord48: that is fine i was just trying to avoid making a var that was not gonna be used too often

nbolton: Yeah I know what you mean.

src/apps/deskflow-gui/MainWindow.cpp
```diff
@@ -465,6 +495,26 @@ void MainWindow::resetCore()
...
 }
 
+void MainWindow::updateSize()
+{
+#ifdef Q_OS_MAC
+// On mac os the titlebar is part of the height so we need to adjust our Y coord to avoid moving the window up
+  const auto kTitleBarOffset = 28
+#else
+  const auto kTitleBarOffset = 0;
+#endif
```
nbolton, line 505: Nice, yeah that makes sense to have the const in the ifdef, as it makes crystal clear what is different for macOS.

nbolton (approved):

nbolton (approved):

PR #8019, opened 2024-12-28
fix: Unify both Quit actions to use the same function

sithlord48:
Fixes the mac os tray quit not working (forgot to connect it to something) 
  Added a single method `MainWindow::quitApp` for both quit actions to call


nbolton:
Will test when next on my Mac.

nbolton (approved):

PR #8020, opened 2024-12-28
refactor: Copy the full version string on the about dialog

sithlord48:
Set it so the AboutDialog's copy version button copies the entire version number shown. 

Before it would capture only the kVersion (x.y.z.t) now its also will include any SHA if added to the dialo. This will make it easier to compare dev versions that are still PRs in cases where the generated version would be the same as another PR version due to them having the same number of commits to add on top of master.


nbolton (approved):
Nice, I kept meaning to report this.

PR #8027, opened 2024-12-31
Allow only one instance of the gui

### LINKED ISSUE #5304: Gnome StatusIcon API is becoming deprecated
Hello,

As of [2014](https://bugzilla.gnome.org/show_bug.cgi?id=734826), the Gtk StatusIcon API has been deprecated and will soon stop being included in the desktop environment.

I am currently running on elementary OS Loki (the current development version) and the team has decided to drop the StatusIcon already.  

This GNOME [document](https://wiki.gnome.org/Design/Whiteboards/StatusIcons) discusses possible migration paths for different purposes.

When I brought up the issue with the elementary OS team. One of the suggestions was that synergy use a Switchboard plug (an entry in the elementary OS settings panel)

Ideally, you would change your UI to be in accordance with the [GNOME guidelines](https://wiki.gnome.org/Design/Whiteboards/StatusIcons/Guidelines) or the [elementary OS guidelines](https://elementary.io/docs/human-interface-guidelines).

I will be able to help with this issue in about a month time. If the decision is made to ignore it, I will fork synergy's UI.

There seems to be a pull request similar to this issue: https://github.com/symless/synergy/pull/4194

### LINKED ISSUE #7975: Do not allow more then one instance of deskflow running at a time
### Feature Request

Deskflow currently allows any number of gui / server / clients to be run . 

If gui / server or client is running prevent a new instance of the same from running at the same time 
Related #7974 (more to add here as well)

sithlord48:
- allow only one instance of the gui to be running 
 - When in the background opening a new instance should pop up the current one , this should work nicely in places with out tray support. 
 - fixes #7975 
 - fixes #5304 (as a tray icon is not really needed to get the running instance back now)

Test Results: 

 -  [x] Mac OS:
    -  [X] macOS Restores when window is minimized
    -  [X] macOS Restores when window is hidden
    -  [X] macOS Restores when window was closed, Needs: https://github.com/deskflow/deskflow/pull/8034
 
 -  [X] Linux (Wayland):
    -  [X] Wayland Restores when window is minimized
        *Wayland doesn't allow for this on all De's the Icon flashes to request you look at the window instead
    -  [X] Wayland Restores when window is hidden
    -  [X] Wayland Restores when window was closed
 
 -  [X] Linux (X11):
    -  [X] X11 Restores when window is minimized
    -  [X] X11 Restores when window is hidden
    -  [X] X11 Restores when window was closed

  - [X] Windows:
    -  [X] Windows Restores when window is minimized
    -  [X] Windows Restores when window is hidden
    -  [X] Windows Restores when window was closed

### DIFF
--- a/src/apps/deskflow-gui/main.cpp
+++ b/src/apps/deskflow-gui/main.cpp
@@ -35,8 +35,10 @@
 #include <QApplication>
 #include <QDebug>
 #include <QGuiApplication>
+#include <QLocalSocket>
 #include <QMessageBox>
 #include <QObject>
+#include <QSharedMemory>
 #include <QtGlobal>
 
 #if defined(Q_OS_MAC)
@@ -89,6 +91,29 @@ int main(int argc, char *argv[])
   QCoreApplication::setOrganizationDomain(kOrgDomain);
 
   QApplication app(argc, argv);
+
+  // Create a shared memory segment with a unique key
+  // This is to prevent a new instance from running if one is already running
+  QSharedMemory sharedMemory("deskflow-gui");
+
+  // Attempt to attach first and detach in order to clean up stale shm chunks
+  // This can happen if the previous instance was killed or crashed
+  if (sharedMemory.attach())
+    sharedMemory.detach();
+
+  // If we can create 1 byte of SHM we are the only instance
+  if (!sharedMemory.create(1)) {
+    // Ping the running instance to have it show itself
+    QLocalSocket socket;
+    socket.connectToServer("deskflow-gui", QLocalSocket::ReadOnly);
+    if (!socket.waitForConnected()) {
+      // If we can't connect to the other instance tell the user its running.
+      // This should never happen but just incase we should show something
+      QMessageBox::information(nullptr, QObject::tr("Deskflow"), QObject::tr("Deskflow is already running"));
+    }
+    socket.disconnectFromServer();
+    return 0;
+  }
 
 #if !defined(Q_OS_MAC)
   // causes dark mode to be used on some DE's
--- a/src/apps/deskflow-gui/MainWindow.cpp
+++ b/src/apps/deskflow-gui/MainWindow.cpp
@@ -43,6 +43,8 @@
 #include <QApplication>
 #include <QDesktopServices>
 #include <QFileDialog>
+#include <QLocalServer>
+#include <QLocalSocket>
 #include <QMenu>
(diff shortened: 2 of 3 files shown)


src/apps/deskflow-gui/main.cpp
```diff
@@ -90,6 +92,29 @@ int main(int argc, char *argv[])
 
   QApplication app(argc, argv);
 
+  // Create a shared memory segment with a unique key
+  // This is to prevent a new instance from running if one is already running
+  QSharedMemory sharedMemory("deskflowGUI");
```
nbolton, line 97:
```suggestion
  QSharedMemory sharedMemory("deskflow");
```

src/apps/deskflow-gui/main.cpp
```diff
@@ -90,6 +92,29 @@ int main(int argc, char *argv[])
...
+  // Attempt to attach first and detach in order to clean up stale shm chunks
+  // This can happen if the previous instance was killed or crashed
+  if (sharedMemory.attach())
+    sharedMemory.detach();
+
+  // If we can create 1 byte of SHM we are the only instance
+  if (!sharedMemory.create(1)) {
+    // Ping the running instance to have it show itself
+    QLocalSocket socket;
+    socket.connectToServer("dfGui", QLocalSocket::WriteOnly);
```
nbolton, line 108:
```suggestion
    socket.connectToServer("deskflow", QLocalSocket::WriteOnly);
```

nbolton (commented):
Could we use "deskflow" for the shared mem and socket name instead? 
- Adding "GUI" suggests there are other named shared memory which isn't the case
- The name "dfGui" seems overly cryptic

src/apps/deskflow-gui/main.cpp
sithlord48: I would rather keep the GUI part so we know what it is.. eventually we can do this for the other items as well

src/apps/deskflow-gui/main.cpp
sithlord48: This only will effect the GUI so its named `dfGUI` no one is gonna see this name its internal.  i wanted to avoid using the same tag as the shared memory but I if they can be the same ill unify the names at least

nbolton:
Let's land #8034 first so I can test properly.

nbolton (approved):

PR #8038, opened 2025-01-03
Fix windows debug builds

sithlord48:
- Run `windeployqt` at install time on package stage
 - Prevent package creation when the build is configured for debug


nbolton (commented):
Retrospective approve, seems like the right approach.

PR #8039, opened 2025-01-03
build:  set macOS RPATH for the gui via property

sithlord48:
#8037 now breaks mac os CI builds as they are made w/o rpath this fixes that by defining the rpath in the gui's properties.


nbolton (commented):
Retrospective approve.

Issue #8028, opened 2025-01-01
macos always ask me to allow Deskflow to control my computer


FAAAAT:
### Project

Deskflow

### Deskflow version number

1.18.0

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [ ] Windows
- [X] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [ ] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [ ] I am not using Wayland on Linux

### Signing on macOS

- [X] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### Continuous build

- [ ] I have tried the latest continuous build and the issue persists
- [ ] I am unable to try the latest continuous build

### OS versions/distros

macOS 15.2

### Additional information

I already add it to the white list through system settings. But I still got that alert and can not run deskflow

I already tried remove the old version and reinstall it but doesn't work.

sithlord48:
@FAAAAT Can you try the build for #8042  and see if this works better ? it should have only now one permission entry.

ace56:
Still have this same problem where Deskflow install continuously asks to allow Deskflow to control this computer. Happened on 4 Mac OSX systems w the following during an upgrade to the new version of Deskflow:

Deskflow v 1.20.1 and OSX Sequoia v 15.2 w Open Core Legacy Patcher OCLP 2.2.0 on 2010 or 2011 macbook pro or iMac

Workaround

1) Quit Deskflow so it is not running (In my experience no need to uninstall deskflow or reboot)
2) open System Settings
3) click Privacy & Security
4) click Accessibility
5) Highlight Deskflow and click the "-" button. Type your admin password if it asks
5) download the latest version of Deskflow dmg and remember where it is
7) double click the dmg Deskflow file you just downloaded 
8) drag the Deskflow icon to the applications folder when it asks
9) It will ask to allow Deskflow to control. Click allow. It will open system settings and show Deskflow.
10) new version of Deskflow should work now.

sithlord48:
Its a bug in mac os... not much we can do about it.

sithlord48:
@FAAAAT  Can you confirm this is fixed for you as well i mark this as stale as we have reports of it being fixed.

anerathil:
[...]

PR #8042, opened 2025-01-04
Macos Sign the App with local cert

sithlord48:
TLDR:
 1. We can use "Open Anyway" again
 2. Removes the need for the second Deskflow permission in accessibility 
 3. Fingers crossed No more permission removal and reassignment after "upgrading" Deskflow

Sign the app with a local dummy cert when building. This allows us to "distribute" and have users see the old this is untrusted dialog where they can use "Open Anyway" in setting to bypass it (no longer do they have to use xattr -c as the only method to initially run the application

This also seams to have the added side effect of the accessibility panel only showing ONE Deskflow entry and only asking once not neeed to ask again when connected to a server or when switching to server mode.


sithlord48:
Could help with issues like https://github.com/deskflow/deskflow/issues/8028

PR #8045, opened 2025-01-05
Improve the Action Dialog UX

sithlord48:
a Modified port the updates i made to the Action Dialog in Input Leap
 
- Replace user facing string of `screen` -> `computer`

https://github.com/input-leap/input-leap/pull/1837
https://github.com/input-leap/input-leap/pull/1838
https://github.com/input-leap/input-leap/pull/1830

(FROM IL PR)

This Pr will be used to Improve the UX for the Action Dialog

 Before: 

After Ui:

Key press, toggle and release look the same but different action strings

### DIFF
--- a/src/apps/deskflow-gui/dialogs/ActionDialog.cpp
+++ b/src/apps/deskflow-gui/dialogs/ActionDialog.cpp
@@ -38,6 +38,9 @@ ActionDialog::ActionDialog(QWidget *parent, const ServerConfig &config, Hotkey &
   connect(
       ui->comboActionType, QOverload<int>::of(&QComboBox::currentIndexChanged), this, &ActionDialog::actionTypeChanged
   );
+  connect(ui->listScreens, &QListWidget::itemChanged, this, [&] {
+    ui->buttonBox->button(QDialogButtonBox::Ok)->setEnabled(canSave());
+  });
   connect(ui->buttonBox, &QDialogButtonBox::accepted, this, &ActionDialog::accept);
   connect(ui->buttonBox, &QDialogButtonBox::rejected, this, &ActionDialog::reject);
 
@@ -78,8 +81,7 @@ ActionDialog::ActionDialog(QWidget *parent, const ServerConfig &config, Hotkey &
 
 void ActionDialog::accept()
 {
-  if (!ui->keySequenceWidget->valid() && ui->comboActionType->currentIndex() >= 0 &&
-      ui->comboActionType->currentIndex() < 3)
+  if (!canSave())
     return;
 
   m_action.setKeySequence(ui->keySequenceWidget->keySequence());
@@ -115,10 +117,12 @@ void ActionDialog::accept()
 void ActionDialog::keySequenceChanged()
 {
   ui->listScreens->setEnabled(!ui->keySequenceWidget->keySequence().isMouseButton());
+  ui->buttonBox->button(QDialogButtonBox::Ok)->setEnabled(canSave());
 }
 
 void ActionDialog::actionTypeChanged(int index)
 {
+  ui->buttonBox->button(QDialogButtonBox::Ok)->setEnabled(canSave());
   ui->keySequenceWidget->setVisible(isKeyAction(index));
   ui->groupScreens->setVisible(isKeyAction(index));
   ui->listScreens->setEnabled(!ui->keySequenceWidget->keySequence().isMouseButton());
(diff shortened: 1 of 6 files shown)


nbolton:
I think I can see @p12tic's POV (to a point): https://github.com/input-leap/input-leap/pull/1830#issuecomment-2407218218

Perhaps the resistance there was that it's radically different and that can take some time to get used to.

I'll chat to some end-users and see what they think.

For me personally, I like the new design. 

The old design was kind of a sea of buttons (normally a sign of bad UX).

Old design:


Reminds me of:
image

In my experience, the old design is a result of stuff getting bolted on over time and is a good example of "programmer art".

sithlord48:
> Might be worthwhile to consider changing the language from "screen" to "computer", as this language can become confusing (today it's more common to have more than one monitor/screen per computer).
> 
> Additional things to consider:
> 
>     1. "Screen" appears in other places in the GUI
> 
>     2. If more granular monitor arrangement support is developed later, we may need to revert the language again to monitor/screen/display (if the hotkey functionality is also enhanced).

Yes Agree we should consolidate our use of screen / computer / client so we have a single language

nbolton:
Is it possible to keep the drop down in the same place when changing between actions? It jumps around a bit which is a bit irritating. Maybe the dialog size can be static, or perhaps we can keep the top of the dialog in one place?

sithlord48:
> Is it possible to keep the drop down in the same place when changing between actions? It jumps around a bit which is a bit irritating. Maybe the dialog size can be static, or perhaps we can keep the top of the dialog in one place?

I do not see any jumping around in the placements.

sithlord48:
I tried on 

Window 
Mac oS
Kde plasma
Gnome 

it seams the theme on gnome resizes the dialog from its center.. all others don't.

nbolton:
> I just tried on GNOME , and it seams the theming on GNOME resizes the dialog from its center. If that is what your talking about then that is what the creators of the GNOME theming intend for it to "feel" like when you resize a dialog.

Yes, I tested on KDE, Windows, and macOS. How it's designed right now makes sense for how those DMs resize the dialog; the bottom grows and the title bar stays in the same place, so, the combobox stays in the same place and it feels right.

So, the issue is GNOME as you rightly say. On GNOME it feels jarring and disruptive to the user experience; since the dialog is always centered, when resized the titlebar moves and so the position of the combo box jumps around too.

I love KDE, but most Linux users are on GNOME\*, so surely we can come up with a solution that makes the design work with GNOME too.

As you suggest, we probably can't do much about the dialog position shifting around (that's how GNOME designed it to be), but what if we prevent the dialog from resizing by setting a fixed height?

\*85% of Synergy Linux users are on Ubuntu which defaults to GNOME. Deskflow stats are likely to be very close. Most people don't change the defaults.

sithlord48:
I Don't care about a users bad choice in distro or DE that is not our problem to fix. And you will go crazy trying to ...
This makes a better UX for Windows / Mac and all other DE's that are not GNOME.

nbolton:
Thought you'd say that. You really hate GNOME, don't you? 😄

OK, well in any case the new dialog design is certainly an improvement from what we have now. Maybe in future someone more experienced with UI design than us will come along and point out an obvious solution for the GNOME problem.

I'll do a code review in the morning but it'll probably just be nits. You know me.

sithlord48:
> Thought you'd say that. You really hate GNOME, don't you? 😄
> 
> OK, well in any case the new dialog design is certainly an improvement from what we have now. Maybe in future someone more experienced with UI design than us will come along and point out an obvious solution for the GNOME problem.
> 
> I'll do a code review in the morning but it'll probably just be nits. You know me.

I plan to spend sometime in GNOME to look into what can be done better for a few things on GNOME

nbolton (approved):
Big improvement, and beautifully done. Great work.

I'm unable to see from the perspective of this being a bad UX design or not user friendly. From my perspective, this new design seems much more user friendly.

Approved; only found a couple of minor nits, up to you if you want to correct.

Oh, and it seems to be resizable... did you want that?
image

nbolton:
> Might be worthwhile to consider changing the language from "screen" to "computer"

Good suggestion :+1:

sithlord48:
> Big improvement, and beautifully done. Great work.
> 
> On further reflection, I'm now struggling to see from the perspective of this being a bad UX design or not user friendly. From my perspective, this new design seems much more user friendly than the old design.
> 
> Approved; only found a couple of minor nits, up to you if you want to correct.
> 
> Oh, and it seems to be resizable... did you want that? image

Yes Resize of the dialog is up to the user if they want to see more of the screen list I don't want to block that.

nbolton:
> Modified to only allow resize for actions with the screen list.

Good idea.

nbolton (approved):

PR #8052, opened 2025-01-07
fix: Improve precision of grabbed scroll events on OSX server

sithlord48:
-  recreate https://github.com/deskflow/deskflow/pull/5673 . 
 
 - Long since landed in both barrier and input-leap (and remains unchanged)


sithlord48:
Does not change anything when testing scrolling on mac os   Possibly superseeded by other changes since then.

PR #8077, opened 2025-01-09
Tray menu include show hide

sithlord48:
When testing XFCE i noticed that their tray impl blocks the "activated" state of the tray so its not possible to hide / show by clicking on it (like mac os). This PR fixes this by

 - Add `Open Deskflow` and `Minimize to tray` to the tray menu
 - Added a second action for the tray's Quit action (mac os will move the first and this is incase others do too)
 - Remove the window menu (the hide show happens on window close or via the tray icon now)
 - Remove old code used on mac os to hide the dock icon when the window is not visible (it does not work and our new native code is less code and works) 

fixes: #8095  (it seams not all GNOME Tray exts allow you "activate" either)

### DIFF
--- a/src/apps/deskflow-gui/MainWindow.cpp
+++ b/src/apps/deskflow-gui/MainWindow.cpp
@@ -85,9 +85,10 @@ MainWindow::MainWindow(ConfigScopes &configScopes, AppConfig &appConfig)
       m_actionAbout{new QAction(this)},
       m_actionClearSettings{new QAction(tr("Clear settings"), this)},
       m_actionHelp{new QAction(tr("Report a Bug"), this)},
-      m_actionMinimize{new QAction(tr("&Hide"), this)},
+      m_actionMinimize{new QAction(tr("&Minimize to tray"), this)},
       m_actionQuit{new QAction(tr("&Quit"), this)},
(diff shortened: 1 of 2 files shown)


nbolton:
Is it best if I test this on XFCE?

sithlord48:
I does not matter where you test it the menu should contain the same entries on all platforms

nbolton (approved):
Tested and working well on GNOME:
- Double click will show/hide
- 'Open Deskflow' shows the window
- 'Minimize to tray' hides the window

nbolton:
Actually, I'll also test on KDE, Windows, and macOS before merging.

nbolton:
> but clicking on mac os does not restore the window (and that is normal)

Yeah, that's normal. Clicking the tray icon should only show the menu. 

However, 'Minimize to tray' doesn't change to 'Open Deskflow' any more, so no way to unhide.

nbolton (approved):
Works great now on macOS, as well as GNOME, KDE, and Windows. 
Tested:
- Close app window normally then restore from tray by clicking
- Show/hide app with menu item (it toggles correctly on all OS now)
- Show hide app from clicking tray icon (except for macOS, which is expected)

PR #8086, opened 2025-01-10
fix: Parse only the patch number from version check result

nbolton:


sithlord48 (approved):
ok with me if you have tested it

PR #8094, opened 2025-01-14
docs: Merged the 'OS versions/distros' and 'configuration' fields on the issue template

nbolton:
When reporing bugs, in the configuration field, I noticed people mostly tend to repeat what they said in the OS field. This is telling me that the configuration field is redundant and generates noise. People tend to better describe their setup in the 'Additional information' field.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -94,28 +94,17 @@ body:
     attributes:
       label: OS versions/distros
       description: |
-        Please provide the version number of your operating system (OS).
-        If you're using Linux, please provide the name of the distribution.
+        Please provide the your operating system (OS) and version.
+        If you're using Linux, please tell us the name of the distribution.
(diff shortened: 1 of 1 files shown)


sithlord48:
With the new copy of info in the about dialog we should have them paste that for us

nbolton:
> With the new copy of info in the about dialog we should have them paste that for us

Sure, I was thinking about this. Do we need a different field entirely where they can just paste in from the about dialog on each of their computers? I worry this could be difficult for people to follow, since they might not be signed into GitHub on all of their computers.

Most people manually type stuff in my experience, especially if copy pasting is hard (or if copy/paste is broken in Deskflow). I would prefer they copy/pasted things like this but that's people for you.

sithlord48:
We have ppl just making feature requests in place of bugs.. we can only guide them they have to take the time and fill it out. If they can't provide info for us to help them with then idk what we can do for them.

nbolton:
Sounds like this idea needs to brew a bit more.

PR #8096, opened 2025-01-15
More cmake Simplification

sithlord48:
Changes to the cmake project files

 - Template based RC file generation
 - Create more cmake project level variables and use them thru out the build files
 - Remove the two remaining use of globing for src  `deskflow-gui` and the `gui` library
 - Remove some no longer needed project wide `include_directories`
 - Remove unused `AppDelegate` (mac only only)


src/apps/deskflow-server/CMakeLists.txt
nbolton, line 1: I'm planning to remove the legacy tray icon code (it's way overdue), so I wonder if we might be as well to not include `EXE_ICON`, `EXE_RC_DIALOG`, `EXE_RC_MENU`, etc.

src/apps/deskflow-client/CMakeLists.txt
nbolton, line 1: Same as comment for `deskflow-server`.

nbolton (approved):

src/apps/deskflow-server/CMakeLists.txt
sithlord48: When you remove it you can remove this new code.

sithlord48:
I would say land it now as the code to remove those icons and the switching is not done yet.

PR #8099, opened 2025-01-16
feat: bake in light and dark icon theme

sithlord48:
- bake a simple light and dark icon theme based on kde breeze.
 - Use SVG for all our icons 

Should use system provided icons if they exist if not fallback to our baked in icon set.


nbolton (changes requested):
Secure connection icon doesn't seem to work on Windows.
image

nbolton (changes requested):
On GNOME the secure icon is different to KDE/macOS.
image

nbolton:
> On GNOME the secure icon is different to KDE/macOS.

Though I kind of like the GNOME low-secure icon :D 
image

sithlord48:
It uses the platform icons and ours as a fallback

nbolton (changes requested):
Trash icon not rendering on Windows.

image

Hmm, logo doesn't render either.

image

nbolton:
> It uses the platform icons and ours as a fallback

Yeah I see that now. 

How come the KDE and macOS icons are the same?

Some of the GNOME icons kind of suck... are they legacy?

sithlord48:
If there is no icon it should fall back to out baked in set of breeze
icons  (the kde default)

sithlord48:
Ill have to lookin to that ..

nbolton:
> If there is no icon it should fall back to out baked in set of breeze icons (the kde default)

I feel like the GNOME icons it uses are super legacy... can we not use breeze icons for all platforms? They're the best ones.

sithlord48:
I don't want to override a users picked theme . I realize it's not a thing
on gnome but that is their choice.

nbolton:
> I don't want to override a users picked theme . I realize it's not a thing on gnome but that is their choice.

Got it. Well in that case I'm changing my GNOME icon theme to Breeze! :+1: 

image

sithlord48:
:)

nbolton:
>> Trash icon not rendering on Windows.
> Ill have to lookin to that ..

Interestingly, the file open/browse icons are working on Windows, but no other icons are.

image

sithlord48:
I found the issue the icons windows can not open are sim links so i will need to just make copies of the items in "common" for each theme

sithlord48:
Should be better on windows now. (It was at least in my vm correct. 

@nbolton  you want breeze-dark if you have a "dark " color scheme.

nbolton (approved):
Tested and working with both light and dark mode on: KDE, GNOME, Windows, macOS.

Excellent PR.

PR #8100, opened 2025-01-16
feat: Handle empty update response and improve logging

nbolton:
- Shows warning when empty response back from update server. 
- Improved logging for when new version available/not available.

### DIFF
--- a/src/apps/deskflow-gui/VersionChecker.cpp
+++ b/src/apps/deskflow-gui/VersionChecker.cpp
@@ -48,11 +48,17 @@ void VersionChecker::replyFinished(QNetworkReply *reply)
   reply->deleteLater();
   qDebug("version check response: %s", qPrintable(newestVersion));
 
-  if (!newestVersion.isEmpty() && compareVersions(kVersion, newestVersion) > 0) {
-    qDebug("update found");
+  if (newestVersion.isEmpty()) {
+    qWarning() << "version check is response is empty";
+    return;
+  }
+
+  if (compareVersions(kVersion, newestVersion) > 0) {
+    qWarning().noquote() //
+        << QStringLiteral("current version %1 out of date, update available: %2").arg(kVersion, newestVersion);
     Q_EMIT updateFound(newestVersion);
   } else {
-    qDebug("no updates found");
+    qDebug().noquote() << QStringLiteral("current version %1 is upto date").arg(kVersion);
   }
 }
 


src/apps/deskflow-gui/VersionChecker.cpp
```diff
@@ -63,11 +63,16 @@ void VersionChecker::replyFinished(QNetworkReply *reply)
   const auto newestVersion = QString(reply->readAll());
   qDebug("version check response: %s", qPrintable(newestVersion));
 
-  if (!newestVersion.isEmpty() && compareVersions(kVersion, newestVersion) > 0) {
-    qDebug("update found");
+  if (newestVersion.isEmpty()) {
+    qWarning("version check response is empty");
```
sithlord48, line 67: qWarning () << "version check is response is empty";

src/apps/deskflow-gui/VersionChecker.cpp
```diff
@@ -63,11 +63,16 @@ void VersionChecker::replyFinished(QNetworkReply *reply)
...
+    return;
+  }
+
+  if (compareVersions(kVersion, newestVersion) > 0) {
+    qWarning("current version %s out of date, update available: %s", qPrintable(kVersion), qPrintable(newestVersion));
```
sithlord48, line 72: qWarning() << QStringLiteral("current version %1 out of date, update available: %2").arg(kVersion, newestVersion);

src/apps/deskflow-gui/VersionChecker.cpp
```diff
@@ -63,11 +63,16 @@ void VersionChecker::replyFinished(QNetworkReply *reply)
...
     Q_EMIT updateFound(newestVersion);
   } else {
-    qDebug("no updates found");
+    qDebug("current version %s up to date", qPrintable(kVersion));
```
sithlord48, line 75: qDebug() << QStringLiteral("current version %1 is upto date").arg(kVersion);

sithlord48 (changes requested):
Don't put the strings in the () for qDebug() , qWarning, and qInfo() . Also this can be done better by using one string perhaps 

auto string = QStringLiteral("current version %1").arg (kversion)
if outof date
string.append(QStringLiterla("is out of date") ... 
etc..  before printing to the console.

nbolton:
Thanks for the review, I'll set to draft until I can carve out some time to fix.

nbolton:
Needs finishing. Not very urgent but I will complete.

src/apps/deskflow-gui/VersionChecker.cpp
```diff
@@ -52,11 +52,17 @@ void VersionChecker::replyFinished(QNetworkReply *reply)
   const auto newestVersion = QString(reply->readAll());
   qDebug("version check response: %s", qPrintable(newestVersion));
 
-  if (!newestVersion.isEmpty() && compareVersions(kVersion, newestVersion) > 0) {
-    qDebug("update found");
+  if (newestVersion.isEmpty()) {
+    qWarning() << "version check is response is empty";
```
sithlord48, line 56: don't you want `.noquote()` here ?

sithlord48 (commented):
Do we really need to log its out of date the gui will show this via the icon.

src/apps/deskflow-gui/VersionChecker.cpp
sithlord48: no but i know you dont like to see the quotes hence the suggestion  if your not getting them then no reason to add or change the line.

nbolton: > if your not getting them then no reason to add or change the line.

Correct. When you use a C-style string literal with the stream insertion operator instead of `QStringLiteral`, quotes are not added; quotes are only added when you use a `QStringLiteral` (or `QString`) which I am of course not doing.

This:
```cpp
  qWarning() << "no quotes here :)";
```

Outputs:
```
[2025-03-21T10:20:18] WARNING: no quotes here :)
```

nbolton:
> Do we really need to log its out of date the gui will show this via the icon.

Think about this from a troubleshooting perspective. When a user sends us information, they often send us either a screenshot or a log but rarely both. If we receive only the log, having all the info there saves time for everyone.

Logs aren't just for the user, they're also for people helping the user.

nbolton:
@sithlord48 Ready for your review.

sithlord48 (approved):

PR #8101, opened 2025-01-17
CI: Build Arm Packages for Debian, Fedora and Opensuse

### LINKED ISSUE #7785: Re-implement Linux ARM64 runners with QEMU

sithlord48:
resolve #7785 

 This happened finally https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

#### Building Arm Packages
  - Debian 13
  - Fedora 41
  - Opensuse tumbleweed

#### Not building arm packages: 

 - Windows
    - Github has not released windows arm runners
 - Arch
    - There is no official archarm image for us to build with . dockerhub's newest is over a year old
 - flatpak 
   - All the parts are in place within the CI job
   - The flatpak builder image we use does not support arm.
   - We need to either set up ubuntu with flatpak builder ourselves or wait for an image.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -141,8 +141,22 @@ jobs:
             timeout: 20
             config-args: "-G Ninja -DCMAKE_INSTALL_PREFIX=/usr"
 
+          - name: "debian-13-arm64"
+            runs-on: ubuntu-24.04-arm
+            container: debian:trixie-slim
+            like: "debian"
+            timeout: 20
+            config-args: "-G Ninja -DCMAKE_INSTALL_PREFIX=/usr"
+
           - name: "fedora-41-x86_64"
             runs-on: ubuntu-latest
+            container: fedora:41
+            like: "fedora"
+            timeout: 20
+            config-args: "-G Ninja -DCMAKE_INSTALL_PREFIX=/usr"
+
+          - name: "fedora-41-arm64"
+            runs-on: ubuntu-24.04-arm
             container: fedora:41
             like: "fedora"
             timeout: 20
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -141,27 +141,55 @@ jobs:
             timeout: 20
             config-args: "-G Ninja -DCMAKE_INSTALL_PREFIX=/usr"
 
+          - name: "debian-13-arm64"
+            runs-on: ubuntu-24.04-arm
+            container: debian:trixie-slim
```
nbolton, line 146: Oh, we don't need to use `arm64v8/debian:trixie-slim` any more?

sithlord48: nope

.github/workflows/continuous-integration.yml
```diff
@@ -276,7 +304,17 @@ jobs:
...
+    runs-on: ${{matrix.flatpak.runs-on}}
+    strategy:
+      fail-fast: false
+      matrix:
+        flatpak:
+          - runs-on: ubuntu-latest
+            arch: x86_64
+# Action does not provide arm image
+#          - runs-on: ubuntu-24.04-arm
+#            arch: aarch64
```
nbolton, line 317:
```suggestion
# TODO: Action does not yet provide arm image; re-enable when available
#          - runs-on: ubuntu-24.04-arm
#            arch: aarch64
```

nbolton (approved):
Looks good but I think the comment could be improved.

.github/workflows/continuous-integration.yml
sithlord48: done!

nbolton (approved):

PR #8109, opened 2025-01-19
fix: #8107,  Remove Old tray code

### LINKED ISSUE #8107: Remove legacy tray icon code
### Feature Request

While reviewing #8096, I realised we're dragging around a lot of dead code/RC stuff into new CMake config.

We should remove all legacy tray icon (`[platform]ServerTaskBarReceiver`) code for client and server as it's dead code which is wasting time.

Edit: Dead code
```
MSWindowsClientTaskBarReceiver.cpp
MSWindowsClientTaskBarReceiver.h
OSXClientTaskBarReceiver.cpp
OSXClientTaskBarReceiver.h
XWindowsClientTaskBarReceiver.cpp
XWindowsClientTaskBarReceiver.h
MSWindowsServerTaskBarReceiver.cpp
MSWindowsServerTaskBarReceiver.h
OSXServerTaskBarReceiver.cpp
OSXServerTaskBarReceiver.h
XWindowsServerTaskBarReceiver.cpp
XWindowsServerTaskBarReceiver.h
```

sithlord48:
Fixes #8107 
Removes all old tray related code


nbolton (commented):
I think we should also change the `--no-tray` arg handler and `--help` so it mentions this arg is deprecated and has no effect.

sithlord48:
The tray will still show up for windows just won't change

nbolton (changes requested):
This PR does _not_ fix #8107.

We should remove the core legacy tray entirely. It is not used anymore; we use the GUI tray now instead.

Making changes to `*TaskBarReceiver.cpp` is a waste of time.

To fix fix #8107 we need to remove the dead code:
```
MSWindowsClientTaskBarReceiver.cpp
MSWindowsClientTaskBarReceiver.h
OSXClientTaskBarReceiver.cpp
OSXClientTaskBarReceiver.h
XWindowsClientTaskBarReceiver.cpp
XWindowsClientTaskBarReceiver.h
MSWindowsServerTaskBarReceiver.cpp
MSWindowsServerTaskBarReceiver.h
OSXServerTaskBarReceiver.cpp
OSXServerTaskBarReceiver.h
XWindowsServerTaskBarReceiver.cpp
XWindowsServerTaskBarReceiver.h
```

We should also do this:
> I think we should also change the `--no-tray` arg handler and `--help` so it mentions this arg is deprecated and has no effect.

sithlord48:
When you run the daemon it will show the tray on icon on windows

nbolton:
> When you run the daemon it will show the tray icon on windows for that reason we can not remove the tray icon

I disagree.

The daemon does not have a tray icon (when I created the daemon I knew this would be pointless since it runs in the system context). When we send the client or server command to the daemon via IPC from the GUI, we always use `--no-tray`. If we didn't then we'd have 2 trays. Nobody has used the legacy core tray since like... 2011 (when we switched from the classic GUI to QSynergy).

The legacy tray code needs to go. It's dead code.

sithlord48:
This removes the icons but not those tray classes completely you want them completely removed now ?

nbolton:
I figured maybe a deprecation warning might be more helpful:
```
[2025-01-23T17:15:12] FATAL: deskflow-server: unrecognized option `--no-tray'
Try `deskflow-server --help' for more information.
        /home/nick/Projects/deskflow/src/lib/deskflow/ArgParser.cpp:58
```

i.e. we handle `--no-tray` as before but print something like:
`the --no-tray option is deprecated and will soon be removed`

nbolton (approved):

sithlord48:
> I figured maybe a deprecation warning might be more helpful:
> 
> ```
> [2025-01-23T17:15:12] FATAL: deskflow-server: unrecognized option `--no-tray'
> Try `deskflow-server --help' for more information.
>         /home/nick/Projects/deskflow/src/lib/deskflow/ArgParser.cpp:58
> ```
> 
> i.e. we handle `--no-tray` as before but print something like: `the --no-tray option is deprecated and will soon be removed`
> 
> Thoughts?
I just figured it was easier to remove them then print a message and should not do much harm. we don't ship any scripts or service files externally that would need this as an example

Issue #8106, opened 2025-01-19
Text on macOS about screen is too close together


nbolton:
### Project

Deskflow

### Deskflow version number

1.18.0.95

### Build type

Local developer build (built it myself)

### Operating systems (OS)

- [ ] Windows
- [x] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### OS versions/distros

macOS 14.5

### What steps will reproduce the problem?

Open about screen

Expected: All text is nicely spaced and aligned
Actual: Looks messy; the group label touches the text above

More spacing might be the answer.

Also, the more I think about it, the text not being left aligned with the logo looks messy too.

### Additional information

macOS:
<img width="457" alt="Image" src="https://github.com/user-attachments/assets/65615484-f81d-446c-8751-08c1b5191da1" />

Looks fine on GNOME:
![Image](https://github.com/user-attachments/assets/96cc5ff6-26e6-461e-8391-2a6af8a68e2e)

nbolton:
I know this is just the way it renders on macOS, but the way we're using it is still an issue.

PR #8111, opened 2025-01-20
fixes 8106, by adjusting the spacing of the about dialog

sithlord48:
Fix 8106 adjust vertical spacing. 
Screenshot 2025-01-19 at 11 20 56 PM


nbolton (commented):
Better but I think we could improve by aligning the logo and text on the left.
> Also, the more I think about it, the text not being left aligned with the logo looks messy too. -- #8106

image

nbolton (approved):

Issue #7886, opened 2024-11-12
Everything typed on client is in ALL CAPS (Wayland)


jwakely:
💵 **Bounty**: $200 ([details](https://github.com/deskflow/deskflow/issues/7886#issuecomment-2593078642))

### Project

Deskflow

### Deskflow version number

1.17.1

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [ ] Linux (X11)
- [X] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [X] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [ ] I am not using Wayland on Linux

### Signing on macOS

- [ ] I have authorized the app to run on my Mac
- [X] I am not using macOS

### Continuous build

- [X] I have tried the latest continuous build and the issue persists
- [ ] I am unable to try the latest continuous build

### OS versions/distros

Fedora 40, running KDE/plasma desktop on Wayland. Using Fedora 40 RPMs downloaded from github.com/deskflow/deskflow.
UK keyboard, KDE region settings are for UK.

### Deskflow configuration

Two Fedora 40 systems. The deskflow server is a laptop with an external display attached (so two screens for the server). The client is a tower with a single display attached. Connecting over a direct cat-5 cable between the two machines, with manually assigned IP addresses.

### What steps will reproduce the problem?

Start Deskflow on both systems, click "Allow" in "Input capture requested" and "Remote control requested" dialogs. Move mouse from server to client and type using the server's keyboard. All letters typed are uppercase. Holding Shift doesn't change anything. Turning caps lock on doesn't change it.

Other characters are wrong too:
Typing `[` enters `8` on the client, `]` enters `9`, backslash enters `-` and `|` enters `¬`

This doesn't always happen. After rebooting both systems and restarting deskflow it seems to be OK again for a while. Once it starts happening, no amount of restarting deskflow on server and client helps, nor does logging out of KDE on the client and logging in again and restarting deskflow.

Anything typed directly on the keyboard connected to the server works as normal, only characters typed on the server's keyboard and shared with the client via deskflow are incorrect.

### Log output

```shell
Server logs:

[2024-11-12T14:19:21] INFO: starting core server process (desktop mode)
[...]

PR #8121, opened 2025-01-21
Fix keymap not correctly loading for the Ei backend

BMagnu:
On some systems, the keymap file provided by Ei required to create the XKB context is not properly seeked to the start.
This will cause reading the file to fail (or, more accurately, just cause 0 bytes to be read), causing the keymap to just default to a US-keymap.
This PR fixes this, causing the correct keymap to be loaded.
See input-leap/input-leap#2158

### DIFF
--- a/src/lib/platform/EiKeyState.cpp
+++ b/src/lib/platform/EiKeyState.cpp
@@ -60,6 +60,7 @@ void EiKeyState::init_default_keymap()
 void EiKeyState::init(int fd, size_t len)
 {
   auto buffer = std::make_unique<char[]>(len + 1);
+  lseek(fd, 0, SEEK_SET);
   auto sz = read(fd, buffer.get(), len);
 
   if ((size_t)sz < len) {


nbolton:
I can confirm, on `master` I hit this line on client in `src/lib/platform/EiKeyState.cpp`:
```
LOG_DEBUG("failed to create xkb context: %s", strerror(errno));
```

I was expecting this only to hit only if I had FR layout enabled on the client, but it still happens with GB or US layout. Also tried turning layouts off entirely in KDE and I still see it.

I wonder if this fixes #7886...

nbolton:
To confirm, on KDE client I saw `failed to create xkb context` on `master`. 

I'm thinking this might fix #7886 as the caps issue goes away when I switch to your branch and comes back when I go to master.

jwakely:
> I can confirm, on `master` I hit this line on KDE client in `src/lib/platform/EiKeyState.cpp`:
> 
> ```
> LOG_DEBUG("failed to create xkb context: %s", strerror(errno));
> ```
> 
> This PR appears to fix this.
> 
> Edit: I'm pretty sure this fixes #7886 too...

Great news! 

Should the LOG_DEBUG be a warning or error instead?

nbolton:
> Should the LOG_DEBUG be a warning or error instead?

I made it a warning in #8122

PR #8127, opened 2025-01-23
chore: enable cpack as DEB for deepin|uos

re2zero:
The deepin and uos are debian liked OS, but they do define "ID_LIKE".

Log: enable cpack as DEB for deepin|uos.


sithlord48 (commented):
In order to check this i need you to post the full output of 

/etc/os-release

for each of the distros you are trying to add

sithlord48 (approved):
I wish they would

Issue #8126, opened 2025-01-23
SPDX License: Add missing license files to comply with REUSE Specification


re2zero:
### Feature Request

When running reuse lint to verify SPDX compliance ([REUSE Spec 3.3](https://reuse.software/spec/)), the project shows critical licensing gaps:

```
# SUMMARY
* Missing licenses: MIT, GPL-2.0, GPL-2.0-only
* Files with license information: 17/885 (only 1.9%)
* Compliance status: ❌ Fails REUSE Specification 3.3
```

Key Issues:

    Missing License Files

        Required licenses not found in LICENSES/ directory:
        MIT.txt, GPL-2.0-only.txt, GPL-2.0.txt

    Low License Annotation Coverage

        Only 17/885 files (1.9%) contain SPDX headers

    Copyright Notice Gaps

        188 files lack copyright declarations

Recommended Actions:
✅ Add missing license texts to LICENSES/ directory
✅ Apply SPDX headers to source files (e.g., SPDX-License-Identifier: MIT)
✅ Add copyright notices using SPDX-FileCopyrightText: format
🔧 Run reuse download --all to fetch standard license texts

Verification:
After fixes, run:

`reuse lint`

Target output:
Congratulations! Your project is compliant with version 3.3 of the REUSE Specification :-)

References:
    [REUSE Quick Start](https://reuse.software/tutorial/)
    [SPDX License List](https://spdx.org/licenses/)

Let me know if I can help prepare the license files or SPDX headers!

sithlord48:
Yes thanks for pointing that our we are NOT reuse compliant at this time.. its on my list of things todo thanks for making a task.

PR #8134, opened 2025-01-24
refactor: move License info to LICENSES folder

sithlord48:
Baby step towards #8126  , move our license info to where it will be after compliance. 
Since we only need to install the license for our code this will be all the changes needed for installs.


sithlord48:
replaced by #8135

PR #8098, opened 2025-01-15
Add FingerPrint DB

varshith257:
Part of #7806 

Blocks: #7931

PR #8139, opened 2025-01-25
chore: add missing String methods

sithlord48:
Clean up some of our String methods
 

- `String.cpp` does not need the access to `ARCH` 
- modified `void String::toHex(...)` -> `std::string String::toHex(...)` will now return a std::string;
- added `int String::fromHexChar(char c)`
- added `std::vector<uint8_t> string::fromHex(const std::string&)`
- Update copyright to SPDX format

#8098  Should base upon this and remove modifications to the `String` Class

### DIFF
--- a/src/lib/base/String.cpp
+++ b/src/lib/base/String.cpp
@@ -16,7 +16,6 @@
  */
 
 #include "base/String.h"
-#include "arch/Arch.h"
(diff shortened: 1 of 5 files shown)


nbolton (approved):
LGTM, can't test today due to technical issue.

PR #8145, opened 2025-01-28
chore: reset screen xywh on update_shape

sithlord48:
based on https://github.com/input-leap/input-leap/pull/2167


nbolton:
I noticed a bug while testing this, but it also happens on master: #8147

nbolton (approved):
Appears to have no negative effects, but I'm unsure how or unable to test the positive scenario (due to bug #8147).

PR #8149, opened 2025-01-29
Revert "feat!: Add support for FingerPrintDatabase"

sithlord48:
This reverts commit 8f219b735e44056edf7ef2529a8948e35592869d.

 something has slipped thru testing and this breaks tls connections.


sithlord48:
Landing w/o Review to prevent breakage for ppl . Added the commit to #7931 so we can test it more

nbolton:
I agree. We should not have landed #8098, that was a mistake.

PR #8160, opened 2025-01-30
Added mention of lack of Wayland clipboard-sharing support

edgan:
I just discovered Deskflow, and I am coming from input-leap. I found the [README.md](https://github.com/deskflow/deskflow/blob/master/README.md) very misleading, as is. This is to make it crystal clear the status quo of clipboard-sharing for Wayland.

 I switched with the hope that Deskflow would have Wayland clipboard-sharing support. I am staying for now. Deskflow seems to be a little slicker. It also doesn't crash for me on Fedora 41, unlike input-leap.


sithlord48:
I would rather you link / mention the known wayland issues https://github.com/deskflow/deskflow/discussions/7499
instead of linking a specific issue.

README.md
```diff
@@ -154,7 +154,8 @@ free to [start a discussion](https://github.com/deskflow/deskflow/discussions) o
 ### Is clipboard sharing supported?
 
 Absolutely. The clipboard-sharing feature is a cornerstone feature of the product and we are 
-committed to maintaining and improving that feature.
+committed to maintaining and improving that feature. Yet it is not yet support for Wayland.
+See [#8031](https://github.com/deskflow/deskflow/issues/8031).
```
sithlord48, line 158: Please do not link the issue We will remove this note when the time comes.  I do not want to link any specific issues since its possible this will be closed and the clipboard still not work.

README.md
```diff
@@ -167,6 +168,8 @@ who is the author of `libei`, a major contributor to `libportal`, and the author
 implementation in Deskflow. Others such as Olivier Fourdan and Povilas Kanapickas helped with the
 Wayland implementation.
 
+See above where lack of clipboard-sharing is mentioned.
```
sithlord48, line 171: > [!Note]
> wayland is still evolving please consult the https://github.com/deskflow/deskflow/discussions/7499.

sithlord48 (changes requested):

README.md
edgan: I very much disagree with not linking to the issue. Just telling people it isn't supported is very incomplete information. Yes, people are more likely to come make noise in the issue, but that is the nature of the beast.

Giving the link lets people track with a subscribe to the issue for emails, and follow the progress as it is made. It could also encourage donations. Which is definitely a good thing. In addition the delay between updates to the issue, and updates to the `README.md` that users will likely prefer the former.

README.md
edgan: Not sure what you are trying to say here. Are you suggesting a text change, or trying to inform me?

README.md
nbolton: Since clipboard support needs to be implemented in upstream libs, how about my suggestion: https://github.com/deskflow/deskflow/pull/8160#discussion_r1937037561

nbolton:
Thanks for the PR, I hope my review was helpful!

README.md
sithlord48: yes i am suggesting you use this as the text.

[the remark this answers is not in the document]
sithlord48: This is where the text should be put.

README.md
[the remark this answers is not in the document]
sithlord48: This should be in the wayland section not the copy and paste supported section.

sithlord48: We do not want to track specific Issues in the read me as they close and change quicker then we update the read me also we will have issues on wayland for a while so tracking them all in one place is more sane then linking a bunch of issues from the read me.

nbolton:
@sithlord48 Is this PR superseded by #8163?

PR #8162, opened 2025-02-01
Gui cleanup.server config

sithlord48:
- Fix: Bug in 1.19.0 where the add client dialog does not show up 
 - Use the naming convention 
 - DO NOT use auto-connections
 - Cleanup the existing slots.
 - Add icons for `list-add` `list-remove` and `document-edit`


src/apps/deskflow-gui/dialogs/ServerConfigDialog.ui
nbolton, line 1: The help text looks a bit odd now that it's centered. Either the title should be centered as before, or the help text should be left aligned.

image

nbolton (changes requested):

nbolton:
> * DO NOT use auto-connections

:+1:

src/apps/deskflow-gui/dialogs/ServerConfigDialog.ui
sithlord48: done

nbolton (approved):

PR #8199, opened 2025-02-14
chore: update to fancy checkout 1.0.1

sithlord48:
Update fancy checkout to 1.0.1

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -243,7 +243,7 @@ jobs:
       # Fancy checkout gets all the tags
       # it also makes sure we can use git --describe correctly
       - name: Fancy Checkout
-        uses: sithlord48/fancy-checkout@v1.0.0
+        uses: sithlord48/fancy-checkout@v1.0.1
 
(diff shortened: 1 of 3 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -243,7 +243,7 @@ jobs:
       # Fancy checkout gets all the tags
       # it also makes sure we can use git --describe correctly
       - name: Fancy Checkout
-        uses: sithlord48/fancy-checkout@v1.0.0
+        uses: sithlord48/fancy-checkout@v1.0.1
```
nbolton, line 246: For GitHub actions, it is conventional to target `v1`, which should point to the latest patch (i.e. `v1.0.1`). This way you don't need to update our CI every time you patch something in `fancy-checkout` action. You simply point the `v1` tag at the latest `v1.x.x` tag. Locking to a particular patch is not normally a good idea unless you want to use an older patch as a temporary workaround.

```suggestion
        uses: sithlord48/fancy-checkout@v1
```

nbolton (changes requested):

PR #8209, opened 2025-02-17
Clean up `MainWindow` spacing in header and source

nbolton:
~~A static inline member makes sense for the TLS fingerprint path as it only needs to be determined once.~~

PR #8211 supersedes some of the commits in this PR, but the spacing fixes are still relevant.

It also adds debug messages related to peer database read/write.


src/apps/deskflow-gui/MainWindow.h
```diff
@@ -186,11 +187,20 @@ class MainWindow : public QMainWindow
   deskflow::gui::TlsUtility m_tlsUtility;
   QSize m_expandedSize = QSize();
 
+  //
```
sithlord48, line 190: Is a single line not enough .. Do i really need a comment to tell me this is a UI file ?

src/apps/deskflow-gui/MainWindow.h
```diff
@@ -186,11 +187,20 @@ class MainWindow : public QMainWindow
...
+  // Qt UI
+  //
+  std::unique_ptr<Ui::MainWindow> ui;
+
+  //
```
sithlord48, line 195: Does * not tell me these are pointers ?

src/apps/deskflow-gui/MainWindow.h
```diff
@@ -165,18 +161,23 @@ class MainWindow : public QMainWindow
   void updateStatus();
   void showAndActivate();
 
-  QString getTlsPath();
-
-  // Generate prints if they are missing
-  // Returns true if successful
+  /**
+   * @brief Generate prints if they are missing.
+   *
```
sithlord48, line 166: Remove this space from the doxygen comment it will be ignored anyway.

src/apps/deskflow-gui/MainWindow.cpp
```diff
@@ -1072,12 +1069,6 @@ void MainWindow::showAndActivate()
   m_actionMinimize->setVisible(true);
 }
 
-QString MainWindow::getTlsPath()
```
sithlord48, line 1075: we will need this method later to get us the server and client paths.

src/lib/gui/CMakeLists.txt
```diff
@@ -54,8 +54,8 @@ add_library(${target} STATIC
   core/CommandProcess.h
   core/CoreProcess.cpp
   core/CoreProcess.h
-  core/CoreTool.cpp
-  core/CoreTool.h
+  core/core_utils.cpp
```
sithlord48, line 57: why do you need to change this file name ?
tell me WHO cares  and how this file being in snake_case is something I as a NEW developer would even know ? Please do not change the name of this file.

src/lib/gui/core/core_utils.cpp
```diff
@@ -1,11 +1,11 @@
 /*
  * Deskflow -- mouse and keyboard sharing utility
  * SPDX-FileCopyrightText: (C) 2024 Chris Rizzitello <sithlord48@gmail.com>
- * SPDX-FileCopyrightText: (C) 2015 Symless Ltd.
+ * SPDX-FileCopyrightText: (C) 2015 - 2025 Symless Ltd.
```
sithlord48, line 4: Did you make NON trivial changes Every year since 2015 on THIS file ? if not then you use the form 

`2015, 2025` for the date.

src/lib/gui/core/core_utils.h
```diff
@@ -0,0 +1,17 @@
...
+ * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2015 - 2025 Symless Ltd.
+ * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
+ */
+
+#pragma once
+
+#include <QString>
+
+namespace deskflow::gui::core_utils {
```
sithlord48, line 11: Why do need to do this we do NOT need over qualified namespaces. also this would be in
`deskflow::net` or `deskflow::tls`  since that is what this covers... 

And Why is it snake_case if our vars are camelCase that is not consistent.. NO on will know if its a class or not that is why this snake_case naming for some files is not gonna be good for new devs.

sithlord48 (changes requested):
I wrote something simiar but forgot to push it. 

I do not think a const is good here as the user can change the path of the certs. in the settings if the user changes the path this will not Update not the new path.

src/lib/gui/CMakeLists.txt
nbolton: It's our code style. I am happy to talk bout your idea of changing our code style in the appropriate discussion: #8188

If we decide to change the code style, we should apply it universally.

> tell me WHO cares

Apparently, we both care a great deal.

src/lib/gui/core/core_utils.cpp
nbolton: That feels a bit much—I’ve maintained this code for decades, and listing every single year I made changes seems unnecessary. I haven’t seen that done in a copyright header before. I don't see what value that would add.

nbolton: By that logic, we wouldn't use years, but exact dates instead. I didn't make changes every day for the whole year. I think people understand this. It's common sense.

src/apps/deskflow-gui/MainWindow.cpp
nbolton: No worries, I'll set this to draft until #7931 has landed.

src/apps/deskflow-gui/MainWindow.h
nbolton: Intent: To prevent it from getting lost among the rest of the members. 

Do you have any alternative ideas?

src/apps/deskflow-gui/MainWindow.h
nbolton: Sure. Will do later.

src/lib/gui/CMakeLists.txt
sithlord48: This is bad style, simply because no developer cares if its a class or not they just want the include making it the style of the file name different just makes it harder to know what the include will be.

src/lib/gui/core/core_utils.h
nbolton: Ah, let's go with: `deskflow::gui::core` as we already use that.

https://github.com/deskflow/deskflow/blob/7724178fa69e4c0bffc6fecaf3f65a65593de188/src/lib/gui/core/WaylandWarnings.cpp#L13

sithlord48: can we just use deskflow... if must put it in a namespace.. i only have so many chars allowed in a line and this is about 1/5 of them once i call a method w/ that namespace..

nbolton:
I believe this is probably superseded now.

PR #8215, opened 2025-02-18
Create GitHub recommended security policy file

nbolton:
On the [security page](https://github.com/deskflow/deskflow/security), the 'Set up a security policy' button creates a file called `SECURITY.md`:
image


nbolton:
> Not sure what if anything else needs to be in here but a start of a policy in this case is better then no policy.

Yeah it did feel a bit light.

Issue #8010, opened 2024-12-26
Pentest to verify TLS CVE fixes


nbolton:
💵 **Bounty:** $500 ([details](https://github.com/deskflow/deskflow/issues/8010#issuecomment-2562857625))

**Blocker:** 
- #7806

**Task:** Penetration Tester/Ethical Hacker/security expert wanted to pentest Deskflow and check the fixes for TLS CVEs:
- CVE-2021-42072
- CVE-2021-42073
- CVE-2021-42074
- CVE-2021-42075
- CVE-2021-42076

We have a bounty for fixing #7806. A number of PRs were opened to fix this. We want to verify these PRs do indeed solve the CVEs so that we can award the bounty.

CyberPro010:
how can i participate ? i want payment in bitcoin or eth

nbolton:
> how can i participate ? i want payment in bitcoin or eth

Right now, @algora-pbc does not support cryptocurrency. If nobody accepts the bounty within the next week or so, perhaps we can talk on Matrix or IRC: https://github.com/deskflow/deskflow/wiki/Chat-with-us

ologbonowiwi:
I'm OOO for the holidays but I'd love to try this out.

What is the desired output here? code attempting to explore the mentioned vulnerabilities?

sithlord48:
Correct we would want pen testing done to show that these are resolved. Perhaps by running the exploit on a known bad version 1.17.2 and then a suspected good version 1.18.0. Ideally this should be something we (or another 3rd party) could also try locally, Maybe even as a test run at build time to make sure we do not regress 

Please note: that  @varshith257 has a bit to finish before the first two CVEs on the list are resolved. These should be resolved with 1.18.0 
 - [CVE-2021-42074](https://github.com/advisories/GHSA-97hp-rfmm-2c7w)
 - [CVE-2021-42075](https://github.com/advisories/GHSA-c58c-q645-27mc)
 - [CVE-2021-42076](https://github.com/advisories/GHSA-g73p-h925-4x6w)

zelosleone:
@sithlord48 @nbolton since prs got merged, is it okay to start?

sithlord48:
YES!

nbolton:
@mgerstner Perhaps also of interest to you.

zelosleone:
[...]

PR #8226, opened 2025-02-19
Pentesting files with Python and Nuclei

zelosleone:
/claim #8010

Mentioned CVE's were:
CVE-2021-42072
CVE-2021-42073
CVE-2021-42074
CVE-2021-42075
CVE-2021-42076

None of them exists in Deskflow versions from 1.70 to 1.90. The only problem that exists, and is not solved in new versions was this sonar scanning result: https://sonarcloud.io/project/issues?open=AZH_1rPbUm1X5WafpxUI&id=deskflow_deskflow

It still accepts the client even if they are nonsense, and Deskflow logs in the Deskflow shows "accepted client" and accepts the socket as well. Though, it doesnt let you execute commands at all, so these 5 CVE's are not affecting deskflow.

To run the nuclei templates use command:
nuclei -t ****absolute path to file here**** -u localhost:24800 -debug -v

Python scripts are obvious.

Also i have to mention running file descriptor and hello client requests back to back in one single python script breaks deskflow to the point you have to reinstall it.

If there is somethings you guys want me to check, or modify in my scripts and templates, would be glad to hear them.


sithlord48:
We do not allow merge commits in our history so please rebase and do not correct pending commits with additional new commits, 
We also moved all our python optional script to our [scripts repo](https://github.com/deskflow/scripts) perhaps this would be better as a pr there

zelosleone:
> We do not allow merge commits in our history so please rebase and do not correct pending commits with additional new commits, We also moved all our python optional script to our [scripts repo](https://github.com/deskflow/scripts?rgh-link-date=2025-02-19T16%3A57%3A04Z) perhaps this would be better as a pr there

Oh alright, i will be opening the pr there then.

PR #8233, opened 2025-02-20
Make a proper status area for the mainwindow

sithlord48:
- Setup a proper status area in the main window
 - track when we should reset the server to "Listening" from connected. 
 - Remove redunant client and server status labels
 - remove unused `lblNotify`

Update:  
- Added "Update available" to the update button
- Added a commit to use QStringLiteral in the `lib/gui/consts.h`
 
Screenshot_20250219_195333


nbolton (changes requested):
This needs to be more noticeable to ensure users don’t miss the update (especially less experienced users who aren't tuned into these things). As well as the tooltip, consider adding the text "Update available" or "Update" next to the icon. This change is important for visibility. Let’s adjust it before landing.

image

sithlord48:
> This needs to be more noticeable to ensure users don’t miss the update (especially less experienced users who aren't tuned into these things). As well as the tooltip, consider adding the text "Update available" or "Update" next to the icon. This change is important for visibility. Let’s adjust it before landing.
> 
> image

My thought about this icon was to things. 
  1. I was worried the old text would make the main window grow in width to fit the longer strings.
  2. I was hoping a new user seeing a new icon would at least have them hover over it to see what it is making its function naturally discoverable without the need the need for a label.

src/apps/deskflow-gui/MainWindow.cpp
```diff
@@ -445,6 +461,11 @@ void MainWindow::openHelpUrl() const
   QDesktopServices::openUrl(QUrl(kUrlHelp));
 }
 
+void MainWindow::openGetNewVersionUrl() const
+{
+  QDesktopServices::openUrl(QStringLiteral("https://github.com/deskflow/deskflow/releases/latest"));
```
nbolton, line 466: URLs should be constants so we can keep the URLs consistent across the app. For consistency, we should use `deskflow.org` as we do for other URLs.

```suggestion
  QDesktopServices::openUrl(kUrlDownload);
```

This will produce the URL: `https://deskflow.org/download?source=gui`
```patch
diff --git a/src/lib/gui/constants.h b/src/lib/gui/constants.h
index d6add12a..5825dd37 100644
--- a/src/lib/gui/constants.h
+++ b/src/lib/gui/constants.h
@@ -18,6 +18,7 @@ const auto kOrgDomain = QStringLiteral("deskflow.org");
 const auto kUrlSourceQuery = "source=gui";
 const auto kUrlApp = QStringLiteral("https://deskflow.org");
 const auto kUrlHelp = QString("%1/help?%2").arg(kUrlApp, kUrlSourceQuery);
+const auto kUrlDownload = QString("%1/download?%2").arg(kUrlApp, kUrlSourceQuery);
 
 #if defined(Q_OS_LINUX)
 const auto kUrlGnomeTrayFix = "https://extensions.gnome.org/extension/615/appindicator-support/";
```

src/lib/gui/constants.h
```diff
@@ -15,12 +15,9 @@ namespace deskflow::gui {
 // notation (rdn), e.g. org.deskflow
 const auto kOrgDomain = QStringLiteral("deskflow.org");
 
-const auto kLinkDownload = R"(<a href="%1" style="color: %2">Download now</a>)";
-
 const auto kUrlSourceQuery = "source=gui";
 const auto kUrlApp = QStringLiteral("https://deskflow.org");
 const auto kUrlHelp = QString("%1/help?%2").arg(kUrlApp, kUrlSourceQuery);
-const auto kUrlDownload = QString("%1/download?%2").arg(kUrlApp, kUrlSourceQuery);
```
nbolton, line 23: Please revert this change. Embedding URLs all over the app makes it harder to keep them consistent.

src/lib/gui/core/CoreProcess.cpp
nbolton, line 1: Please revert this change or find a way to do it without counting connections.

I tried this before, and it doesn't work. When you close the GUI on Windows, connections are still able to drop or connect (because it's running in the background via the daemon). When you open the GUI again, it will be out of sync with reality and won't make sense.

The only way to do this properly is to use IPC to query the server so we know for sure how many connections there actually are.

nbolton:
> * I was worried the old text would make the main window grow in width to fit the longer strings.
> * I was hoping a new user seeing a new icon would at least have them hover over it to see what it is making its function naturally discoverable without the need the need for a label.

I see your point about keeping the window compact and leveraging hover for discoverability. However, for accessibility, text is crucial. Not everyone can easily interpret icons (especially small ones), and colorblind users may miss subtle visual cues. Adding the text 'Update Available' makes it more readable and inclusive for all users.

If width is the concern, we could test a slightly abbreviated version like 'Update'.

Let's make this adjustment before merging to ensure it's accessible to everyone.

src/apps/deskflow-gui/MainWindow.cpp
sithlord48: i don't see any reason to even use that abstraction. We only use that link once and if you do go to 
deskflow.org/download?source=gui its the release page not release latest. The update will only show on new stable release and they will always be our latest tag.

src/lib/gui/constants.h
sithlord48: its used in one place , one time and a static there hiding things here makes it harder to find what this value means. also this string is way less efficient then a single QStringLiteral use.

src/lib/gui/core/CoreProcess.cpp
sithlord48: If i am reading this correctly then the main problem  here is if the gui was run and the daemon had a connection. 

If that is the case our current server label should suffer from the same issue ?

nbolton: Currently, it's not an issue because--for better or worse--when you start the GUI on Windows, it tells the daemon to restart the process. Actually, thinking about it, wouldn't the behavior solve the counting out of sync issue I described?

I'll do some experimenting and see if my knowledge of my previous attempt is inaccurate or perhaps there's a different reason why counting might not work.

In any case, ultimately IPC is the best solution.

src/lib/gui/constants.h
nbolton: > its used in one place

Even if it's currently used once, future updates or features might require the same URL. Centralizing now prevents duplication later. It also keeps the pattern consistent across the codebase, making it easier for new contributors to understand the structure.

> hiding things here makes it harder to find

Using descriptive constants like kUrlDownload makes the code self-documenting. The name itself conveys the intent, which is more meaningful than hardcoding the URL directly. 

In modern development environments, navigating to a constant's definition is just a click away. The concern about "hiding things" is outdated when tools like IntelliJ, VSCode, and Qt Creator make it trivial to trace constants back to their source.

I’ve encountered broken links in the past because URLs were scattered throughout the codebase. Without a centralized approach, updates were made inconsistently, leading to outdated or mismatched links. Centralizing URLs enforces a discipline that prevents this issue, ensuring that when one link is changed, all related links are reviewed and updated together, maintaining consistency and reliability across the code.

> string is way less efficient then a single QStringLiteral use

The efficiency difference is negligible in this context because URLs are typically static and only initialized once. Maintainability should take priority over micro-optimizations. If performance is a concern, a compromise could be using QStringLiteral but still centralizing the constants.

src/lib/gui/core/CoreProcess.cpp
sithlord48: yes ipc is the best solution (and i think that is that is for another pr its more involved) 

This one just moves the counting from the status widget to the core process so its the same behavior we had before but now the connection message changes when you disconnect that is what im trying to confirm

nbolton: Oh really? OK, will test it thoroughly tomorrow to see if I can break it.

nbolton (approved):

PR #8251, opened 2025-02-26
No stop on print check

sithlord48:
- Fixup the message for when a fingerprint db does not yet exist
 - Do not stop the server when a new fingerprint is connecting
 - Do not show the fingerprint dialog for a print we are all ready checking.


src/apps/deskflow-gui/MainWindow.cpp
```diff
@@ -777,6 +777,7 @@ void MainWindow::checkFingerprint(const QString &line)
     return;
   }
 
+  static QString checkFor;
```
nbolton, line 780: I suspect that multiple clients connecting will break this mechanism. I will check.

nbolton: I would also add that in C++, relying on static variables goes against the grain of object-oriented design and breaks encapsulation. While they’re useful for shared constants or utility functions, they introduce hidden dependencies and global state, making code harder to maintain and test. Unlike C, where statics are a common pattern due to the lack of objects, using statics in C++ is more often than not, a code smell.

src/lib/net/SecureSocket.cpp
```diff
@@ -677,7 +677,9 @@ bool SecureSocket::verifyCertFingerprint(const deskflow::fs::path &fingerprintDb
   db.read(fingerprintDbPath);
   if (!db.fingerprints().empty()) {
     LOG((CLOG_NOTE "read %d fingerprints from %s", db.fingerprints().size(), fingerprintDbPath.c_str()));
-  } else {
+  }
+
+  if (file.good() && db.fingerprints().empty()) {
     LOG((CLOG_ERR "failed to open trusted fingerprints file: %s", fingerprintDbPath.c_str()));
     return false;
   }
```
nbolton, line 685: An early return with two different error messages would make more sense (see earlier suggestion).

```suggestion
```

nbolton (changes requested):

src/apps/deskflow-gui/MainWindow.cpp
```diff
@@ -786,16 +787,16 @@ void MainWindow::checkFingerprint(const QString &line)
       deskflow::fingerprintTypeToString(deskflow::FingerprintType::SHA256),
       deskflow::string::fromHex(match.captured(2).toStdString())
   };
+  if (checkFor == match.captured(2))
+    return;
```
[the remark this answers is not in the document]
nbolton: I updated my comment. I maintain my position about silent returns. They hide problems.

nbolton:
> No stop on print check

A more helpful title would be:
- Remove Core stop/start for fingerprint dialog

nbolton:
> The fingerprint dialog should have an 'Ignore' button where you can ignore the client for at least the duration of the running GUI process. The `QList` solution above would also solve this.

Any thoughts on changing "No" to "Ignore" to make the purpose of the button clearer?

src/apps/deskflow-gui/MainWindow.h
```diff
@@ -209,6 +209,8 @@ class MainWindow : public QMainWindow
   QToolButton *m_btnFingerprint = nullptr;
   QPushButton *m_btnUpdate = nullptr;
 
+  QStringList m_checkedPeers;
```
nbolton, line 212: Please don't add members with random spacing. Group them together.
```suggestion
```

The members in this class declaration are driving me mad. It's so disorganized.

Here's how I want it to look:
```cpp
  inline static const auto m_guiSocketName = QStringLiteral("deskflow-gui");

  VersionChecker m_versionChecker;
  bool m_secureSocket = false;
  deskflow::gui::config::ServerConfigDialogState m_serverConfigDialogState;
  bool m_saveOnExit = true;
  deskflow::gui::core::WaylandWarnings m_waylandWarnings;
  deskflow::gui::ConfigScopes &m_configScopes;
  AppConfig &m_appConfig;
  ServerConfig m_serverConfig;
  deskflow::gui::CoreProcess m_coreProcess;
  deskflow::gui::ServerConnection m_serverConnection;
  deskflow::gui::ClientConnection m_clientConnection;
  deskflow::gui::TlsUtility m_tlsUtility;
  QSize m_expandedSize = QSize();
  QStringList m_checkedPeers;
  QSystemTrayIcon *m_trayIcon = nullptr;
  QLocalServer *m_guiDupeChecker = nullptr;
  QLabel *m_lblSecurityStatus = nullptr;
  QLabel *m_lblStatus = nullptr;
  QToolButton *m_btnFingerprint = nullptr;
  QPushButton *m_btnUpdate = nullptr;

  // Window Actions
  QAction *m_actionAbout = nullptr;
  QAction *m_actionClearSettings = nullptr;
  QAction *m_actionReportBug = nullptr;
  QAction *m_actionMinimize = nullptr;
  QAction *m_actionQuit = nullptr;
  QAction *m_actionTrayQuit = nullptr;
  QAction *m_actionRestore = nullptr;
  QAction *m_actionSave = nullptr;
  QAction *m_actionSettings = nullptr;
  QAction *m_actionStartCore = nullptr;
  QAction *m_actionStopCore = nullptr;
  QAction *m_actionTestCriticalError = nullptr;
  QAction *m_actionTestFatalError = nullptr;
```

As a patch:
```diff
diff --git a/src/apps/deskflow-gui/MainWindow.h b/src/apps/deskflow-gui/MainWindow.h
index ef279637..bf5422c8 100644
--- a/src/apps/deskflow-gui/MainWindow.h
+++ b/src/apps/deskflow-gui/MainWindow.h
@@ -185,12 +185,13 @@ private:
   // Returns true if successful
   bool regenerateLocalFingerprints();
 
+  inline static const auto m_guiSocketName = QStringLiteral("deskflow-gui");
+
   VersionChecker m_versionChecker;
   bool m_secureSocket = false;
   deskflow::gui::config::ServerConfigDialogState m_serverConfigDialogState;
   bool m_saveOnExit = true;
   deskflow::gui::core::WaylandWarnings m_waylandWarnings;
-
   deskflow::gui::ConfigScopes &m_configScopes;
   AppConfig &m_appConfig;
   ServerConfig m_serverConfig;
@@ -199,18 +200,14 @@ private:
   deskflow::gui::ClientConnection m_clientConnection;
   deskflow::gui::TlsUtility m_tlsUtility;
   QSize m_expandedSize = QSize();
-
+  QStringList m_checkedPeers;
   QSystemTrayIcon *m_trayIcon = nullptr;
   QLocalServer *m_guiDupeChecker = nullptr;
-  inline static const auto m_guiSocketName = QStringLiteral("deskflow-gui");
-
   QLabel *m_lblSecurityStatus = nullptr;
   QLabel *m_lblStatus = nullptr;
   QToolButton *m_btnFingerprint = nullptr;
   QPushButton *m_btnUpdate = nullptr;
 
-  QStringList m_checkedPeers;
-
   // Window Actions
   QAction *m_actionAbout = nullptr;
   QAction *m_actionClearSettings = nullptr;
```

src/apps/deskflow-gui/MainWindow.cpp
```diff
@@ -782,20 +782,24 @@ void MainWindow::checkFingerprint(const QString &line)
...
+  }
+  m_checkedPeers.append(sha256Text);
 
   // Only Save the sha256
   deskflow::FingerprintDatabase db;
   db.read(trustedFingerprintDb().toStdString());
   if (db.isTrusted(sha256)) {
     return;
   }
```
nbolton, line 802: > Lack of spacing here makes code harder to read.

```suggestion
  }
  
```

src/lib/net/SecureSocket.cpp
```diff
@@ -675,13 +675,17 @@ bool SecureSocket::verifyCertFingerprint(const deskflow::fs::path &fingerprintDb
...
+  const bool emptyDB = db.fingerprints().empty();
+
+  if (file.good() && emptyDB) {
     LOG((CLOG_ERR "failed to open trusted fingerprints file: %s", fingerprintDbPath.c_str()));
     return false;
   }
 
+  if (!emptyDB) {
+    LOG((CLOG_NOTE "read %d fingerprints from %s", db.fingerprints().size(), fingerprintDbPath.c_str()));
+  }
```
nbolton, line 687: Your latest change makes for confusing reading.

Please use simple early returns instead, like this:
```suggestion
  if (!file.good()) {
    LOG((CLOG_ERR "failed to open trusted fingerprints file: %s", fingerprintDbPath.c_str()));
    return false;
  }

  if (db.fingerprints().empty()) {
    LOG((CLOG_ERR "trusted fingerprints file is empty: %s", fingerprintDbPath.c_str()));
    return false;
  }

  LOG((CLOG_NOTE "read %d fingerprints from: %s", db.fingerprints().size(), fingerprintDbPath.c_str()));
```

nbolton (changes requested):

src/lib/net/SecureSocket.cpp
sithlord48: This is wrong is not an error if we don't have the file that is the reason i changed the code around.

src/apps/deskflow-gui/MainWindow.h
sithlord48: I am not going to move unreleased lines.

src/lib/net/SecureSocket.cpp
sithlord48: if the file exists and its empty thats is a problem 
if the file does not (!file.good()) is not an error we just didn't make it yet
its only a problem if the files exists and is empty

nbolton: I see. In that case, please change the error message to reflect reality:
```cpp
  if (file.good() && emptyDB) {
    LOG((CLOG_ERR "trusted fingerprints file is empty: %s", fingerprintDbPath.c_str()));
    return false;
  }
```

sithlord48: if the file is good (i.e it exists) but our database is empty after trying to read that is a read file error. There is no way for this to happen otherwise.

nbolton: OK, understood.

src/apps/deskflow-gui/MainWindow.h
```diff
@@ -184,16 +184,18 @@ class MainWindow : public QMainWindow
    */
   QString trustedFingerprintDb();
 
+  inline static const auto m_nameRegEx = QRegularExpression(QStringLiteral("^[\\w\\-_\\.]{0,255}$"));
+
   // Generate prints if they are missing
   // Returns true if successful
   bool regenerateLocalFingerprints();
+  inline static const auto m_guiSocketName = QStringLiteral("deskflow-gui");
```
nbolton, line 192: Please group static member variables together. The `m_nameRegEx` var needs to be moved down under the member functions. We should group it with `m_guiSocketName`.

Also, lack of space makes `regenerateLocalFingerprints` and `m_guiSocketName` appear related, which they are not.

```suggestion
  // Generate prints if they are missing
  // Returns true if successful
  bool regenerateLocalFingerprints();
  
  inline static const auto m_nameRegEx = QRegularExpression(QStringLiteral("^[\\w\\-_\\.]{0,255}$"));
  inline static const auto m_guiSocketName = QStringLiteral("deskflow-gui");
```

nbolton (changes requested):
You're almost there but your last change didn't completely solve the class decl chaos.

nbolton (approved):

PR #8275, opened 2025-03-01
fix: Do stop the client on fingerprint dialog

### LINKED ISSUE #8267: Error dialog shows in front of fingerprint dialog
### Project

Deskflow

### Deskflow version info

Deskflow: 1.19.0.109 (63fa0f5c)
Qt: 6.8.1
System: Arch Linux
Session: KDE (wayland)

### Build type

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [ ] Linux (X11)
- [x] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### OS versions/distros

Arch Linux - KDE

### Deskflow configuration

Linux client

1. Clear settings on both client and server
2. Connect to server

Expected: Only fingerprint dialog
Actual: Fingerprint dialog shows, then error dialog shows in front of it (this will definitely confuse users)

Image

### Additional information

Proposed solution: Mute client connection errors in `src/lib/gui/core/ClientConnection.cpp` while the fingerprint dialog is being shown.

sithlord48:
Stops the core process in client mode while the client it validating fingerprints of the server, this has two benefits
 1. The Client no longer repeatably try to connect while the user is looking at the fingerprints of the server. So the Connection failed message mentioned in #8267 is no longer spawned.
 2. The Server will not be spammed with connection attempts at this time from the connecting client.

I fail to see any downside to this since the any client may only be connected to a single server.  I tested this on the with a client using auto connect to a name with two servers using the same name one accepted and one not and the client only talks to the proper one the fingerprint dialog is not even show, so it does not look like a rouge server can use this as a DDOS of the client.

fixes #8267

### DIFF
--- a/src/apps/deskflow-gui/MainWindow.cpp
+++ b/src/apps/deskflow-gui/MainWindow.cpp
@@ -809,18 +809,19 @@ void MainWindow::checkFingerprint(const QString &line)
     return;
   }
 
-  if (isClient) {
-    m_checkedServers.append(sha256Text);
-  } else {
-    m_checkedClients.append(sha256Text);
-  }
-
   deskflow::FingerprintDatabase db;
   db.read(trustedFingerprintDb().toStdString());
 
   if (db.isTrusted(sha256)) {
     qDebug("fingerprint is trusted");
     return;
+  }
+
+  if (isClient) {
+    m_checkedServers.append(sha256Text);
+    m_coreProcess.stop();
+  } else {
+    m_checkedClients.append(sha256Text);
   }
 
   auto dialogMode = isClient ? FingerprintDialogMode::Client : FingerprintDialogMode::Server;
@@ -833,6 +834,7 @@ void MainWindow::checkFingerprint(const QString &line)
     db.write(trustedFingerprintDb().toStdString());
     if (isClient) {
       m_checkedServers.removeAll(sha256Text);
+      m_coreProcess.start();
     } else {
       m_checkedClients.removeAll(sha256Text);
     }


src/apps/deskflow-gui/MainWindow.cpp
```diff
@@ -811,6 +811,7 @@ void MainWindow::checkFingerprint(const QString &line)
 
   if (isClient) {
     m_checkedServers.append(sha256Text);
+    m_coreProcess.stop();
```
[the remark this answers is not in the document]
sithlord48: No but i suspect that maybe related to #8278  ?

sithlord48:
Have you tried the CI generated artifact ? Test in release mode NOT debug

nbolton:
> Have you tried the CI generated artifact ? Test in release mode NOT debug

_Both_ release and debug builds must work. One type is not enough.

sithlord48:
Not a race condition, in Debug the stack / heap is literately different and all the additional debug messages are changing the timing

src/apps/deskflow-gui/MainWindow.cpp
```diff
@@ -811,6 +811,7 @@ void MainWindow::checkFingerprint(const QString &line)
...
   } else {
     m_checkedClients.append(sha256Text);
   }
```
nbolton, line 817:
```suggestion
```

nbolton: This patch fixes it for me. I cannot add a change suggestion outside of your changed lines, so it has to be a patch.

```diff
diff --git a/src/apps/deskflow-gui/MainWindow.cpp b/src/apps/deskflow-gui/MainWindow.cpp
index 36180ce73..d68c07d1f 100644
--- a/src/apps/deskflow-gui/MainWindow.cpp
+++ b/src/apps/deskflow-gui/MainWindow.cpp
@@ -809,13 +809,6 @@ void MainWindow::checkFingerprint(const QString &line)
     return;
   }
 
-  if (isClient) {
-    m_checkedServers.append(sha256Text);
-    m_coreProcess.stop();
-  } else {
-    m_checkedClients.append(sha256Text);
-  }
-
   deskflow::FingerprintDatabase db;
   db.read(trustedFingerprintDb().toStdString());
 
@@ -824,6 +817,15 @@ void MainWindow::checkFingerprint(const QString &line)
     return;
   }
 
+  if (isClient) {
+    m_checkedServers.append(sha256Text);
+
+    qDebug("stopping client before showing fingerprint dialog");
+    m_coreProcess.stop();
+  } else {
+    m_checkedClients.append(sha256Text);
+  }
+
   auto dialogMode = isClient ? FingerprintDialogMode::Client : FingerprintDialogMode::Server;
 
   FingerprintDialog fingerprintDialog(this, {sha1, sha256}, dialogMode);
```

nbolton (changes requested):

src/apps/deskflow-gui/MainWindow.cpp
sithlord48: is this message worth logging because qDebug is only added in debug mode.

sithlord48: I do not understand how a log message could possibly fix this for you....

nbolton: It's not the log message that fixes it. The logic is entirely different.

sithlord48:
Once again I fail to see how a message on debug mode 'fixes' this for you ...

nbolton:
> Once again I fail to see how a message on debug mode 'fixes' this for you ...

The debug message is not relevant. Please re-read the patch. The logic is different.

sithlord48:
It not remotely noticeable in what is shown here :(

nbolton (approved):
Works great! 👍

Issue #8248, opened 2025-02-25
Use state machine for watchdog loop in Windows daemon


nbolton:
### Enhancement Request

The watchdog in the Windows daemon is overcomplicated and uses many member variables unnecessarily.

The logic would be far easier to follow if it was a state machine.

Blocked by:
- #8180

nbolton:
Note to self: I have a local stash for this improvement that I made while working on #8180.

Issue #8278, opened 2025-03-01
Massive spike in cpu use with daemon running and stop pressed on gui


sithlord48:
### Project

Deskflow

### Deskflow version info

Deskflow 1.19.0.154+

### OS versions/distros

 Windows 

### What steps will reproduce the problem?

1. Install deskflow 
2. Check the daemon is running
3. start the gui 
4. pick a mode
5. start the mode
6. stop the mode
7. log spam seen below
8. also note cpu use spikes as we spam the log.

Only happens if the daemon is running

### Log output

```shell
DEBUG: process started but command is empty, shutting down
DEBUG: shutting down existing processes
FOR EVER
```

sithlord48:
This happens ever time i try touch the stop button then i have to stop the service and hit #8279

nbolton:
Please test PR:
- #8281

If it's still happening for you, maybe we could pair up.

Attaching your `C:\Program Files\Deskflow\deskflow-daemon.log` file might be helpful.

PR #8281, opened 2025-03-03
Use state machine for Windows daemon watchdog and cleanup

nbolton:
Fixes: #8248


sithlord48 (commented):
why is this 28 commits ?

nbolton:
> why is this 28 commits ?

Splitting a refactor into 28 commits makes the PR easier to review by keeping each change focused and isolated, and it also helps with maintenance in the future. It allows reviewers to verify small, logical changes, making it easier to understand and spot issues. 

As you probably already realise, `git bisect` can pinpoint the problematic commit, and individual changes can be reverted without undoing unrelated improvements. Logical grouping makes the history clear and avoids monolithic commits that are harder to review and understand. Smaller commits also reduce the risk of merge conflicts.

If you're finding this PR hard to review then the discreet commits have not fulfilled their purpose. Would you like me to squash some of the commits together? For example, I could squash commits with "Remove unused" messages.

nbolton:
> make a pr w/ a single commit

I see your point—some of the commits (like log level changes) could have been in a separate PR, and I can be more mindful of that in the future. The intent behind splitting this into multiple commits was to keep changes granular and reviewable, but if the commit structure is making it harder to review rather than easier, I can squash related commits. Would you prefer a single commit for all "remove unused" changes, or should I split them into a separate PR? Happy to adjust to whatever makes review smoother.

sithlord48:
Are they all safe to remove unused items then yes one commit is fine. If one you think is gonna break something else do that one alone.. 

Quick test on windows this seams to resolve #8278  so i added it to the list.

sithlord48 (approved):
This does work as intended, Please be more mindful of not including unrelated commits in Prs in the future.

nbolton:
> * Test if this fixes [Massive spike in cpu use with daemon running and stop pressed on gui #8278](https://github.com/deskflow/deskflow/issues/8278)

Un-readied while I test.

Issue #8227, opened 2025-02-19
Help -> 'Clear settings' does not clear Windows daemon settings


nbolton:
### Project

Deskflow

### Deskflow version number

1.19.0

### Build type

Local developer build (built it myself)

### Operating systems (OS)

- [x] Windows
- [ ] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### OS versions/distros

Any OS

### What steps will reproduce the problem?

1. Go to Preferences -> Advanced
2. Change settings profile to 'All users'
3. Help -> 'Clear settings'

Expected: On Windows the daemon registry entries are deleted (in `HKEY_LOCAL_MACHINE\SOFTWARE`)
Actual: The Windows daemon registry entries persist

### Additional information

The clear settings feature is important for development so that we can easily reproduce bugs on fresh installs, e.g.: https://github.com/deskflow/deskflow/pull/8194#issuecomment-2669206576

sithlord48:
I think you need to have the daemon clear its settings as well since its running as another user the user running deskflow gui may not be able to remove those keys.

nbolton:
> I think you need to have the daemon clear its settings as well since its running as another user the user running deskflow gui may not be able to remove those keys.

Yes that's right, due to access permissions the weakness here is that the daemon has to be running and the GUI has to be connected to it. If the daemon is broken for some reason (eg won't start), or the IPC connection fails, the registry keys have to be deleted manually.

PR #8284, opened 2025-03-03
fix: Delete `HKEY_LOCAL_MACHINE\SOFTWARE\Deskflow` when clearing settings

nbolton:
Fixes: #8227

To test: 
1. Use Windows
2. Click 'Help'
3. Click 'Clear settings'

Expect: Registry key does not exist: `HKEY_LOCAL_MACHINE\SOFTWARE\Deskflow`

### DIFF
--- a/src/lib/arch/win32/ArchMiscWindows.cpp
+++ b/src/lib/arch/win32/ArchMiscWindows.cpp
@@ -165,6 +165,15 @@ void ArchMiscWindows::deleteValue(HKEY key, const TCHAR *name)
   RegDeleteValue(key, name);
 }
 
+void ArchMiscWindows::deleteKeyTree(HKEY key, const TCHAR *name)
+{
+  assert(key != NULL);
+  assert(name != NULL);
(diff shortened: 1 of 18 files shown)


src/lib/arch/win32/ArchMiscWindows.cpp
```diff
@@ -165,6 +165,15 @@ void ArchMiscWindows::deleteValue(HKEY key, const TCHAR *name)
   RegDeleteValue(key, name);
 }
 
+void ArchMiscWindows::deleteKeyTree(HKEY key, const TCHAR *name)
+{
+  assert(key != NULL);
+  assert(name != NULL);
+  if (key == NULL || name == NULL)
+    return;
```
nbolton, line 173: Not really happy with this. I mindlessly copied the code above... if either were null this'd be confusing in release.

sithlord48: Agree.

sithlord48 (approved):

PR #8298, opened 2025-03-05
Create a proper settings object

### LINKED ISSUE #8293: Improve settings
### Feature Request

Improve our settings.

### LINKED ISSUE #8296: Create a settings object
### Feature Request

Create a settings object to manage the settings the object should 

- Use QSettings
- Singleton we can use anywhere 
- Handle all the settings 
- Make it easier to add new settings
- Clean up old settings
- provide default values for the settings
- Replaces All of the various settings object.
- Use User scope by default
- Have an option to use a specific override file, User provided 
- Emits when `SettingsChange`.

The file is broken into several sections
   
  - `core` - These settings are used by more then one Think `screenName` 
  - `gui` - These settings are used only by the gui 
  - `client` - Client based settings
  - `server` - Server sepecific settings
  - `security` - Holds our security released info used by the client and the server.

Thinking of using `DeskflowSettings` but this is a tad log so may call it `Settings` or `AppSettings` 
key access will be via static members like `Setting::Gui::AutoHide`  or `Settings::Core::ScreenName` 

`DeskflowSettings::value(Settings::Gui::AutoHide).toBool()` for example would be how to get the value of the autohide and treat it like a bool . 

`DeskflowSettings::setValue(Setting::Gui::AutoHide, true); would set the value to true

sithlord48:
- resolve #8293

File Support: 
 - [X] Use Ini Format for all platforms
 - [X] Use `Deskflow.conf` for the setting file name
   - [X] All: allow a config file in the same path as the exe to be used as an override settings file
   - [X] macOS: Write all the settings into `~/Library/Deskflow` or `/Library/Deskflow`
   - [X] Windows:  Write settings to `C:\Users\$USER\AppData\Local\Deskflow` or `C:\ProgramData\Deskflow`
   - [X] Linux: Use `~/.config/Deskflow`  or `/etc/Deskflow`
   - [x] Make sure TLS and server config paths are updated based on the current config path
 - [X] Port all AppConfig options

Things This pr does not do 
 - Migration Tool no porting of settings, Test with reset settings. (may want to suggest users do reset
 - Support Screens
 - Support Server Config
 - Support Hotkeys
 - Support Actions 
 - Retire QSettingsProxy


sithlord48:
@nbolton  might need some help w/ the tests i 've broken

src/lib/common/DeskflowSettings.cpp
```diff
@@ -0,0 +1,160 @@
...
+void DeskflowSettings::setSettingFile(const QString &settingsFile)
+{
+  if (instance()->m_portableSettingsFile == settingsFile)
+    return;
+
+  instance()->m_portableSettingsFile = settingsFile;
+  if (instance()->m_settings)
+    instance()->m_settings->deleteLater();
+  instance()->m_settings = new QSettings(instance()->m_portableSettingsFile, QSettings::IniFormat);
+  qInfo() << "Settings file" << instance()->m_settings->fileName();
```
nbolton, line 27: Please use all lower case for consistency with other log lines. The log output would also be more readable if you used a colon:
```cpp
qInfo() << "settings file:" << instance()->m_settings->fileName();
```

There are 3 instances that need fixing.

nbolton (changes requested):

src/lib/common/settings.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <QDir>
+#include <QObject>
+
+/**
+ * @brief The Settings class contains
+ *  all the keys, enum types and other settings data
+ *  to be used by DeskflowSettings
+ */
+class Settings
```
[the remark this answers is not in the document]
sithlord48: Why cause its just more maintainable. 

Settings has to be a class to expose the enums correctly. 
I would reather not have this all in one huge file (and that still does not make it one class) 
So what names would you recommend ?

src/lib/common/DeskflowSettings.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+DeskflowSettings *DeskflowSettings::instance()
+{
+  static DeskflowSettings m;
+  return &m;
+}
+
+void DeskflowSettings::setSettingFile(const QString &settingsFile)
+{
+  if (instance()->m_portableSettingsFile == settingsFile)
```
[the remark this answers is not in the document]
sithlord48: We need to stop doing this and start using [QLoggingCatagories](https://doc.qt.io/qt-6/qloggingcategory.html) properly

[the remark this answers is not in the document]
nbolton: Is there an issue for this?

src/lib/common/settings.h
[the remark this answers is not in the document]
nbolton: OK, I see your point, but the responsibility should be clear from the name of a class and right now it is not clear.

It seems that it's mostly about where to keep the _keys_, so:
- `Settings` -> `SettingsKeys`
- `DeskflowSettings` -> `Settings`

This would be far easier to understand.

src/lib/common/DeskflowSettings.cpp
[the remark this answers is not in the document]
sithlord48: Don't think so

src/lib/common/settings.h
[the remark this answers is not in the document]
sithlord48: Just gonna do this in one class as they keys have both keys and values type (i.e enums)

nbolton:
> * Support Screens
> * Support Server Config
> * Support Hotkeys
> * Support Actions

I'm sure I'll figure this out when reviewing, but does this mean the old code is still in use?

src/lib/common/settings.h
```diff
@@ -0,0 +1,69 @@
...
+
+#include <QDir>
+
+#include "common/constants.h"
+
+class Settings : public QObject
+{
+  Q_OBJECT
+public:
+  // KEY VALUES AND ENUMS
```
nbolton, line 20: No all caps rage comments, please.

If you want comments to stand out, this style is very common in our code base:
```
//
// Key values and enums
//
```

However, I am wondering what this comment really adds. 

Comments should really explain why and not what; code should be self-documenting. If you have to explain what code is doing, then it's probably not well written.

src/lib/common/settings.h
```diff
@@ -0,0 +1,69 @@
...
+#if defined(Q_OS_WIN)
+  inline const static auto UserSettingFile =
+      QStringLiteral("%1/AppData/Local/%2/%3.conf").arg(QDir::homePath(), kAppName, kAppName);
```
nbolton: Static is redundant when using inline, because inline already ensures a single instance across translation units (In C++17 and later).

Calling `.arg()` on a `QStringLiteral` converts it to a temporary `QString` anyway. So using `QStringLiteral` in this case does not provide any real benefit. The optimization that `QStringLiteral` provides (avoiding runtime string construction) is lost once you apply the .arg() function, so there's no performance advantage in this particular usage.

src/lib/common/settings.h
```diff
@@ -0,0 +1,69 @@
...
+  inline const static auto UserSettingFile =
+      QStringLiteral("%1/.config/%2/%3.conf").arg(QDir::homePath(), kAppName, kAppName);
+  inline const static auto SystemSettingFile = QStringLiteral("/etc/%2/%3.conf").arg(kAppName, kAppName);
+#endif
+  struct Core
+  {
+    inline static const auto Scope = QStringLiteral("core/loadFromSystemScope");
+  };
+
+  // METHODS
```
nbolton, line 129: Same as above, is this a useful comment?

src/lib/common/settings.h
```diff
@@ -0,0 +1,69 @@
...
+private:
+  explicit Settings(QObject *parent = nullptr);
+  Settings *operator=(Settings &other) = delete;
+  Settings(const Settings &other) = delete;
+  ~Settings() = default;
+  static bool isPortableSettings();
+  void cleanSettings();
+  void initSettings();
+
+  // Vars
```
nbolton, line 157: > Same as above, is this a useful comment?

src/lib/common/settings.h
```diff
@@ -0,0 +1,69 @@
...
+  QSettings *m_settings = nullptr;
+  QString m_portableSettingsFile = QStringLiteral("%1.conf").arg(kAppName);
+  inline static const QStringList m_validKeys = {Settings::Core::Scope};
```
nbolton: The inline static comment also applies here. The static keyword is redundant when using inline for this class member.

src/lib/common/settings.h
```diff
@@ -0,0 +1,69 @@
...
+#elif defined(Q_OS_MAC)
+  inline const static auto UserSettingFile =
+      QStringLiteral("%1/Library/%2/%3.conf").arg(QDir::homePath(), kAppName, kAppName);
+  inline const static auto SystemSettingFile = QStringLiteral("/Libaray/%2/%3.conf").arg(kAppName, kAppName);
+#else
+  inline const static auto UserSettingFile =
+      QStringLiteral("%1/.config/%2/%3.conf").arg(QDir::homePath(), kAppName, kAppName);
+  inline const static auto SystemSettingFile = QStringLiteral("/etc/%2/%3.conf").arg(kAppName, kAppName);
+#endif
+  struct Core
```
nbolton, line 41: Space above this `struct` would help with readability.

src/lib/common/settings.cpp
```diff
@@ -0,0 +1,158 @@
...
+  if (instance()->m_portableSettingsFile == settingsFile) {
+    qDebug().noquote() << "settings file already in use";
+    return;
+  }
+
+  instance()->m_portableSettingsFile = settingsFile;
+  if (instance()->m_settings)
+    instance()->m_settings->deleteLater();
+  instance()->m_settings = new QSettings(instance()->m_portableSettingsFile, QSettings::IniFormat);
+  qInfo().noquote() << "settings file" << instance()->m_settings->fileName();
```
nbolton, line 29: I previously suggested `"settings file:"` (with a colon) rather than `"settings file"`.

The colon is important as it improves readability of the log line.

There are several instances of this.

src/test/unittests/main.cpp
```diff
@@ -25,6 +26,8 @@ int main(int argc, char **argv)
   // HACK: Unit tests should not use the filesystem.
   std::filesystem::create_directories(testDir);
 
+  Settings::setSettingFile("tmp/test/settings.ini");
```
nbolton, line 29: The hack comment also applies here; using the filesystem in unit tests is a hack.
```
// HACK: Unit tests should not use the filesystem.
```

When a unit test relies on systems outside of the binary, it becomes an integration test. 

I understand this is temporary as we want to use Qt Test for GUI tests in future, but temporary things have a habit of becoming permanent.

nbolton (dismissed):
Tested and works well. Only nits found in the review. I'm happy to land this as-is and fix the nits later.

Overall this is a big improvement!

nbolton:
I just realised that the "settings file" line is printed before main window is shown, so you'd only ever see that log line if you watch stdout (e.g. start it from the command line on Unix).

```
[2025-03-13T11:39:51] INFO: settings file C:/Users/nick/AppData/Local/Deskflow/Deskflow.conf
```

I can see it on Windows through attaching a debugger (release build), but since we're using info level, I'm guessing this is intended for users to see, so maybe it'd be more useful to print this line from `MainWindow`?

src/lib/common/settings.h
sithlord48: The literal part is created at compile time and at runtime the substution happens... It does not remove any proformance advantage.

src/lib/common/settings.h
sithlord48: `inline static` is an intilizer for static added in C++17 to allow you to make static member inline not in the class. It does not put the member "inline" like an inline methods

src/test/unittests/main.cpp
sithlord48: The settings have to go somewhere 
I don't see a problem with a test that needs to read / write a file writing a temp file. todo so in a temp directory.

sithlord48:
> I just realised that the "settings file" line is printed before main window is shown, so you'd only ever see that log line if you watch stdout (e.g. start it from the command line on Unix).
> 
> ```
> [2025-03-13T11:39:51] INFO: settings file C:/Users/nick/AppData/Local/Deskflow/Deskflow.conf
> ```
> 
> I can see it on Windows through attaching a debugger (release build), but since we're using info level, I'm guessing this is intended for users to see, so maybe it'd be more useful to print this line from `MainWindow`?

The setttings are made before the main windows.  You will see this if you change the scope and it should be wrtting to a text log ...

nbolton (approved):
One very minor nit. Your call if you want to land or fix.

nbolton (approved):

Issue #8288, opened 2025-03-04
Windows daemon access violation (and IPC connection fails)


renzhenxuexi0:
### Project

Deskflow

### Deskflow version info

Deskflow: v1.20.0.0

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [x] Windows
- [ ] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [ ] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [ ] I am not using Wayland on Linux

### Signing on macOS

- [ ] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### Continuous build

- [x] I have tried the latest continuous build and the issue persists
- [x] I am unable to try the latest continuous build

### OS versions/distros

- Windows 11

### Deskflow configuration

- Windows 11

### What steps will reproduce the problem?

1. Start Deskflow
2. make this computer the server
3. Start

### Log output

```shell
[2025-03-04T15:22:20] INFO: starting daemon
[2025-03-04T15:22:22] INFO: stopping core process (service mode)
[2025-03-04T15:22:22] INFO: starting core server process (service mode)
[2025-03-04T15:22:22] INFO: core config file: C:/Users/wuhaiyue/AppData/Local/Deskflow/deskflow-server.conf
[2025-03-04T15:22:22] INFO: running command: D:/Tools/Deskflow/deskflow-server.exe -f --debug INFO --name wuhy --stop-on-desk-switch --profile-dir C:/Users/wuhaiyue/AppData/Local --address :8688 -c C:/Users/wuhaiyue/AppData/Local/Deskflow/deskflow-server.conf
[2025-03-04T15:22:22] INFO: connecting to daemon ipc
[2025-03-04T15:22:22] WARNING: ipc client error: "QLocalSocket::connectToServer: Invalid name"
[2025-03-04T15:22:22] WARNING: ipc client failed to connect to server: deskflow-daemon
[2025-03-04T15:22:22] WARNING: ipc client keep alive failed to connect
[2025-03-04T15:22:22] CRITICAL: cannot start process, ipc command failed
[2025-03-04T15:22:23] INFO: starting daemon
```

gavrochelegnou:
Yep on the last version i get the same result.
[...]

PR #8301, opened 2025-03-05
fix: Remove mutex locking in Windows daemon watchdog (solves access violation bug)

nbolton:
⚠️ **Draft** (not ready for review)

Fixes: #8288

Blocked by: 
- #8303
- #8305

~~This is a bit of a backward step, but for some reason the C++11 mutex was causing an access violation in the release build on some computers.~~

Use of either C++11 mutex or native Win32 mutex causes access violation. The mutex was added as it looked safe to do so but perhaps we can live without it.

This probably a symptom of a much bigger problem like stack or heap corruption, making this more of a workaround than a proper fix.

Note: I tried using `QMutex`, but linking `platform` against Qt causes many things to break, so we should do this some other time.

### DIFF
--- a/src/lib/common/CMakeLists.txt
+++ b/src/lib/common/CMakeLists.txt
@@ -5,7 +5,7 @@ set(CMAKE_INCLUDE_CURRENT_DIR_IN_INTERFACE ON)
 
 configure_file(constants.h.in constants.h @ONLY)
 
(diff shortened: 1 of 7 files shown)


nbolton:
Closing as removing the mutex is not the right solution.

Issue #8311, opened 2025-03-06
Program allows incorrect VC++ runtime DLL to be loaded


nbolton:
### Project

Deskflow

### Deskflow version info

Deskflow: 1.20.0.7 (a7ce936d)
Qt: 6.8.2
System: Windows 11 Version 24H2

### Build type

Local developer build (built it myself)

### Operating systems (OS)

- [x] Windows
- [ ] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [ ] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [ ] I am not using Wayland on Linux

### Signing on macOS

- [ ] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### Continuous build

- [ ] I have tried the latest continuous build and the issue persists
- [ ] I am unable to try the latest continuous build

### OS versions/distros

Windows 11

### What steps will reproduce the problem?

Run Deskflow on a fresh Windows 11 install where VC++ runtime v14.26.28808.1 is installed

Expected: Error should show because the program was compiled against a newer VC++ runtime (v14.42.34433.0)
Actual: Program runs without error until it encounters difficult to debug issues (e.g. #8288)

### Log output

```shell

```

### Additional information

Rather than failing in some random way, the program should fail fast with a VC++ runtime version error.

sithlord48:
I think i know what causes this  Just added a"fix" to #8306  lets see if that builds / works

PR #8312, opened 2025-03-06
fix: Guard MSVC runtime version depending on compiler version

nbolton:
Fixes: #8311

PR shows an error dialog and exits the program (server, client, and daemon) if the MSVC++ runtime is too old. This avoids silent crashes which are very confusing and waste a lot of time. Windows doesn't tell you when a binary was linked against an older library, it just loads whatever DLL it finds and hopes for the best.

Wrong version:
image

Missing runtime (default Windows message):
image


sithlord48:
Should the gui do this check as well ?

nbolton:
> Should the gui do this check as well ?

We could if we link the `gui` bin to the `arch` lib. Thoughts?

That said, the user should never be able to get to the point of running the GUI because this new feature prevents the MSI from completing the installation.

sithlord48 (approved):

PR #8325, opened 2025-03-07
fix: Allow compiling on VS2019 for contributors with old IDEs

nbolton:
Some contributors have only VS2019 and have not yet updated to VS2022. We should allow them to build as well (so they aren't forced to update their IDE and risk breaking their dev env).


sithlord48 (approved):
Great Note

PR #8360, opened 2025-03-14
refactor: mainwindow move save config to a button

sithlord48:
-  Remove from the main window the "save Configuration as" menu item and replace it with a button that is only shown when in server mode. 
 - Add "configure" icon to the "Configure Server" button


nbolton:
While testing, I found another problem with #8372:
Image

I'll also post there since I think it's not caused by this PR.

nbolton:
Interesting, when I click the save icon it also causes bug #8372:

image

nbolton (approved):

nbolton:
Still getting that not-so-transient Flatpak error:
```
fatal: unable to access 'https://gitlab.freedesktop.org/libinput/libei/': The requested URL returned error: 503
Failed to download sources: module libei: Child process exited with code 128
```

PR #8361, opened 2025-03-14
refactor: move daemon settings to Settings

### LINKED ISSUE #8353: Port Daemon settings to use Settings object
### Feature Request

The daemon is using its own methods to save settings, it should be using the `Settings` object

1. Use existing values in the deamon where possible from `Settings`
2. Add to `Settings` new needed keys in a `daemon`  section.

sithlord48:
fixes #8353


src/lib/deskflow/DaemonApp.cpp
```diff
@@ -50,39 +49,21 @@ void DaemonApp::saveLogLevel(const QString &logLevel) const
...
   LOG_DEBUG("elevate value changed: %s", elevate ? "yes" : "no");
   m_elevate = elevate;
-
-  try {
-    // saves setting for next time the daemon starts.
-    ARCH->setting("Elevate", std::string(elevate ? "1" : "0"));
-  } catch (XArch &e) {
-    LOG_ERR("failed to save elevate setting: %s", e.what());
-  }
+  Settings::setValue(Settings::Daemon::Elevate, m_elevate);
```
nbolton, line 59: Minor nit: Slightly inconsistent with `setCommand`. In `setCommand` you use the param value, but there you use member. Functionally it's the same, but looks like a mistake. Unless it was deliberate?

src/lib/platform/XWindowsPowerManager.cpp
```diff
@@ -32,16 +40,69 @@ XWindowsPowerManager::~XWindowsPowerManager()
...
 {
-  if (!sleepInhibitCall(false, ArchSystemUnix::InhibitScreenServices::kScreenSaver) &&
-      !sleepInhibitCall(false, ArchSystemUnix::InhibitScreenServices::kSessionManager)) {
+  if (!sleepInhibitCall(false, XWindowsPowerManager::InhibitScreenServices::kScreenSaver) &&
+      !sleepInhibitCall(false, XWindowsPowerManager::InhibitScreenServices::kSessionManager)) {
     LOG((CLOG_WARN "failed to enable system idle sleep"));
   }
 }
+
+#ifndef __APPLE__
```
nbolton, line 57: > You shouldn't need this `#ifndef` now that you're in an `XWindows*` class, which is only compiled on Linux.

```suggestion
```

src/lib/platform/XWindowsPowerManager.cpp
```diff
@@ -32,16 +40,69 @@ XWindowsPowerManager::~XWindowsPowerManager()
...
+bool XWindowsPowerManager::DBusInhibitScreenCall(InhibitScreenServices serviceID, bool state, std::string &error)
```
nbolton, line 58: While we're here...

```suggestion
bool XWindowsPowerManager::inhibitScreenCall(InhibitScreenServices serviceID, bool state, std::string &error)
```

I know this is moved code (someone else wrote it) but since `DBus` is an implementation, now might be a good opportunity to remove it from the function name, if you want to.

src/lib/platform/XWindowsPowerManager.cpp
```diff
@@ -9,13 +9,21 @@
 #include "arch/Arch.h"
 #include "base/Log.h"
 
+#ifndef __APPLE__
```
nbolton, line 12: You shouldn't need this #ifndef now that you're in an XWindows* class, which is only compiled on Linux.
```suggestion
```

src/lib/platform/XWindowsPowerManager.cpp
nbolton, line 58: As an aside: The `DBusInhibitScreenCall` function should never have been added to `arch`. Whoever added this didn't know what they were doing. The `platform` is the correct place for Mac/Linux-specific code. The `arch` lib is supposed to only be for Unix/Win32 code.

src/lib/platform/XWindowsPowerManager.h
```diff
@@ -7,6 +7,10 @@
 
 #pragma once
 
+#ifndef __APPLE__
```
nbolton, line 10: > You shouldn't need this `#ifndef` now that you're in an `XWindows*` class, which is only compiled on Linux.

```suggestion
```

src/lib/platform/XWindowsPowerManager.h
```diff
@@ -25,4 +29,13 @@ class XWindowsPowerManager
 
   XWindowsPowerManager(const XWindowsPowerManager &) = delete;
   XWindowsPowerManager &operator=(const XWindowsPowerManager &) = delete;
+
+#ifndef __APPLE__
```
nbolton, line 33: > You shouldn't need this `#ifndef` now that you're in an `XWindows*` class, which is only compiled on Linux.

```suggestion
```

nbolton (approved):

src/lib/platform/XWindowsPowerManager.cpp
sithlord48: This was moved from arch unix system so the checks for apple were needed ill go ahead and remove them from this part.

src/lib/deskflow/DaemonApp.cpp
sithlord48: I have no idea what your trying to tell me here

sithlord48: OHH yeah no this is gonna be fixed now it had to be updated after #8364

PR #8346, opened 2025-03-12
Remove the old CoreTool and its use

sithlord48:
- Fix the Settings override to use `settings/Deskflow.conf`
 - Remove the use of CoreTool
 - should fix #7789
 - Unify the ConfigDir and the "SettingsDir", Do not allow the app to write outside of its settings dir.

PR #8363, opened 2025-03-15
refactor: unify Coredir to SettingsPath

sithlord48:
This was listed as part of #8346 but i forgot to add the commit so here is it . 

 - Unify the ConfigDir and the "SettingsDir", Do not allow the app to write outside of its settings dir.
 - Remove the now unused lib/gui/paths.h


nbolton (approved):

nbolton:
Confirmed, this PR caused: #8371

Bad testing on my part.

PR #8373, opened 2025-03-17
Fix typos (avilable, recieve)

kenhys:
avilable =>
available

recieve =>
receive

NOTE: There are many ..Recieve... named methods, but not touched.


sithlord48:
you can add the method name fixes as well if you like

sithlord48:
Oh i wonder if its safe to fix that revieve typo in methods as one is the protocol naeed methods . @nbolton  what do u think ?

sithlord48 (approved):

nbolton:
> Oh i wonder if its safe to fix that revieve typo in methods as one is the protocol naeed methods . @nbolton what do u think ? the typos are from this event `REGISTER_EVENT(File, fileRecieveCompleted)`

I already replied on Matrix but for posterity, I'll share here: These should be safe to rename, but incidentally, we plan to remove them as they are related to file drag and drop.

PR #8385, opened 2025-03-20
Remove the Common/Std*.h and  platform/Win32Hook.h files

sithlord48:
Changes to `lib/common`
 - Remove all the `Std*.h` files and uses the types directly
 - Does not Re add all the pragma from StdPre.h
 - Removes `_NOEXCEPT` define and use `throw()` 
 - Removes unused `MacOSXPrecomp.h`

Changes to `lib/platform`
 - Move `Win32Hook.h` parts to `MSWindowsHook.h`

Changes to `apps/deskflow-gui`
 - remove unused `QUtility`
 - remove unused `ProcessorArch.h`


src/lib/platform/MSWindowsHook.h
```diff
@@ -7,11 +7,47 @@
...
+#include <stdint.h>
 
 #define WIN32_LEAN_AND_MEAN
 #include <Windows.h>
 
+#if defined(dfwhook_EXPORTS)
+#define CDESKFLOWHOOK_API __declspec(dllexport)
+#else
+#define CDESKFLOWHOOK_API __declspec(dllimport)
+#endif
```
nbolton, line 19: I believe this can be removed.

nbolton (approved):
Looks good, could land as is but there is some dead code that can go.

src/lib/platform/MSWindowsHook.h
sithlord48: yeah lets try it  its a static lib so import / export should not be used anyway

nbolton (approved):

PR #8386, opened 2025-03-21
Move more lib/gui => apps/deskflow-gui

sithlord48:
- move `lib/gui/validators` => `apps/deskflow-gui/validators`
 - move `lib/gui/StyleUtils.h` => `apps/deskflow-gui/StyleUtils.h`
 - replace `lib/gui/EnvVars.h` with a new Settings `Settings::Core::UpdateUrl`
 - move `lib/gui/Diagnostic` => `apps/deskflow-gui/Diagnostic`
 - move `lib/gui/Constants.h` => `lib/common/UrlConstants.h`
 - remove `lib/gui/StringUtils`
   - `StringUtils::trimEnd` => `QString::trimmed`
   - `StringUtils::StrToBool` => `QVariant().toBool()`
 - MainWindow remove test menu


src/lib/common/Settings.cpp
```diff
@@ -110,6 +110,9 @@ QVariant Settings::defaultValue(const QString &key)
   if (key == Core::ElevateMode)
     return Settings::ElevateMode::Always;
 
+  if (key == Core::UpdateUrl)
+    return QStringLiteral("https://api.deskflow.org/version");
```
nbolton, line 114: Please move this to `src/lib/gui/Constants.h`. See discussion we had a few weeks back about scattering URLs in the source code.

nbolton (changes requested):

src/lib/common/Settings.cpp
sithlord48: We do not want that then common needs gui .. 
Since this is  a setting I plan to move all the settings defaults here. The only Reason i left some in common/Constants is that file gets preconfigured.

nbolton (approved):

Issue #8409, opened 2025-03-28
Settings are read-only unless running as admin


shira-374:
### Project

Deskflow

### Deskflow version info

Deskflow: 1.21.0.0 (f781ac98)
Qt: 6.8.2
System: Windows 11 Version 23H2

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [x] Windows
- [x] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [ ] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [ ] I am not using Wayland on Linux

### Signing on macOS

- [ ] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### Continuous build

- [ ] I have tried the latest continuous build and the issue persists
- [ ] I am unable to try the latest continuous build

### OS versions/distros

Windows 11

### Deskflow configuration

Fresh installation

### What steps will reproduce the problem?

1. install deskflow
2. go to edit - preferences
3. you get a warning about read-only settings

![Image](https://github.com/user-attachments/assets/b0511013-8b6c-4ab6-9ca3-b85ea726fb22)

### Log output

```shell

```

### Additional information

- since 1.21.0 settings don't load or save when starting the app normally
- a workaround is to run the app as admin
- this happens even on a fresh install and after a full restart
- the previous version (1.20.1) works fine as-is

nbolton:
I thought regular users had access to C:\ProgramData by default.

Does your user account have admin privileges or do you run as admin with a different account?

sithlord48:
This is where applications on windows are expected to write their application specific settings.  I don't see this when testing on win11 my normal user can r/w settings @nbolton  have you seen this ?

shira-374:
I got admin privileges. I was running it via right click - run as admininstrator.

I've even tried removing that ProgramData/Deskflow folder between reinstalls or changing its privileges. It didn't help.

sithlord48:
make a file in the install path "settings/settings.ini" and the app should use the settings folder there instead

nbolton:
> I don't see this when testing on win11 my normal user can r/w settings [@nbolton](https://github.com/nbolton) have you seen this ?

[...]

PR #8416, opened 2025-03-29
Fix: settings releated issues, prepare it for testing

sithlord48:
Cleanup Settings:
  - Remove unused private `Settings::isPortableSettings`
  - Remove unused signal `Settings::scopeChanged`
  - Remove unused signal `Settings::writableChanged`
  - Add `Settings::validKeys` method to get the list of valid keys
  - Use the same logic for the settings file on all platforms
  - Change the default for the daemon log file on windows to the path of the daemon
  - Update the configuration.md file include a section on the gui config
  - Use `XDG_CONFIG_DIR` if set.
  - Fix: mixmatch set / read for `UpdateCheck`

Fix Settings Issues:
  - fixes: #8418 
  - fixes: #8409 
  - fixes: #8421 

Wix Installer Changes:
  - Run deskflow when the finish button is clicked
  - Run deskflow as a normal user post install

Tested on a Win11 CleanVm to make sure setting worked and daemon was picked up after mode selection
Tested flatpak via action build

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -2,6 +2,7 @@
 
  The search order for a setting file is:
  1. `<install-path>/settings/Deskflow.conf`
+ 1. `<XDG_CONFIG_HOME>/Deskflow/Deskflow.conf`
  1. A user settings file
  1. A system settings file
  
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -1,13 +1,12 @@
-# Gui Config
+# GUI Config
 
-The search path for settings ar as follows:
+ The search order for a setting file is:
+ 1. `<install-path>/settings/Deskflow.conf`
+ 1. A user settings file
+ 1. A system settings file
  
- - Check Install_path/settings/Deskflow.conf
- - Check if User Settings Path
- - Check if System Settings File
- - Make a User Setting file
- 
- The path of the settings file will be used as the base for all other config files
+ A new settings file will be created in the user path if no settings file is found.
+ The path of the settings file will be used as the base for all other config files.
  
 ### Windows
  - System: `C:\ProgramData\Deskflow\Deskflow.conf`
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -1,3 +1,26 @@
+# Gui Config
+
+The search path for settings ar as follows:
+ 
+ - Check Install_path/settings/Deskflow.conf
+ - Check if User Settings Path
+ - Check if System Settings File
+ - Make a User Setting file
+ 
+ The path of the settings file will be used as the base for all other config files
+ 
+### Windows
+ - System: `C:\ProgramData\Deskflow\Deskflow.conf`
+ - User: `C:\Users\userName\AppData\Local\Deskflow\Deskflow.conf`
+ 
+### Linux
+ - System: `/etc/Deskflow/Deskflow.conf`
+ - User: `~/.config/Deskflow/Deskflow.conf`
+ 
+### macOS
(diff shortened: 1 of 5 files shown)


TheMadman:
The flatpak version from https://github.com/deskflow/deskflow/actions/runs/14158887852 appears to have resolved this issue for me.

doc/configuration.md
```diff
@@ -1,3 +1,26 @@
+# Gui Config
+
+The search path for settings ar as follows:
+ 
+ - Check Install_path/settings/Deskflow.conf
+ - Check if User Settings Path
+ - Check if System Settings File
+ - Make a User Setting file
+ 
+ The path of the settings file will be used as the base for all other config files
```
nbolton, line 10:
```suggestion
# GUI Config

The search path for settings is as follows:
 
 - Check <install-path>/settings/Deskflow.conf
 - Check in the user settings path
 - Check in the system settings file
 - Make a user setting file
 
The path of the settings file will be used as the base for all other config files.
```

doc/configuration.md
```diff
@@ -1,3 +1,26 @@
+# Gui Config
+
+The search path for settings ar as follows:
+ 
+ - Check Install_path/settings/Deskflow.conf
+ - Check if User Settings Path
+ - Check if System Settings File
+ - Make a User Setting file
```
nbolton, line 8: I'm slightly confused about this last check. Perhaps it should be separate to the list but under it, e.g.:

> If no settings file is found then the settings are saved to the user settings path.

src/lib/common/Settings.cpp
```diff
@@ -125,7 +120,11 @@ QVariant Settings::defaultValue(const QString &key)
     return defaultProcessMode;
 
   if (key == Daemon::LogFile) {
+#ifdef Q_OS_WIN
+    return QStringLiteral("%1/%2").arg(QCoreApplication::applicationDirPath(), kDaemonLogFilename);
```
nbolton, line 124: What's the idea behind moving the daemon log back to the install dir? It seemed to work fine in ProgramData.

nbolton (commented):
Haven't tested yet, but had some thoughts. I'll test first thing tomorrow.

src/lib/common/Settings.cpp
sithlord48: Having the log here on windows makes the portable version easier while giving us a place we can always find it that is not based on a user related path.

doc/configuration.md
sithlord48: anything in   < > markdown will eat so i'll quote the path here too

src/lib/common/Settings.cpp
sithlord48: Its my under standing that everyone needs to know the path of this file and read it is that correct ?

doc/configuration.md
```diff
@@ -1,3 +1,26 @@
+# GUI Config
+
+ The search order for a setting file is:
+ 1. `<install-path>/settings/Deskflow.conf`
+ 1. `<XDG_CONFIG_HOME>/Deskflow/Deskflow.conf`
+ 1. A user settings file
+ 1. A system settings file
+ 
+ If no settings are found a new user settings file will be used
```
nbolton, line 9: Very minor nit: Grammar/phrasing
```suggestion
A new settings file will be created in the user path if no settings file is found.
```

nbolton (approved):

nbolton:
Forgot to mention, also tested on fresh Windows VM. Works great.

nbolton (approved):

Issue #8345, opened 2025-03-12
Core on Windows is elevated after logging in when 'As needed' is selected


nbolton:
### Project

Deskflow

### Deskflow version info

Deskflow: 1.20.1.20 (eb84c4ae)
Qt: 6.8.2
System: Windows 11 Version 24H2

### Build type

Local developer build (built it myself)

### Operating systems (OS)

- [x] Windows
- [ ] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### OS versions/distros

Windows 11

### What steps will reproduce the problem?

1. Ensure Core server/client is running through Daemon
2. Ensure 'As needed' elevation mode is selected
3. Restart computer
4. Login
5. Check processes

Expected: Process is not elevated and is running as user
Actual: Process is elevated and running as SYSTEM

This is a potential security vulnerability, depending on your views on 'As needed'.

### Additional information

The problem is most likely in the `MSWindowsDesks::checkDesk` function:
https://github.com/deskflow/deskflow/blob/bb38ad0766751140d75942dd4e641ff8082f08c2/src/lib/platform/MSWindowsDesks.cpp#L744

robot3498712:
Seems related to what I'm observing with the latest official release (which is entirely broken on windows 10 here: deskflow-1.21.0-win-x64.msi). 
Config location appears having changed to programdata, and most typically **permissions are wrong**; trying to change settings via edit > preference showing a "read-only" error. No matter if I change permissions via windows built in facilities, it eventually resets to non-working permissions. 

I assume the general connection failure is related to that.

Upgrade attempted from deskflow-1.19.0-win-x64.msi (which is working fine).
I would assume releases are actually tested on a basic level, even though windows may not be a top priority.

nbolton:
> Seems related to what I'm observing with the latest official release

@robot3498712 I think you're talking about a different issue:
- #8409

robot3498712:
Thanks!
The relevant read-only issue was fixed (latest continuous build). 


I have another comment. I'm not sure if this is the right place, but please bear with me.

Advanced > Service != "Never elevate"

With TLS error:

[2025-03-31T16:39:33] ERROR: secure socket error: tls certificate doesn't exist: C:/WINDOWS/system32/config/systemprofile/AppData/Local/Deskflow/tls/deskflow.pem
[2025-03-31T16:39:33] ERROR: could not load client certificates

[...]

Issue #8350, opened 2025-03-13
Split Windows daemon elevate combobox into two checkboxes


nbolton:
### Feature Request

Problem: I added the 'As needed' option 15 years ago when I was first implementing the Windows daemon, as I thought it'd be a good idea to not always run the Core process as elevated. This conflates two options, `--stop-on-desk-switch` and elevate.

Solution: Replace elevate combobox (containing 'As needed' option) with an 'Elevate' checkbox which is on by default and a 'Stop on desk switch' checkbox (which is a workaround for sleep bugs, clipboard bugs, etc).

PR #8428, opened 2025-03-31
refactor: settings for windows 

sithlord48:
fixes: #8429 
fixes: #8432 
fixes: #8434
 - On windows use the registry to save the items normally in `Deskflow.conf`
 - On windows use the registry unless the portableSettings file `[install-dir]/settings/Deskflow.conf` if found
   - On Windows Use either `ProgramData\Deskflow`  to save tls and other non "settings" when registry is in use
   - On Windows Use either `[installdir]/settings`  to save tls and other non "settings" when registry is not in use
 - Set the default process Type based on the kind of settings loaded
     -  `InI Format` (unix / mac / windows portable) the mode defaults to `Desktop`
     -  `Native Format` settings (only windows installed mode) we default to `Service`
 - Set the default elevation mode based on type also 
     -  `InI Format` (unix / mac / windows portable) the mode defaults to `Never`
     -  `Native Format` settings (only windows installed mode) we default to `Always` 
 - Hide the service section when its not able to be used 
 - Daemon log is always saved in the install dir
 - Make sure if using portable settings the a new empty file is created if settings are cleared.

Windows artifact to test with is [here](https://github.com/deskflow/deskflow/actions/runs/14184097971/artifacts/2855504724) generated via [latest build of this pr](https://github.com/deskflow/deskflow/actions/runs/14184097971/)

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -29,7 +29,7 @@
 
 Windows will save to the install dir if settings are loaded from there. If not, it saves any other config files in: `C:\ProgramData\Deskflow\`
 
-When using settings from the install dir its not recommend to have elevation enabled.
+When using settings from the install dir, the service mode will not be available.
 
 # Server Config Examples
 
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -1,18 +1,17 @@
 # GUI Config
 
+ Deskflow will automaticlly figure out where to save settings and other files.
+
+
+## Unix Systems
  The search order for a setting file is:
- 1. `<install-path>/settings/Deskflow.conf`
  1. `<XDG_CONFIG_HOME>/Deskflow/Deskflow.conf`
  1. A user settings file
  1. A system settings file
  
  A new settings file will be created in the user path if no settings file is found.
  The path of the settings file will be used as the base for all other config files.
- 
-### Windows
- - System: `C:\ProgramData\Deskflow\Deskflow.conf`
- - User: `C:\Users\userName\AppData\Local\Deskflow\Deskflow.conf`
- 
+
 ### Linux
  - System: `/etc/Deskflow/Deskflow.conf`
  - User: `~/.config/Deskflow/Deskflow.conf`
@@ -20,6 +19,17 @@
 ### macOS
  - System: `/Library/Deskflow/Deskflow.conf`
  - User: `~/Library/Deskflow/Deskflow.conf`
+ 
+
+## Windows
+
+ The search order for a setting file is:
+ 1. `<install-path>/settings/Deskflow.conf`
+ 1. Windows Registry `HKCU\Software\Deskflow\Deskflow`
+
+Windows will save to the install dir if settings are loaded from there. If not, it saves any other config files in: `C:\ProgramData\Deskflow\`
+
+When using settings from the install dir its not recommend to have elevation enabled.
 
 # Server Config Examples
 
--- a/src/lib/common/Settings.cpp
+++ b/src/lib/common/Settings.cpp
@@ -109,8 +109,9 @@ QVariant Settings::defaultValue(const QString &key)
   if (key == Server::Binary)
     return kServerBinName;
 
-  if (key == Core::ElevateMode)
+  if (key == Core::ElevateMode) {
     return Settings::ElevateMode::Always;
+  }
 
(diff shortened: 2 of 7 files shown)


robot3498712:
Deskflow: 1.21.1.5 (3b8c872c)
Qt: 6.8.2
System: Windows 10 Version 22H2

[2025-04-01T01:54:08] ERROR: secure socket error: tls certificate doesn't exist: C:/WINDOWS/system32/config/systemprofile/AppData/Local/Deskflow/tls/deskflow.pem
[2025-04-01T01:54:08] ERROR: could not load client certificates

Cert path: C:/Users/Admin/AppData/Local/Deskflow/tls/deskflow.pem (autofilled)

Daemon: always elevate (default setting)
UAC: second highest

automatic (as needed): "works" initially
As soon there is an elevation prompt it stops working (loss of control) and fails subsequently (TLS error as outlined). The workaround is Stop > Start, which fixes the TLS path. 

.. so from my point of view it looks like there is a path error which may be rather trivial to fix - "just" use the correct path as written to config? Log path in malfunctional state tells me somewhere it's trying to read (as SYSTEM) from C:/WINDOWS/system32/config/systemprofile , which is wrong - this path does not exist. 
Alternatively create the TLS keys somewhere else "world viewable"; I don't care about local compromise. What's important is that the clipboard is not published on LAN without any protection.

sithlord48:
That seams like the wrong version to me if you check click on the hash its another PR..

robot3498712:
Thanks! That seems to have done it.
Automatic (as needed): check
Always: check

(I had to manually restart the deskflow service.)

Deskflow: 1.21.1.7 (b2185350)
Qt: 6.8.2
System: Windows 10 Version 22H2

sithlord48:
yeah the service is a bit odd on reinstall sometimes.

doc/configuration.md
```diff
@@ -1,25 +1,34 @@
...
+ 
+
+## Windows
+
+ The search order for a setting file is:
+ 1. `<install-path>/settings/Deskflow.conf`
+ 1. Windows Registry `HKCU\Software\Deskflow\Deskflow`
+
+Windows will save to the install dir if settings are loaded from there if not it save any other config files in `C:\ProgramData\Deskflow\`
+When using settings from the install dir service mode will not be availble
```
nbolton, line 31:
```suggestion
Windows will save to the install dir if settings are loaded from there. If not, it saves any other config files in: `C:\ProgramData\Deskflow\`

When using settings from the install dir, the service mode will not be available.
```

src/lib/common/Settings.cpp
```diff
@@ -157,16 +166,25 @@ const QStringList Settings::validKeys()
 
 bool Settings::isWritable()
 {
+  if (instance()->isUsingRegistry())
+    return true;
   return instance()->m_settings->isWritable();
 }
 
+bool Settings::isUsingRegistry()
```
nbolton, line 174: I wonder if this might be confusing to a Linux developer who doesn't know this is referring to the Windows registry... perhaps we could surround it in an `#ifdef Q_OS_WIN` or rename the function to `isUsingWindowsRegistry` or perhaps `isNativeFormat`?

nbolton (commented):
Quick review. I will test this shortly.

nbolton:
> fixes: #8345

How can this be fixed? This would require changes to the daemon. I think there is some confusion here cause by the off-topic comments which should have been raised as a new issue.

nbolton:
Fresh install. Daemon log tail isn't working for me.

image

sithlord48:
> Fresh install. Daemon log tail isn't working for me.

Clean machine ? i wonder if the path does not exsist yet . ill check that

nbolton (approved):
Strange. I managed to get the daemon to freeze up by setting elevate mode to 'As needed' then running an elevated process to show UAC. The client didn't re-launch and the daemon became unresponsive. I then saw an IPC error. Clicking 'Apply' in the GUI seemed to un-freeze it, but after trying a 2nd time I wasn't able to repro. 

Edit: This was on a fresh VM.

I noticed you made 'Always elevate' the default which will certainly make things simpler. At this point I'm think we just need to get rid of the 'As needed' option (maybe in a separate PR). 

Related: 
- #8350

PR #8439, opened 2025-04-02
Settings service update

sithlord48:
fixes: #8350 

 - Removes `Settings::Core::ElevateMode` and sets `Daemon::Elevate` Instead (default: true)
 - New Settings: `Settings::Core::StopOnDeskSwitch` (default: false) 

You can only select one of the option

### DIFF
--- a/src/lib/gui/dialogs/SettingsDialog.ui
+++ b/src/lib/gui/dialogs/SettingsDialog.ui
@@ -23,7 +23,7 @@
    <item>
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>1</number>
+      <number>0</number>
      </property>
      <widget class="QWidget" name="tabRegular">
       <property name="sizePolicy">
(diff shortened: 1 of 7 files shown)


src/lib/gui/dialogs/SettingsDialog.ui
```diff
@@ -541,110 +541,40 @@
...
-              &lt;ul&gt;
-              &lt;li&gt;Start the server or client automatically when the computer starts.&lt;/li&gt;
-              &lt;li&gt;Run the server or client in an elevated mode (e.g. on login screen, UAC dialogs, etc).&lt;/li&gt;
-              &lt;/ul&gt;
-              </string>
+             <string/>
             </property>
             <property name="text">
-             <string>Use background service (daemon)</string>
+             <string>Stop on desk switch</string>
```
nbolton, line 569: I tested with this and it does still auto-elevate, but only when you disable the other 'elevate' option. So, when you enable the elevate option, it becomes redundant (the setting effectively does nothing). 

If we keep this option, I think it should be grayed out when the elevate option is checked.

```suggestion
             <string>Run as system only when at login screen and UAC</string>
```

However, in it's current state it feels like a bit of a hack. The behavior is, when not elevated, if the `winlogon.exe` process is found, it will elevate the process automatically:

https://github.com/deskflow/deskflow/blob/fbc9168c041bce26dd309ec82a47baa3ffdc33bc/src/lib/platform/MSWindowsWatchdog.cpp#L156

Therefore, the other elevate option in it's current wording and behavior is a bit misleading; if you de-select system/elevate and select stop on desk switch, it will still elevate as needed.

nbolton (changes requested):
TL;DR: We could land this but the checkbox wording and behavior should change. If we keep the hacky 'stop on desk switch' behavior then we need to explain how it works and that it will still elevate the process (as needed) if selected.

Removing it is an option but this will create dead code that we should remove, possibly in a new PR. The code we'd need to remove would be in the watchdog:
https://github.com/deskflow/deskflow/blob/fbc9168c041bce26dd309ec82a47baa3ffdc33bc/src/lib/platform/MSWindowsWatchdog.cpp#L156

It's one of those... 'clever but stupid' features.

nbolton (approved):

nbolton:
FTR, we decided to remove the `--stop-on-desk-switch` feature in a later PR:
- ##8458

PR #8454, opened 2025-04-06
Rm switchon desk

### LINKED ISSUE #8458: Remove Windows auto-elevate option through `--stop-on-desk-switch`
Problem: After discussing on [Matrix](https://matrix.to/#/#deskflow-dev:matrix.org), we decided that auto-elevating for UAC and login is pointless. I originally implemented this (15 years ago) because I didn't want the Core always to run as elevated, and wanted it to only run as elevated when necessary (UAC dialog and login screen). This seemed like it would be more secure, but this doesn't make sense because it is still running elevated when the system is most vulnerable, so it might give the user a false sense of security. Auto-elevating also causes problems like the connection abruptly being dropped when the UAC dialog shows or the user switches to the login screen. On top of this, the complexity of the auto-elevate code also causes some bugs:
- #8228
- #8345

Solution: It's much simpler for the daemon to either elevate the process all the time or none of the time. This makes sense because users who are security conscious wouldn't even want the Core process running elevated at the login screen.

sithlord48:
requires: https://github.com/deskflow/deskflow/pull/8519
fixes #8458

 A Start to removing the switch on desk option.

 Done:
   - Remove from settings
   - Remove from Args
   - Remove from MSWindowsScreen 
   - Remove from MSWindowsWatchdog

### DIFF
--- a/src/lib/platform/MSWindowsWatchdog.cpp
+++ b/src/lib/platform/MSWindowsWatchdog.cpp
@@ -149,13 +149,9 @@ MSWindowsWatchdog::getUserToken(LPSECURITY_ATTRIBUTES security, bool elevatedTok
 {
   m_session.updateActiveSession();
 
-  // always elevate if we are at the vista/7 login screen. we could also
-  // elevate for the uac dialog (consent.exe) but this would be pointless,
-  // since deskflow would re-launch as non-elevated after the desk switch,
-  // and so would be unusable with the new elevated process taking focus.
-  if (elevatedToken || m_session.isProcessInSession("logonui.exe", nullptr)) {
+  if (elevatedToken) {
 
-    LOG_DEBUG("getting elevated token, %s", (elevatedToken ? "elevation required" : "at login screen"));
+    LOG_DEBUG("getting elevated token");
 
     HANDLE process;
     if (!m_session.isProcessInSession("winlogon.exe", &process)) {
@@ -290,16 +286,9 @@ void MSWindowsWatchdog::startProcess()
   } else {
     LOG_DEBUG("starting new process in user session");
 
-    LOG_DEBUG("getting active desktop name");
-    const auto activeDesktopName = runActiveDesktopUtility();
-
-    LOG_DEBUG("active desktop name: %s", activeDesktopName.c_str());
-    // When we're at a UAC prompt, lock screen, or the login screen, Windows switches to the Winlogon desktop.
-    const auto isOnSecureDesktop = activeDesktopName == "Winlogon";
-
     SECURITY_ATTRIBUTES sa;
     ZeroMemory(&sa, sizeof(SECURITY_ATTRIBUTES));
-    HANDLE userToken = getUserToken(&sa, isOnSecureDesktop || m_elevateProcess);
+    HANDLE userToken = getUserToken(&sa, m_elevateProcess);
 
     // set UIAccess to fix Windows 8 GUI interaction
     DWORD uiAccess = 1;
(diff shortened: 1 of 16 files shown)


nbolton:
Ah, nice, didn't see this until after I raised the issue:
- #8458

nbolton:
There is also some code in the watchdog to remove on or near this line:
https://github.com/deskflow/deskflow/blob/fbc9168c041bce26dd309ec82a47baa3ffdc33bc/src/lib/platform/MSWindowsWatchdog.cpp#L156

I think `m_session.isProcessInSession("logonui.exe", nullptr)` can go, and one of the lines below:
`if (!m_session.isProcessInSession("winlogon.exe", &process)) {`

There's probably lots of other code in the Daemon/watchdog we can remove too, but it may take some time to discover.

sithlord48:
Feel free to take this over @nbolton

nbolton:
Thinking on this, I'll do some testing and land as is, then we can remove the rest of the dead code later in another PR.

src/lib/platform/MSWindowsWatchdog.cpp
```diff
@@ -438,53 +431,6 @@ void MSWindowsWatchdog::shutdownExistingProcesses()
...
-std::string MSWindowsWatchdog::runActiveDesktopUtility()
-{
-  char fileNameBuffer[MAX_PATH];
-  GetModuleFileName(NULL, fileNameBuffer, MAX_PATH);
-  std::string fileName(fileNameBuffer);
-  size_t lastSlash = fileName.find_last_of("\\");
-  const auto installDir = fileName.substr(0, lastSlash);
-
-  const auto coreBinPath = installDir + "\\deskflow-server.exe";
-  std::string utilityCommand = "\"" + coreBinPath + "\" --active-desktop";
```
nbolton, line 450: I think we can also get rid of this now:
https://github.com/deskflow/deskflow/blob/5b4ba198d547b1b616728a796a01315bc7e64225/src/apps/deskflow-server/deskflow-server.cpp#L33-L55

nbolton (approved):
A bit more dead code to remove, but we can do that in a new PR if you want.

src/lib/platform/MSWindowsWatchdog.cpp
```diff
@@ -153,7 +153,7 @@ MSWindowsWatchdog::getUserToken(LPSECURITY_ATTRIBUTES security, bool elevatedTok
   // elevate for the uac dialog (consent.exe) but this would be pointless,
   // since deskflow would re-launch as non-elevated after the desk switch,
   // and so would be unusable with the new elevated process taking focus.
-  if (elevatedToken || m_session.isProcessInSession("logonui.exe", nullptr)) {
+  if (elevatedToken) {
 
     LOG_DEBUG("getting elevated token, %s", (elevatedToken ? "elevation required" : "at login screen"));
```
nbolton, line 158:
```suggestion
    LOG_DEBUG("getting elevated token");
```

nbolton (changes requested):
Actually, I think this log could be improved.

src/lib/platform/MSWindowsWatchdog.cpp
```diff
@@ -153,7 +153,7 @@ MSWindowsWatchdog::getUserToken(LPSECURITY_ATTRIBUTES security, bool elevatedTok
   // elevate for the uac dialog (consent.exe) but this would be pointless,
   // since deskflow would re-launch as non-elevated after the desk switch,
   // and so would be unusable with the new elevated process taking focus.
```
nbolton, line 155: Oh, and this entire comment is now redundant.

sithlord48:
I've removed those additional items, If you see anymore lmk

nbolton (approved):
Tested Windows client at:
- startup login screen
- lock screen
- UAC prompt

Working well :+1:

Issue #8456, opened 2025-04-07
Source paths are no longer indented in debug builds


nbolton:
### Project

Deskflow

### Deskflow version info

Deskflow: 1.21.1.27 (1ace03d4)
Qt: 6.8.2
System: Fedora Linux 41 (Workstation Edition)
Session: GNOME (wayland)

### Build type

Local developer build (built it myself)

### What steps will reproduce the problem?

1. Build debug
2. Look at log in GUI

Expected: Neat output
Actual: Messy; source paths not indented

### Log output

```
[2025-04-07T10:33:07] INFO: network encryption protocol: TLSv1.3
/home/nick/Projects/deskflow/src/lib/net/SslLogger.cpp:110
[2025-04-07T10:33:07] NOTE: accepted client connection
/home/nick/Projects/deskflow/src/lib/server/ClientListener.cpp:158
[2025-04-07T10:33:09] NOTE: new client disconnected
/home/nick/Projects/deskflow/src/lib/server/ClientProxyUnknown.cpp:278
[2025-04-07T10:33:10] NOTE: peer fingerprint: (SHA1) D7:A3:E3:E6:8D:3D:31:55:00:C5:DA:04:FD:2A:07:52:2A:3D:BB:B9 (SHA256) 63:68:81:DB:C6:B3:D1:97:B0:3D:93:DA:F2:AF:4E:D3:0B:D2:5A:D1:7B:67:5E:81:E6:2D:F1:41:D7:F8:E9:52
[2025-04-07T10:33:10] DEBUG: fingerprint is trusted
	/home/nick/Projects/deskflow/src/lib/gui/MainWindow.cpp:791
/home/nick/Projects/deskflow/src/lib/net/SecureSocket.cpp:663
[2025-04-07T10:33:10] INFO: accepted secure socket
/home/nick/Projects/deskflow/src/lib/net/SecureSocket.cpp:459
[2025-04-07T10:33:10] INFO: network encryption protocol: TLSv1.3
/home/nick/Projects/deskflow/src/lib/net/SslLogger.cpp:110
[2025-04-07T10:33:10] NOTE: accepted client connection
/home/nick/Projects/deskflow/src/lib/server/ClientListener.cpp:158
```

PR #8457, opened 2025-04-07
fix: Remove unnecessary log line endings and trimming

nbolton:
Fixes: #8456


src/lib/gui/Logger.cpp
```diff
@@ -32,10 +32,8 @@ QString printLine(FILE *out, const QString &type, const QString &message, const
   auto logLine = QString("[%1] %2: %3").arg(datetime, type, message);
 
   QTextStream stream(&logLine);
-  stream << Qt::endl;
-
   if (!fileLine.isEmpty()) {
-    stream << "\t" + fileLine << Qt::endl;
+    stream << Qt::endl << "\t" + fileLine;
```
sithlord48, line 36: use '\n' in place of 'Qt::endl' , 'Qt::endl' also flushes the output

Issue #8032, opened 2025-01-02
Persist permission dialog in XDG Desktop Portal (Wayland, etc)


nbolton:
### Feature Request

The [XDG Desktop Portal](https://github.com/flatpak/xdg-desktop-portal) dialog shows every time Deskflow needs input capture or remote desktop access. This is annoying to users, who want the dialog to show only once.

Part of the work needs to be done in XDG Desktop Portal, but also needs to be done in compositors like [Mutter](https://en.wikipedia.org/wiki/Mutter_(software)) and [KWin](https://en.wikipedia.org/wiki/KWin)).

Progress has been made already by @whot in XDG Desktop Portal as a draft PR, which will need to be completed:
- https://github.com/flatpak/xdg-desktop-portal/pull/1186

> the xdg-desktop-portal part of this is effectively done, but the compositor
> integration of this is what is missing. That's also the hardest bit and
> requires re-doing in every compositor (well, at least mutter, kwin, and
> sway to make most ppl happy) to make it work across the board.

Maintainers: This issue represents the entire work (including work to be done outside of this repo) so that a bounty can be created. Do not close as "not our bug".

nbolton:
/bounty $2500

> [!IMPORTANT]
> This bounty has been split in two, counterpart: 
> - https://github.com/deskflow/deskflow/issues/8869

> [!IMPORTANT]
> The bounty split ($2,500 per issue) is pending a support request open with Algora.

This bounty encompasses landing all necessary PRs in:
- [XDG Desktop Portal](https://flatpak.github.io/xdg-desktop-portal/)
- [Mutter](https://gitlab.gnome.org/GNOME/mutter) and [KWin](https://invent.kde.org/plasma/kwin)
- Deskflow (this project)

The bounty is for all work needed to get the XDG Desktop Portal permission dialog to persist for input capture ~~and remote desktop~~ on Mutter and KWin.

Anshgrover23:
/assign me  /attempt #8032

<div id="algora-attempt" />

| [Algora profile](https://console.algora.io/@/Anshgrover23) | Completed bounties | Tech | Active attempts | Options |
| --- | --- | --- | --- | --- |
| @Anshgrover23 | 29 bounties from 4 projects | <div align="center">TypeScript, JavaScript, <br />Go & more</div> | <div align="center"></div> | [Cancel attempt](https://console.algora.io/api/bounties/cm5fc43ss000jl80337fu5wqc/cancel-attempt) |

[...]

PR #8503, opened 2025-04-17
Fix: Cursor stuck off server screen (Wayland, attempt)

mrnicegyu11:
### tl;dr

It can happen that the libeis implementation (read: the wayland compositor) sends a `start_emulating` event to a device already emulating. This is a protocol violation of libei/libeis. The client<-->server connection is severed (which is correct behavior), but deskflow never releases the captured xdg-desktop-portal-input-capture. Thus, the cursor remains perpetually stuck. The fix is to release the portal-input-capture when deskflow handles the `EI_EVENT_DISCONNECT` event.

#### The issue as mentioned in #8005

The bug causes the mouse cursor to become stuck at the edge of the server screen (most often the top, but sometimes the sides) when using Deskflow on Wayland. When this happens, all mouse and keyboard input on the server is frozen, and the only way to recover is to SSH in and kill the deskflow-server process. Occasionally, instead of freezing, the cursor may jump from one edge to another or move along the edge but remains trapped. Logs show the system repeatedly trying and failing to move the cursor offscreen.

#### Reproducing the issue

I am working on KDE/plasma and arch-linux.

For me, the bug in a production build of deskflow, and also on a debug build, happens very rarely (~once per day or less). Thus, I first sought to reproduce the issue more reliably.

At some point in my investigation I have decided to attach valgrind to deskflow to look for memory leaks. Valgrind seems to slow the execution of deskflow, and by chance on my system makes the issue more reproducible. When valgrind is attached to deskflow-server, the transmission of events to the client is slowed. 

Here, there is an interesting finding: I think there is probably some blocking code in deskflow, and when this code is executed deskflow will not respond to the client with heartbeats.
Explanation: When setting short heartbeat times (<1 s) in the deskflow-server, the client will often disconnect due to the server being unresponsive within the heartbeat, and looking at the server's `DEBUG2` level logs shows that the server is busy processing the incoming mouse-events, unable to respond. In this case, the bug *cannot* be reproduced at all. But, meanwhile deskflow is barely usable due to the frequent resets. 

But, if one sets longer heartbeats, like "5000" (==5s), and attaches valgrind to a debug build of deskflow to slow the server down a bit, one can reproduce the rare issue by "wiggling" the cursor with small (or circular) movements between the two screens (back-and-forth). Then, it shows as follows: *Sometimes*, the cursor is stuck, sometime it is *not* stuck, but when trying to enter the client's screen the server & client connections are capped, and the server idles and logs the characteristic log-lines "event: motion on primary" and "event queue write result: 1" ([see](https://github.com/deskflow/deskflow/issues/8005#issuecomment-2590678945)).

Understanding the root-cause of this bug (well, the one that shows in my valgrind&debug-build), has taken me some time. My setup is a debug build of deskflow, executed with the command `DEBUGINFOD_URLS="https://debuginfod.archlinux.org/" valgrind --leak-check=full --show-leak-kinds=all --track-origins=yes --log-file=valgrind.log --verbose ./build/bin/deskflow-server -f --debug DEBUG2 --name linux --log deskflow.log --disable-client-cert-check --address 0.0.0.0:24800 --no-daemon --config ./deskflow-server.conf` .From the `DEBUG2` level logs of deskflow I could see that for this reproducible variant of the bug, the log-lines 
```
[2025-04-16T14:26:40] DEBUG1: ei: dispatching ei_device.start_emulating() on object 0xff00000000000007
	/home/dustin/git/personalForks/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-16T14:26:40] ERROR: ei: Invalid device state 3 for a start_emulating event
	/home/dustin/git/personalForks/deskflow/src/lib/platform/EiScreen.cpp:105
[2025-04-16T14:26:40] WARNING: ei: Connection error: Invalid device state 3 for a start_emulating event.
```
are always logged. 

Here, it starts getting messy: I don't know why the other people reporting the bug in the bugticket did not get these log-lines. @pruriggro [here](https://github.com/deskflow/deskflow/issues/8005#issuecomment-2590678945) has gotten two subsequent `EI_EVENT_DEVICE_START_EMULATING` events, which could be consistent with what I observed. My logs were more verbose though. `libei` might not log as part of deskflow on other people's build? I don't fully get this, in the hope of not chasing a different bug here, I attribute this to my quite custom environment. 

OK, back to the topic: Device state Nr. 3 is "EMULATING" (see enum here https://gitlab.freedesktop.org/libinput/libei/-/blob/main/src/libei-device.h?ref_type=heads#L36 ), and libei(s) also specify in their spec that calling `start_emulating()` on a device that is already emulating is a protocol violation (see https://libinput.pages.freedesktop.org/libei/interfaces/ei_device/index.html#ei_devicestart_emulating). So this should not ever happen. I still couldn't fully wrap my head around the full inner workings of xdg-desktop-portal-InputCapture and libei(s), but as I understand it, the `xdg-desktop-portal` via a (gtk)callback tells `libeis` (read: the wayland compositor) to start capturing the input device. So at this point deskflow is not really involved, *but*: Deskflow captures and uses the file-descriptors of the xdg-desktop-portal and handles the libei (client-side) implementation. Reading the cautious comments in `EiEventQueueBuffer.cpp`, I think that due to the "akward" and specifically not ordering-retaining handling of the queue/buffer there, we might get into the situation of start_emulating() being called on a device that is already emulating... 

In any case, the `deskflow-server` code does not itself call `start_emulating()` on the device, the wayland compositor in the end causes this libei(s) protocol violation. So, at this point, the best `deskflow-server` can do is handle it properly, and this is where I think something is not right. 

One key step for understanding the behaviour was obtaining debug logs from the xdg-desktop-portal by running `dbus-monitor --session "interface='org.freedesktop.portal.InputCapture'"` in a seperate terminal. This showed that when the issue (calling `start_emulating()` on an already emulating device) happens, the xdg-portal-input-capture, which causes the pointer to vanish from the server's screen, is never released. It perpetually stays in the "activated" state. See this video that I awkwardly took with my phone because I screwed up my ssh connection: 

https://github.com/user-attachments/assets/a27e8ca4-dd60-438b-a93b-e6b26e03f7d5
(note in the video that I can't click at all, but when I wiggle the cursor the plasma/KDE animation [huge cursor] still triggers)

In the video, one can see that the xdg-portal-InputCapture was activated with a negative (?) pixel value, which is for me so far consistent in the case when the cursor on the server-screen disappears. I tried to find out if negative values are explicitly (dis)allowed or reasonable, but didnt find conclusive evidence. Maybe someone here knows how this is in xdg/wayland? According to https://wayland.app/protocols/xdg-shell#xdg_positioner , `xdg_positioner` disallows negative values. However, wayland subsurfaces seem to allow negative values (see https://wayland-client-d.dpldocs.info/wayland.client.protocol.wl_subsurface_set_position.html). I am a bit confused here tbh.

After reading a lot of code, it occurred to me that when the protocol violation occurs, the libeis (read: wayland compositor-side) semi-gracefully stops the connection to the libei-client (read: deskflow), by sending the following signals/events:

`EI_DEVICE_STATE_EMULATING → EI_DEVICE_STATE_REMOVED_FROM_CLIENT → EI_DEVICE_STATE_DEAD → EI_EVENT_DISCONNECT`

Here, this is where I see the root cause of the bug. In the handling of `EI_EVENT_DISCONNECT` on the side of libei/deskflow, the xdg-portal-InputCapture is not released, and thus the cursor stays stuck. If we release it as part of the handling of `EI_EVENT_DISCONNECT`, together with some other minor code improvements there, we can have deskflow handle the severed EI<-->EIS connection gracefully. The mouse/keyboard on the server is usable, and deskflow will try to re-establish the InputCapture (on KDE/plasma popping up the dialog window again) when one tries to enter the client screen.

### Further issues identified

One time (when the discussed bug was not hit at all actually), valgrind caught a segfault in deskflow (without any codechanges form my side, debug built). So, it is still possible that I caught a different bug that only is visible in slow debug builds, and the actual "cursor stuck" bug comes simply from undefined behavior due to a corrupted heap. This was a KDE/plasma wayland server with a windows client. For the record, here is what valgrind logged before dying due to the segfault:

```
==31020== Invalid read of size 8
==31020==    at 0x25086A: ClientProxy1_6::handleClipboardSendingEvent(Event const&, void*) (ClientProxy1_6.cpp:54)
==31020==    by 0x250D51: TMethodEventJob<ClientProxy1_6>::run(Event const&) (TMethodEventJob.h:49)
==31020==    by 0x13E633: EventQueue::dispatchEvent(Event const&) (EventQueue.cpp:242)
==31020==    by 0x13DD68: EventQueue::loop() (EventQueue.cpp:107)
==31020==    by 0x1B8578: ServerApp::mainLoop() (ServerApp.cpp:750)
==31020==    by 0x1B8B88: ServerApp::standardStartup(int, char**) (ServerApp.cpp:801)
==31020==    by 0x1BEDAC: standardStartupStatic(int, char**) (AppUtilUnix.cpp:36)
==31020==    by 0x1B89B4: ServerApp::runInner(int, char**, int (*)(int, char**)) (ServerApp.cpp:782)
==31020==    by 0x1BEDF7: AppUtilUnix::run(int, char**) (AppUtilUnix.cpp:41)
==31020==    by 0x15B1EB: App::run(int, char**) (App.cpp:99)
==31020==    by 0x131908: main (deskflow-server.cpp:58)
==31020==  Address 0x9419ad0 is 0 bytes inside a block of size 456 free'd
==31020==    at 0x48498DD: operator delete(void*, unsigned long) (vg_replace_malloc.c:1181)
==31020==    by 0x251378: ClientProxy1_8::~ClientProxy1_8() (ClientProxy1_8.h:15)
==31020==    by 0x23FC41: Server::handleClientDisconnected(Event const&, void*) (Server.cpp:1336)
==31020==    by 0x24B4E9: TMethodEventJob<Server>::run(Event const&) (TMethodEventJob.h:49)
==31020==    by 0x13E633: EventQueue::dispatchEvent(Event const&) (EventQueue.cpp:242)
==31020==    by 0x13DD68: EventQueue::loop() (EventQueue.cpp:107)
==31020==    by 0x1B8578: ServerApp::mainLoop() (ServerApp.cpp:750)
==31020==    by 0x1B8B88: ServerApp::standardStartup(int, char**) (ServerApp.cpp:801)
==31020==    by 0x1BEDAC: standardStartupStatic(int, char**) (AppUtilUnix.cpp:36)
==31020==    by 0x1B89B4: ServerApp::runInner(int, char**, int (*)(int, char**)) (ServerApp.cpp:782)
==31020==    by 0x1BEDF7: AppUtilUnix::run(int, char**) (AppUtilUnix.cpp:41)
==31020==    by 0x15B1EB: App::run(int, char**) (App.cpp:99)
==31020==  Block was alloc'd at
==31020==    at 0x4845F93: operator new(unsigned long) (vg_replace_malloc.c:487)
==31020==    by 0x214526: ClientProxyUnknown::initProxy(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int, int) (ClientProxyUnknown.cpp:197)
==31020==    by 0x214958: ClientProxyUnknown::handleData(Event const&, void*) (ClientProxyUnknown.cpp:240)
==31020==    by 0x214F09: TMethodEventJob<ClientProxyUnknown>::run(Event const&) (TMethodEventJob.h:49)
==31020==    by 0x13E633: EventQueue::dispatchEvent(Event const&) (EventQueue.cpp:242)
==31020==    by 0x291805: StreamFilter::filterEvent(Event const&) (StreamFilter.cpp:89)
==31020==    by 0x26DF77: PacketStreamFilter::filterEvent(Event const&) (PacketStreamFilter.cpp:182)
==31020==    by 0x291859: StreamFilter::handleUpstreamEvent(Event const&, void*) (StreamFilter.cpp:94)
==31020==    by 0x291993: TMethodEventJob<StreamFilter>::run(Event const&) (TMethodEventJob.h:49)
==31020==    by 0x13E633: EventQueue::dispatchEvent(Event const&) (EventQueue.cpp:242)
==31020==    by 0x13DD68: EventQueue::loop() (EventQueue.cpp:107)
==31020==    by 0x1B8578: ServerApp::mainLoop() (ServerApp.cpp:750)
==31020== 
==31020== 
==31020== Process terminating with default action of signal 11 (SIGSEGV): dumping core
==31020==  Bad permissions for mapped region at address 0x29F6D8
==31020==    at 0x29F6D8: ??? (in /home/REDACTED_USERNAME/git/personalForks/deskflow/build/bin/deskflow-server)
==31020==    by 0x250D51: TMethodEventJob<ClientProxy1_6>::run(Event const&) (TMethodEventJob.h:49)
==31020==    by 0x13E633: EventQueue::dispatchEvent(Event const&) (EventQueue.cpp:242)
==31020==    by 0x13DD68: EventQueue::loop() (EventQueue.cpp:107)
==31020==    by 0x1B8578: ServerApp::mainLoop() (ServerApp.cpp:750)
==31020==    by 0x1B8B88: ServerApp::standardStartup(int, char**) (ServerApp.cpp:801)
==31020==    by 0x1BEDAC: standardStartupStatic(int, char**) (AppUtilUnix.cpp:36)
==31020==    by 0x1B89B4: ServerApp::runInner(int, char**, int (*)(int, char**)) (ServerApp.cpp:782)
==31020==    by 0x1BEDF7: AppUtilUnix::run(int, char**) (AppUtilUnix.cpp:41)
==31020==    by 0x15B1EB: App::run(int, char**) (App.cpp:99)
==31020==    by 0x131908: main (deskflow-server.cpp:58)
```

I would really hope that this actually is not the real root cause, and we can simply ignore this one-time find for now, as debugging this heap corruption will be fairly painful imho...

If you read all of this, thanks for your attention guys, let me know if you have further thoughts/ideas!

/claim #8005

### Suggested next steps

I have some confidence that this MR fixes an actual (albeit rare) bug. I have personally not observed the wayland-server cursor getting stuck bug with my code, but since it was anyway rare for me on the release build with KDE/plasma, I can't be sure. So I would suggest to merge and test-drive this PR and see if it might fix the bug :--)

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -711,8 +711,9 @@ void EiScreen::handle_connected_to_eis_event(const Event &event, void *)
 
 void EiScreen::handle_portal_session_closed(const Event &event, void *)
 {
-  // Portal may or may EI_EVENT_DISCONNECT us before sending the DBus Closed
-  // signal Let's clean up either way.
+  // Portal may or may not EI_EVENT_DISCONNECT us before sending the DBus Closed
+  // signal. Let's clean up either way.
+  LOG_DEBUG("eis screen handling portal session closed");
   cleanup_ei();
   init_ei();
 }
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -145,7 +145,7 @@ void EiScreen::cleanup_ei()
     ei_device_set_user_data(ei_abs_, nullptr);
     ei_abs_ = ei_device_unref(ei_abs_);
   }
-  ei_seat_unref(ei_seat_);
+  ei_seat_ = ei_seat_unref(ei_seat_);
   for (auto it = ei_devices_.begin(); it != ei_devices_.end(); it++) {
     free(ei_device_get_user_data(*it));
     ei_device_set_user_data(*it, nullptr);
@@ -721,7 +721,6 @@ void EiScreen::handle_portal_session_closed(const Event &event, void *)
 void EiScreen::handleSystemEvent(const Event &sysevent, void *)
 {
   std::lock_guard lock(mutex_);
-  bool disconnected = false;
 
   // Only one ei_dispatch per system event, see the comment in
   // EiEventQueueBuffer::addEvent
@@ -767,10 +766,20 @@ void EiScreen::handleSystemEvent(const Event &sysevent, void *)
       // We're using libei which emulates the various seat/device remove events
       // so by the time we get here our EiScreen should be in a neutral state.
       //
-      // We don't do anything here, we let the portal's Session.Closed signal
-      // handle the rest.
-      LOG_WARN("disconnected from eis");
-      disconnected = true;
+      // We must release the xdg-portal InputCapture in case it is still active
+      // so that the cursor is usable and not stuck on the deskflow server.
+      LOG_WARN("disconnected from eis, will afterwards commence attempt to reconnect");
+      if (is_primary_) {
+        LOG_DEBUG("re-allocating portal input capture connection and releasing active captures");
+        if (portal_input_capture_) {
+          if (portal_input_capture_->is_active()) {
+            portal_input_capture_->release();
+          }
+          delete portal_input_capture_;
+          portal_input_capture_ = new PortalInputCapture(this, this->events_);
+        }
+      }
+      this->handle_portal_session_closed(sysevent, nullptr);
       break;
     case EI_EVENT_DEVICE_PAUSED:
       LOG_DEBUG("device %s is paused", ei_device_get_name(device));
@@ -826,9 +835,6 @@ void EiScreen::handleSystemEvent(const Event &sysevent, void *)
     }
     ei_event_unref(event);
   }
-
-  if (disconnected)
-    ei_ = ei_unref(ei_);
 }
 
 void EiScreen::updateButtons()


prurigro:
FYI: The build fails for me unless I add a semicolon after `LOG_DEBUG("re-allocating portal input capture connection and releasing active captures")`

nbolton:
@whot This PR may be of interest to you.

nbolton:
This might be a red herring, since I accidentally ran the Core twice (killed the GUI but did not also kill the Core) but the bug is still happening for me:
https://github.com/user-attachments/assets/f9747e2b-79ca-4fa3-80a9-803743286494

I will keep testing and see if the issue happens without multiple Core processes running (as I know that the bug still happens when only one process is running).

mrnicegyu11:
@nbolton if you want you could potentially run `dbus-monitor --session "interface='org.freedesktop.portal.InputCapture'"` in a background shell (maybe piping the output to a file). It would be interesting to know if when the bug happens the xdg input portal capture is still active ;)

sithlord48:
This seams to include some commits it should not, can you make sure to rebase this upon the current master branch as well as correct the linting.

mrnicegyu11:
> This seams to include some commits it should not, can you make sure to rebase this upon the current master branch as well as correct the linting.

Will do of course, I merged upstream master, that can be undone. And I'll make the linting pass:--) thanks for highlighting

sithlord48:
We do not allow merges you must rebase.

mrnicegyu11:
@sithlord48 yes, sorry about this, I will correct it and force push a cleaned version to this PR/fork

mrnicegyu11:
OK, in the force-push I have:
- rebase'd onto upstream/master (@sithlord48 )
- Corrected the old comment `we do nothing` --> `we release the input capture if it is active` ( @prurigro )
- Changed the code formatting to make the linter happy

:--)

nbolton:
I've been testing this fork for a week and the only time I saw the cursor getting stuck off screen was when I had 2 processes running by accident.

@sithlord48 Before we land this PR, I think we should implement Core process de-dupe with Qt `LocalServer`/`LocalSocket`. Thoughts?

sithlord48:
> @sithlord48 Before we land this PR, I think we should implement Core process de-dupe with Qt LocalServer/LocalSocket. Thoughts?

Yeah that is a good idea idk that it has to happen before this but that should be the final part.  Honestly i think todo it and not  make it annoying we should first switch to the unified core

mrnicegyu11:
@nbolton I didnt use conventional commit messages before, this is new to me tbh :) I will adjust the messages and force push once more

src/lib/platform/EiScreen.cpp
```diff
@@ -145,7 +145,9 @@ void EiScreen::cleanup_ei()
     ei_device_set_user_data(ei_abs_, nullptr);
     ei_abs_ = ei_device_unref(ei_abs_);
   }
-  ei_seat_unref(ei_seat_);
+  if (ei_seat_) {
+    ei_seat_unref(ei_seat_);
+  }
```
whot, line 150: all of libei's `unref` functions are NULL-safe, this one isn't needed.

However what you probably *do* want to do is:
```c
ei_seat_ = ei_seat_unref(ei_seat_);
```

because this will set it to NULL, otherwise you potentially have a dangling pointer.

whot:
> sends a start_emulating event to a device already emulating.

fwiw this really shouldn't be possible using the libeis API, can you file a bug in libei please and I'll have a look at it.

This issue may also hint at a bug in the compositor because if the ei connection is terminated it *should* send the `Closed` signal and this issue should've never shown up in deskflow. As it is from a review of the patches (I've paged most of the details out by now, sorry) this approach looks correct though. Do make sure though all the signals pertaining to the current session are disconnected too (probably done in the destructor anyway, but worth double-checking) - this way we won't get a delayed Closed event on an already destroyed session.

src/lib/platform/EiScreen.cpp
mrnicegyu11: good finding thx, I addressed this and forcepushed the PR

mrnicegyu11:
@whot @nbolton 
I changed the commits to:
- have the messages adhere to conventional commits
- include the suggestion from @whot for cleaning up the ei_seat

I will open another bug-ticket in libei for the "emulating-while-already-emulating" behaviour :--)

mrnicegyu11:
I have opened a bug ticket in the `libei` repo for the "start-emulating-on-already-emulating-device" observation: https://gitlab.freedesktop.org/libinput/libei/-/issues/78

mrnicegyu11:
@sithlord48 

tldr: ready to review/merge.


From my side this PR is ready to be reviewed/merged. I am confident that this PR fixes an actual bug, and on my setup I didn't encounter the cursor being stuck anymore with this fix. I guess that @nbolton and maybe @sithlord48 also tested this for a bit? Anyway,  for the hypothetical upstream patch in libei, as I understood it, they will for now only check "why" the technically not-allowed thing of "START_EMULATING" being called on a already emulating device is happening. That is either in libeis code, or in the composer (said @whot ). While this might fix the same cursor-being-stuck bug (from the other side so to say), I would still suggest to merge my PR, as what it actually does is properly clean-up and release the xdg-input-capture-portal when it is necessary during tear-down of the libei<-->libeis connection severance. And also, if there are further issues that cause similar bugs where the cursor is stuck, neither my PR nor the upstream libei patch would fix that at this point.

So from my side, green light go ahead :--)

nbolton:
@prurigro Do you still see the bug with this PR?

sithlord48:
i will test it when Its ready for review.

nbolton:
> If you read all of this

It was an interesting read. Thank you for the detail. It's interesting that using `valgrind` makes the bug more reproducible.

Annoyingly, I have not seen this bug in `master` for a while, so it's difficult to prove that this fixes the bug, hence why I need some anecdotal support from users. 

I'll try the valgrind trick to see if I can get it to repro.

sithlord48 (approved):
Can't say for sure this will fix the issue but testing its not making anything worse..  that is a +1 from me

nbolton:
@sithlord48 I want to try with valgrind first before we land. 
My plan is: if I can repro the bug with using valgrind on master but not on the branch, then we'll land it.

nbolton:
> * Debug server build of deskflow
> * Valgrind attached to deskflow-server with full options (see my initial post)
> * Client screen connected at top of server screen
> * Wiggle cursor in circular across screen-border
> * Within 1-3min I could usually get the issue

Perfect, thanks, will try that first thing tomorrow.

whot:
A related side-note: libeis is a fairly simple library and using [dbusmock](https://github.com/martinpitt/python-dbusmock) you can mock DBus interfaces to emulate a portal. (liboeffis, xdg-desktop-portal and libportal all use this approach, x-d-p has the more modern code)

libei also has a python wrapper around the C library in its test suite which can be used for ideas on how to create a custom python binding and of course there's [snegg](https://gitlab.freedesktop.org/libinput/snegg).

Since you don't care about the output bits much it would be relatively trivial to create a tool and/or test suite that scaffolds the required settings, pretends to be a compositor and you can then run deskflow against it. This would allow you to recreate whatever custom behaviour you need to reproduce a bug. Sounds like a PITA at first but long term it'll save you quite a bit of time[^1].


[^1]: coincidentally, that's also one of the motivations for my old request to document the protocol: if you have a test suite that can speak the protocol you can test each component individually.

whot:
A note on this comment from the OP:
> @pruriggro https://github.com/deskflow/deskflow/issues/8005#issuecomment-2590678945 has gotten two subsequent EI_EVENT_DEVICE_START_EMULATING events, which could be consistent with what I observed.

Checking that comment: the events are on two different objects (`0xff00000000000007` and `0xff00000000000003`, likely mouse and keyboard) so this is valid. The emulating state is per-device and deskflow will simply start/stop both keyboard and mouse.

Anyway, I built 35f0e9e6e44c940a41df0438d5eaacf95ec4fb2d today on two fedora 42 machines and I've failed to reproduce the issue (client on top of server, with release/debug builds, with/without valgrind and 5s heartbeats). So at least for the libei parts of this issue I'd need the full deskflow log at `DEBUG2` of a sequence that triggered the issue.

tbc: I still think that this PR is *a* correct fix but I can't comment on whether it's *the* correct fix :)

nbolton:
Managed to reproduce it on `master`. I actually had to attempt this 3 times... and after 4m30s I was just about to give up and then it happened.

@whot Here's the log:
```
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:677
[2025-04-30T10:38:39] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:39] DEBUG: event: motion on primary x=565 y=0)
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:677
[2025-04-30T10:38:39] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:39] DEBUG: event: motion on primary x=565 y=0)
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:677
[2025-04-30T10:38:39] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:39] DEBUG: event: motion on primary x=565 y=0)
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:677
[2025-04-30T10:38:39] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:39] DEBUG: event: motion on primary x=565 y=0)
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:677
[2025-04-30T10:38:39] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:39] DEBUG: event: motion on primary x=565 y=0)
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:677
[2025-04-30T10:38:39] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:39] DEBUG: event: motion on primary x=565 y=0)
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:677
[2025-04-30T10:38:39] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:39] DEBUG: event: motion on primary x=565 y=0)
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:677
[2025-04-30T10:38:39] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:39] DEBUG: event: motion on primary x=565 y=0)
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:677
[2025-04-30T10:38:39] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:39] DEBUG: event: motion on primary x=565 y=0)
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:677
[2025-04-30T10:38:39] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:39] DEBUG2: find neighbor on up of "spock-tux"
	/home/nick/Projects/deskflow/src/lib/server/Server.cpp:606
[2025-04-30T10:38:39] DEBUG2: "tuvok-win" is on up of "spock-tux" at 0.164390
	/home/nick/Projects/deskflow/src/lib/server/Server.cpp:629
[2025-04-30T10:38:39] DEBUG1: try to leave "spock-tux" on up
	/home/nick/Projects/deskflow/src/lib/server/Server.cpp:794
[2025-04-30T10:38:39] INFO: switch from "spock-tux" to "tuvok-win" at 315,1079
	/home/nick/Projects/deskflow/src/lib/server/Server.cpp:461
[2025-04-30T10:38:39] INFO: leaving screen
	/home/nick/Projects/deskflow/src/lib/deskflow/Screen.cpp:136
[2025-04-30T10:38:39] DEBUG1: modifiers on update: 0x0000
	/home/nick/Projects/deskflow/src/lib/deskflow/KeyState.cpp:770
[2025-04-30T10:38:39] DEBUG: ignored screen "spock-tux" update of clipboard 0 (unchanged)
	/home/nick/Projects/deskflow/src/lib/server/Server.cpp:1488
[2025-04-30T10:38:39] DEBUG: ignored screen "spock-tux" update of clipboard 1 (unchanged)
	/home/nick/Projects/deskflow/src/lib/server/Server.cpp:1488
[2025-04-30T10:38:39] DEBUG1: send enter to "tuvok-win", 315,1079 459 0000
	/home/nick/Projects/deskflow/src/lib/server/ClientProxy1_0.cpp:238
[2025-04-30T10:38:39] DEBUG2: writef(CINN%2i%2i%4i%2i)
	/home/nick/Projects/deskflow/src/lib/deskflow/ProtocolUtil.cpp:74
[2025-04-30T10:38:39] DEBUG: portal cb activated, id=493
	/home/nick/Projects/deskflow/src/lib/platform/PortalInputCapture.cpp:272
[2025-04-30T10:38:39] DEBUG2: wrote 14 bytes
	/home/nick/Projects/deskflow/src/lib/deskflow/ProtocolUtil.cpp:124
[2025-04-30T10:38:39] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:39] DEBUG2: writef(CALV)
	/home/nick/Projects/deskflow/src/lib/deskflow/ProtocolUtil.cpp:74
[2025-04-30T10:38:39] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:39] DEBUG2: wrote 4 bytes
	/home/nick/Projects/deskflow/src/lib/deskflow/ProtocolUtil.cpp:124
[2025-04-30T10:38:39] DEBUG2: event queue read result: 64
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:85
[2025-04-30T10:38:39] DEBUG1: ei: dispatching ei_device.start_emulating() on object 0xff00000000000003
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] ERROR: ei: Invalid device state 3d for a start_emulating event
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:105
[2025-04-30T10:38:39] WARNING: ei: Connection error: Invalid device state 3d for a start_emulating event
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:102
[2025-04-30T10:38:39] DEBUG1: ei: device 0xff00000000000003: EI_DEVICE_STATE_EMULATING → EI_DEVICE_STATE_REMOVED_FROM_CLIENT
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: sending: object 0xff00000000000004 (ei_pointer@v1:release(0)) signature ''
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: sending: object 0xff00000000000005 (ei_scroll@v1:release(0)) signature ''
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: sending: object 0xff00000000000006 (ei_button@v1:release(0)) signature ''
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: sending: object 0xff00000000000003 (ei_device@v2:release(0)) signature ''
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: deregistering ei_pointer v1 object 0xff00000000000004
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: deregistering ei_scroll v1 object 0xff00000000000005
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: deregistering ei_button v1 object 0xff00000000000006
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: deregistering ei_device v2 object 0xff00000000000003
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: queuing event type EI_EVENT_DEVICE_REMOVED (6)
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: device 0xff00000000000003: EI_DEVICE_STATE_REMOVED_FROM_CLIENT → EI_DEVICE_STATE_DEAD
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: device 0xff00000000000007: EI_DEVICE_STATE_EMULATING → EI_DEVICE_STATE_REMOVED_FROM_CLIENT
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: sending: object 0xff00000000000008 (ei_keyboard@v1:release(0)) signature ''
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: sending: object 0xff00000000000007 (ei_device@v2:release(0)) signature ''
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: deregistering ei_keyboard v1 object 0xff00000000000008
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: deregistering ei_device v2 object 0xff00000000000007
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: queuing event type EI_EVENT_DEVICE_REMOVED (6)
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: device 0xff00000000000007: EI_DEVICE_STATE_REMOVED_FROM_CLIENT → EI_DEVICE_STATE_DEAD
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: queuing event type EI_EVENT_SEAT_REMOVED (4)
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: deregistering ei_seat v1 object 0xff00000000000002
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: sending: object 0xff00000000000000 (ei_connection@v1:disconnect(1)) signature ''
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: queuing event type EI_EVENT_DISCONNECT (2)
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG1: ei: Connection dispatch: CONNECTED -> DISCONNECTING
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:38:39] DEBUG: removing device captured relative pointer
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:492
[2025-04-30T10:38:39] DEBUG: logical output size: 1x1@-1.-1
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:436
[2025-04-30T10:38:39] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:39] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:39] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:39] DEBUG: removing device captured keyboard
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:492
[2025-04-30T10:38:39] DEBUG: logical output size: 1x1@-1.-1
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:436
[2025-04-30T10:38:39] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:39] WARNING: disconnected from eis
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:771
[2025-04-30T10:38:39] DEBUG2: msg from "tuvok-win": CNOP
	/home/nick/Projects/deskflow/src/lib/server/ClientProxy1_0.cpp:133
[2025-04-30T10:38:39] DEBUG2: no-op from
	/home/nick/Projects/deskflow/src/lib/server/ClientProxy1_0.cpp:186
[2025-04-30T10:38:39] DEBUG2: msg from "tuvok-win": CALV
	/home/nick/Projects/deskflow/src/lib/server/ClientProxy1_0.cpp:133
[2025-04-30T10:38:39] DEBUG2: msg from "tuvok-win": CNOP
	/home/nick/Projects/deskflow/src/lib/server/ClientProxy1_0.cpp:133
[2025-04-30T10:38:39] DEBUG2: no-op from
	/home/nick/Projects/deskflow/src/lib/server/ClientProxy1_0.cpp:186
[2025-04-30T10:38:39] DEBUG: screen "spock-tux" shape changed
	/home/nick/Projects/deskflow/src/lib/server/Server.cpp:1165
[2025-04-30T10:38:39] DEBUG2: onMouseMoveSecondary initial +0,+0
	/home/nick/Projects/deskflow/src/lib/server/Server.cpp:1795
[2025-04-30T10:38:39] DEBUG1: DESKFLOW_MOUSE_ADJUSTMENT not set, using original values +0,+0
	/home/nick/Projects/deskflow/src/lib/server/Server.cpp:1810
[2025-04-30T10:38:39] DEBUG: screen "spock-tux" shape changed
	/home/nick/Projects/deskflow/src/lib/server/Server.cpp:1165
[2025-04-30T10:38:39] DEBUG2: onMouseMoveSecondary initial +0,+0
	/home/nick/Projects/deskflow/src/lib/server/Server.cpp:1795
[2025-04-30T10:38:39] DEBUG1: DESKFLOW_MOUSE_ADJUSTMENT not set, using original values +0,+0
	/home/nick/Projects/deskflow/src/lib/server/Server.cpp:1810
[2025-04-30T10:38:39] DEBUG2: event queue read result: 64
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:85
[2025-04-30T10:38:39] DEBUG2: event queue read result: 64
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:85
[2025-04-30T10:38:39] DEBUG2: event queue read result: 64
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:85
[2025-04-30T10:38:39] DEBUG2: event queue read result: 41
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:85
[2025-04-30T10:38:41] DEBUG2: writef(CALV)
	/home/nick/Projects/deskflow/src/lib/deskflow/ProtocolUtil.cpp:74
[2025-04-30T10:38:41] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:41] DEBUG2: wrote 4 bytes
	/home/nick/Projects/deskflow/src/lib/deskflow/ProtocolUtil.cpp:124
[2025-04-30T10:38:41] DEBUG2: event queue read result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:85
[2025-04-30T10:38:41] DEBUG2: event queue read result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:85
[2025-04-30T10:38:41] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:41] DEBUG2: writef(CALV)
	/home/nick/Projects/deskflow/src/lib/deskflow/ProtocolUtil.cpp:74
[2025-04-30T10:38:41] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:38:41] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
```

nbolton:
Continging from my last, I'll start a new comment for results testing `mrnicegyu11:2025/fix/waylandServerCursorStuck` branch:

I think I managed to "reproduce" the condition and get different behavior. Instead of the cursor being stuck off screen I now see the portal dialog:
image

And this log:
```
[2025-04-30T10:59:23] DEBUG1: ei: queuing event type EI_EVENT_DISCONNECT (2)
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:59:23] DEBUG1: ei: Connection dispatch: CONNECTED -> DISCONNECTING
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:59:23] DEBUG: removing device captured relative pointer
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:492
[2025-04-30T10:59:23] DEBUG: logical output size: 1x1@-1.-1
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:436
[2025-04-30T10:59:23] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:59:23] DEBUG: removing device captured keyboard
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:492
[2025-04-30T10:59:23] DEBUG: logical output size: 1x1@-1.-1
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:436
[2025-04-30T10:59:23] DEBUG2: event queue write result: 1
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:128
[2025-04-30T10:59:23] WARNING: disconnected from eis, will afterwards commence attempt to reconnect
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:771
[2025-04-30T10:59:23] DEBUG: re-allocating portal input capture connection and releasing active captures
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:773
[2025-04-30T10:59:23] DEBUG: releasing input capture session, id=230
	/home/nick/Projects/deskflow/src/lib/platform/PortalInputCapture.cpp:233
[2025-04-30T10:59:23] DEBUG: shutting down glib thread
	/home/nick/Projects/deskflow/src/lib/platform/PortalInputCapture.cpp:380
[2025-04-30T10:59:23] DEBUG1: thread 0x00000003 exit
	/home/nick/Projects/deskflow/src/lib/mt/Thread.cpp:135
[2025-04-30T10:59:23] DEBUG: eis screen handling portal session closed
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:716
[2025-04-30T10:59:23] DEBUG: adopting new buffer
	/home/nick/Projects/deskflow/src/lib/base/EventQueue.cpp:154
[2025-04-30T10:59:23] DEBUG: discarding 2 event(s)
	/home/nick/Projects/deskflow/src/lib/base/EventQueue.cpp:159
[2025-04-30T10:59:23] DEBUG1: ei: deregistering ei_connection v1 object 0xff00000000000000
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T10:59:23] DEBUG: adopting new buffer
	/home/nick/Projects/deskflow/src/lib/base/EventQueue.cpp:154
[2025-04-30T10:59:23] DEBUG1: thread 0x00000004 entry
	/home/nick/Projects/deskflow/src/lib/mt/Thread.cpp:133
[2025-04-30T10:59:23] DEBUG: glib thread running
	/home/nick/Projects/deskflow/src/lib/platform/PortalInputCapture.cpp:373
[2025-04-30T10:59:23] DEBUG: setting up input capture session
	/home/nick/Projects/deskflow/src/lib/platform/PortalInputCapture.cpp:198
```

So, I'm pretty damn confident this PR is good to land. Just a bit more testing to be sure.

nbolton (changes requested):
This PR is a significant improvement because the behavior is now different (portal dialog shows on `EI_EVENT_DISCONNECT`), so we're much closer to the fix. _Before landing this PR_, we should try to fix the cursor getting stuck intermittently on `EI_EVENT_DISCONNECT` (at the point where the portal dialog shows).

mrnicegyu11:
Hey guys I will catch up on this in a few hours sorry. Very happy that this is actually reproducible, thanks for the verbose logs & feedback @nbolton . I can confirm that the portal dialogue showing is what my PR "intended", so this is what i was expecting and observing as well. I guess this https://github.com/deskflow/deskflow/issues/8032 *might* address the portal dialogue showing again, but anyway since the portal-input-capture is deactivated, it is at least from the procedural code-flow imho ok that the pop-up shows again, albeit this is a bit annoying for the user still.
I never managed to get the cursor fully stack, I'll be happy to read through your second attempt logs :--)

nbolton:
> I guess this #8032 _might_ address the portal dialogue showing again

Bounty is still up for grabs ;-)

whot:
> Invalid device state 3d for a start_emulating event

that error message tells me you don't have libei 1.4.x installed yet (*oops, just saw your edit*). I don't *think* it makes a difference but worth trying. Also a few ei-related fixes went into mutter with [mutter!3839](https://gitlab.gnome.org/GNOME/mutter/-/issues/3839) so that is also worth testing (this is all in Fedora 42)

Any chance of getting the full log file? Because it's still just when the issue occurs but whatever gets out of sync is going to happen well before that, probably the previous time when you leave/enter the screen. Also, which host is the server and which one is the client - I'm guessing `spock-tux` is your linux server?

The issue starts at `[2025-04-30T10:46:25]` but after staring at libei I don't know how this message can arrive twice without a previous message getting lost - and that bit is in the cut-off part of the log file.

The other thing I noticed:
```
[2025-04-30T10:46:25] DEBUG: portal cb activated, id=34
...
[2025-04-30T10:46:25] DEBUG: releasing input capture session, id=34
...
[2025-04-30T10:46:25] DEBUG: portal cb activated, id=35
```
This looks wrong - the activated/release should trigger a corresponding START_EMULATING/STOP_EMULATING on the ei connection but it doesn't show up in the log. How that can happen I'm not sure though. libeis doesn't let you send two start_emulating events and that code hasn't changed since the beginning.

nbolton:
> ### ❌ 8th attempt: Stuck!!!
> I was about to give up and it finally happened again

@mrnicegyu11 2 out of 8 times, it was stuck while the dialog was shown. 

Perhaps you could also try running `stress` on top of your `valgrind` command to try and repro the 'stuck while portal dialog shown' bug.

> Any chance of getting the full log file?

@whot Sure, attached. I should have done that earlier.

[deskflow.log](https://github.com/user-attachments/files/19976184/deskflow.log)

IIRC, this 1st disconnect on L1615 did not cause the mouse to be stuck.
```
[2025-04-30T12:24:14] DEBUG1: ei: queuing event type EI_EVENT_DISCONNECT (2)
```

2nd time on L7661 the cursor became stuck while the portal dialog was shown:
```
[2025-04-30T12:27:36] DEBUG1: ei: queuing event type EI_EVENT_DISCONNECT (2)
```

The 3rd time on L8042 was due to me calling `kill` on my `valgrind` process via SSH.
```
[2025-04-30T12:28:04] DEBUG1: ei: queuing event type EI_EVENT_DISCONNECT (2)
```

>  Also, which host is the server and which one is the client - I'm guessing spock-tux is your linux server?

Server: `spock-tux` (Fedora 41)
Client: `tuvok-win` (Windows 11)

sithlord48:
i dont' think this has anything to do  with the dialog being persistant or not. you lost EI its asking to give it back.. You cursor should be free at that time. could be a bug in EI or the compositor not sure. but its not because the dialog does not save access...

mrnicegyu11:
I don't want to add noise, but @whot since you have mentioned that (let me paraphrase to see if I understood correctly) some signals or messages seem to arrive out-of-order, I remember reading this https://github.com/deskflow/deskflow/blob/59df2db7b7d892846e8774a593e82fc378c60a10/src/lib/platform/EiEventQueueBuffer.cpp#L95 comment in the `EiEventQueueBuffer.cpp` code, which seems to admit that things in the buffer for wayland/ei(s) specifically might not be handled according to the order in which they arrive. If I read the comment correctly. So if the queue, or the associated `fd` , gets .. "full?" maybe this causes the weird shuffling of messages? 

I will think about it later a bit, might edit this comment if it gets clearer to me ;)

sithlord48:
maybe worth noting today with libei 1.4.0-1 (arch package) i have lots my cursor twice already.. 

Deskflow: 1.21.2.95 (59df2db7)
Qt: 6.9.0
System: Arch Linux
Session: KDE (wayland)

whot:
Thanks for the logs, unfortunately, I'm still stuck.


Something is wrong with the way our pipe works: we tickle the pipe with a single character but the read func doesn't seem to be called (esp. at the beginning of the log), see:

```console
$ grep "event queue read result" deskflow.*
deskflow.1.log:[2025-04-30T12:23:49] DEBUG2: event queue read result: 64
deskflow.1.log:[2025-04-30T12:23:52] DEBUG2: event queue read result: 64
deskflow.1.log:[2025-04-30T12:23:55] DEBUG2: event queue read result: 64
deskflow.1.log:[2025-04-30T12:23:57] DEBUG2: event queue read result: 64
deskflow.1.log:[2025-04-30T12:24:03] DEBUG2: event queue read result: 64
deskflow.1.log:[2025-04-30T12:24:07] DEBUG2: event queue read result: 64
deskflow.1.log:[2025-04-30T12:24:11] DEBUG2: event queue read result: 64
deskflow.log:[2025-04-30T12:24:14] DEBUG2: event queue read result: 64
deskflow.log:[2025-04-30T12:24:16] DEBUG2: event queue read result: 4
deskflow.log:[2025-04-30T12:24:16] DEBUG2: event queue read result: 1
... mostly 1-4 events, but there's another 64 event block later
```

This indicates that we called `getEvent()` and `addEvent()` 64+ times until we got to the `waitForEvent()` callback. I've fixed the draining in #8556 but I'm not sure if this is an indication of ... anything. And admittedly I have since forgotten the details of the `getEvent()`/`waitForEvents()` etc. interactions.

In the logs in https://github.com/deskflow/deskflow/pull/8503#issuecomment-2841702394:

```
... the previous event comes in:

[2025-04-30T12:24:09] DEBUG1: ei: dispatching ei_device.start_emulating() on object 0xff00000000000003
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T12:24:09] DEBUG1: ei: queuing event type EI_EVENT_DEVICE_START_EMULATING (200)
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96

... and is processed by us:

[2025-04-30T12:24:10] DEBUG: device captured relative pointer started emulating
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:801
[2025-04-30T12:24:10] DEBUG: device captured keyboard started emulating
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:801

... InputCapture.Activated signal id=9

[2025-04-30T12:24:11] DEBUG: portal cb activated, id=9
	/home/nick/Projects/deskflow/src/lib/platform/PortalInputCapture.cpp:272

... released by us in 

[2025-04-30T12:24:12] DEBUG: releasing input capture at x=606 y=1
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:346
[2025-04-30T12:24:12] DEBUG: releasing input capture session, id=9 x=606.0 y=1.0
	/home/nick/Projects/deskflow/src/lib/platform/PortalInputCapture.cpp:240

===== but there is no stop_emulating here =====

... and then eventually we fail:

[2025-04-30T12:24:14] DEBUG2: event queue read result: 64
	/home/nick/Projects/deskflow/src/lib/platform/EiEventQueueBuffer.cpp:85
[2025-04-30T12:24:14] DEBUG1: ei: dispatching ei_device.start_emulating() on object 0xff00000000000003
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:96
[2025-04-30T12:24:14] ERROR: ei: Invalid device state 3d for a start_emulating event
	/home/nick/Projects/deskflow/src/lib/platform/EiScreen.cpp:105

... with the corresponding (but later arriving):

[2025-04-30T12:24:14] DEBUG: portal cb activated, id=10
	/home/nick/Projects/deskflow/src/lib/platform/PortalInputCapture.cpp:272
```
So based on these logs it's very definitely a protocol error on the ei side, there should be a `DEBUG1: ei: dispatching ei_device.stop_emulating()` at the point marked above. I just can't figure out how this could ever land on the wire given the current mutter and libeis code.


There's a bug in xdg-desktop-portal where a `Deactivated` would prevent the next `Activated` from being sent but this doesn't seem to be triggered here since the ids in the log are still incrementally going up -> https://github.com/flatpak/xdg-desktop-portal/pull/1703. If we had triggered this one we'd see the activation ids skipping values.

And I have a [PR for libei](https://gitlab.freedesktop.org/libinput/libei/-/merge_requests/330) to aid with debugging, this should print the sequence which may make it easier to correlate (should be the same as the activation id in mutter).

But at this point it looks like a libei bug, I just can't figure out how this happens (not with mutter using libeis, anyway).

> some signals or messages seem to arrive out-of-order, I remember reading this 

yeah, some of that is expected, some of that is part of deskflow's internal design.

What is expected is that ei events and portal signals arrive in any order. Mutter *does* send the signal first (in its code) but that DBus signal goes through the portal to deskflow. Meanwhile the `start_emulating` event is sent over the fd between mutter and deskflow so it's almost guaranteed to arrive first. This is effectively unfixable and a result of having two channels for communication - the good news is that it rarely matters: the activation id is mostly just used as info. In theory you could hold device events back until you see the signal but why bother.

What's harder to fix is the `getEvent()` approach with libei. deskflow assumes that there's a single queue that it can get the next event from and somehow those are ordered? In the XWindows implementation `addEvent()` sends a ClientMessage to the server so that data arrives back at deskflow as part of the normal event sequence, mixed in at the right-ish order. Then we pop the next event, pass it to the screen. So `getEvents()` sees, roughly, X events and Added events:
```
[X][X][A][X][X][X][A]...
```

In libei `getEvents()` sees, either "Ei has events" or an Addded events:
```
[EI][A][EI][EI][A]
```
And *each* of those `EI` events is not a single event but a notification that "ei has events". When that event is popped, we run `ei_get_event()` until the dispatch is empty but that *can* mess up our sequence - we cannot guarantee that the ei events processed are only those that happened before whatever next `A` event was put in the queue (and note: DBus portal events are some of those added events).

Realistically there are two ways around this: timestamp comparisons or take the ei events out of the queue during `getEvent()`, convert them into custom deskflow structs and pass those on, releasing the `ei_event`. Given the number of events we care about is quite small (see `EiScreen::handleSystemEvent`) this is the better approach.

nbolton:
> Realistically there are two ways around this: timestamp comparisons or take the ei events out of the queue during `getEvent()`, convert them into custom deskflow structs and pass those on, releasing the `ei_event`. Given the number of events we care about is quite small (see `EiScreen::handleSystemEvent`) this is the better approach.

Sounds reasonable. Is this best done in a new PR or in this PR?

FWIW, after rebasing in PR #8556, I reproduced the new 'cursor stuck while dialog portal shown' bug immediately without much difficulty today (with `valgrind` and `stress`). 

[deskflow.log](https://github.com/user-attachments/files/19998578/deskflow.log)

IDK if helpful, but here's the new code in action:
```
$ grep 'total drained: [2-9]' deskflow.log
[2025-05-01T10:27:58] DEBUG2: event queue read result: 0 (total drained: 2)
[2025-05-01T10:28:01] DEBUG2: event queue read result: 0 (total drained: 2)
[2025-05-01T10:28:03] DEBUG2: event queue read result: 0 (total drained: 2)
[2025-05-01T10:28:04] DEBUG2: event queue read result: 0 (total drained: 2)
[2025-05-01T10:28:12] DEBUG2: event queue read result: 0 (total drained: 2)
[2025-05-01T10:28:15] DEBUG2: event queue read result: 0 (total drained: 3)
```

whot:
fwiw, it's a race condition, #8556 shouldn't change much when it happens, I think it was coincidence that you reproduced it immediately.

> Edit: I suspect that the cursor gets stuck if it's off the screen when the bug happens.

base on the logs: when the cursor moves *away* from the server. compositor starts capturing, ei events start flowing (i.e. the compositor routes everything to us). Then the bug is triggered, ei disconnects and eventually(?) the compositor should disconnect us and Disable the portal. That appears to be a separate bug in mutter - if the eis connection is disconnected we should still complete the portal dance to shut down, we shouldn't get stuck (IOW this PR *shouldn't* be needed if the compositor works correctly but it's still good to have).

It looks like in this log we actually go the `Activated` signal before the ei events:
```
[2025-05-01T10:28:15] DEBUG: portal cb activated, id=51
	/home/nick/Projects/deskflow/src/lib/platform/PortalInputCapture.cpp:272
```

I recommend installing [this libei PR](https://gitlab.freedesktop.org/libinput/libei/-/merge_requests/330) so the error also includes the sequence id so we can match it to the Activated signal.

Looking at the code again, all these `motion on primary` are wrong - we should only get one (or a few if under stress) of these because the code immediately calls `portal->release()`. That should trigger the release in the compositor and a `stop_emulating` on the ei protocol. e.g. `id=47` where it still works there's nearly *a second* between those two events when really it should be a few ms. And in between those we get loads of ei events delivered.

I kinda wonder if mutter has priority on input events and prefers those over handling the dbus signal (but that's just idle speculation).

> Sounds reasonable. Is this best done in a new PR or in this PR?

New PR I guess? It's completely unrelated to this one, akin to code cleanup. It may help with this bug, it may not, but it should at least make debugging easier.

mrnicegyu11:
Hey @whot that's super interesting thanks a lot for all the background information. I tried 2 times for a few hours with stress and related tools but in my setup, on my PR code I cannot reproduce the cursor being stuck, even with tools like `stress` or trying to decrease network bandwidth. I still can manage to hit the known bug, and then in my PR's code the "Input Capture Requested"  windows pops up. I can move the mouse in the server and initiate the input capture again, and continue using 2 screens. So actually for it's well behaved atm 🙈 

So if the *remaining* part of the bug is a race condition in the libei(s) or compositor, I am not hitting it at all in my setup. For the record here is my system/build info:
```
- Server: Archlinux 6.14.4, Thinkpad X1 G11, wayland 1.23.1, kde-plasma 6.3.4., gcc15.1.1, xdg-desktop-portal 1.20.0-2
- Client: Windows 11, vanilla deskflow binary
```

Maybe @nbolton you can check if you are using different stuff. I am very curious, but I am bound to reading logs for this issue now since I can't manage to hit it. I am considering installing gnome though to check if it hits me then, after all this talk about mutter :D

sithlord48:
after using this i did not lose my cursor for the rest of the day ..

nbolton (approved):
Sounds like we're all pretty confident that this 'stuck while portal dialog' is an upstream race condition bug and cannot be fixed in this PR. Therefore I think it's time to land.

> after using this i did not lose my cursor for the rest of the day ..

It happens for me very rarely. Like @whot said, the fact that it happened to me first thing is probably pure coincidence.

FWIW, my plan is to build and use `libei` with [PR 330](https://gitlab.freedesktop.org/libinput/libei/-/merge_requests/330) and upload the logs ~here~ (to issue #8559) for @whot. But I don't think that should block the land.

whot:
> the compositor should disconnect us and Disable the portal. That appears to be a separate bug in mutter 

Filed as https://gitlab.gnome.org/GNOME/mutter/-/issues/4088

PR #8518, opened 2025-04-21
Save the host name if the lineEdit is not empty

### LINKED ISSUE #8517: Client hostname is not saved in config on exit
### Project

Deskflow

### Deskflow version info

Deskflow: 1.21.2.19 (f5f3bc04)
Qt: 6.8.2
System: Windows 11 Version 24H2

### Build type

### Operating systems (OS)

- [x] Windows
- [ ] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### OS versions/distros

Windows 11

### Deskflow configuration

Windows client

1. Choose client option
2. Enter hostname
3. Quit GUI
4. Start GUI

Expect: Hostname is as before quit
Actual: Hostname field is empty (has to be typed again on each GUI start)

### Additional information

~~Maybe not Windows bug, just found it on Windows.~~

Edit: Seems to be just Windows client? Odd.

sithlord48:
requires: https://github.com/deskflow/deskflow/pull/8519
fixes #8517

the previous logic would only save if the setting was already not empty, when we want to only save if the lineHostname is not empty

### DIFF
--- a/src/lib/gui/MainWindow.cpp
+++ b/src/lib/gui/MainWindow.cpp
@@ -697,7 +697,7 @@ void MainWindow::saveSettings()
   } else if (ui->rbModeServer->isChecked()) {
     Settings::setValue(Settings::Core::CoreMode, Settings::CoreMode::Server);
   }
-  if (!Settings::value(Settings::Client::RemoteHost).isNull())
+  if (!ui->lineHostname->text().isEmpty())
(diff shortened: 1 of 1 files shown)


nbolton (approved):
Oh yeah, that makes more sense. Hah.

PR #8540, opened 2025-04-25
fix: add reuse info to **.md by default

sithlord48:
- reuse lint is upset we landed a non reuse compliant md file 
 - rather then have ever md file trigger a rebuild we will auto assign `**/*.md` to MIT this is inline with our ci filter to not run on those files.


sithlord48:
Landing to fix the build

PR #8549, opened 2025-04-27
Use nullptr where possible

sithlord48:
- replaces 'NULL' with `nullptr` where we can


src/lib/arch/win32/ArchNetworkWinsock.cpp
```diff
@@ -713,21 +713,21 @@ std::vector<ArchNetAddress> ArchNetworkWinsock::nameToAddr(const std::string &na
...
-  assert(addr != NULL);
+  assert(addr != nullptr);
 
   char host[1024];
   char service[20];
   int ret =
       getnameinfo(TYPED_ADDR(struct sockaddr, addr), addr->m_len, host, sizeof(host), service, sizeof(service), 0);
 
-  if (ret != NULL) {
+  if (ret != 0) {
```
nbolton, line 730: Good catch. `NULL` is bad here.

PR #8553, opened 2025-04-30
chore: Silence two compiler signedness warnings

whot:
Both trigger:
  warning: comparison of integer expressions of different signedness: ‘int’ and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]

In the first hunk we already checked for i < 0 so casting to size_t is fine. In the second hunk we don't use i so we can pick the right type.

### DIFF
--- a/src/lib/base/String.cpp
+++ b/src/lib/base/String.cpp
@@ -56,7 +56,7 @@ std::string vformat(const char *fmt, va_list args)
           index.push_back(i);
           pos.push_back(static_cast<size_t>((scan - 1) - fmt));
           width.push_back(static_cast<size_t>((end - scan) + 2));
-          if (i > maxIndex) {
+          if (static_cast<size_t>(i) > maxIndex) {
             maxIndex = i;
           }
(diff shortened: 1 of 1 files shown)


src/lib/base/String.cpp
```diff
@@ -56,7 +56,7 @@ std::string vformat(const char *fmt, va_list args)
           index.push_back(i);
           pos.push_back(static_cast<size_t>((scan - 1) - fmt));
           width.push_back(static_cast<size_t>((end - scan) + 2));
-          if (i > maxIndex) {
+          if ((size_t)i > maxIndex) {
```
sithlord48, line 59: I will be working to remove these "C" style casts, Can we do `static_cast<size_t>(i)` instead ?

sithlord48 (dismissed):

src/lib/base/String.cpp
whot: sure, done now.

nbolton (approved):

PR #8563, opened 2025-05-02
Deskflow-gui add `reset` option

sithlord48:
Remove the use of the environmental variable `DESKFLOW_RESET_ALL` and reset instead when --reset is passed as an option.


nbolton (changes requested):
The 'Help' -> 'Clear settings' stopped working because it uses `--no-reset`
https://github.com/deskflow/deskflow/blob/50fc1b15d04e2ddae9681483d789092d3100f5e5/src/lib/gui/Diagnostic.cpp#L22-L23

nbolton (approved):

PR #8567, opened 2025-05-03
Better detect windows Arm64 on both CI and local builds

sithlord48:
When testing locally with win arm64 the detection was incorrect. I have adjusted the check to look for the CI's ENV var then try to use the CMAKE_SYSTEM_PROCESSOR when detecting the cpu type on windows

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,8 +96,14 @@ set(REQUIRED_LIBPORTAL_VERSION 0.8)
 set(REQUIRED_QT_VERSION 6.7.0)
 
 if (WIN32)
+  # VSCMD_ARG_TGT_ARCH is set on CI
   if ("$ENV{VSCMD_ARG_TGT_ARCH}" STREQUAL "")
-    set (BUILD_ARCHITECTURE x64)
+    # NOT on CI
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -96,8 +96,14 @@ set(REQUIRED_LIBPORTAL_VERSION 0.8)
 set(REQUIRED_QT_VERSION 6.7.0)
 
 if (WIN32)
-  if ("$ENV{VSCMD_ARG_TGT_ARCH}" STREQUAL "")
-    set (BUILD_ARCHITECTURE x64)
+  # VSCMD_ARG_TGT_ARCH is set on CI
+  if (NOT $ENV{VSCMD_ARG_TGT_ARCH})
```
nbolton, line 100: It seems this does _not_ evaluate to true when the `VSCMD_ARG_TGT_ARCH` env var isn't set. Did you mean `NOT DEFINED` instead? In any case, does this handle when the env is set but to empty string? It seems to me that `"$ENV{VSCMD_ARG_TGT_ARCH}" STREQUAL ""` is safer, though not as pretty.

sithlord48: No should be the same but we can just do STREQUAL "" to be safe .

nbolton (approved):

Issue #8459, opened 2025-04-07
Remove drag and drop


sithlord48:
### Feature Request

Drag and drop is broken on all platforms in one case or another. Remove this broken feature

PR #8569, opened 2025-05-05
Remove drag and drop support

sithlord48:
fixes: #8459 
 - Remove Drag and Drop
 - Rename OSXDragSimulator => OSXCocoaApp 
    - Keep the parts to start / stop the cocoa app (needed for client / server to work) 
    - Remove the parts related to DnD

Tested: 
 -  [X] Linux host (wayland)
 -  [x] Linux client (wayland)
 -  [x] Linux host (X)
 -  [x] Linux client (X)
 -  [x] Windows host
 -  [X] Windows client 
 -  [X] Mac host
 -  [X] Mac client


nbolton:
Tested and working:
- [x] macOS host
- [ ] macOS client
- [ ] Windows host
- [ ] Windows client
- [ ] Linux host
- [ ] Linux client

src/lib/platform/OSXCocoaApp.m
```diff
@@ -0,0 +1,45 @@
...
+void runCocoaApp()
+{
+  NSAutoreleasePool *pool = [[NSAutoreleasePool alloc] init];
+
+  [NSApplication sharedApplication];
+
+  NSWindow *window = [[NSWindow alloc] initWithContentRect:NSMakeRect(0, 0, 3, 3)
+                                                 styleMask:NSBorderlessWindowMask
+                                                   backing:NSBackingStoreBuffered
+                                                     defer:NO];
```
nbolton, line 28: IIRC, `NSWindow *window` was a dummy window that was positioned under the cursor when the cursor went to the edge of the screen. Its purpose was to capture dragged files when they hit the screen edge.

Can we remove this? It seems redundant.

src/lib/platform/OSXCocoaApp.m
```diff
@@ -0,0 +1,45 @@
...
+
+#import <Cocoa/Cocoa.h>
+#import <CoreData/CoreData.h>
+#import <Foundation/Foundation.h>
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+
+NSWindow *g_dragWindow = NULL;
+NSString *g_ext = NULL;
```
nbolton, line 17: Unused.
```suggestion
```

nbolton (dismissed):
Fine to land as it's working but maybe we can reduce `src/lib/platform/OSXCocoaApp.m` or remove it entirely...

src/lib/platform/OSXCocoaApp.m
sithlord48: i have to check more but removal of the run / stop is what was causing mac os client / server to not work correctly.

nbolton: Odd. Seems like a code smell. You'll notice this dummy window is only 3px in height and width. Maybe we can try to remove in another PR if it's going to be some work to figure out why it's "needed" for normal mouse and keyboard sharing behavior to work.

sithlord48: I agree. We should look to minimize this as much as we can .

sithlord48: Up to you if you want to land this now and do that in another pr or attempt it here. I would say another PR maybe eaiser to follow the "Fix"

nbolton (approved):

nbolton (approved):

PR #8580, opened 2025-05-08
Fix code smells in GUI

tcanabrava:


sithlord48:
Can you fix them to use our style https://github.com/deskflow/deskflow/wiki/Hacking-Guide
Ill check the code asap.

src/lib/gui/MainWindow.cpp
```diff
@@ -1014,24 +1014,14 @@ void MainWindow::updateLocalFingerprint()
...
-      showConfigureServer(
-          tr("Please add the server (%1) to the grid.").arg(Settings::value(Settings::Core::ScreenName).toString())
-      );
-      break;
-    }
+  int res = m_serverConfig.autoAddScreen(name);
+  const auto possible_vals = QList<int>({kAutoAddScreenManualServer, kAutoAddScreenManualClient});
+  if (possible_vals.contains(res)) {
+    showConfigureServer(
+        tr("Please add the server (%1) to the grid.").arg(Settings::value(Settings::Core::ScreenName).toString())
```
nbolton, line 1023: I think the original intent here was to say either "add server" or "add client", and not always "add server" (which wouldn't make sense when adding a client).

sithlord48: That is correct the in the Old code it should never happen for a in the case of a client adding a server. 


The text should be "Please add the client (%1) to the grid." ...

sithlord48: i modified this please recheck

nbolton: Yeah new code makes sense.

sithlord48 (approved):
Looks good to me 
@nbolton  any idea what those methods that were not used in mainwindow were for ?

sithlord48:
> > Looks good to me @nbolton any idea what those methods that were not used in mainwindow were for ?
> 
> Probably leftovers of removed code. Since compilers don't complain about unimplemented methods, they just stuck for years.

Well yeah but what were they for when they were implemented. (i.e do we need to implement them again )

nbolton (approved):
Not tested, ETIME, but LGTM

PR #8592, opened 2025-05-12
fix: Disable the save button when config not writable

nbolton:
Todo:
- [x] Test

### DIFF
--- a/src/lib/gui/dialogs/SettingsDialog.cpp
+++ b/src/lib/gui/dialogs/SettingsDialog.cpp
@@ -252,6 +252,8 @@ void SettingsDialog::updateControls()
   const bool serviceChecked = ui->groupService->isChecked();
   const bool logToFile = ui->cbLogToFile->isChecked();
 
+  ui->buttonBox->button(QDialogButtonBox::Save)->setEnabled(writable);
+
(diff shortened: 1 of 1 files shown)


nbolton:
I noticed if I make the file read only, then only the preferences are shown as read only. 

I do wonder if we can simplify all of this by showing an error on save or even on load if the settings are read only:
> Your settings are read-only and you will be unable to save your settings, please check your permissions.

src/lib/gui/dialogs/SettingsDialog.cpp
```diff
@@ -252,6 +252,12 @@ void SettingsDialog::updateControls()
   const bool serviceChecked = ui->groupService->isChecked();
   const bool logToFile = ui->cbLogToFile->isChecked();
 
+  const auto saveButton = ui->buttonBox->button(QDialogButtonBox::Save);
+  if (!saveButton) {
```
sithlord48, line 256: You only need to do `ui->buttonBox-button(QDialogButtonBox::Save)->setEnabled(writable);`
The button will always exist its made on construction and its around until destruction.

sithlord48 (changes requested):

sithlord48 (approved):
Ok

PR #8593, opened 2025-05-13
build(vcpkg): Set `github-binarycache` on `vcpkg-action`

nbolton:
I wonder if we should be setting `github-binarycache`?

Both examples in the readme say to set it:
https://github.com/johnwason/vcpkg-action

And it appears to be reccomended:
```
    # "Use vcpkg built-in GitHub binary caching if "true". If not specified, will use the dry-run based file cache."
    # Recommended set to "true"
    github-binarycache: ''
```

### DIFF
--- a/.github/actions/install-dependencies/action.yml
+++ b/.github/actions/install-dependencies/action.yml
@@ -94,6 +94,7 @@ runs:
         extra-args: --classic
         triplet: x64-windows-release
         token: ${{ github.token }}
+        github-binarycache: true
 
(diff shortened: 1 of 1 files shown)


sithlord48:
I was not able to get that to work using the binary cache built in to vcpkg (we used to do this, but in two manual steps ). When that failed i forked the action an used the "normal" github cache instead.

sithlord48:
Consider deleting the branch

Issue #8587, opened 2025-05-10
ERROR: cannot read configuration [path]: read error: line 7: unknown argument "none"


fengqing11:
### Project

Deskflow

### Deskflow version info

https://github.com/deskflow/deskflow/releases/tag/v1.21.2

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [x] Windows
- [ ] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [x] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [x] I am not using Wayland on Linux

### Signing on macOS

- [ ] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### Continuous build

- [ ] I have tried the latest continuous build and the issue persists
- [ ] I am unable to try the latest continuous build

### OS versions/distros

win11

### Deskflow configuration

```
section: screens
	DESKTOP-TGKAA6T:
		ctrl = shift
		alt = shift
		meta = shift
		super = shift
		none = shift
		halfDuplexCapsLock = false
		halfDuplexNumLock = false
		halfDuplexScrollLock = false
		xtestIsXineramaUnaware = false
		switchCorners = none
		switchCornerSize = 0
	DESKTOP-QTEJK2B:
		ctrl = shift
		alt = shift
		meta = shift
		super = shift
		none = shift
		halfDuplexCapsLock = false
		halfDuplexNumLock = false
		halfDuplexScrollLock = false
		xtestIsXineramaUnaware = false
		switchCorners = none
		switchCornerSize = 0
end


section: aliases
end

section: links
	DESKTOP-TGKAA6T:
		right = DESKTOP-QTEJK2B
	DESKTOP-QTEJK2B:
		left = DESKTOP-TGKAA6T
end

section: options
	protocol = barrier
	relativeMouseMoves = false
	win32KeepForeground = false
	disableLockToScreen = false
	clipboardSharing = true
	clipboardSharingSize = 3072
	switchCorners = none 
	switchCornerSize = 0
end
```

### Log output

```shell
[2025-05-10T17:28:33] ERROR: cannot read configuration "C:/ProgramData/Deskflow/deskflow-server.conf": read error: line 7: unknown argument "none"
[...]

PR #8594, opened 2025-05-13
Do not use SHA1

sithlord48:
Remove most support for SHA1

No Longer:
 - Write SHA1 hash to localDB
 - Send a SHA1 when sending peer fingerprints.
- Preview SHA1 in the Fingerprint preview

We still support SHA1 As a valid digest type in fingerprint. We can recognize them but don't use them anymore.

### DIFF
--- a/src/lib/gui/widgets/FingerprintPreview.cpp
+++ b/src/lib/gui/widgets/FingerprintPreview.cpp
@@ -12,57 +12,36 @@
 
 #include <net/SecureUtils.h>
 
-FingerprintPreview::FingerprintPreview(QWidget *parent, const QList<Fingerprint> &fingerprints) : QFrame(parent)
+FingerprintPreview::FingerprintPreview(QWidget *parent, const Fingerprint &fingerprint) : QFrame(parent)
 {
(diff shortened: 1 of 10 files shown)


src/lib/gui/widgets/FingerprintPreview.cpp
```diff
@@ -12,57 +12,36 @@
...
-  frameSha1->setLayout(sha1Layout);
+  setLayout(fingerprint.type == Fingerprint::Type::SHA256 ? sha256Layout(fingerprint) : emptyLayout());
 
+  adjustSize();
+  setFixedSize(size());
+}
+
+QLayout *FingerprintPreview::emptyLayout()
+{
+  auto *label = new QLabel(tr("Invalid Hash format"));
```
nbolton, line 29: Why the random caps?
```suggestion
  auto *label = new QLabel(tr("Invalid hash format"));
```

nbolton:
I'm blocked by: #8587

nbolton (dismissed):

nbolton (approved):

PR #8596, opened 2025-05-14
fix settings modifiers initilization, move initization to header

sithlord48:
fixes: #8587

My mistake there was a small error i missed at review time in https://github.com/deskflow/deskflow/pull/8580 that initialized our modifiers to 0.
This fixes that initialization moves them to the header and removes the now no needed init() method


nbolton:
Unable to test, blocked:
- #8598

PR #8603, opened 2025-05-17
Directy use of console output in logOutputters

sithlord48:
Between the 3 `IArchConsole` subclasses only one hand any code. The only implemented method was `ArchStdConsole::write`. To simplify the code base, I've removed all the `IArchConsole` subclasses and now do what `ArchStdConsole::write` was doing directly in the `ConsoleLogOutputter::write` method.

Removes: 
 - `IArchConsole`
 - `ArchStdConsole`
 - `ArchConsoleUnix`
 - `ArchConsoleWindows`

Added a few `// do nothing` comments to empty overrides in logoutputters.


nbolton (approved):
That was a lot of boilerplate.

PR #8607, opened 2025-05-20
More Sonar cleanup

sithlord48:
Fix more sonar issues
  - Daemon and subclasses should pass `const & func`
  - Use C++17 style if / switch initilizers where we can 
  - more Const methods
  - start to use even more `auto` / `const auto`
  - replace use of `typedef` with `using`
  - minor typo fix from `colom` to `colon`
  - Define only one var per line
  - Add more `// do nothing` for empty methods.

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -30,9 +30,11 @@
 #include <unistd.h>
 #include <vector>
 
+// Values are in pixels
 struct ScrollRemainder
 {
-  double x, y; // scroll remainder in pixels
+  double x;
+  double y;
 };
 
 namespace deskflow {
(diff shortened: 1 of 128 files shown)


src/lib/platform/EiScreen.cpp
```diff
@@ -32,7 +32,8 @@
 
 struct ScrollRemainder
 {
-  double x, y; // scroll remainder in pixels
+  double x;
+  double y; // scroll remainder in pixels
```
nbolton, line 36:
```suggestion
  double x;
  double y;
```

src/lib/platform/EiScreen.cpp
```diff
@@ -32,7 +32,8 @@
 
 struct ScrollRemainder
```
nbolton, line 33:
```suggestion
// Values are in pixels
struct ScrollRemainder
```

nbolton (dismissed):

nbolton (approved):

Issue #8598, opened 2025-05-15
Unable to accept TLS fingerprint (endless loop of dialogs)


nbolton:
### Project

Deskflow

### Deskflow version info

TODO - master

### Build type

Local developer build (built it myself)

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [ ] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [ ] I am not using Wayland on Linux

### Signing on macOS

- [ ] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### Continuous build

- [ ] I have tried the latest continuous build and the issue persists
- [ ] I am unable to try the latest continuous build

### OS versions/distros

Any

### What steps will reproduce the problem?

1. Fresh setup
2. Accept TLS fingerprint on dialog

Expected: Connection works
Actual: Same TLS dialog keeps popping

Same on all 4 computers.

### Log output

```shell

```

### Additional information

Seen while testing #8596 - maybe caused by #8594

Yinghe-5:
When I use the Windows laptop as the server and the MacBook Pro as the client, this situation occurs. However, when the MacBook Pro is used as the server and the Windows laptop is used as the client, this situation does not happen.

sithlord48:
What versions are being used please copy and paste the info from the about box here.

PR #8613, opened 2025-05-22
Log Error and stop the core process if print database is not writable

sithlord48:
If for some reason writing to the fingerprint database fails we log a message and also stop the core process so the user can address the problem related to #8598 Should at least stop a loop in the case of an un-writable db file.

### DIFF
--- a/src/lib/gui/MainWindow.cpp
+++ b/src/lib/gui/MainWindow.cpp
@@ -811,7 +811,11 @@ void MainWindow::checkFingerprint(const QString &line)
 
   if (fingerprintDialog.exec() == QDialog::Accepted) {
     db.addTrusted(sha256);
-    db.write(trustedFingerprintDatabase());
+    if (!db.write(trustedFingerprintDatabase())) {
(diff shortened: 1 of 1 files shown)


src/lib/gui/MainWindow.cpp
```diff
@@ -811,7 +811,11 @@ void MainWindow::checkFingerprint(const QString &line)
 
   if (fingerprintDialog.exec() == QDialog::Accepted) {
     db.addTrusted(sha256);
-    db.write(trustedFingerprintDatabase());
+    if (!db.write(trustedFingerprintDatabase())) {
+      qWarning().noquote() << QString("Unable to write fingerprint to db: %1").arg(trustedFingerprintDatabase());
```
nbolton, line 815:
```suggestion
      qWarning().noquote() << QString("unable to write fingerprint to db: %1").arg(trustedFingerprintDatabase());
```

Or

```suggestion
      qWarning().noquote() << "unable to write fingerprint to db:" << trustedFingerprintDatabase();
```

nbolton:
Looks good.

image

nbolton (approved):

PR #8639, opened 2025-05-31
Cleanup Windows classes unused code and includes

sithlord48:
- Remove unused `platform/MSWindowsUtils` class
 - Remove unused windows includes


src/lib/arch/win32/ArchDaemonWindows.cpp
```diff
@@ -12,7 +12,9 @@
 #include "arch/win32/ArchMiscWindows.h"
 #include "arch/win32/XArchWindows.h"
 #include "base/Log.h"
+#include "common/Constants.h"
 
+inline static const auto DEFAULT_DAEMON_NAME = _T(kAppName);
```
nbolton: Screaming snake case means macro, which this is not.
```suggestion
inline static const auto kDefaultDaemonName = _T(kAppName);
```

nbolton (changes requested):
Minor nit.

src/lib/arch/win32/ArchDaemonWindows.cpp
sithlord48: i just moved the define i did not make a new variable here.

nbolton: Are you sure?
```diff
- #define DEFAULT_DAEMON_NAME _T(kAppName)
```

sithlord48: yeah  i moved it from the header but i like the new name better anyway and updated its use here in the CPP file

sithlord48: Oh i see you nit here .. you don't consider the DEFINE the same.. .. either way fixed up

nbolton (approved):

PR #8643, opened 2025-06-01
fix: core not starting if gui started minimized

### LINKED ISSUE #8641: CoreProcess does not start when preference set to "Hide the window when the app starts"
### Project

Deskflow

### Deskflow version info

Deskflow: 1.22.0.0 (00962837) 
Qt: 6.9.0
System: Arch Linux
Session: X-Cinnamon (x11)

### Build type

Community package (apt, dnf, brew, etc.)

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [x] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [x] I am not using Wayland on Linux

### Signing on macOS

- [x] I am not using macOS

### Continuous build

- [x] I have tried the latest continuous build and the issue persists

### OS versions/distros

Arch
Cinnamon - 6.4.10
Linux kernel - 6.14.7-arch2-1
Display Server - X11

### Deskflow configuration

Arch client - cinnamon desktop in X // fresh install via the archinstall

Specific to version 1.22.0.0 (00962837) - continous tested, but is the same version

Deskflow loaded at boot on client and is correctly displayed in the tray.

However, it does not connect to the server. When I VNC into the client, and click on the Deskflow tray icon, Deskflow opens and the log starts to fill out from that moment; connecting to the server as per norm.

### Log output

```shell
N/A - no errors to list
```

### Additional information

Workarounds
- Uncheck "Hide the window when the app starts" in preferences. Reboot system and Deskflow connects to the server correctly (even before the desktop has been displayed).

- Manually install release: 1.21.2.0 (daa09546) - does not have this issue.

// EDIT-UPDATE //
- running the deskflow-client command found in Log at startup, rather than the deskflow application itself, will also bypass this issue.

sithlord48:
fixes #8641, In https://github.com/deskflow/deskflow/pull/8493/commits/bc8e7322a37d87019a4a397003be4bc49b9df6e0 the starting of the core was moved out of  `MainWindow::open` and into `MainWindow::firstShown`. This prevents the core from starting when hidden since we only call hide in this case and `MainWindow::shown` event happens to trigger the logic.

 - Remove `MainWindow::firstShown`
 - Remove `MainWindow::shown`
 - Remove `MainWindow::ShowEvent` Override. 
 - Move the previous logic from `MainWindow::firstShown` -> `MainWindow::open`

Now it does the same thing with less code and works as expected.


src/lib/gui/MainWindow.cpp
```diff
@@ -635,6 +600,25 @@ void MainWindow::serverConnectionConfigureClient(const QString &clientName)
 void MainWindow::open()
 {
   Settings::value(Settings::Gui::Autohide).toBool() ? hide() : showAndActivate();
+
+  // incase any messages are under the mainwindow
+  messages::raiseCriticalDialog();
```
nbolton: Using this as a test:
```patch
diff --git a/src/apps/deskflow-gui/deskflow-gui.cpp b/src/apps/deskflow-gui/deskflow-gui.cpp
index 0dbb6a7c4..74a31b1f2 100644
--- a/src/apps/deskflow-gui/deskflow-gui.cpp
+++ b/src/apps/deskflow-gui/deskflow-gui.cpp
@@ -136,6 +136,8 @@ int main(int argc, char *argv[])
     diagnostic::clearSettings(false);
   }
 
+  qCritical() << "test";
+
   MainWindow mainWindow;
   mainWindow.open();
 ```

The call to `raise` doesn't seem to work (the error isn't raised up on top of main window). The critical error dialog is shown under the main window. 

_However_, I just tested this on `master` and it didn't work there either. 

Since complexity causes bugs: We could just get rid of this call to `raise` and simplify the logic in `src/lib/gui/Messages.cpp` and get rid of `s_criticalMessage`.

**Review by nbolton — CHANGES_REQUESTED** [review]

File: src/lib/gui/MainWindow.cpp (hunk 1 above)
Comment (sithlord48) [reply src/lib/gui/MainWindow.cpp]: qCritical() << "test"; <= does not make a dialog , that just prints to the console

try 
```
 MainWindow mainWindow;
 messages::showWaylandLibraryError(mainWindow);
 mainWindow.open();
```

nbolton: > qCritical() << "test"; <= does not make a dialog

It does.

https://github.com/deskflow/deskflow/blob/410b67f2ae09105c57032b8f7727a4828dc8b7c0/src/apps/deskflow-gui/deskflow-gui.cpp#L112

https://github.com/deskflow/deskflow/blob/410b67f2ae09105c57032b8f7727a4828dc8b7c0/src/lib/gui/Messages.cpp#L98-L105

sithlord48: it does not make one when i do it here ? i used the messages  wayland dialog to check when i wrote this.

sithlord48: Removed the commit that changes the raise timer, let just focus on the bug with this pr and we can decide how best to handle the raise issue then make a pr for that.

Issue #8167, opened 2025-02-04
Potentially drop `libpoprtal` and implement it directly


nbolton:
### Feature Request

From @whot, the author of [`libei`](https://gitlab.freedesktop.org/libinput/libei) and a major [`libportal`](https://github.com/flatpak/libportal) contributor:

https://github.com/deskflow/deskflow/issues/8031#issuecomment-2632520460:
> > Are we committed to using libportal, or does rewriting PortalInputCapture and PortalRemoteDesktop with QDBus make more sense? (Ditching libportal seems like a better plan to me).
> 
> long term dropping libpoprtal and implementing it directly will be the better option. That would also allow dropping the rather clunky glib-mainloop-in-a-thread (and drop glib altogether). At the time it was the faster way to get something working. There are a few awkward behaviours on how portal's sessions and requests work but once you figure those out a Qt only implementation is the better option for deskflow.
> 
> Having said that, adding clipboard support to libportal should be relatively easy and mostly an exercise in copy/paste together with test cases.
> 
> > if you are able to do this w/ only one for each "portal" then i say go for it
> 
> FTR, portals are split into "Portals" which are application facing and are what you target. And there are "implementations" (imp.portal on dbus) which are the compositor-specific backend implementations. As an application you never care about the impls, that's the job of the respective desktop environment/toolkits/... Same way as you don't need to care if sqlite runs on ext4 or btrfs, you just talk SQL to it.
> 
> > XDG Desktop Portal is a Flatpak lib
> 
> Not really :) the XDG Desktop Portal is a process that exposes a few DBus interfaces that Do Things. The only flatpak-y thing is that _if_ your application runs in a flatpak _sandbox_, the sandbox has a automatic hole to allow applications to talk to the portal (i.e. that particular DBus name is always permitted by default). This allows you to Do Things that the sandbox would otherwise prevent. If you're running outside a sandbox you can still let the portal Do Things for you instead of doing them yourself.

cnlohr:
Related to this: The dependence on libportal>=.8 means that the .deb's don't work on anyone using anything except the very latest edge of debian, etc.  Would be really nice if it could work with older libportals, or just not need libportal at all.  (This bit me, on Linux Mint)

sithlord48:
[...]

Issue #8031, opened 2025-01-02
Clipboard support in XDG Desktop Portal (Wayland)


nbolton:
> [!WARNING]
> **Bounty hunters:** We are not accepting bounties on this PR. We tried removing the bounty on Algora.io, but their app is buggy and errors when we try to delete the bounty. We tried reaching out to Algora.io, but they have become unresponsive.

### Feature Request

The [XDG Desktop Portal](https://github.com/flatpak/xdg-desktop-portal) is missing clipboard support, meaning that we cannot copy and paste between a Deskflow server or client running on Wayland.

Part of the work needs to be done in XDG Desktop Portal, but also needs to be done in the [Mutter](https://en.wikipedia.org/wiki/Mutter_(software)) and [KWin](https://en.wikipedia.org/wiki/KWin) compositors. Other compositors are beyond the scope of this issue (but this does not limit someone attempting implementation).

**Maintainers & bounty:** This issue and related bounty represents the entire work (including work to be done outside of this repo) so that a bounty can be created. Do not close as "not our bug".

### Wip PR 

we have much of this staged but are waiting on libportal to land the needed pr so we can finalize  our impl 

https://github.com/deskflow/deskflow/pull/9431

Related:
- https://github.com/deskflow/deskflow/discussions/7499
- https://github.com/deskflow/deskflow/issues/8032

> [!note]
> Flatpaks are also effected by this as they must talk via the portal system to the host.

### Where's the bounty?

We decided to remove the bounty from this PR as it was attracting only low-effort untested AI-generated PRs. We are still interested in sponsoring this PR. See: [Funding](https://github.com/deskflow/deskflow/wiki/Funding)

nbolton:
/bounty $5000

This bounty encompasses landing all necessary PRs in:
- [XDG Desktop Portal](https://flatpak.github.io/xdg-desktop-portal/)
- [Mutter](https://gitlab.gnome.org/GNOME/mutter) and [KWin](https://invent.kde.org/plasma/kwin)
- Deskflow (this project)

The bounty is for all the work needed to make clipboard support when using Wayland on both Mutter and KWin.

nbolton:
OK, please note the updated bounty comment: https://github.com/deskflow/deskflow/issues/8031#issuecomment-2567734972

nbolton:
Surfacing: https://github.com/deskflow/deskflow/discussions/7499#discussioncomment-11789316
> Does this newish data control portal help with the clipboard issue at all?
> 
[...]

PR #8648, opened 2025-06-03
add wayland clipboard support

Mic92:
This is currently based on wl-copy/wl-paste. One could probably just re-implement what wl-copy/wl-paste does without adding much complexity and better performance.

Tested between macOS and Linux. I have not any converters for macOS's html/richtext stuff, only plaintext for now. Not super familiar with the code base, so I might have re-implemented something that is already available elsewhere. Tried to keep things self-contained for now.

Haven't tested actually I broke anything on other platforms. But this will be work for another day.

### DIFF
--- a/src/lib/platform/EiClipboard.cpp
+++ b/src/lib/platform/EiClipboard.cpp
@@ -1,0 +1,134 @@
+/*
+ * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2025 Deskflow Developers
+ * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
+ */
+
+#include "platform/EiClipboard.h"
+
+#include "base/Log.h"
+#include "deskflow/ClipboardTypes.h"
+
+namespace deskflow {
+
+EiClipboard::EiClipboard()
+{
+  initialize();
+}
+
+EiClipboard::~EiClipboard()
+{
+  cleanup();
+}
+
+bool EiClipboard::isAvailable() const
+{
+  return m_available;
+}
+
+IClipboard* EiClipboard::getClipboard(ClipboardID id) const
+{
+  if (!m_available || id >= m_clipboards.size()) {
+    return nullptr;
+  }
+
+  return m_clipboards[id].get();
+}
+
+bool EiClipboard::hasChanged() const
+{
+  if (!m_available) {
+    return false;
+  }
+
+  for (const auto& clipboard : m_clipboards) {
+    if (clipboard && clipboard->hasChanged()) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
+void EiClipboard::startMonitoring()
+{
+  if (!m_available || m_monitoring) {
+    return;
+  }
+
+  for (auto& clipboard : m_clipboards) {
+    if (clipboard) {
+      clipboard->startMonitoring();
+    }
+  }
+
+  m_monitoring = true;
+}
+
+void EiClipboard::stopMonitoring()
+{
+  if (!m_available || !m_monitoring) {
+    return;
+  }
+
+  for (auto& clipboard : m_clipboards) {
+    if (clipboard) {
+      clipboard->stopMonitoring();
+    }
+  }
+
+  m_monitoring = false;
+}
+
+void EiClipboard::resetChanged()
+{
+  if (!m_available) {
+    return;
+  }
+  
+  for (auto& clipboard : m_clipboards) {
+    if (clipboard) {
+      clipboard->resetChanged();
+    }
+  }
+}
+
+void EiClipboard::initialize()
+{
+  // Check if Wayland clipboard is available
+  if (!WaylandClipboard::isAvailable()) {
+    LOG_WARN("wl-clipboard tools not found, clipboard functionality disabled");
+    return;
+  }
+
+  // Create clipboard instances for each clipboard type
+  m_clipboards.resize(kClipboardEnd);
+
+  try {
+    // Primary clipboard (selection)
+    m_clipboards[kClipboardSelection] = std::make_unique<WaylandClipboard>(kClipboardSelection);
+
+    // Standard clipboard
+    m_clipboards[kClipboardClipboard] = std::make_unique<WaylandClipboard>(kClipboardClipboard);
+
+    m_available = true;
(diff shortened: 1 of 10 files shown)


sithlord48:
does this change the depends on wayland ? 
can you update your PR so the EIClipboard files end with an empty new line 
When should i test this ?

Mic92:
> does this change the depends or wayland ? 

It's the clipboard protocol in wayland. So yes.

> When should i test this ?

You can already test it. It's functional for me - I may just get rid of calling the CLI. I don't have a Windows handy for example, if you can test that.

sithlord48:
>This is currently based on wl-copy/wl-paste

 does this require having wl-copy / wl-paste (i.e adding a dependeny on wl-clipboard ) We need to access the clipboard via/ the input portal ultimately. I have been very hesitant  to add a depend on wl-clipboard as its the wrong aproach 
we do not want to be calling external applicitons.

Mic92:
> We need to access the clipboard via/ the input portal ultimately
Why is this?

Like I said in the beginning, I will probably just just re-implement what wl-copy is doing. I consider this a solution until all the portal stuff is actually fully implemented. For now this work fine. And actually fixes keyboard input as well on macOS.

Mic92:
I can otherwise also just close this pull request again and just use it myself, your choice.

sithlord48:
> Like I said in the beginning, I will probably just just re-implement what wl-copy is doing. 

A better approach since we won't need any external execuitables, I would suggest when your doing the re implementation to use the xdg-clipboard portal in place of the wlroots specific one wl-clipboard is using. This should then work on all compositors.  Since wl-clipboard is not using the standard xdg-clipboard interface it do wonder if it work on all compositors or just those with supporting the wlr-data extension . I do think this would be a better solution until the input-capture and remote-desktop portals are fully hooked into the clipboard one.

> And actually fixes keyboard input as well on macOS.

I am not sure how this changes anything in the macOS build since non of the LibEi / Libportal stuff is built on macOS.

nbolton:
FYI, there's a bounty on:
- #8031

Mic92:
> > And actually fixes keyboard input as well on macOS.
> 
> I am not sure how this changes anything in the macOS build since non of the LibEi / Libportal stuff is built on macOS.

Without this patch none, keyboard inputs are not handled correctly from Wayland server to macOS client. I also found an issue for this, but forgot where this was.

Mic92:
> > Like I said in the beginning, I will probably just just re-implement what wl-copy is doing.
> 
> A better approach since we won't need any external execuitables, I would suggest when your doing the re implementation to use the xdg-clipboard portal in place of the wlroots specific one wl-clipboard is using. This should then work on all compositors. Since wl-clipboard is not using the standard xdg-clipboard interface it do wonder if it work on all compositors or just those with supporting the wlr-data extension . I do think this would be a better solution until the input-capture and remote-desktop portals are fully hooked into the clipboard one.

I don't really want to argue with you what is better or not. wl-clipboard worked on all wayland environments that I have used so far where this portal things seems brand new and not fully integrated everywhere. For me this patch works - I don't need flatpack support, which seems the major reason to want to use it. I can wait until upstream deskflow has catched up.

Mic92:
> FYI, there's a bounty on:
> 
>     * [Clipboard support in XDG Desktop Portal (Wayland) #8031](https://github.com/deskflow/deskflow/issues/8031)

This one is for portal integration, I only implemented wayland clipboard support.

nbolton:
> This one is for portal integration, I only implemented wayland clipboard support.

Right, just thought I'd mention. Also this:
- #8167

sithlord48:
When you do update this make sure to rebase not merge update.

sithlord48:
> wl-clipboard worked on all wayland environments that I have used so far. 

Can you tell me what environments you have used this  on ?

Mic92:
> > wl-clipboard worked on all wayland environments that I have used so far.
> 
> Can you tell me what environments you have used this on ?

KDE, Gnome, various wl-roots based ones like sway, qtile, hyprland....

src/lib/platform/EiClipboard.cpp
```diff
@@ -0,0 +1,134 @@
...
+{
+  if (m_monitoring) {
+    stopMonitoring();
+  }
+  
+  m_clipboards.clear();
+  m_available = false;
+}
+
+} // namespace deskflow
```
sithlord48, line 134: Please end this file with an empty new line.

src/lib/platform/EiClipboard.h
```diff
@@ -0,0 +1,61 @@
...
+  //! Cleanup clipboard backends
+  void cleanup();
+
+private:
+  std::vector<std::unique_ptr<WaylandClipboard>> m_clipboards;
+  bool m_available = false;
+  bool m_monitoring = false;
+};
+
+} // namespace deskflow
```
sithlord48, line 61: Please end this file with an empty new line

src/unittests/platform/WaylandClipboardTests.h
```diff
@@ -0,0 +1,74 @@
...
+like opening/closing the clipboard, adding/getting data in different formats,
+and monitoring clipboard changes.
+
+Note: These tests require wl-clipboard tools to be installed and will be
+skipped if they are not available or if Wayland support is not compiled in.
+*/
+class WaylandClipboardTests : public QObject
+{
+  Q_OBJECT
+private slots:
```
sithlord48, line 32: We recently enabled QT_NO_KEYWORDS so when you rebase upon master you have to change this to `private Q_SLOTS`

sithlord48 (changes requested):
Did a quick test and noticed a few things

This needs to be rebased upon master

When testing i was able to copy from my server (wayland) to client (macos)  but not from my client to my server.
* I only tried with text

scraprats:
This works, tested on F42, KDE, Wayland <---> Barrier Debian 10, KDE.

This seems like a good temporary solution until a more definite solution is decided.

lxe:
I've rebased and fixed all the issues:

https://github.com/lxe/deskflow

Works on arch/hyprland host and osx guest.

sithlord48:
Please rebase upon master.

Mic92:
Sorry stopped using deskflow fow now. But looks like @lxe can open a PR?

lxe:
> Sorry stopped using deskflow fow now. But looks like @lxe can open a PR?

Done. #8971

Issue #8568, opened 2025-05-03
MacOS client causes high power usage when in sleep


gawtDamn:
### Project

Deskflow

### Deskflow version info

Deskflow 1.21.2

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [x] Windows
- [x] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [ ] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [ ] I am not using Wayland on Linux

### Signing on macOS

- [x] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### Continuous build

- [ ] I have tried the latest continuous build and the issue persists
- [ ] I am unable to try the latest continuous build

### OS versions/distros

Windows 11
MacOS 15.4.1

### Deskflow configuration

Server: Windows 11
client: MacOS 15.4.1

Windows on the left, two screens, Mac on the right, one screen.

Mac Hardware: Mac Mini M4

### What steps will reproduce the problem?

1. be able to measure the power draw of the Mac
2. Measure idle power of the Mac with the deskflow client running
3. Sleep the mac. 
4. Measure the sleep power draw
5. Wake it, kill deskflow
6. sleep the mac
7. measure power draw again

### Log output

```shell
[2025-05-02T20:38:33] INFO: entering screen
[2025-05-02T20:38:49] INFO: suspend
[2025-05-02T20:38:49] INFO: leaving screen
[2025-05-02T20:38:49] WARNING: cursor may not be visible
[2025-05-02T20:38:49] WARNING: cursor may not be visible
[2025-05-02T20:38:49] NOTE: disconnected from server
[2025-05-02T20:44:08] INFO: resume
[2025-05-02T20:44:08] NOTE: connecting to '192.168.1.222': 192.168.1.222:24800
[2025-05-02T20:44:08] NOTE: peer fingerprint: (SHA1)
```

### Additional information

[...]

PR #8658, opened 2025-06-06
fix: disable App Nap on macOS

iameli:
Been trying to diagnose a random disconnect (Windows server, Mac Client) and found a bunch of macOS console messages related to App Sleep right around the time that the disconnect happened. I'm still testing whether this causes it... but surely Deskflow is an app that would want to out out of App Sleep backgrounding in any case, yes? [Some similar discussion around QEMU here](https://gitlab.com/qemu-project/qemu/-/issues/334).

### DIFF
--- a/src/apps/res/deskflow.plist.in
+++ b/src/apps/res/deskflow.plist.in
@@ -35,6 +35,10 @@
 	<key>NSSupportsSuddenTermination</key>
 	<false/>
 
+	<!-- Don't allow the app to be background throttled -->
+	<key>NSAppSleepDisabled</key>
+	<true/>
+
 	<!-- Local Lan Entitlements -->
 	<key>NSLocalNetworkUsageDescription</key>
 	<string>This app uses local network permissions for communication between a Deskflow server and clients.</string>


sithlord48:
It seams to be that allowing the app to block sleep would contribute to #8568

iameli:
Despite the name, I don't believe this feature has anything to do with macOS sleep mode - correct me if I'm wrong. Renamed the PR to make that clearer

sithlord48:
My only hesitance is this allows the application to not be put to appnap under some conditions. Reading the apple guide it seams we may not be able to enter this state taking some IO Abilitys like input capture, Im now looking to see when / if appnap is entered for deskflow on my machine.

nbolton:
I wonder if we can somehow make this optional/configurable... I fear not.

sithlord48:
> I wonder if we can somehow make this optional/configurable... I fear not.

A user could add the key to the plist file manually.

sithlord48:
after not touch my mbp for a few days and leaving the activitiy monitor on i've yet to see deskflow (gui or server) enter app nap state. @iameli  how long do you leave it idle before you see it enter app nap and under what conditions ?

iameli:
> after not touch my mbp for a few days and leaving the activitiy monitor on i've yet to see deskflow (gui or server) enter app nap state. @iameli how long do you leave it idle before you see it enter app nap and under what conditions ?

I was basing it off of a `[2025-06-07T20:53:43] NOTE: server is dead` message in the Deskflow logs correlated with lots of App Sleep-related messages in the Console; it's the only thing I could find that seemed semi-related. But I may be completely wrong about the root cause.

> I wonder if we can somehow make this optional/configurable... I fear not.

@nbolton What's the use case for Deskflow entering App Nap? If it's in-use, throttling its resources would presumably lead to a very bad UX. If it's not in-use I guess you could make an argument for App Nap but presumably Deskflow could also just be careful and use as few resources as possible in that state.

sithlord48:
@iameli  are you saying you only are seeing this on the serving mac ? i have been in client mode let me switch it over and see if im able to appnap that way

nbolton:
> I guess you could make an argument for App Nap

IDK enough about it. If it's only ever detrimental, then maybe disabling it is best.

nbolton:
@sithlord48 Any thoughts on what to do with this?

sithlord48:
I don't want to add this to the plist until we at least know we can enter app sleep.. I think here we need some more testing

nbolton:
Is it a draft? Since it's arguably not ready to land.

sithlord48:
The pr it self is ready code wise we need to really just test it to see if this helps or not

iameli:
FWIW this did not fix my problem AFAICT. It may still be true that there's no use case for App Nap in Deskflow though.

sithlord48:
> FWIW this did not fix my problem AFAICT. It may still be true that there's no use case for App Nap in Deskflow though.

Ok I will close this in that case.

PR #8660, opened 2025-06-08
ci: Update install-qt-action to v4.3.0

sithlord48:
Update the install-qt-action to v 4.3.0


.github/actions/install-dependencies/action.yml
```diff
@@ -68,11 +64,10 @@ runs:
 
     - name: Install Qt
       if: ${{runner.os != 'Linux' }}
-      uses: jurplel/install-qt-action@v4
+      uses: jurplel/install-qt-action@v4.3.0
```
nbolton: Using the `v4` tag already uses the latest version. We generally only pin the version when a later version causes a bug, and that tends to be a temporary hack.

https://github.com/jurplel/install-qt-action/tree/v4

See how the latest commit for the `v4` tag is "Build v4.3.0"?

```suggestion
      uses: jurplel/install-qt-action@v4
```

sithlord48: Valid ill revert that part.

nbolton (approved):

PR #8667, opened 2025-06-09
chore: correct appstream license to gpl2.0 NOT gpl2.0+

sithlord48:
The Appstream file incorrectly says the project is `gpl2.0+`. the project is `gpl2.0`


nbolton:
Only applies to Flatpak, right? Out of curiosity, where did the problem manifest? On the Flatpak page?

sithlord48:
It was in the appstream data so distros that use that info in their package manger would see it in places like discover and gnome software as well as the flatpak page.

PR #8668, opened 2025-06-09
ci: Fix windows random build failures

sithlord48:
- windows has random build failues with 6.9.1 so reverted it to 6.8.3
 - reverted the vcpkg action to no longer use my fork since johnwason's now is updated.
 - A few files ha included QApplication, QGuiApplicaiton or QCoreApplication when it was not needed this removes those includes.


sithlord48:
Bypass merged to fix builds.

nbolton:
> Bypass merged to fix builds.

I believe a review before land is still useful even if the build is broken.

PR #8669, opened 2025-06-10
docs: Re-introduce chat links to grow community

nbolton:
I want our community to grow, and I noticed a drop off in new Matrix members. This is because we dropped the Matrix links from the README. I have also noticed some friction getting people on Matrix ("I have my IRC client just how I want it").

Since I'm on both Matrix and IRC, it makes sense to send people to the chat platform they prefer rather than trying to force people. Putting links on the README is the best place since people don't tend to hunt these links down in the wiki.


nbolton:
image

sithlord48:
Idk if like this idea one place to chat makes things easier to follow.. If ppl don't want to join the matrix room they can discuss here.  Imho it seams like your making more work for yourself since now in irc you will be relaying messages between irc matrix and here. IMHO if your gonna do that you had better have a working bridge between the two.

nbolton:
> Idk if like this idea one place to chat makes things easier to follow.. 

Only if everyone abandons IRC and moves to Matrix which will never happen.

> If ppl don't want to join the matrix room they can discuss here.

They won't do that.

> Imho it seams like your making more work for yourself since now in irc 

Not really, it's pretty easy.

> you will be relaying messages between irc matrix and here.

It won't be necessary to relay info.

> IMHO if your gonna do that you had better have a working bridge between the two.

I dislike bridges; they don't work very well in my experience.

sithlord48:
> It won't be necessary to relay info.

It will if someone in irc is trying to talk to me or asking you questions you then have to ask me.

> I dislike bridges; they don't work very well in my experience.

Well we should try to find one, as that would unify the two and make it so people only need to join one not both to see if anyone in either can help.

sithlord48:
@nbolton You can land this if you like. But we should at least look into a bridge to unify them and then it does not matter where ppl want to join.

PR #8674, opened 2025-06-10
Begin to use std::mutex in place of ArchMutex

sithlord48:
Begin to replace `ArchMutex` with `std::mutex`. In this first step all cases of `ArchMutex` or `Mutex` that were not part of a conditional var have been replaced. 

Tested locally for a few days didn't see any issues.

Backport from https://github.com/debauchee/barrier/pull/410:
 - use `std::mutex` in `arch/unix/ArchMutithreadPosix`
 - use `std::mutex` in `arch/ArchStrings`
 - use `std::mutex` in `base/Log`
 - use `std::mutex` in `base/EventQueue`

Additionally:
- Links arch to base
- use `std::mutex` in `arch/win32/ArchMultiThreadWindows`
- use `std::mutex` in `arch/unix/ArchNetworkBSD`
- use `std::mutex` in `arch/unix/ArchNetworkWinSock`
- use `std::mutex` in `deskflow/PacketStreamFilter`
- use `std::mutex` in `net/TCPListenSocket`
- use `std::mutex` in `platform/XWindowsEventQueue`

### DIFF
--- a/src/lib/arch/ArchString.cpp
+++ b/src/lib/arch/ArchString.cpp
@@ -6,28 +6,21 @@
  */
 
 #include "arch/ArchString.h"
-#include "arch/Arch.h"
 
 #include <climits>
 #include <cstdlib>
 #include <cstring>
+#include <mutex>
 
-static ArchMutex s_mutex = nullptr;
+std::mutex s_mutex;
 
 //
 // use C library non-reentrant multibyte conversion with mutex
 //
 
-ArchString::~ArchString()
-{
-  if (s_mutex != nullptr) {
-    ARCH->closeMutex(s_mutex);
-    s_mutex = nullptr;
-  }
-}
-
 int ArchString::convStringWCToMB(char *dst, const wchar_t *src, uint32_t n, bool *errors) const
 {
+  std::lock_guard<std::mutex> lock(s_mutex);
   ptrdiff_t len = 0;
 
   bool dummyErrors;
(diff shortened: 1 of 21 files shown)


nbolton (approved):

src/lib/arch/ArchString.cpp
```diff
@@ -6,28 +6,21 @@
...
 #include "arch/ArchString.h"
-#include "arch/Arch.h"
 
 #include <climits>
 #include <cstdlib>
 #include <cstring>
+#include <mutex>
 
-static ArchMutex s_mutex = nullptr;
+std::mutex s_mutex;
```
jwakely, line 15: Should this still be `static`, or in an anonymous namespace? It doesn't need to be globally visible (and possibly clash with an `s_mutex` in other translation units)

src/lib/arch/unix/ArchNetworkBSD.cpp
```diff
@@ -602,18 +591,18 @@ std::string ArchNetworkBSD::addrToString(ArchNetAddress addr)
...
   case kINET6: {
     char strAddr[INET6_ADDRSTRLEN];
     const auto *ipAddr = TYPED_ADDR(struct sockaddr_in6, addr);
-    ARCH->lockMutex(m_mutex);
-    inet_ntop(AF_INET6, &ipAddr->sin6_addr, strAddr, INET6_ADDRSTRLEN);
-    ARCH->unlockMutex(m_mutex);
+    {
+      std::lock_guard<std::mutex> lock(m_mutex);
+      inet_ntop(AF_INET6, &ipAddr->sin6_addr, strAddr, INET6_ADDRSTRLEN);
+    }
```
jwakely, line 605: I assume the purpose of the extra braces is so that the mutex is unlocked before constructing the std::string return value from the array `strAddr`? In that case, why is the `kINIT` case above different? That one constructs the string `s` (and move constructs it into the return value) while the mutex is held.

sithlord48: Honestly not sure this was ported from downstream i would think they should both have braces. 

Step 1 was to Port, Now we can improve it feel free to propose a PR

jwakely: Submitted as #8922

Issue #411, opened 2014-10-13
Portable version


nbolton:
**Imported issue:**
- Author: sjrmay (Google user)
- Date: 2010-03-22 00:00:00
- Legacy ID: 421

Hi folks,  

I was wondering if anyone was thinking about making a "portable" version of  
Synergy-plus for Windows - ala portableapps.com?  

My use scenario is that I've got a work PC that I can't install Apps on (I  
can run portable apps though :) )  Getting Synergy-Plus working in that way  
would be amazing for me...sure others are the same.  

Any chance?  

Si

velvitonator:
Consider this another vote.

We're a small games studio with our own tools, and Synergy is a great way to get mouse + keyboard going on devices that are either finicky to get to support it or don't have it at all (e.g. consoles, mobile).
I'd be really happy if I could just bundle Synergy with our tools packages and get it to run when the tools start, as opposed to it being in the ether or not.

jrhortian:
Any update on this?

telorand:
I'd also like a portable version. I would like to use it with my work computer (to which I don't have admin rights) and my own personal computer, but without those rights, I'm out of luck. Cameyo is also of no avail.

Gitoffthelawn:
+1

pdemro:
Please give us a portable version!  Client laptops always come locked down and thanks to Nick I am addicted to Synergy

pdemro:
To whoever finds their way here - I was able to find a workaround to enable Synergy on clients without administrator access.  Check out this forum thread here:

http://synergy-project.org/forum/viewtopic.php?f=17&t=547&p=4008#p4008

bobpaul:
pdemro, do [these instructions](https://portableapps.com/comment/190414#comment-190414) look like what you had? 

FWIW, I've tried installing various versions (I haven't yet found the old 1.3.7 build for windows; I bet that would work as protable) and can't get any of them to work on their own.

paul-suarez:
Elsni from forums:

I'd like to run synergy on my company laptop to control my personal mac when I work in home office. But I have no admin rights.

Is it technically possible to provide a synergy version for windows which runs without installation and admin rights?

Thanks!

[Use without install/admin privileges](https://symless.com/forums/topic/2975-use-without-installadmin-privileges/)

tandr3i:
+1 for portable version or at least a workaround to be able to use it on a pc without admin

PR #8680, opened 2025-06-12
backport: barrier #411

sithlord48:
backports: https://github.com/debauchee/barrier/pull/411

ported-by: sithlord48

- Failing on GCC Compilers :?!?!?!

### DIFF
--- a/src/lib/net/ISocketMultiplexerJob.h
+++ b/src/lib/net/ISocketMultiplexerJob.h
@@ -1,5 +1,7 @@
 /*
  * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2025 Deskflow Developers
+ * SPDX-FileCopyrightText: (C) 2019 Barrier Contributors
  * SPDX-FileCopyrightText: (C) 2012 - 2016 Symless Ltd.
  * SPDX-FileCopyrightText: (C) 2002 Chris Schoeneman
  * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
@@ -9,6 +11,21 @@
 
 #include "arch/IArchNetwork.h"
 #include "common/IInterface.h"
+
+#include <memory>
+
+class ISocketMultiplexerJob;
+struct MultiplexerJobStatus
+{
+  MultiplexerJobStatus(bool cont, std::unique_ptr<ISocketMultiplexerJob> &&nj)
+      : continueServicing(cont),
+        newJob(std::move(nj))
+  {
+    // do nothing
+  }
+  bool continueServicing = false;
+  std::unique_ptr<ISocketMultiplexerJob> newJob;
+};
 
 //! Socket multiplexer job
 /*!
@@ -22,20 +39,19 @@ public:
 
   //! Handle socket event
   /*!
-  Called by a socket multiplexer when the socket becomes readable,
-  writable, or has an error.  It should return itself if the same
-  job can continue to service events, a new job if the socket must
-  be serviced differently, or nullptr if the socket should no longer
-  be serviced.  The socket is readable if \p readable is true,
-  writable if \p writable is true, and in error if \p error is
-  true.
+  Called by a socket multiplexer when the socket becomes readable, writable, or has an error.
+  The socket is readable if \p readable is true, writable if \p writable is true, and in error
+  if \p error is true.
+  The method returns false as the continue_servicing member of the returned struct if the socket
+  should no longer be served and true otherwise. Additionally, if the newJob member of the
+  returned pair is not empty, the socket should be serviced differently with the specified job.
 
   This call must not attempt to directly change the job for this
   socket by calling \c addSocket() or \c removeSocket() on the
   multiplexer.  It must instead return the new job.  It can,
   however, add or remove jobs for other sockets.
   */
-  virtual ISocketMultiplexerJob *run(bool readable, bool writable, bool error) = 0;
+  virtual MultiplexerJobStatus run(bool readable, bool writable, bool error) = 0;
 
   //@}
   //! @name accessors
@@ -60,6 +76,11 @@ public:
   becomes writable.
   */
   virtual bool isWritable() const = 0;
+
+  virtual bool isCursor() const
+  {
+    return false;
+  }
 
   //@}
 };
--- a/src/lib/net/SecureSocket.cpp
+++ b/src/lib/net/SecureSocket.cpp
@@ -67,6 +67,11 @@ SecureSocket::SecureSocket(
 
 SecureSocket::~SecureSocket()
 {
+  isFatal(true);
+  // take socket from multiplexer ASAP otherwise the race condition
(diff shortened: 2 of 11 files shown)


src/lib/net/ISocketMultiplexerJob.h
```diff
@@ -10,6 +12,21 @@
 #include "arch/IArchNetwork.h"
 #include "common/IInterface.h"
 
+#include <memory>
+
+class ISocketMultiplexerJob;
```
sithlord48, line 17: @nbolton  Any ideas on this one ? 
Pretty sure the failure to build on GCC comes from this class being mixed in w/ the struct.

nbolton: Hmm, yes we’re in WTF territory. That circular dependency between `MultiplexerJobStatus` and `ISocketMultiplexerJob` gives me the creeps (sure, it should work, but seems smelly). I've been fiddling with this for like 30 mins now and I'm still stumped. The problem seems to be that the codegen can't do the delete op because that needs a full type. I wonder if maybe we can break the coupling and come up with a design that doesn't involve needing to forward declare.

sithlord48: In c++ 20  move needs a fullly implimented type in c++14 it does not. so yeah might have to decouple a bit here to get it

nbolton: Ah yeah, IL/Barrier has an older C++ spec requirement, doesn't it?

sithlord48: Yes C++14

nbolton:
Please make this PR title more helpful, e.g.:
> Backport: Use explicit memory ownership in SocketMultiplexer (Barrier)

src/lib/net/ISocketMultiplexerJob.h
```diff
@@ -10,6 +12,21 @@
...
+struct MultiplexerJobStatus
+{
+  MultiplexerJobStatus(bool cont, std::unique_ptr<ISocketMultiplexerJob> &&nj)
```
tcanabrava, line 20: This is a struct, everything is public, just remove the constructor and access it via the default initializer.

src/lib/net/SecureSocket.cpp
```diff
@@ -92,29 +97,33 @@ void SecureSocket::connect(const NetworkAddress &addr)
   TCPSocket::connect(addr);
 }
 
-ISocketMultiplexerJob *SecureSocket::newJob()
+std::unique_ptr<ISocketMultiplexerJob> SecureSocket::newJob()
```
tcanabrava, line 92: this seems incorrect.
if there is a possibility of returning an invalid unique_ptr, then, this is not a unique_ptr. it's a optional.

what would be the problem of returning a std::optional<std::unique_ptr<ISocketMultiplierJob>> - why not sleep untill m_secureReady, and return the newJob so this method never fails?

src/lib/net/SecureSocket.cpp
```diff
@@ -670,7 +679,7 @@ bool SecureSocket::verifyCertFingerprint(const QString &FingerprintDatabasePath)
   return true;
 }
 
-ISocketMultiplexerJob *SecureSocket::serviceConnect(ISocketMultiplexerJob *job, bool, bool write, bool error)
+MultiplexerJobStatus SecureSocket::serviceConnect(ISocketMultiplexerJob *job, bool, bool write, bool error)
```
tcanabrava, line 675: I'm not sure I like the ergonomics of this.
the status also contains the job?
shouldn't just the job have a status?
because the return values are `bool, job` I believe this is better served as an std::optional<job>

src/lib/net/SocketMultiplexer.h
```diff
@@ -52,7 +53,7 @@ class SocketMultiplexer
 private:
   // list of jobs.  we use a list so we can safely iterate over it
   // while other threads modify it.
-  using SocketJobs = std::list<ISocketMultiplexerJob *>;
+  using SocketJobs = std::list<std::unique_ptr<ISocketMultiplexerJob>>;
```
tcanabrava, line 56: why list and not a vector here?

src/lib/net/TCPListenSocket.cpp
```diff
@@ -55,11 +55,10 @@ void TCPListenSocket::bind(const NetworkAddress &addr)
     ARCH->setReuseAddrOnSocket(m_socket, true);
     ARCH->bindSocket(m_socket, addr.getAddress());
     ARCH->listenOnSocket(m_socket);
-    m_socketMultiplexer->addSocket(
-        this, new TSocketMultiplexerMethodJob<TCPListenSocket>(
-                  this, &TCPListenSocket::serviceListening, m_socket, true, false
-              )
+    auto job = std::make_unique<TSocketMultiplexerMethodJob<TCPListenSocket>>(
```
tcanabrava, line 58: there are a few of those calls around, and they are *huge*.
why not this?
```
template<typename T> new_job(args) -> std::unique_ptr<TSocketMultiplexerMethodJob<T>> {
    return std::make_unique<TSocketMultiplexerMethodJob<TCPListenSocket>>(args);
}
```

then we could use it when needed as:

`auto job = new_job<TCPListenSocket>( args... )`
}

src/lib/net/TCPSocket.cpp
```diff
@@ -470,27 +476,35 @@ ISocketMultiplexerJob *TCPSocket::serviceConnecting(ISocketMultiplexerJob *job,
...
   }
 
   if (write) {
     sendEvent(EventTypes::DataSocketConnected);
     onConnected();
-    return newJob();
+    auto nJob = newJob();
+    if (nJob)
+      return {true, std::move(nJob)};
+    else
```
tcanabrava, line 495: no else after return, and decrease the scope of njob.
```
if (auto nJob = newJob()) {
    return {true, std::move(njob)};
}
return {false, {}};
```

but the ergonomics are not great.

sithlord48:
@tcanabrava  Thanks for looking at this! its an attempt at a porting this directly from https://github.com/debauchee/barrier/pull/411 . So i can't speak to why some of the choices were made in that pr, I was hoping to port it directly then improve it in a follow up commit so it was easier to track the code thru the forks.

src/lib/net/SecureSocket.cpp
sithlord48: > shouldn't just the job have a status?

I think this could be its own pr that may remove the need for this one, @nbolton  any idea why the job does not have just have the status and we use this type instead ?

nbolton: I'll have a think, long time since I touched that code.

nbolton: Ah, I see, new code. It does seem a bit counterintuitive that `MultiplexerJobStatus` contains the job.

@p12tic Do you remember the rationale for this? To support some legacy code, I imagine...

sithlord48:
Closing this we can do better.

PR #8691, opened 2025-06-16
Minor smell cleanup

sithlord48:
- Use `std::scoped_lock` in place of `std::lock_guard` 
 - use Arch::time in place of ARCH->time()
 - use Arch::sleep() in place of ARCH->sleep()

### DIFF
--- a/src/lib/arch/ArchString.cpp
+++ b/src/lib/arch/ArchString.cpp
@@ -20,7 +20,7 @@ std::mutex s_mutex;
 
 int ArchString::convStringWCToMB(char *dst, const wchar_t *src, uint32_t n, bool *errors) const
 {
-  std::lock_guard<std::mutex> lock(s_mutex);
+  std::scoped_lock lock{s_mutex};
   ptrdiff_t len = 0;
 
   bool dummyErrors;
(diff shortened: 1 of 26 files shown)


sithlord48:
> Purely academic, but: What is the rationale for using `scoped_lock` on single mutexes? As I understand it, performance-wise, `lock_guard` is the same for single mutexes but some consider `scoped_lock` to be less safe:

From Sonarscan: 

std::scoped_lock basically provides the same feature as std::lock_guard, but is more generic: It can lock several mutexes at the same time, with a deadlock prevention mechanism (see [S5524](https://sonarcloud.io/organizations/deskflow/rules?open=cpp%3AS5524&rule_key=cpp%3AS5524)). The equivalent code to perform simultaneous locking with std::lock_guard is significantly more complex. Therefore, it is simpler to use std::scoped_lock all the time, even when locking only one mutex (there will be no performance impact).std::scoped_lock basically provides the same feature as std::lock_guard, but is more generic: It can lock several mutexes at the same time, with a deadlock prevention mechanism (see [S5524](https://sonarcloud.io/organizations/deskflow/rules?open=cpp%3AS5524&rule_key=cpp%3AS5524)). The equivalent code to perform simultaneous locking with std::lock_guard is significantly more complex. Therefore, it is simpler to use std::scoped_lock all the time, even when locking only one mutex (there will be no performance impact).

nbolton:
> simpler to use std::scoped_lock all the time

Yes, I agree with this. What do you think about what Howard Hinnant is saying? He is saying that `scoped_lock` is less safe because you can accidentally lock 0 mutexes. TBH, I don't want to get bogged down in academic discussion, just curious about your perspective. From the reviewer's perspective: It's more work to review because I have to check that the mutex was passed to the lock in each code change, and there is a possibility that we missed one.

nbolton (approved):
Tested and working well.

PR #8692, opened 2025-06-16
fix: Error text is not shown for Qt CLI parser

nbolton:
Related to: #8685

We should print the parser error text to stderr and return non-zero exit code so that it's clear when you enter an invalid option.

As it is, it's a bit confusing how it doesn't tell you when you enter an invalid arg.

### DIFF
--- a/src/apps/deskflow-gui/deskflow-gui.cpp
+++ b/src/apps/deskflow-gui/deskflow-gui.cpp
@@ -36,6 +36,8 @@ using namespace deskflow::gui;
 bool checkMacAssistiveDevices();
 #endif
 
+const static auto kHeader = QStringLiteral("%1: %2\n").arg(kAppName, kDisplayVersion);
+
 int main(int argc, char *argv[])
 {
 #if defined(Q_OS_UNIX) && defined(QT_DEBUG)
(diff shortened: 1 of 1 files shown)


sithlord48:
we don't need the method  for this but otherwise seams ok will test out later

src/apps/deskflow-gui/deskflow-gui.cpp
```diff
@@ -63,10 +70,15 @@ int main(int argc, char *argv[])
   parser.addOption(resetOption);
   parser.parse(QCoreApplication::arguments());
 
+  if (!parser.errorText().isEmpty()) {
+    qCritical().noquote() << parser.errorText() << Qt::endl;
+    showHelp(parser);
```
sithlord48: 1. do we want to show help even or just bail ? 
2. You seam to have removed the "unknownOptionsNames" handling i think it shoudl be part of this block

nbolton: > 1. do we want to show help even or just bail ?

Pros and cons either way; if we just bail (which I did think of), then we don't need an extra function, but it's a minor user inconvenience as they have to run it again with `-h`... though it's much more deliberate if they're made to do this.

> 2. You seam to have removed the "unknownOptionsNames" handling i think it shoudl be part of this block

Unless I'm missing something... `errorText` always returns a value when `unknownOptionsNames` has values, so I couldn't see the use of checking both.

sithlord48: > Pros and cons either way; if we just bail (which I did think of), then we don't need an extra function, but it's a minor user inconvenience as they have to run it again with -h... though it's much more deliberate if they're made to do this.

OK then lets make it bail they can re run with -h

> Unless I'm missing something... errorText always returns a value when unknownOptionsNames has values, so I couldn't see the use of checking both.

Edit: double checked, It does

src/apps/deskflow-gui/deskflow-gui.cpp
```diff
@@ -63,15 +65,19 @@ int main(int argc, char *argv[])
   parser.addOption(resetOption);
   parser.parse(QCoreApplication::arguments());
 
-  const auto header = QStringLiteral("%1: %2\n").arg(kAppName, kDisplayVersion);
-  if (parser.isSet(helpOption) || !parser.unknownOptionNames().isEmpty() || !parser.errorText().isEmpty()) {
-    QTextStream(stdout) << header << QStringLiteral("  %1\n\n").arg(kAppDescription)
+  if (!parser.errorText().isEmpty()) {
+    qCritical().noquote() << parser.errorText() << Qt::endl << "Use --help for more information.";
```
sithlord48: Qt::endl will flush use "\nUse..." instead

sithlord48 (changes requested):
One tiny change request.

src/apps/deskflow-gui/deskflow-gui.cpp
nbolton: Always forget.

sithlord48 (approved):

PR #8699, opened 2025-06-18
Add Main Input-leap / Barrier Devs to our Important dev list

sithlord48:
- Added @p12tic  and @shymega to our list of important developers. 
 - Adjusted the about dialog's default size

PR #8702, opened 2025-06-18
refactor: Improve display of names on about screen with nbsp

nbolton:
Related to: #8699

### DIFF
--- a/src/lib/gui/dialogs/AboutDialog.cpp
+++ b/src/lib/gui/dialogs/AboutDialog.cpp
@@ -32,10 +32,18 @@ AboutDialog::AboutDialog(QWidget *parent) : QDialog(parent), ui{std::make_unique
   connect(ui->btnCopyVersion, &QPushButton::clicked, this, &AboutDialog::copyVersionText);
 
   ui->lblVersion->setText(kDisplayVersion);
-
   ui->lblDescription->setText(kAppDescription);
   ui->lblCopyright->setText(kCopyright);
-  ui->lblImportantDevs->setText(QStringLiteral("%1\n").arg(s_awesomeDevs.join(", ")));
+
+  // Use non-breaking space in each awesome dev name so names are not split across lines.
+  QStringList devsNbsp;
+  for (const auto &dev : s_awesomeDevs) {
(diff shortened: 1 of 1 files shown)


src/lib/gui/dialogs/AboutDialog.cpp
```diff
@@ -32,10 +32,18 @@ AboutDialog::AboutDialog(QWidget *parent) : QDialog(parent), ui{std::make_unique
   connect(ui->btnCopyVersion, &QPushButton::clicked, this, &AboutDialog::copyVersionText);
 
   ui->lblVersion->setText(kDisplayVersion);
-
   ui->lblDescription->setText(kAppDescription);
   ui->lblCopyright->setText(kCopyright);
-  ui->lblImportantDevs->setText(QStringLiteral("%1\n").arg(s_awesomeDevs.join(", ")));
+
+  // Use non-breaking space in each awesome dev name so names are not split across lines.
+  QStringList devsNbsp;
```
sithlord48, line 39: I was hoping to change this format to be something like 
```
Cool Dev: They did a cool thing
Awesome Dev: did some awesome thing
...
```
So each dev has a bit of info about why they are cool. Add Joining them with \n makes a nice list. 
If we need more the one line we can make each person their own little box w/ label inside.  Neither of these will fit into the window without scrolling but i think would be much nicer over all.

nbolton: Ah yeah that's a nice idea.

sithlord48: Its up to you if you want to land this as is or work on fleshing out that info a bit more.

sithlord48: on Second thought let's land this for now and we can fix it when / if we do later on

sithlord48 (approved):

src/lib/gui/dialogs/AboutDialog.cpp
nbolton: Yeah that's cool, don't mind if this code gets replaced whenever there's a better approach.

PR #8707, opened 2025-06-19
Sonar cleanup 3

sithlord48:
Continue Cleaning up the sonar warnings

### DIFF
--- a/src/lib/base/EventQueue.h
+++ b/src/lib/base/EventQueue.h
@@ -122,7 +122,6 @@ private:
   // event handlers
   HandlerTable m_handlers;
 
-private:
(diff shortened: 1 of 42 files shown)


src/lib/base/EventQueue.h
```diff
@@ -122,7 +122,6 @@ class EventQueue : public IEventQueue
   // event handlers
   HandlerTable m_handlers;
 
-private:
```
nbolton, line 125: Why make these public?

nbolton (approved):

src/lib/base/EventQueue.h
nbolton: Never mind, I see private is already on line 71.

PR #8718, opened 2025-06-24
chore: simplify icon theme path fallback path

sithlord48:
Simplify how we resolve the fallback search paths. 
 1. We do not need to to search any themes child dirs. `StyleUtil::checkSubDir` is not needed  and has been removed
 2. Set our fallback search to always include the icon theme path `:icons/deskflow-light` or `:/icons/deskflow-dark` `StyleUtil::SetFallbackThemePath` has been removed and just is set in deskflow-gui.cpp 
 3. Remove the Theme settings in MainWindow and do it deskflow-gui

### DIFF
--- a/src/apps/deskflow-gui/deskflow-gui.cpp
+++ b/src/apps/deskflow-gui/deskflow-gui.cpp
@@ -111,8 +111,13 @@ int main(int argc, char *argv[])
   }
 #endif
 
-  // Sets the fallback icon path
-  setIconFallbackPaths();
+  // Sets the fallback icon path and fallback theme
+  const auto themeName = QStringLiteral("deskflow-%1").arg(iconMode());
+  if (QIcon::themeName().isEmpty())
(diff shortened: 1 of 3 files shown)


nbolton:
On `master`, some buttons are broken in light mode on GNOME:
image

- Fingerprint in bottom left and next to 'Configure server'
- Restart button icon

Looks like not introduced by your PR and it slipped through at some point.

p.s. Damn you for making me use light mode :p (my eyes, my eyes...)

nbolton (approved):

sithlord48:
@nbolton  updated i think its got confused by both icons being there.

nbolton (approved):
Looks the same, but like I said, I think it's a separate issue/PR as the problem exists in `master` (edit: i.e. this PR doesn't make the problem worse, so it's good to land)

Raised:
- #8722

image

image

PR #8728, opened 2025-06-30
chore: fix W:useless-whatis-entry lintian warnings

kenhys:
By default, help2man generates "manual page for foobar" for manual files.

It is useless because there is no addtional information which explains what deskflow is.

With --name option, .SH NAME section will be set correctly.

Before:

```
  .\" DO NOT MODIFY THIS FILE!  It was generated by help2man 1.49.3.
  .TH DESKFLOW-CLIENT "1" "June 2025" "deskflow-client v1.22.0.109, protocol v1.8" "User Commands"
  .SH NAME
  deskflow-client \- manual page for deskflow-client v1.22.0.109, protocol v1.8
  .SH SYNOPSIS
  .B deskflow-client
  [\fI\,OPTIONS\/\fR]
  .SH DESCRIPTION
```

After:

```
  .\" DO NOT MODIFY THIS FILE!  It was generated by help2man 1.49.3.
  .TH DESKFLOW-CLIENT "1" "June 2025" "deskflow-client v1.22.0.110, protocol v1.8" "User Commands"
  .SH NAME
  deskflow-client \- Keyboard and mouse sharing client
  .SH SYNOPSIS
  .B deskflow-client
  [\fI\,OPTIONS\/\fR]
  .SH DESCRIPTION
```


sithlord48:
Can you show me an example of the output  before and after this

kenhys:
Updated commit message.

nbolton:
Next steps?

sithlord48:
I would like to see if we can come up with better descriptions then just reusing the project description to describe each application.

nbolton:
@kenhys Any interest in improving your PR?

sithlord48 (approved):
I don't think this is the best description but its at least using the cmake project description for the string now (and not a new duplicated string). While not ideal i think its a good start If we think of better descriptive strings we can improve them later on.

PR #8738, opened 2025-07-03
Sonar cleanup 6

sithlord48:
Squash more sonar warnings
  - noexcept for swap methods (sonar considers this a "BLOCKING" issue)
  - do not use nested `new`  in calls (sonar considers this a "BLOCKING" issue)
  - comment empty method bodies 
  - prefer in-class initializers over initializer lists.
  - use constexpr / static const in place of defines
  - More enum class use
  - Remove unused < operator for Timer type
  - add default cases for switches
  - Mark a few missed methods noreturn

### DIFF
--- a/src/lib/deskflow/App.h
+++ b/src/lib/deskflow/App.h
@@ -130,59 +130,53 @@ private:
 #else
 #define DAEMON_RUNNING(running_)
 #endif
+constexpr static auto s_helpGeneralArgs = //
+    "  -d, --debug <level>      filter out log messages with priority below level.\n"
+    "                             level may be: FATAL, ERROR, WARNING, NOTE, INFO,\n"
+    "                             DEBUG, DEBUG1, DEBUG2.\n"
+    "  -n, --name <screen-name> use screen-name instead the hostname to identify\n"
+    "                             this screen in the configuration.\n"
+    "  -1, --no-restart         do not try to restart on failure.\n"
+    "*     --restart            restart the server automatically if it fails.\n"
+    "  -l  --log <file>         write log messages to file.\n"
+    "      --enable-crypto      enable TLS encryption.\n"
+    "      --tls-cert           specify the path to the TLS certificate file.\n";
 
-#define HELP_COMMON_INFO_1                                                                                             \
-  "  -d, --debug <level>      filter out log messages with priority below "                                            \
-  "level.\n"                                                                                                           \
-  "                             level may be: FATAL, ERROR, WARNING, NOTE, "                                           \
-  "INFO,\n"                                                                                                            \
-  "                             DEBUG, DEBUG1, DEBUG2.\n"                                                              \
-  "  -n, --name <screen-name> use screen-name instead the hostname to "                                                \
-  "identify\n"                                                                                                         \
-  "                             this screen in the configuration.\n"                                                   \
-  "  -1, --no-restart         do not try to restart on failure.\n"                                                     \
-  "*     --restart            restart the server automatically if it fails.\n"                                         \
-  "  -l  --log <file>         write log messages to file.\n"                                                           \
-  "      --enable-crypto      enable TLS encryption.\n"                                                                \
-  "      --tls-cert           specify the path to the TLS certificate file.\n"
+constexpr static auto s_helpVersionArgs = //
+    "  -h, --help               display this help and exit.\n"
+    "      --version            display version information and exit.\n";
 
-#define HELP_COMMON_INFO_2                                                                                             \
-  "  -h, --help               display this help and exit.\n"                                                           \
-  "      --version            display version information and exit.\n"
-
-#define HELP_COMMON_ARGS                                                                                               \
-  " [--name <screen-name>]"                                                                                            \
-  " [--restart|--no-restart]"                                                                                          \
-  " [--debug <level>]"
+constexpr static auto s_helpCommonArgs = //
+    " [--name <screen-name>]"
+    " [--restart|--no-restart]"
+    " [--debug <level>]";
 
 // system args (windows/unix)
 #if SYSAPI_UNIX
 
 // unix daemon mode args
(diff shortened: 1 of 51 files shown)


src/lib/deskflow/App.h
```diff
@@ -130,59 +130,53 @@ class App : public IApp
...
-
-#define HELP_COMMON_INFO_2                                                                                             \
-  "  -h, --help               display this help and exit.\n"                                                           \
-  "      --version            display version information and exit.\n"
-
-#define HELP_COMMON_ARGS                                                                                               \
-  " [--name <screen-name>]"                                                                                            \
-  " [--restart|--no-restart]"                                                                                          \
-  " [--debug <level>]"
+constexpr static auto s_HelpCommon_1 = //
```
nbolton:
```suggestion
constexpr static auto s_helpCommon1 = //
```

nbolton:
> * Win Server

Works :+1:

sithlord48:
updated the name lmk if you find anything else @nbolton

src/lib/deskflow/ProtocolUtil.cpp
```diff
@@ -255,6 +255,9 @@ uint32_t ProtocolUtil::getLength(const char *fmt, va_list args)
         case 4:
           len = 4 * (uint32_t)(va_arg(args, std::vector<uint32_t> *))->size() + 4;
           break;
+
+        default:
+          break;
```
nbolton: A warning or at least debug log message here would be prudent.

src/lib/gui/KeySequence.cpp
```diff
@@ -163,6 +163,8 @@ QString KeySequence::keyToString(int key)
       return "2";
     case Qt::MiddleButton:
       return "3";
+    default:
```
nbolton, line 166: Same here, warn or debug would be wise.

src/lib/gui/ScreenSetupModel.cpp
```diff
@@ -57,6 +57,9 @@ QVariant ScreenSetupModel::data(const QModelIndex &index, int role) const
 
   case Qt::DisplayRole:
     return screen(index).name();
+
+  default:
+    return QVariant();
   }
   return QVariant();
```
nbolton: Maybe we can get away with not logging default case here as that'd probably just be noise in this case, but the return at end of function is possibly redundant.
```suggestion

  default:
    return QVariant();
}
```

src/lib/platform/XWindowsScreen.cpp
```diff
@@ -1333,6 +1333,9 @@ void XWindowsScreen::handleSystemEvent(const Event &event)
         LOG((CLOG_INFO "group change: %d", xkbEvent->state.group));
         m_keyState->setActiveGroup((int32_t)xkbEvent->state.group);
         return;
+
+      default:
```
nbolton, line 1337: 50/50 here whether log is noise, but worth discussing.

src/lib/platform/XWindowsScreenSaver.cpp
```diff
@@ -160,6 +160,8 @@ bool XWindowsScreenSaver::handleXEvent(const XEvent *xevent)
       return true;
     }
     break;
+  default:
+    break;
```
nbolton, line 164: Happy to skip log here, as X is deprecated anyway.

src/lib/server/ClientProxyUnknown.cpp
```diff
@@ -181,6 +181,9 @@ void ClientProxyUnknown::initProxy(const std::string &name, int major, int minor
     case 8:
       m_proxy = new ClientProxy1_8(name, m_stream, m_server, m_events);
       break;
+
+    default:
+      break;
```
nbolton, line 186: Warn/error log here would be a good idea.

nbolton (approved):
Happy to land but some minor nits could be addressed first if you feel like it.

src/lib/deskflow/ProtocolUtil.cpp
sithlord48: This case should be impossible due to the above assert but ill add one.

src/lib/gui/ScreenSetupModel.cpp
sithlord48: It should be, If i remove it i suspect a this function could not return a value warning so leaving it for that.  so ill just break instead

src/lib/platform/XWindowsScreen.cpp
sithlord48: > Happy to skip log here, as X is deprecated anyway.

src/lib/server/ClientProxyUnknown.cpp
sithlord48: i was thinking that but  the logging would be redundant.  If we leave this switch we throw an unknown Protocol error that contains the info we would have logged as err / warn.

src/lib/deskflow/ProtocolUtil.cpp
nbolton: Because `assert` is optimized out in release builds (only compiled in debug builds) we should certainly handle this case, and it should be an error or at least warning.

Also the message will be hard to grok for newcomers:
`I message unexpected length, 5`

Devs unfamiliar with this code are unlikely to realize that "I" means "format specifier %I5" (if 5 is the length) and I think using the word "message" here is a bit confusing as it's referring to a particular part of the format string (the format specifier).

Something like this would be better:
`format specifier %I5 has invalid length`

So, I guess this would be something like:
`LOG((CLOG_ERR "format specifier %%I%d has invalid length", len));`

src/lib/server/ClientProxyUnknown.cpp
nbolton: Ah, yeah I didn't zoom out.

Issue #7865, opened 2024-11-06
100 ways to leave your Segfault 


nPHYN1T3:
### Project

Deskflow

### Deskflow version number

1.17.0.176

### Build type

Local developer build (built it myself)

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [X] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [ ] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [X] I am not using Wayland on Linux

### Signing on macOS

- [ ] I have authorized the app to run on my Mac
- [X] I am not using macOS

### Continuous build

- [ ] I have tried the latest continuous build and the issue persists
- [X] I am unable to try the latest continuous build

### OS versions/distros

Arch (pure) 6.11.6-arch1-1

### Deskflow configuration

Server is single XScreen
Client is triple XScreen, deskflow instance per

### What steps will reproduce the problem?

Launch GUI on client, configure and when you click connect - segfault.
Launch GUI on server, click regenerate TLS cert - segfault

### Log output

```shell
2024-11-01T14:38:00] INFO: Deskflow v1.17.0.169
[2024-11-01T14:38:00] INFO: no .env file in dir: CWD
[2024-11-01T14:38:00] INFO: no .env file in app config dir: $HOME/.config/Deskflow
[2024-11-01T14:38:00] INFO: starting core client process (desktop mode)
Failed to compute left/right minimum bearings for "Courier"
[2024-11-01T14:38:00] INFO: running command: /usr/bin/deskflow-client -f --no-tray --debug INFO --name name --enable-crypto --sync-language pillar:24800
Failed to compute left/right minimum bearings for "Courier"
Failed to compute left/right minimum bearings for "Courier"
[2024-11-01T14:38:00] WARNING: Failed to compute left/right minimum bearings for "Courier"
Failed to compute left/right minimum bearings for "Courier"
Failed to compute left/right minimum bearings for "Courier"
Failed to compute left/right minimum bearings for "Courier"
Failed to compute left/right minimum bearings for "Courier"
Segmentation fault (core dumped)

[...]

Issue #8332, opened 2025-03-09
Add Ctrl+F for log section in main window


Slient1802:
### Feature Request

# Issue Description:
Currently, DeskFlow lacks an efficient way to search log files using Ctrl + F, making it difficult for developers and support teams to track past occurrences of bugs. When troubleshooting recurring issues, manually scanning through large log files is time-consuming and inefficient. Implementing a searchable log system would greatly enhance debugging efficiency.

## Steps to Reproduce:

Attempt to use Ctrl + F to search for a specific error or keyword in log section.

Observe that there is no direct or efficient way to locate previous bug entries.

## Expected Behavior:
Users should be able to use Ctrl + F or an equivalent search function to quickly locate relevant error messages within the log files.

## Actual Behavior:
Manual scrolling is required to find specific log entries, significantly slowing down the debugging process.

## Suggested Solution:

Implement a built-in search function within DeskFlow's log viewer.

Enable Ctrl + F functionality if logs are displayed in a UI component.

![Image](https://github.com/user-attachments/assets/0df7d1fd-b85d-4696-8b7c-c45de5c6bff2)

Provide filtering options to refine searches based on error types, timestamps, or keywords.

## Impact:

Improves accuracy in identifying recurring bugs.

Enhances overall efficiency finding old bug being detected.

Additional Notes:
This enhancement will streamline the debugging process and improve the overall usability of DeskFlow for development and support teams.

### Voting Instructions

React with :+1: or :-1: to vote on this request

nbolton:
This would be a very nice feature.

RSlient007:
> This would be a nice feature. Currently to find in the log you have to copy/paste to a text editor, which is a bit tedious.

I want to strongly support this feature request! The ability to use Ctrl + F and implement robust filtering options (by timestamp, level, or keyword) would be a massive efficiency gain for both development and support teams.

# To make this feature even more valuable, I suggest a complementary function focused on security:

# Feature Extension: Log Redaction/Replacement 🔒
Could we include an option to redact or replace sensitive data in the log viewer? Logs often contain highly sensitive information like TLS keys, peer fingerprints, etc.

## Refinement for Redaction Implementation:
When implementing the replacement function, it's critical that the substitution maintains context for developers while securing the data. Instead of a generic [REDACTED], the replacement string should convey the original data's length or size.

# This could be achieved via two methods:

## Fixed-Length Replacement: 
[...]

PR #8750, opened 2025-07-14
chore: MainWindow, Use system monospace font in log

sithlord48:
Use the system's monospace font for the log.
Bump the font size up a bit for macOS to 12px. Its default is way to small.

fixes : maybe some of https://github.com/deskflow/deskflow/issues/7865 , Others?

windows runner was updated to ninja 1.13.1 so reverting the commit to update it with our action

### DIFF
--- a/src/lib/gui/MainWindow.cpp
+++ b/src/lib/gui/MainWindow.cpp
@@ -85,6 +85,14 @@ MainWindow::MainWindow()
 {
   ui->setupUi(this);
 
+  // setup the log font
+  ui->textLog->setFont(QFontDatabase::systemFont(QFontDatabase::FixedFont));
+#ifdef Q_OS_MAC
+  auto f = ui->textLog->font();
+  f.setPixelSize(12);
+  ui->textLog->setFont(f);
+#endif
+
   // Setup Actions
   m_actionAbout->setText(tr("About %1...").arg(kAppName));
   m_actionAbout->setMenuRole(QAction::AboutRole);
--- a/.github/actions/install-dependencies/action.yml
+++ b/.github/actions/install-dependencies/action.yml
@@ -27,10 +27,9 @@ runs:
 
   steps:
     - name: Install Depends
+      if: ${{ runner.os != 'Windows' }}
       run: |
-          if [ "$RUNNER_OS" == "Windows" ]; then
-            choco install ninja
-          elif [ "$RUNNER_OS" == "macOS" ]; then
+          if [ "$RUNNER_OS" == "macOS" ]; then
             brew install googletest openssl --quiet
           elif [ "$RUNNER_OS" == "Linux" ]; then
             if [ ${{inputs.like}} == "debian" ]; then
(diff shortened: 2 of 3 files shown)


nbolton:
Hmm, doesn't seem to be monospaced on GNOME.

Before:
image

With `QFont::Monospace`:
image

sithlord48:
@nbolton fixed.

nbolton:
> @nbolton fixed.

Is this monospace? I can't quite tell... It's a bit harder to read than the original font.

image

sithlord48:
Its not forcing a font it should match the font the gui with a 14px height.  We can do more about this later on since im gonna have to make a proper log widget if we want to tackle #8332. I think then the log widget may gain a few optional settings like font selection (at least)

edit: it seams to not be as Bold as the previous font to me, its gonna depend on the font your system uses for the log.

src/lib/gui/MainWindow.cpp
```diff
@@ -85,6 +85,12 @@ MainWindow::MainWindow()
 {
   ui->setupUi(this);
 
+  auto f = font();
+  f.setFamily("");
+  f.setStyleHint(QFont::Monospace);
+  f.setKerning(true);
```
nbolton: Please help me understand why you are setting kerning to `true`. I can't see why kerning makes sense here; monospace fonts have fixed-width glyphs and do not use kerning.

nbolton: I don't think the style hint is working for me. This does not look like a monospace font to me:

Image


```suggestion
  f.setFamily("monospace");
```

If I set the family to monospace, that actually does give me a monospace font.

Image

sithlord48: I was just matching how the old font was set on the UI . it was a mono space font w/ kerning on ..i will disable it if you prefer.

nbolton:
Looks good (actually better) on GNOME, KDE, and Windows... but I'm not so sure about macOS. I find the smaller font harder to read.

Here's how default Qt fixed width looks against the macOS Terminal app default:

Screenshot 2025-07-18 at 11 24 15

Screenshot 2025-07-18 at 11 25 38

What do you think about setting a specific font (`SF Mono`) just for macOS? I believe this font will always be available on macOS since it's the default that Apple uses. Alternatively, `Menlo` is a popular choice on macOS for consoles.

Edit: Maybe macOS is fine and I'm just not used to it. Open to feedback on that one as it's pretty subjective. Land as-is if you feel like it's fine.

----

# macOS

## Before

Screenshot 2025-07-18 at 11 30 26

## After

Screenshot 2025-07-18 at 11 27 49

----

# Windows

## Before

image


## After

image

----

# GNOME

Quite similar, but a bit nicer after.

## Before

image

## After

image

----

# KDE

## Before
image

## After

image

----

sithlord48:
Leave it to mac os to have a Smaller then needed font for the system default monospace ..  let me see if i can bump it up a bit

sithlord48:
Gonna land. Approval got removed when addressing mac os log size.

PR #8758, opened 2025-07-16
Install less windows Dlls

sithlord48:
- Modify the run of windeploy qt to only run once
  - place a copy of the depends in the bin folder for devs
  - use install to make sure qt depends are in the package when deploying
  - deploy qt plugins into plugins folder
- Do not include these dll files for windows deployments
  - `dxil.dll`(~1.5Mb)  we do not use dx shaders
  - `dxcompiler.dll` (~14Mb)  we do not use dx shaders
  - `opengl32sw.dll` (~20mb) we don't use Qml or OpenGLWidgets


nbolton:
I'll test this on a fresh VM.

nbolton (approved):
Tested on fresh Windows VM, works.

PR #8790, opened 2025-07-28
Fix mouse ghost gap (delay cursor show) bug on Windows client

nbolton:
Previously we were toggling the 'MouseKeys' Windows accessibility feature on and off, but this was causing many bugs; cursor show delay while Windows brings up the mouse stack(?) and often forever lost mouse cursor when toggling off the 'MouseKeys' feature. After a few days of hacking away at this, it seems safest to just leave it on.


sithlord48 (dismissed):
Looks good only one minor suggestion.

sithlord48 (approved):

PR #8794, opened 2025-07-30
Sonar Cleanup 10

sithlord48:
Next round of Sonar updates, This round includes the following
 - Replace use of Synergy name in clipboard test with Deskflow
 - IClipboard::EFormat => Enum class `IClipbpoard::Format`
 - ServerApp:EServerState => Enum class `ServerApp::ServerState`
 - AddResults => enum class `ScreenAddResults`
 - Remove Screen Corner enums and replace with a set of statics
 - Unify ENetworkProtocol and ServerProtocol in a new enum class NetworkProtocol in file base/Networkprotocol.h

### DIFF
--- a/src/lib/base/CMakeLists.txt
+++ b/src/lib/base/CMakeLists.txt
@@ -25,6 +25,7 @@ add_library(base STATIC
   Log.cpp
   Log.h
   LogLevel.h
+  NetworkProtocol.h
   Path.cpp
   Path.h
   PriorityQueue.h
--- a/src/lib/base/NetworkProtocol.h
+++ b/src/lib/base/NetworkProtocol.h
@@ -1,0 +1,13 @@
+/*
+ * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2025 Deskflow Developers
+ * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
+ */
+
+#pragma once
+
+enum class NetworkProtocol
+{
+  Synergy,
+  Barrier
+};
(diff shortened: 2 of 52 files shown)


src/lib/base/Enum_NetworkProtocol.h
nbolton, line 1: Why is the filename not `NetworkProtocol.h`?

sithlord48: I dont want it to be confused for a class that handles network protocol items.. 

We have a few enums  that are not really tied to a class or would be better used outside i.e other files are including for just the enum alone. Spiting some enums to their own headers or making just a enums.h in common / base or something might be a good idea. 

 I had considered the `Type_` prefix as well but in this case `Enum_`  makes it very clear that file is a header for an enum.

nbolton: Yeah, I get the intent behind the `Enum_` prefix, but it feels unconventional and a bit awkward in practice. Header files usually reflect the logical domain and not the internal structure. Using `Enum_` as a prefix leaks implementation detail into the filename and breaks away from common naming conventions. That could create some confusion; some files will follow this pattern, others won’t, and people won’t always know when/whether to use it.

Enforcing it consistently would also be a pain, and we’d likely end up needing extra rules or documentation to maintain it which neither of us want. I'm worried that if someone’s searching for `NetworkProtocol.h`, they could stumble if they don't know to look for `Enum_NetworkProtocol.h`, especially if they’re new to the codebase. Sure, fuzzy finders help, but consistency still matters... at least for keeping mental overhead low. Deviating from well known code styles always adds friction in my experience.

sithlord48: So we renaming this or not then ? I don't really care either way if you don't like the Enum prefix then we can drop it .

nbolton: I would appreciate it if the file was called `NetworkProtocol.h` :)

nbolton (approved):

PR #8804, opened 2025-08-01
Fix: Crash when m_buttons size is smaller than maxButton

corossig:
Issue https://github.com/deskflow/deskflow/issues/8803

This prevent crash of deskflow-server

### DIFF
--- a/src/lib/platform/XWindowsScreen.cpp
+++ b/src/lib/platform/XWindowsScreen.cpp
@@ -1863,7 +1863,7 @@ void XWindowsScreen::updateButtons()
   }
 
   // allocate button array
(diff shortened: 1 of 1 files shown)


sithlord48 (approved):
This seams reasonable.

sithlord48:
Can you fix the commit message so its `fix: ...`. Our release notes are auto generated.   (git commit --amend / git push --force) 

Edit: If I do it i can no longer "approve" the PR since i'll then count as an author for the review check.

sithlord48 (changes requested):
The commit message needs to be change to start with `fix: `

PR #8805, opened 2025-08-01
Sonar cleanup 12

sithlord48:
Next round of sonar "issues"  cleanup 

 - Add `explicit` for a bunch of constructiors
 - Remove unused `StopWatch::double()`
 - Make `App::m_events` private, access it from existing `App::getEvents()` in the sub classes
 - Make `TCPListenSocket` protected var's private and access via protected methods in subclasses
 - Make `TCPSocket::m_events` private access it from getEvents() in sub classes.
     - Note Attempting to make `TcpSocket::outputBuffer` / `TcpSocket::inputBuffer` private with protected member access did build but resulted in failure to connect.  
 - Unify `ClientApp::handleScreenError` and `ServerApp::handleScreenError` into `App::handleScreenError`
 - Make `App::m_bye` protected member
 - Remove `do nothing` methods `ClientApp::updateStatus(...)` and `ServerApp::updateStatus(...)` (was even calling it in places)

### DIFF
--- a/src/lib/net/TCPListenSocket.h
+++ b/src/lib/net/TCPListenSocket.h
@@ -44,10 +44,24 @@ public:
 protected:
   void setListeningJob();
 
-  IEventQueue *m_events;
-  ArchSocket m_socket;
-  SocketMultiplexer *m_socketMultiplexer;
+  ArchSocket socket() const
+  {
+    return m_socket;
+  }
+
+  IEventQueue *events() const
+  {
+    return m_events;
+  }
+
+  SocketMultiplexer *socketMultiplexer() const
+  {
+    return m_socketMultiplexer;
+  }
 
 private:
+  ArchSocket m_socket;
+  IEventQueue *m_events;
+  SocketMultiplexer *m_socketMultiplexer;
   std::mutex m_mutex;
 };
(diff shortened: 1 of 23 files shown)


src/lib/net/TCPListenSocket.h
```diff
@@ -44,10 +44,24 @@ class TCPListenSocket : public IListenSocket
 protected:
   void setListeningJob();
 
-  IEventQueue *m_events;
-  ArchSocket m_socket;
-  SocketMultiplexer *m_socketMultiplexer;
+  ArchSocket archSocket() const
+  {
+    return m_socket;
+  }
```
nbolton: The accessor should match the underlying member. Either of these would do, but I prefer `socket()`.

```suggestion
  ArchSocket socket() const
  {
    return m_socket;
  }
```

```suggestion
  ArchSocket archSocket() const
  {
    return m_archSocket;
  }
```

nbolton (dismissed):
OK to land, but one minor nit if you want to fix.

src/lib/net/TCPListenSocket.h
sithlord48: socket is used as a name in methods so that is why I did not use it. Let  me check if its one the one spot.

nbolton (approved):

PR #8806, opened 2025-08-03
fix: server: initialize the m_protocol member with a default value

chrschmidt:
It is likely random whether the server works or not without a manually configured protocol option, which is neither documented nor enforced to be set.
With a random uninitialized value in m_protocol, this can happen: 
[2025-08-03T12:39:57] NOTE: accepted client connection
[2025-08-03T12:39:57] FATAL: a runtime error occurred: XInvalidProtocol

Reports of "it works with debug enabled" are likely due to having parts of the memory space the server object is allocated into being cleared, and thus having a 0 (Synergy) startup value.

### DIFF
--- a/src/lib/server/Server.h
+++ b/src/lib/server/Server.h
@@ -358,7 +358,7 @@ private:
   };
 
   // used in hello message sent to the client
-  NetworkProtocol m_protocol;
+  NetworkProtocol m_protocol = NetworkProtocol::Barrier;
 
   // the primary screen client
   PrimaryClient *m_primaryClient;


src/lib/server/Server.h
```diff
@@ -358,7 +358,7 @@ class Server : public INode
   };
 
   // used in hello message sent to the client
-  NetworkProtocol m_protocol;
+  NetworkProtocol m_protocol = NetworkProtocol::Synergy;
```
sithlord48: Our default is Barrier.

sithlord48 (changes requested):
We generally do not want people to use non deskflow server config. This should only be an issues if your doing that. That being said this is not a bad idea but there is only one small issue our default protocol is Barrier.

src/lib/server/Server.h
chrschmidt: Updated default to Barrier as requested.

chrschmidt:
> We generally do not want people to use non deskflow server config and that is when this should be an issue. Internally our settings object defaults to Barrier. This is not a bad idea however there is one small issue with this pr, our default protocol is Barrier not Synergy.

The whole [wiki page](https://github.com/deskflow/deskflow/wiki/Text-Config) on this subject does not mention the need, or even the option, though. I was told "the issue is known for quite some time", and at least at that point, the person who wrote that did not know the correct solution, but recommend to just run it in debug mode as a solution.

nbolton (approved):

sithlord48:
> > We generally do not want people to use non deskflow server config and that is when this should be an issue. Internally our settings object defaults to Barrier. This is not a bad idea however there is one small issue with this pr, our default protocol is Barrier not Synergy.
> 
> The whole [wiki page](https://github.com/deskflow/deskflow/wiki/Text-Config) on this subject does not mention the need, or even the option, though. I was told "the issue is known for quite some time", and at least at that point, the person who wrote that did not know the correct solution, but recommend to just run it in debug mode as a solution.

That entire page needs to be updated.

sithlord48 (approved):
Looks good to me!

Issue #8807, opened 2025-08-04
Print version with Git SHA to stdout in debug mode


nbolton:
### Feature Request

I can see the Git SHA from the about screen but not from stdout. This is inconsistent and means I have to open the GUI to see what Git SHA the build is from.

Whenever we print the version to stdout, also print the Git SHA.

sithlord48:
.... This is why i didn't want to show the a sha (and we have unique versions numbers) .. next its gonna be the continuous packages does not have a sha in it name so i have to run it to see what version...

PR #8811, opened 2025-08-04
refactor: show display version for --version

sithlord48:
fixes: #8807 by using the display version when calling `--version`


src/lib/deskflow/App.cpp
nbolton, line 1: FYI it's also printed here:
https://github.com/deskflow/deskflow/blob/8182df40e86fc891d2abc651c8611edacb37998a/src/apps/deskflow-gui/deskflow-gui.cpp#L124

I'm guessing you want to axe that `qPrintable`

sithlord48: yeah i don't see why not lol

nbolton (approved):

Issue #8168, opened 2025-02-05
Some chars fake the wrong character on Wayland client


jwakely:
### Project

Deskflow

### Deskflow version number

1.19.0

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [ ] Linux (X11)
- [x] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [x] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [ ] I am not using Wayland on Linux

### Signing on macOS

- [ ] I have authorized the app to run on my Mac
- [x] I am not using macOS

### Continuous build

- [x] I have tried the latest continuous build and the issue persists
- [ ] I am unable to try the latest continuous build

### OS versions/distros

Fedora 41 (for both server and client)

### Deskflow configuration

_Maintainer edit:_ Bug happens when using Windows and macOS servers; bug is isolated to Wayland client.

Fedora 41 server, laptop with external screen and external USB mouse+keyboard (all connected through thunderbird3 dock).
Fedora 41 client, desktop with a single screen.
Server on the left, client on the right.

### What steps will reproduce the problem?

Start Deskflow, move mouse to client, type:
<kbd>[</kbd> <kbd>]</kbd> <kbd>{</kbd> <kbd>}</kbd> <kbd>\<</kbd> <kbd>\></kbd> <kbd>\\</kbd> <kbd>|</kbd>

On the client, the output is:
<kbd>8</kbd> <kbd>9</kbd> <kbd>&</kbd> <kbd>)</kbd> <kbd>Z</kbd> <kbd>X</kbd> <kbd>-</kbd> <kbd>¬</kbd>

I initially though this was part of #7886 but the problem remains after that has been fixed. Maybe also related to #7499

### Log output

Server
```shell
[2025-02-05T11:55:32] INFO: starting core server process (desktop mode)
[2025-02-05T11:55:32] INFO: core config file: /home/jwakely/.config/Deskflow/deskflow-server.conf
[2025-02-05T11:55:32] INFO: running command: /usr/bin/deskflow-server -f --debug INFO --name carbon.home --enable-crypto --address :24800 -c /home/jwakely/.config/Deskflow/deskflow-server.conf --tls-cert /home/jwakely/.config/Deskflow/tls/deskflow.pem
[...]

PR #8813, opened 2025-08-05
tools: add a protocol debugger

whot:
This adds a Python-based protocol-debugger that can be used as MITM between a server and clients. The easiest invocation of it is to run without argument on the same host as the server and then connect to it via a client on port 24801.

This will simply print out all known messages as they fly between client and server, making it easier to figure out what's going on on the wire.

Note that the message parsing was intentionally hand-implemented to potentially show any discrepancies between implementations that rely on the format strings. The format strings from ProtocolTypes.cpp are included for human verification of the parsing vs what is expected.

Typical output is:

```console
$ ./protocol-debugger.py
10:15:56 - deskflow        - I - Listening on port 24801, connecting to localhost:24800
10:15:56 - deskflow        - I - Client connected from 192.168.1.235:54952
10:15:56 - deskflow        - I - Connected to remote localhost:24800
10:15:56 - server → client - I - Barrier major=1, minor=8
10:15:56 - server ← client - I - Barrier major=1, minor=8
10:15:56 - server → client - I - QINF
10:15:56 - server → client - I - LSYN data=en
10:15:56 - server ← client - I - DINF x=0, y=0, w=1920, h=1080, mx=0, my=960, size=540
10:15:56 - server → client - I - CIAK
10:15:56 - server → client - I - CROP
10:15:56 - server → client - I - DSOP options=30
[...]
10:16:48 - server ← client - I - CNOP
10:16:48 - server → client - I - DMMV x=897, y=297
10:16:48 - server ← client - I - CNOP
10:16:48 - server → client - I - DMMV x=895, y=295
10:16:48 - server ← client - I - CNOP
10:16:48 - server → client - I - DKDL keyid=97, mask=0, button=38, lang= (obfuscated)
10:16:48 - server ← client - I - CNOP
10:16:48 - server → client - I - DKDL keyid=97, mask=0, button=38, lang= (obfuscated)
```
By default keycodes and buttons are obfuscated to always 97/38 (the 'a' key). This is to avoid users accidentally leaking private data when attaching logs, use --show-keycodes if debugging key values is actually neccessary.

Prompted by #8168 
For full disclosure: written with the help of Cursor's LLM.


whot:
ftr, this was a quick-enough hack I did last night and polished this morning and I added it to `tools/` because I spent a grand total of 30s figuring out where to put it :) Happy to change this to anywhere, rename it, etc. Or alternatively even put it in a separate repo though not sure that would have a huge benefit.

sithlord48:
thinking [deskflow/scripts](https://github.com/deskflow/scripts)  might be a good repo for this since that has our additional python scripts .

tools/protocol-debugger.py
```diff
@@ -0,0 +1,845 @@
...
+
+
+@dataclass
+class MessageHelloBarrier(_MessageHello):
+    code: str = "Barrier"
+
+
+@dataclass
+class MessageHelloDeskflow(_MessageHello):
+    code: str = "Deskflow"
```
nbolton: FYI, there is no "Deskflow" protocol; we use either "Synergy" or "Barrier" (with Barrier, of course, just being a clone/rename of the Synergy protocol with no material difference). If a "Deskflow" hello was added for future proofing, then it can be removed as we don't plan to change the protocol hello message.

tools/protocol-debugger.py
```diff
@@ -0,0 +1,845 @@
+#!/usr/bin/env python3
+
+# Deskflow -- mouse and keyboard sharing utility
+# SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
+# SPDX-FileCopyrightText: 2025 Red Hat
+
+# This is a protocol debugger working as a MITM process. Typically you would
+# run it on the same host as the Deskflow server, then connect a client to it
+# on the default port (24801). All messages between server and client are logged.
```
nbolton, line 9: Thanks for the PR.

Maybe a silly question, but: What's the benefit of using this script over Wireshark? 

image

It might be worth adding a "When to use this instead of Wireshark" type explanation at the top.

Off the top of my head:
- Another way to see how the protocol works (i.e. maybe easier than RTFM or reading C++)
- Allows for key obfuscation
- Can be extended in future for message injection
- Any other future plans or ideas that others could explore

tools/protocol-debugger.py
```diff
@@ -0,0 +1,845 @@
...
+        action="store_true",
+        help="Enable verbose logging",
+    )
+    parser.add_argument(
+        "--show-keycodes",
+        action="store_true",
+        default=False,
+        help="Show keycodes in the logging output (by default all keys are shown as A)",
+    )
+    args = parser.parse_args()
```
nbolton, line 764: If you feel like it/have time, an option to filter out certain messages like `CALV` and `CNOP` might be useful. If so, I'd consider turning it on by default. If not, no worries, and we can land as is.

Afterthought: ...unless the tool is intended to be used with grep for filtering, then never mind.

```
12:23:05 - server ← client - I - CALV 
12:23:05 - server ← client - I - CNOP 
12:23:08 - server → client - I - CALV 
12:23:08 - server ← client - I - CALV 
12:23:08 - server ← client - I - CNOP 
12:23:10 - server → client - I - CALV 
12:23:10 - server ← client - I - CALV 
12:23:10 - server ← client - I - CNOP 
12:23:14 - server → client - I - CALV 
12:23:14 - server ← client - I - CALV 
12:23:14 - server ← client - I - CNOP 
12:23:16 - server → client - I - CALV 
12:23:16 - server ← client - I - CALV 
12:23:16 - server ← client - I - CNOP 
````

File: tools/protocol-debugger.py [hunk 4]
```diff
@@ -0,0 +1,845 @@
...
+class MessageDMDN(ProtocolMessage):
+    code: str = "DMDN"
+    format_string: str = "DMDN%1i"
+    button: int = 0
+
+    @classmethod
+    def from_bytes(cls, data: bytes) -> Self:
+        data = data[len(cls.code) :]
+        return cls(
+            button=int.from_bytes(data[0:1], byteorder="big"),
```
nbolton, line 360: I'm wondering whether it would be useful to log this in hex, as that's how it appears in the Deskflow log:
`[2025-08-05T12:08:42] DEBUG1: recv key down id=0x00000038, mask=0x0000, button=0x001d, lang="en"`

nbolton (approved):

tools/protocol-debugger.py
nbolton, line 1: Oops, forgot to say, yes as @sithlord48 mentioned, let's put it in the scripts repo: https://github.com/deskflow/scripts

nbolton (changes requested):

tools/protocol-debugger.py
whot: I see your screenshot and raise you one :)
Screenshot From 2025-08-06 17-02-19

Biggest advantages:
- easier to start and find the various bits than wireshark
- can be used for single-client debugging - just connect to the 24801 port and only that client will be logged (same for server connection)
- easy to pipe and process, or grep for data
- key obfuscation as you mentioned
- custom processing of data for each message (see e.g. the `(obfuscated)` suffix)
- single no-dep python script so easy to download/run
- *could* be extended to modify certain messages, e.g. a few lines of python code to hardcode key events to be certain codes is possible.

Wireshark is great as a generic tool of course but I'm not sure anything beats a custom protocol debugger for any protocol out there :)

nbolton: Makes sense. Let's get this landed in https://github.com/deskflow/scripts -- want to open the PR or shall I?

tools/protocol-debugger.py
whot: removed

tools/protocol-debugger.py
whot: added `--filter=FOO,BAR` support which defaults to `CALV,CNOP`

tools/protocol-debugger.py
whot: This one is a bit more involved, need to find the time to do it. easiest approach is this I think:
```python
class HexInt(int):
   pass
```
and then change all the field that should be hex numbers and detect this in `__str__()`.

whot:
Moved to https://github.com/deskflow/scripts/pull/4 as requested, closing this one

tools/protocol-debugger.py
nbolton: Works great! Thanks.

PR #8815, opened 2025-08-05
Improve logging in `Server.cpp`

nbolton:
We plan to move this elsewhere, but in the meantime, the DEBUG1 logging was too noisy. Also fixed some code style that was bugging me.

### DIFF
--- a/src/lib/server/Server.cpp
+++ b/src/lib/server/Server.cpp
@@ -1423,7 +1423,8 @@ void Server::onClipboardChanged(const BaseClientProxy *sender, ClipboardID id, u
 
   // ignore if data hasn't changed
   if (data == clipboard.m_clipboardData) {
-    LOG((CLOG_DEBUG "ignored screen \"%s\" update of clipboard %d (unchanged)", clipboard.m_clipboardOwner.c_str(), id)
+    LOG(
+        (CLOG_DEBUG "ignored screen \"%s\" update of clipboard %d (unchanged)", clipboard.m_clipboardOwner.c_str(), id)
     );
     return;
   }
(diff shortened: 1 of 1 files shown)


sithlord48 (commented):
Why are some log calls 
` LOG_DEBUG2("mouse move on secondary: %+d,%+d", dx, dy);`
and other like this 
`LOG((CLOG_DEBUG2 "Adjusted to %+d,%+d using multiplier %.2f", adjustedDx, adjustedDy, multiplier));`
Are we able to use one style( I prefer the first one)?

nbolton:
> Are we able to use one style( I prefer the first one)?

Because I forgot to use the new style consistently. `LOG((CLOG_DEBUG2 ...))` is the old style from the original 90s-era code. `LOG_DEBUG2(...)` is the newer style, inspired by @p12tic's changes in Input Leap... unfortunately, we haven't yet implemented the new style quite as thoroughly as he did downstream.

Will fix now.

src/lib/server/Server.cpp
```diff
@@ -1670,21 +1667,19 @@ bool Server::onMouseMovePrimary(int32_t x, int32_t y)
...
-      dx = adjustedDx; // Update dx and dy to adjusted values
-      dy = adjustedDy;
+      double multiplier = std::stod(envVal);
+      dx = static_cast<int32_t>(std::round(dx * multiplier));
+      dy = static_cast<int32_t>(std::round(dy * multiplier));
+      LOG_DEBUG2("adjusted mouse x %.2f: %+d,%+d", multiplier, dx, dy);
     } catch (const std::exception &e) {
-      // Log the error message from the exception
-      LOG((CLOG_ERR "Invalid DESKFLOW_MOUSE_ADJUSTMENT value: %s. Exception: %s", envVal, e.what()));
+      LOG_ERR("invalid %s value: %s", adjustEnv, e.what());
```
nbolton, line 1681: Oops!

PR #8824, opened 2025-08-06
Sonar cleanup 14

sithlord48:
Next batch of sonar clean up  
 - DaemonApp: provide context object for lambda 
 - CoreProcess: fully qualify signal that send an enum
 - ScreenSetupModel::isFull compare const_iterators


src/lib/arch/Arch.cpp
```diff
@@ -30,14 +30,6 @@ Arch::Arch(Arch *arch)
   s_instance = arch;
 }
 
-void Arch::init()
-{
-  ARCH_NETWORK::init();
-#if SYSAPI_WIN32
-  ArchMiscWindows::init();
-#endif
-}
```
nbolton: > * Remove `Arch::init()` let arch call it from the base classes instead

This causes `ArchMiscWindows::init` to not be called.

On this PR:
```
calling arch init
ArchNetworkWinsock::init()
arch init called
```

On master:
```
calling arch init
ArchNetworkWinsock::init()
ArchMiscWindows::init()
arch init called
```

```patch
diff --git a/src/apps/deskflow-client/deskflow-client.cpp b/src/apps/deskflow-client/deskflow-client.cpp
index 68b374e82..3486faaca 100644
--- a/src/apps/deskflow-client/deskflow-client.cpp
+++ b/src/apps/deskflow-client/deskflow-client.cpp
@@ -16,6 +16,8 @@
 #include <QCoreApplication>
 #endif
 
+#include <iostream>
+
 int main(int argc, char **argv)
 {
 #if SYSAPI_WIN32
@@ -30,7 +32,9 @@ int main(int argc, char **argv)
 #endif
 
   Arch arch;
+  std::cout << "calling arch init" << std::endl;
   arch.init();
+  std::cout << "arch init called" << std::endl;
 
   Log log;
   EventQueue events;
diff --git a/src/lib/arch/win32/ArchMiscWindows.cpp b/src/lib/arch/win32/ArchMiscWindows.cpp
index 073e0478c..4cedb26d2 100644
--- a/src/lib/arch/win32/ArchMiscWindows.cpp
+++ b/src/lib/arch/win32/ArchMiscWindows.cpp
@@ -17,6 +17,7 @@
 
 #include <array>
 #include <filesystem>
+#include <iostream>
 
 // Useful for debugging Windows specific bootstrapping code before the logging system is initialized.
 // This output can be viewed by attaching a Microsoft debugger or by using the DebugView program.
@@ -60,6 +61,8 @@ HINSTANCE ArchMiscWindows::s_instanceWin32 = nullptr;
 
 void ArchMiscWindows::init()
 {
+  std::cout << "ArchMiscWindows::init()" << std::endl;
+
   // stop windows system error dialogs from showing.
   SetErrorMode(SEM_FAILCRITICALERRORS);
 }
diff --git a/src/lib/arch/win32/ArchNetworkWinsock.cpp b/src/lib/arch/win32/ArchNetworkWinsock.cpp
index a1f9fea8e..62d13d359 100644
--- a/src/lib/arch/win32/ArchNetworkWinsock.cpp
+++ b/src/lib/arch/win32/ArchNetworkWinsock.cpp
@@ -12,6 +12,7 @@
 #include "arch/win32/ArchMultithreadWindows.h"
 #include "arch/win32/XArchWindows.h"
 
+#include <iostream>
 #include <malloc.h>
 
 static const int s_family[] = {
@@ -100,6 +101,8 @@ ArchNetworkWinsock::~ArchNetworkWinsock()
 
 void ArchNetworkWinsock::init()
 {
+  std::cout << "ArchNetworkWinsock::init()" << std::endl;
+
   static const char *s_library[] = {"ws2_32.dll"};
 
   assert(WSACleanup_winsock == nullptr);

```

src/lib/platform/EiKeyState.cpp
```diff
@@ -185,15 +185,13 @@ void EiKeyState::getKeyMap(deskflow::KeyMap &keyMap)
           LOG_WARN("multiple keysyms per keycode are not supported, keycode %d", keycode);
 
         deskflow::KeyMap::KeyItem item{};
-        xkb_keysym_t keysym = syms[0];
```
nbolton, line 219: @whot What are your thoughts on dropping `keysym` here? I wonder if perhaps it could open us up to bugs where someone changes the index e.g. on [new] L188 and forgets to also update L194 (which could easily slip past a code review). I imagine that's why you had `keysym` in the first place (to prevent that kind of bug).

src/lib/server/ClientProxy1_4.cpp
```diff
@@ -25,25 +25,3 @@ ClientProxy1_4::ClientProxy1_4(const std::string &name, deskflow::IStream *strea
...
-
-void ClientProxy1_4::keyUp(KeyID key, KeyModifierMask mask, KeyButton button)
-{
-  ClientProxy1_3::keyUp(key, mask, button);
-}
-
-void ClientProxy1_4::keepAlive()
-{
-  ClientProxy1_3::keepAlive();
-}
```
nbolton, line 49: Yes, seems like this code does nothing. I wonder why this was done in the first place 🤔

nbolton (changes requested):

src/lib/platform/EiKeyState.cpp
sithlord48: i kept this because the later call wants a xkb_keysym_t not int32_t. i suspect it would convert fine but just to be safe i left it .

sithlord48:
@nbolton  try again .. Arch on windows was not a sub class of ArchMiscWindows (it is now)

src/lib/arch/Arch.cpp
sithlord48: Try this again should be fixed now

nbolton (dismissed):

src/lib/platform/EiKeyState.cpp
sithlord48: All the involved types are just redef int32_t

sithlord48: Re did this to keep keysym .

PR #8814, opened 2025-08-05
fix: Account for more modifiers in the EI keymap calculation

whot:
Some modifiers, notably LevelThree, were not accounted for in the modifier mask calculation. This leads to an incorrect keymap where some key symbols were listed for keys without (or insufficient) modifiers. Emulating those keys can then produce the wrong keysyms.

Reproducible with server and client on English(UK) layout and typing e.g. [. That key has its own key AD11 (right of P) but is also on the third level of AE08 (the 8 key). When the server sends the keyid 91 (for [) the client looks it up in the pre-generated keymap. It is found first on the 8 key with no modifiers (because LevelThree was previously ignored), hence the client simulates a key 8 press without modifiers. This erroneously produced an 8 instead of the wanted [.

Fix this by ensuring that all modifiers are accounted for in the key map. This fix is likely incomplete as it does not account for the full virtual modifier to real modifier mappings possible (e.g. Mod5 *may* be something other than AltGr) but it does push the can a bit further down the road, for someone else to release the worms.

Closes: #8168

---
Big disclaimer: this fixes the immediate bug but it *may* introduce new bugs especially on more exotic layouts where the modifiers are mapped in a more unusual way. Eventually this will need to have a much more complex modifier mapping LUT but for now... this will do?

And for the next bug at least we know where to look first ;)

PR #8825, opened 2025-08-06
Conform EI files to our naming convention

### LINKED ISSUE #8818: Use `camelCase` for variables in EI source
### Feature Request

We're using `snake_case` for many variables, where we should be using `camelCase`.

sithlord48:
fixes: #8818

update our variable and methods names to match our naming conventions

Some `snake_case` remains but these are types or methods outside of our code.

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -680,7 +680,7 @@ void EiScreen::onMotionEvent(ei_event *event)
   if (m_isOnScreen) {
     LOG_DEBUG("event: motion on primary x=%i y=%i)", m_cursorX, m_cursorY);
     sendEvent(EventTypes::PrimaryScreenMotionOnPrimary, MotionInfo::alloc(m_cursorX, m_cursorY));
-    if (m_portalInputCapture->is_active()) {
+    if (m_portalInputCapture->isActive()) {
       m_portalInputCapture->release();
     }
   } else {
@@ -776,7 +776,7 @@ void EiScreen::handleSystemEvent(const Event &sysevent)
       if (m_isPrimary) {
         LOG_DEBUG("re-allocating portal input capture connection and releasing active captures");
         if (m_portalInputCapture) {
-          if (m_portalInputCapture->is_active()) {
+          if (m_portalInputCapture->isActive()) {
             m_portalInputCapture->release();
           }
           delete m_portalInputCapture;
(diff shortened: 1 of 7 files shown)


nbolton:
Maybe land after #8814? So we don't hold it up longer than needed.

sithlord48:
that depend on when #8814 gets landed.. I can update the base after this .

src/lib/platform/EiScreen.cpp
```diff
@@ -606,12 +606,12 @@ void EiScreen::onPointerScrollEvent(ei_event *event)
 {
   // Ratio of 10 pixels == one wheel click because that's what mutter/gtk
   // use (for historical reasons).
-  const int PIXELS_PER_WHEEL_CLICK = 10;
+  static const int PixelsPerWheelClick = 10;
```
nbolton: Minor nit.

I agree that screaming snake is wrong here; it looks like an #ifdef (maybe it was originally).

However, one of these suggestions would be more inline with our existing convention. When I see `PascalCase` I think "Oh, it must be an enumerator/enum value"

```suggestion
  static const int kPixelsPerWheelClick = 10;
```

```suggestion
  static const int s_kPixelsPerWheelClick = 10;
```

```suggestion
  static const int s_pixelsPerWheelClick = 10;
```

nbolton (dismissed):

src/lib/platform/EiScreen.cpp
sithlord48: i just really find the `s_` prefix pointless and often forget ... its static its defined right there and its used in this method ..

sithlord48: updated to `s_pixel...`

src/lib/platform/EiScreen.cpp
```diff
@@ -606,12 +606,12 @@ void EiScreen::onPointerScrollEvent(ei_event *event)
...
   // Ratio of 10 pixels == one wheel click because that's what mutter/gtk
   // use (for historical reasons).
-  const int PIXELS_PER_WHEEL_CLICK = 10;
+  static const int s_pixelsPerWheelClick = 10;
   // Our logical wheel clicks are multiples 120, so we
   // convert between the two and keep the remainders because
   // we will very likely get subpixel scroll events.
-  // This means a single pixel is 120/PIXEL_TO_WHEEL_RATIO in wheel values.
-  const int PIXEL_TO_WHEEL_RATIO = 120 / PIXELS_PER_WHEEL_CLICK;
+  // This means a single pixel is 120/s_PixelToWheelRation in wheel values.
```
nbolton: Very minor nit if you want to fix:
```suggestion
  // This means a single pixel is 120/s_pixelToWheelRation in wheel values.
```

nbolton (dismissed):

nbolton (approved):

PR #8836, opened 2025-08-07
Exception rename

sithlord48:
fixes: #8831 

Uses `Exception` as a post fix for all previous `X` Items


src/lib/arch/win32/ArchDaemonWindows.cpp
```diff
@@ -51,7 +51,7 @@ UINT ArchDaemonWindows::getDaemonQuitMessage()
 void ArchDaemonWindows::daemonFailed(int result)
 {
   assert(s_daemon != nullptr);
-  throw XArchDaemonRunFailed(result);
+  throw ArchDaemonExceptionRunFailed(result);
```
nbolton:
```suggestion
  throw ArchDaemonRunFailedException(result);
```

Or, if that's a bit tricky to read...
```suggestion
  throw ArchDaemonRunException(result);
```

Any more than 4 words in a name and it starts to get chewy.

src/lib/arch/win32/ArchDaemonWindows.cpp
```diff
@@ -64,7 +64,7 @@ void ArchDaemonWindows::installDaemon(
   SC_HANDLE mgr = OpenSCManager(nullptr, nullptr, GENERIC_WRITE);
   if (mgr == nullptr) {
     // can't open service manager
-    throw XArchDaemonInstallFailed(windowsErrorToString(GetLastError()));
+    throw ArchDaemonInstallFailedException(windowsErrorToString(GetLastError()));
```
nbolton: I do wonder if "Failed" is redundant in these names.

```suggestion
    throw ArchDaemonInstallException(windowsErrorToString(GetLastError()));
```

src/lib/deskflow/DeskflowException.h
```diff
@@ -1,39 +1,40 @@
...
 
-#include "base/XBase.h"
+#include "base/BaseException.h"
 
 /**
- * @brief The XDeskflow class Generic deskflow exception class
+ * @brief The DeskflowException class Generic deskflow exception class
  */
-class XDeskflow : public XBase
+class DeskflowException : public BaseException
```
nbolton, line 16: Not now, but in the future we need a better name for the `deskflow` lib and things inside it such as `XDeskflow`/`DeskflowException`

nbolton (dismissed):

src/lib/deskflow/DeskflowException.h
sithlord48: yes cisp.. :D

PR #8837, opened 2025-08-07
fix: MainWindow::setHostName potental error

sithlord48:
Adjust `MainWindow::setHostName`
 - Return early if the new name is the same as the old name 
 - Only check the server config for an existing screen if we are in server mode.
 - If in server mode make sure to call serverConfig()::upateServerName() when the name changes, Failing todo this can break the server layout as the screen config is not updated.


nbolton:
Do you have test/repro steps for this or a related issue? I'm trying to figure out what it fixes.

sithlord48:
I ran in to this when I: 

1. Start server
2. Rename server screen using main window 
3. Look at screen layout
4. Note the old name is still in use
5. Rename to old name Note screen name exists error

nbolton (approved):

PR #8847, opened 2025-08-11
GUI: Move the shown ip location, only show when in server mode

sithlord48:
GUI
 - Use `Link Color` for coloring the ip to use 
 - Move the gui display to the top right 
 - show ONLY the suggested IP on the gui the rest are in a tooltip. no tooltip for a single ip
 - Only show the IP when in server mode. 

Screenshot_20250811_172429

Screenshot_20250811_172518

I don't have a screen shot for the tooltip buts its 

```
If connecting via the hostname fails, Try the following Ips:
 SuggestedIP (in color) 
 other ip 
 other ip
...
```

### DIFF
--- a/src/lib/gui/MainWindow.cpp
+++ b/src/lib/gui/MainWindow.cpp
@@ -1008,10 +1008,10 @@ QString MainWindow::getIPAddresses() const
         !address.isInSubnet(QHostAddress::parseSubnet("169.254.0.0/16"))) {
 
       // usually 192.168.x.x is a useful ip for the user, so indicate
-      // this by making it bold.
+      // this by coloring it in the "link" color.
       if (!hinted && address.isInSubnet(localnet)) {
         QString format = R"(<span style="color:%1;">%2</span>)";
-        result.append(format.arg(kColorTertiary, address.toString()));
+        result.append(format.arg(palette().link().color().name(), address.toString()));
         hinted = true;
       } else {
         result.append(address.toString());
--- a/src/lib/gui/MainWindow.cpp
+++ b/src/lib/gui/MainWindow.cpp
@@ -220,7 +220,8 @@ void MainWindow::setupControls()
   secureSocket(false);
 
   ui->btnConfigureServer->setIcon(QIcon::fromTheme(QStringLiteral("configure")));
-  ui->lblIpAddresses->setText(tr("This computer's IP addresses: %1").arg(getIPAddresses()));
+
+  updateNetworkInfo();
 
   if (Settings::value(Settings::Core::LastVersion).toString() != kVersion) {
     Settings::setValue(Settings::Core::LastVersion, kVersion);
@@ -567,6 +568,7 @@ void MainWindow::coreModeToggled()
 
 void MainWindow::updateModeControls(bool serverMode)
 {
+  ui->lblIpAddresses->setVisible(serverMode);
   ui->serverOptions->setVisible(serverMode);
   ui->clientOptions->setVisible(!serverMode);
   ui->lblNoMode->setVisible(false);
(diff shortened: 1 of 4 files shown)


src/lib/gui/MainWindow.cpp
```diff
@@ -606,6 +608,48 @@ void MainWindow::updateSecurityIcon(bool visible)
...
+        ipList.append(colorText.arg(palette().link().color().name(), suggestedAddress));
+        hinted = true;
+      } else {
+        ipList.append(address.toString());
+      }
+    }
+  }
+
+  if (suggestedAddress.isEmpty()) {
+    ui->lblIpAddresses->setText(colorText.arg(palette().linkVisited().color().name(), tr("No Ip Detected")));
```
nbolton:
```suggestion
    ui->lblIpAddresses->setText(colorText.arg(palette().linkVisited().color().name(), tr("No IP detected")));
```

src/lib/gui/MainWindow.cpp
```diff
@@ -606,6 +608,48 @@ void MainWindow::updateSecurityIcon(bool visible)
...
+    ui->lblIpAddresses->setToolTip(tr("Unable to detect an IP address. Check that your network connection is active"));
```
nbolton:
```suggestion
    ui->lblIpAddresses->setToolTip(tr("Unable to detect an IP address. Check that your network connection is active."));
```

src/lib/gui/MainWindow.cpp
```diff
@@ -606,6 +608,48 @@ void MainWindow::updateSecurityIcon(bool visible)
...
+    return;
+  }
+
+  ui->lblIpAddresses->setText(QStringLiteral("Suggested Ip: %1").arg(suggestedAddress));
```
nbolton:
```suggestion
  ui->lblIpAddresses->setText(QStringLiteral("Suggested IP: %1").arg(suggestedAddress));
```

src/lib/gui/MainWindow.cpp
```diff
@@ -606,6 +608,48 @@ void MainWindow::updateSecurityIcon(bool visible)
...
+
+  if (ipList.count() < 2) {
+    ui->lblIpAddresses->setToolTip("");
+  } else {
+    ui->lblIpAddresses->setToolTip(
+        tr("<p>If connecting via the hostname fails, Try the following Ips:<br/>%1</p>").arg(ipList.join("<br/>"))
```
nbolton:
```suggestion
        tr("<p>If connecting via the hostname fails, try the following IPs:<br/>%1</p>").arg(ipList.join("<br/>"))
```

src/lib/gui/MainWindow.ui
```diff
@@ -131,31 +131,28 @@
...
+       <item>
+        <widget class="QLabel" name="lblIpAddresses">
+         <property name="sizePolicy">
+          <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
+           <horstretch>0</horstretch>
+           <verstretch>0</verstretch>
+          </sizepolicy>
+         </property>
+         <property name="text">
+          <string notr="true"> IP addresses:</string>
```
nbolton: Is the preceding space intended?
```suggestion
          <string notr="true">IP addresses:</string>
```

src/lib/gui/MainWindow.cpp
```diff
@@ -606,6 +608,48 @@ void MainWindow::updateSecurityIcon(bool visible)
...
+
+  const auto addresses = QNetworkInterface::allAddresses();
+  for (const auto &address : addresses) {
+    if (address.protocol() == QAbstractSocket::IPv4Protocol && address != QHostAddress(QHostAddress::LocalHost) &&
+        !address.isInSubnet(QHostAddress::parseSubnet("169.254.0.0/16"))) {
+      // usually 192.168.x.x is a useful ip for the user, so indicate
+      // this by coloring it in the "link" color
+      if (!hinted && address.isInSubnet(QHostAddress::parseSubnet("192.168/16"))) {
+        suggestedAddress = address.toString();
+        ipList.append(colorText.arg(palette().link().color().name(), suggestedAddress));
```
nbolton, line 628: Just a thought: When helping Synergy customers, we often found that highlighting one of the IPs (although seemingly helpful) can be misleading because they think that it's the only IP address they can use. Some even go so far as to assume that it is the IP address that the server is bound to (which, of course, is not the case). It causes them to go fiddle with Preferences -> Advanced -> Network IP and wonder why it doesn't change which is highlighted.

Also, curiously, on macOS, I get the value:

`u"<span style=\"color:#3586ff;\">192.168.0.135</span>"`

But it doesn't appear to be highlighted:

Image

This also seem to be the case for Windows, KDE, and GNOME; the IP is never highlighted. Does it work for you? What am I missing?

sithlord48: The highlight ONLY is shown in the tooltip . Since we only show one suggested Ip on the window there is no reason to highlight it

src/lib/gui/MainWindow.ui
sithlord48: This string does not matter as its replaced by the function I am going to remove the string from the ui

src/lib/gui/MainWindow.cpp
nbolton: Ah, of course...
Screenshot 2025-08-12 at 15 47 25

nbolton (approved):

PR #8849, opened 2025-08-12
feat: remove toml config file only use our QSettings based format

sithlord48:
- Remove support for toml based config files.


nbolton:
I believe this is blocked by: 
- #8852

I will prioritize working on this so we can delete the TOML parser.

Issue #8839, opened 2025-08-08
Unable to type special characters with non-English keyboard


nbolton:
> [!NOTE]
> We originally thought this was a Wayland client bug, but it also happens on X11 clients.
> However, it may actually be two separate bugs (for Wayland and X11).

### Project

Synergy

### Deskflow version info

Synergy v1.18.1 (with EI map calc fix #8814)

### Build type

Other (please specify)

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [ ] Linux (X11)
- [x] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### OS versions/distros

Ubuntu 24.04.1

### Deskflow configuration

Wayland client

### What steps will reproduce the problem?

1. Use any server
2. Mouse to client
3. Type special chars (e.g. below)

From `swede0` on Discord:
> I think you could reproduce this on a UK one, probably.. Alt+2 should produce  @, alt+(key just right of z) should produce |, these are two failures that I suspect would be identical on your keyboard. The åäö keys that are the main weirdness of swedish keyboards alters things mainly on the right side of the main keyboard, not so much over there on the left.

Expected: Characters typed correctly
Actual: Wrong chars typed

### Log output

```shell
[2025-08-08T16:21:49] DEBUG1: recv key down id=0x0000efe3, mask=0x2002, button=0x001d, lang="en"
[2025-08-08T16:21:49] DEBUG1: mapKey efe3 (61411) with mask 2002, start state: 0000, group: 1
[2025-08-08T16:21:49] DEBUG1: language en has group id 0
[2025-08-08T16:21:49] DEBUG1: best key index 1 of 1 (exact)
[2025-08-08T16:21:49] DEBUG1: found key in group 0
[2025-08-08T16:21:49] DEBUG1: language en has group id 0
[2025-08-08T16:21:49] DEBUG1: state: 0000,0000,0000
[2025-08-08T16:21:49] DEBUG1: flip: 0000 (0000 vs 0000 in 0000 - 0000)
[2025-08-08T16:21:49] DEBUG1: desired state: 2002 0000,0000,0000
[2025-08-08T16:21:49] DEBUG1: flip: 0000 (0000 vs 2002 in fffd - 6020)
[2025-08-08T16:21:49] DEBUG1: mapped to 025, new state 0002
[2025-08-08T16:21:49] DEBUG1: keystrokes:
[...]

PR #8856, opened 2025-08-13
fix: resolve issues with some swedish characters

sithlord48:
Partly solves: #8839

### DIFF
--- a/src/lib/deskflow/KeyMap.cpp
+++ b/src/lib/deskflow/KeyMap.cpp
@@ -210,14 +210,13 @@ void KeyMap::finish()
 
 void KeyMap::foreachKey(ForeachKeyCallback cb, void *userData)
 {
-  for (const auto &[keyId, keyGroup] : m_keyIDMap) {
-    const KeyGroupTable &groupTable = keyGroup;
-    for (size_t group = 0; group < groupTable.size(); ++group) {
-      const KeyEntryList &entryList = groupTable.at(group);
+  for (auto &[keyId, keyGroup] : m_keyIDMap) {
+    for (size_t group = 0; group < keyGroup.size(); ++group) {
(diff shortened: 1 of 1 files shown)


nbolton:
@sithlord48 Nice. Do you have some good repro steps so I can see the bug happening on master? Do I need to switch to non-English layout?

sithlord48:
You will need to use a non US layout and try to send @ or | using the 3rd layer compositing.  
And i am unsure if the client or server is the one that messed up here.
@tnn2 might be able to provide more info on what you should be trying.

nbolton:
> And i am unsure if the client or server is the one that messed up here.

Maybe client? But I'll try both.

nbolton:
I don't think this fixes the original bug reported in #8839 (which I originally tagged as a Wayland issue). I'm happy for #8839 to be repurposed as the X11 bug that @tnn2 is seeing. The original issue was for a Synergy 1.18 user which was long before a33574e so I'll simply clone #8839 to a new issue which is Wayland specific as before.

With Wayland client, on both master and this PR, with a French layout on the client, the number row is quite messed up; it's all opposite.
- When pressing `Shift+0` I get: `0`
- When pressing `0` I get: `à`

Whether or not I choose 'Use server language' in preferences, it makes no difference to the behavior. I tried all combinations of server/client layout and it seems to only apply to the client.

image

If this fixes the issue for @tnn2 then I suggest we merge it.

@tnn2 Could you confirm this PR solves your issue?

sithlord48:
It should not do any extra harm its just not using a for each style loop now for two of the four loops involved.

nbolton:
You could land or wait for @tnn2 to confirm. Your call.

sithlord48:
He said in the issue it worked so we can land but i will not close the issue any more

Issue #8840, opened 2025-08-08
Alt and Meta are not faked on KDE+Wayland (but works GNOME+Wayland client)


nbolton:
### Project

Deskflow

### Deskflow version info

Deskflow: 1.23.0.73 (a5c97442)
Qt: 6.9.1
System: Arch Linux
Session: KDE (wayland)

### Build type

Local developer build (built it myself)

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [ ] Linux (X11)
- [x] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### OS versions/distros

Arch + KDE

### Deskflow configuration

KDE+Wayland client

### What steps will reproduce the problem?

1. Mouse to KDE+Wayland client
2. Press Alt+T

Expected: Alt+T faked on client
Actual: Only `T` is faked

### Log output

a5c97442 pressing Alt+T (bug):
```shell
[2025-08-08T13:58:15] DEBUG1: recv key down id=0x0000efe9, mask=0x0004, button=0x003b, lang="en"
[2025-08-08T13:58:15] DEBUG1: mapKey efe9 (61417) with mask 0004, start state: 0000, group: 0
[2025-08-08T13:58:15] DEBUG1: language en has group id 0
[2025-08-08T13:58:15] DEBUG1: best key index 1 of 2 (exact)
[2025-08-08T13:58:15] DEBUG1: found key in group 0
[2025-08-08T13:58:15] DEBUG1: state: 0000,0000,0000
[2025-08-08T13:58:15] DEBUG1: flip: 0000 (0000 vs 0000 in 0000 - 0000)
[2025-08-08T13:58:15] DEBUG1: desired state: 0004 0000,0000,0000
[2025-08-08T13:58:15] DEBUG1: flip: 0000 (0000 vs 0004 in fff3 - 6020)
[2025-08-08T13:58:15] DEBUG1: mapped to 038, new state 000c
[2025-08-08T13:58:15] DEBUG1: fake key: 038 (00000000) down
[2025-08-08T13:58:15] DEBUG1: update key state: keyval=64 pressed=1
[2025-08-08T13:58:15] DEBUG1: ei: sending: object 0xff0000000000000c (ei_keyboard@v1:key(1)) signature 'uu'
[2025-08-08T13:58:15] DEBUG1: ei: sending: object 0xff0000000000000b (ei_device@v2:frame(3)) signature 'ut'
[...]

PR #8860, opened 2025-08-14
feat: Use `xkb_keymap_mod_get_mask` in EiKeyState if available

nbolton:
Uses `xkb_keymap_mod_get_mask` introduced in xkbcommon v1.10.0

TODO: Figure out the CMake config and detect xkbcommon v1.10.0 instead of checking for the function.

### DIFF
--- a/cmake/Libraries.cmake
+++ b/cmake/Libraries.cmake
@@ -162,6 +162,16 @@ macro(configure_unix_libs)
       find_library(LIBM m)
       include_directories(${LIBXKBCOMMON_INCLUDE_DIRS} ${GLIB2_INCLUDE_DIRS}
                           ${LIBM_INCLUDE_DIRS})
+      
+      message(STATUS "xkbcommon version: ${LIBXKBCOMMON_VERSION}")
+
+      # Available since xkbcommon v1.10
+      include(CheckSymbolExists)
+      check_symbol_exists(xkb_keymap_mod_get_mask "xkbcommon/xkbcommon.h" HAVE_XKB_KEYMAP_MOD_GET_MASK)
+      message(VERBOSE "xkb_keymap_mod_get_mask: ${HAVE_XKB_KEYMAP_MOD_GET_MASK}")
+      if (HAVE_XKB_KEYMAP_MOD_GET_MASK)
+        add_definitions(-DHAVE_XKB_KEYMAP_MOD_GET_MASK=1)
+      endif()
     else()
       message(WARNING "pkg-config not found, skipping wayland libraries")
     endif()
@@ -267,6 +277,8 @@ macro(configure_xorg_libs)
   if(HAVE_Xi)
     list(APPEND libs Xi)
   endif()
+
+
 
   add_definitions(-DWINAPI_XWINDOWS=1)
 
--- a/src/lib/platform/EiKeyState.cpp
+++ b/src/lib/platform/EiKeyState.cpp
@@ -117,8 +117,16 @@ std::uint32_t EiKeyState::convertModMask(std::uint32_t xkbMask) const
   std::uint32_t modMask = 0;
 
   for (xkb_mod_index_t xkbmod = 0; xkbmod < xkb_keymap_num_mods(m_xkbKeymap); xkbmod++) {
+#ifdef HAVE_XKB_KEYMAP_MOD_GET_MASK
+    // Available since xkbcommon v1.10
+    const auto modMask = xkb_keymap_mod_get_mask(m_xkbKeymap, xkbmod);
+    if ((xkbMask & modMask) == 0)
+      continue;
+#else
+    // HACK: <= xkbcommon v1.7 we need to check the mask manually.
     if ((xkbMask & (1 << xkbmod)) == 0)
       continue;
+#endif
 
     /* added in libxkbcommon 1.8.0 in the same commit so we have all or none */
 #ifndef XKB_VMOD_NAME_ALT


cmake/Libraries.cmake
```diff
@@ -268,6 +278,8 @@ macro(configure_xorg_libs)
     list(APPEND libs Xi)
   endif()
 
+
+
```
nbolton: IDK what happened there

cmake/Libraries.cmake
```diff
@@ -162,6 +162,16 @@ macro(configure_unix_libs)
       find_library(LIBM m)
       include_directories(${LIBXKBCOMMON_INCLUDE_DIRS} ${GLIB2_INCLUDE_DIRS}
                           ${LIBM_INCLUDE_DIRS})
+      
+      message(STATUS "xkbcommon version: ${LIBXKBCOMMON_VERSION}")
+
```
sithlord48: This really should be done in platform/CMakeLists.txt so we can use `target_compile_definitions`. This will make sure that we only set the define for the target that is going to use it .

in platform/CMakeLists.txt do this (line 183)
```
  if(NOT APPLE)
    target_link_libraries(platform Qt6::DBus)
    link_wayland_libs()
    if(${LIBXKBCOMMON_VERSION} VERSION_GREATER "1.7")
      target_compile_definitions(platform PRIVATE HAVE_XKB_KEYMAP_MOD_GET_MASK=1)
    endif()
  endif()
```

src/lib/platform/EiKeyState.cpp
```diff
@@ -117,8 +117,16 @@ std::uint32_t EiKeyState::convertModMask(std::uint32_t xkbMask) const
...
+    if ((xkbMask & modMask) == 0)
+      continue;
+#else
+    // HACK: <= xkbcommon v1.7 we need to check the mask manually.
     if ((xkbMask & (1 << xkbmod)) == 0)
       continue;
+#endif
 
     /* added in libxkbcommon 1.8.0 in the same commit so we have all or none */
 #ifndef XKB_VMOD_NAME_ALT
```
sithlord48: This would get removed right? (here to line 166/174)

src/lib/platform/EiKeyState.cpp
```diff
@@ -117,8 +117,16 @@ std::uint32_t EiKeyState::convertModMask(std::uint32_t xkbMask) const
   std::uint32_t modMask = 0;
 
   for (xkb_mod_index_t xkbmod = 0; xkbmod < xkb_keymap_num_mods(m_xkbKeymap); xkbmod++) {
+#ifdef HAVE_XKB_KEYMAP_MOD_GET_MASK
```
sithlord48, line 121: IIRC this method works for both X11 and EI screens so we should maybe do this in a class both can use

cmake/Libraries.cmake
sithlord48: i Can not seam to get that check_symbol_exists to work either but This does work works bit much but it does end with the the same results as the check_symbol_exsits macro for our needs at least. (I put this in platfom/CMakeLists.txt) there is a xkb_keymap_mod_get_mask2 in v1.11 of xkbcommon (so i added a check to make sure it was working right

```
file(READ "${LIBXKBCOMMON_INCLUDE_DIRS}/xkbcommon/xkbcommon.h" XKBCONTENTS)
string(REGEX MATCH "xkb_keymap_mod_get_mask\\(" HAVE_XKB_KEYMAP_MOD_GET_MASK ${XKBCONTENTS})
string(REGEX MATCH "xkb_keymap_mod_get_mask2\\(" HAVE_XKB_KEYMAP_MOD_GET_MASK2 ${XKBCONTENTS})

if (HAVE_XKB_KEYMAP_MOD_GET_MASK)
  target_compile_definitions(platform PRIVATE HAVE_XKB_KEYMAP_MOD_GET_MASK=1)
endif()

if (HAVE_XKB_KEYMAP_MOD_GET_MASK2)
  target_compile_definitions(platform PRIVATE HAVE_XKB_KEYMAP_MOD_GET_MASK2=1)
endif()
```

nbolton: @Wismill What do you think might be the most stable approach long-term for CMake config and xkbcommon? Do we check the xkbcommon version or regex the header (see suggestion above)?

src/lib/platform/EiKeyState.cpp
nbolton: ~~Not sure how it's related to this PR.~~

~~Ah, I see what you're saying. Yes, I hope so!~~

Edit: OK, now I _really_ see what you mean. That's the code we need to build our own mask based on these values, so it has to stay:
https://github.com/deskflow/deskflow/blob/90d3192524268f543a40fb6b56c91188adbcd5d5/src/lib/deskflow/KeyTypes.h#L71-L86

src/lib/platform/EiKeyState.cpp
nbolton: Do you know where the equivalent `xkb_keymap_num_mods` call is in the X11 code?

src/lib/platform/EiKeyState.cpp
```diff
@@ -116,9 +115,23 @@ std::uint32_t EiKeyState::convertModMask(std::uint32_t xkbMask) const
...
+    const char *name = xkb_keymap_mod_get_name(m_xkbKeymap, xkbMod);
+
+#ifdef HAVE_XKB_KEYMAP_MOD_GET_MASK
+    // Available since xkbcommon v1.10
+    const auto xkbModMask = xkb_keymap_mod_get_mask(m_xkbKeymap, name);
+    if ((xkbMask & xkbModMask) == 0)
+      continue;
+
+    // HACK: returning here makes alt and shift work, but not ctrl?
+    return xkbModMask;
```
nbolton: @wismill @whot Summoning the experts here 🙂 

I think the issue is that I fundamentally don't understand the relationship between `xkb_keymap_mod_get_mask` and the loop around `xkb_keymap_num_mods`.

I first tried `|=`-ing the result of `xkb_keymap_mod_get_mask` to `modMask` outside the loop, but that resulted in ~~no modifiers working.~~ shift and ctrl being swapped.

Then I thought: Can I just return the first result of `xkb_keymap_mod_get_mask`? Turns out this makes my Alt key work now (maybe that fixes #8840), but damn... the Ctrl key stopped working 🤔 


Edit: As was suggested, I'm taking a look at [how-to-type.c](https://github.com/xkbcommon/libxkbcommon/blob/master/tools/how-to-type.c) to see if I can grasp this.


~~Edit: Found something weird: https://github.com/deskflow/deskflow/pull/8860#discussion_r2293587162~~ (never mind)

src/lib/platform/EiKeyState.cpp
```diff
@@ -114,10 +113,24 @@ void EiKeyState::pollPressedKeys(KeyButtonSet &pressedKeys) const
...
+
+#ifdef HAVE_XKB_KEYMAP_MOD_GET_MASK
+    // Available since xkbcommon v1.10
+    // Note: xkb_keymap_mod_get_mask2 was added in v1.11 which accepts xkb_mod_index_t.
+    const auto _modMask = xkb_keymap_mod_get_mask(m_xkbKeymap, name);
+    if ((xkbMask & _modMask) == 0)
+      continue;
+
+    // TODO: this only works when we run the core directly but not through the gui... why?
+    modMask |= _modMask;
```
nbolton: Hmm, breakthrough... this actually does work when I run the core directly. But why would running through the GUI cause this not to work? Memory bug?

nbolton: Never mind, got tripped up by:
- #8868

src/lib/platform/EiKeyState.cpp
nbolton: Maybe this monster func?
https://github.com/deskflow/deskflow/blob/90d3192524268f543a40fb6b56c91188adbcd5d5/src/lib/platform/XWindowsKeyState.cpp#L329

Edit: Certainly in another PR.

src/lib/platform/EiKeyState.cpp
nbolton: Hmm, if I interpret it correctly, I think we can only use this to replace `if ((xkbMask & (1 << xkbMod)) == 0)` and the rest of the function should remain the same.

src/lib/platform/EiKeyState.cpp
nbolton, line 1: Hmm, I wonder if `xkb_keymap_mod_get_mask` is meant to be used in relation to this code?
https://github.com/deskflow/deskflow/blob/123ba1226dd832202bf43b6b917a91df8785862f/src/lib/platform/EiKeyState.cpp#L184-L198

src/lib/platform/CMakeLists.txt
```diff
@@ -153,6 +153,10 @@ include_directories(${inc})
 # wayland.h is included to check for wayland support
 add_library(platform STATIC ${PLATFORM_SOURCES} Wayland.h)
 
+if(${LIBXKBCOMMON_VERSION} VERSION_GREATER_EQUAL "1.10")
```
sithlord48: guard this with 

if (UNIX AND NOT APPLE)
... 
endif()

or move it down to line 187 (thats what i did locally )

nbolton:
Edit: Decided to raise an issue:
- #8883 

---

Couldn't quite get `xkb_state_mod_name_is_active` working.

Only got this far:
```cmake
    if(${LIBXKBCOMMON_VERSION} VERSION_GREATER_EQUAL "1.8")
      target_compile_definitions(platform PRIVATE HAVE_XKB_STATE_MOD_NAME_IS_ACTIVE=1)
    endif()
```
```cpp
#ifdef HAVE_XKB_STATE_MOD_NAME_IS_ACTIVE
    auto state = xkb_state_new(m_xkbKeymap);
    xkb_state_update_key(state, xkbModMask, XKB_KEY_DOWN);
    const auto isActive = xkb_state_mod_name_is_active(state, name, XKB_STATE_MODS_DEPRESSED);
    xkb_state_unref(state);

    if (isActive == 0)
      continue;
    else if (isActive == -1)
      LOG_WARN("modifier mask %s not found in keymap, this is a bug", name);
#else
    // If this modifier is not active, skip it.
    if ((xkbMaskIn & xkbModMask) == 0)
      continue;
#endif
```

sithlord48 (approved):
This seams to make All my modifiers work correctly on kde wayland client (with xkbcommon 1.11)

There is still the issue of numlock or caps lock is not in sync with the client in all cases but this is also an issue elsewhere tool.

nbolton:
> All my modifiers work correctly on kde wayland client

This PR did not fix my modifiers for some reason... I had to open #8881 for that.

src/lib/platform/EiKeyState.cpp
```diff
@@ -112,12 +110,25 @@ void EiKeyState::pollPressedKeys(KeyButtonSet &pressedKeys) const
...
+    const auto xkbModMask = xkb_keymap_mod_get_mask(m_xkbKeymap, name);
+#else
+    // HACK: in older xkbcommon we need to create the mask manually from the index.
+    const xkb_mod_mask_t xkbModMask = (1 << xkbModIdx);
+#endif
 
-  for (xkb_mod_index_t xkbmod = 0; xkbmod < xkb_keymap_num_mods(m_xkbKeymap); xkbmod++) {
-    if ((xkbMask & (1 << xkbmod)) == 0)
+    // Skip inactive modifiers.
+    if ((xkbModMaskIn & xkbModMask) == 0)
```
wismill, line 131: Late to the party, but I think you should do:
```diff
- if ((xkbModMaskIn & xkbModMask) == 0)
+ if ((xkbModMaskIn & xkbModMask) != xkbModMask)
```

because the modifier encoding is opaque and may set multiple bits.

sithlord48: made a pr for this anything else lmk please https://github.com/deskflow/deskflow/pull/8892

PR #8865, opened 2025-08-18
feat: Support Remote Desktop token persistance

### LINKED ISSUE #8869: Persist Remote Desktop XDG Desktop Portal dialog (Wayland)
### Feature Request

/bounty $2500

Counterpart Input Capture issue:
- https://github.com/deskflow/deskflow/issues/8032

Make Remote Desktop/Control Portal permission dialogs remember your trust preference.

GNOME - Remote desktop (client mode):
image

KDE - Remote control requested:
KDE Remote control requested

### Workaround

> [!WARNING]
> Only use this workaround if you know what you are doing.

The tool https://github.com/deskflow/accept-portal-dialog can be used to automatically accept Portal permission dialogs.

sithlord48:
Fixes #8869 for Remote Desktop (i.e client) 

IMHO There is no reason to ever allow input capture always (i.e server).

Besides, When / If The Input Capture portal gains support for a restore token it should be trivial to use this PR as an example enabling it for that portal. 

Edit by @nbolton: Thanks to @whot for work upstream which made this possible.

### DIFF
--- a/src/lib/common/Settings.h
+++ b/src/lib/common/Settings.h
@@ -37,6 +37,7 @@ public:
     inline static const auto InvertScrollDirection = QStringLiteral("client/invertScrollDirection");
     inline static const auto LanguageSync = QStringLiteral("client/languageSync");
     inline static const auto RemoteHost = QStringLiteral("client/remoteHost");
+    inline static const auto XdpRestoreToken = QStringLiteral("client/xdpRestoreToken");
   };
   struct Core
   {
--- a/src/lib/platform/PortalRemoteDesktop.cpp
+++ b/src/lib/platform/PortalRemoteDesktop.cpp
@@ -1,5 +1,6 @@
 /*
  * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2025 Deskflow Developers
  * SPDX-FileCopyrightText: (C) 2024 Symless Ltd.
  * SPDX-FileCopyrightText: (C) 2022 Red Hat, Inc.
  * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
@@ -8,6 +9,7 @@
 #include "platform/PortalRemoteDesktop.h"
 #include "base/Log.h"
 #include "base/TMethodJob.h"
+#include "common/Settings.h"
 
 namespace deskflow {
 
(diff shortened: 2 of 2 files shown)


nbolton:
I'm looking forward to using this feature!

I get an error on Arch+KDE: `[2025-08-19T09:35:03] ERROR: failed to initialize remote desktop session: GDBus.Error:org.freedesktop.portal.Error.InvalidArgument: Restore token is not a valid UUID string`

Edit: Seems to work on GNOME though, amazing! I wonder... how do I un-remember the decision?

Screenshot From 2025-08-19 09-53-04

sithlord48:
Um to remove remember involves using the desktop to clear it. The easiest way is to (for now) is to remove the xdgRestoreToken entry from the client section of  the settings file.

Ill maybe add a setting to clear it in our gui. But this is something your desktop should provide a settings area for like the kde flatpak permission kcm. 

Odd as its fine on my kde machine the restore key (in settings) should be a uuid .. did you use the QtPortals branch on that machine that branch write a non UUID token for the same value (if so just clear it out)

nbolton (approved):

nbolton:
> did you use the QtPortals branch on that machine that branch write a non UUID token for the same value (if so just clear it out)

Turns out I had a borked `libportal.so`, reinstalling fixed it: `sudo pacman -S xdg-desktop-portal`

> Um to remove remember involves using the desktop to clear it. The easiest way is to (for now) is to remove the xdgRestoreToken entry from the client section of the settings file.

Yes, this clears it. I edited `/home/nick/.config/Deskflow/Deskflow.conf` and removed the `xdpRestoreToken=` line.

> Ill maybe add a setting to clear it in our gui. 

I think manually editing the .conf is good enough, but maybe we should document that at least?

sithlord48:
> I think manually editing the .conf is good enough, but maybe we should document that at least? 

Added to https://github.com/deskflow/deskflow/wiki/Workarounds#Remove-Xdg-Remote-Desktop-permission

whot:
> IMHO There is no reason to ever allow input capture always (i.e server).

depends on your setting but either way [xdg-desktop-portal#1768](https://github.com/flatpak/xdg-desktop-portal/pull/1768) would address that.

That's an orthogonal problem to the token persistence though (token persistence is also used for "allow reconnect" after a previous ack, for some value of "after". As opposed to "allow unquestioned connect", two subtly different problems).

Issue #8258, opened 2025-02-27
Prevent multiple core instances


nbolton:
### Feature Request

# Problem
Due to some bugs, the core (server/client) can start multiple times, e.g.:
- #5408

This may also happen if the GUI crashes or is killed forcefully.

# Proposed solution
Just like how we use Qt prevent the GUI from opening multiple times, we should use QLocalServer/QLocalSocket to prevent the core from running multiple times. 

Perhaps we need to implement IPC or make this part of the IPC feature:
- #8257

sithlord48:
Yeah ipc for all there is no way to communicate the two

PR #8867, opened 2025-08-18
feat: Prevent core, client and or server from running at the same time

sithlord48:
fixes: #8258

 - Prevents `deskflow-core` from running two instances at the same time
       - `deskflow-core --help` will always return the help message
       - `deskflow-core --version` will always return the version info
       -  Duplicate runs will print `An instance of deskflow core is already running` and exit
 - Intercept `--help` and `-h` in `deskflow-core` to call its `showHelp`
 - Intercept `--version` and `-v` in `deskflow-core` to call  its `showVersion`
 - Move the exit codes to a new `common/ExitCodes.h`
 - Use new exit codes for all exits.
 - Add new exit code `s_exitDuplicate` to return when exiting because of an instance is already running.

### DIFF
--- a/src/apps/deskflow-core/deskflow-core.cpp
+++ b/src/apps/deskflow-core/deskflow-core.cpp
@@ -6,6 +6,7 @@
  * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
  */
 
+#include "VersionInfo.h"
 #include "arch/Arch.h"
 #include "base/EventQueue.h"
 #include "base/Log.h"
@@ -20,6 +21,8 @@
 
 #include <QSharedMemory>
 #include <iostream>
+
+const static auto kHeader = QStringLiteral("%1-core: %2\n").arg(kAppId, kDisplayVersion);
 
 void showHelp()
 {
@@ -38,6 +41,20 @@ bool isHelp(int argc, char **argv)
 {
   for (int i = 0; i < argc; ++i) {
     if (argv[i] == std::string("--help") || argv[i] == std::string("-h"))
+      return true;
+  }
+  return false;
+}
+
+void showVersion()
+{
+  std::cout << qPrintable(kHeader) << qPrintable(kCopyright) << std::endl;
+}
+
+bool isVersion(int argc, char **argv)
+{
+  for (int i = 0; i < argc; ++i) {
+    if (argv[i] == std::string("--version") || argv[i] == std::string("-v"))
       return true;
   }
   return false;
@@ -62,6 +79,11 @@ int main(int argc, char **argv)
 
   if (isHelp(argc, argv)) {
     showHelp();
+    return s_exitSuccess;
+  }
+
+  if (isVersion(argc, argv)) {
+    showVersion();
     return s_exitSuccess;
   }
 
--- a/src/apps/deskflow-core/deskflow-core.cpp
+++ b/src/apps/deskflow-core/deskflow-core.cpp
@@ -17,6 +17,7 @@
 #include <QCoreApplication>
 #endif
 
+#include <QSharedMemory>
 #include <iostream>
 
 void showHelp()
@@ -32,6 +33,15 @@ void showHelp()
   cApp.help();
 }
 
+bool isHelp(int argc, char **argv)
+{
+  for (int i = 0; i < argc; ++i) {
+    if (argv[i] == std::string("--help") || argv[i] == std::string("-h"))
+      return true;
+  }
+  return false;
+}
+
 bool isServer(int argc, char **argv)
 {
   return (argc > 1 && argv[1] == std::string("server"));
@@ -44,6 +54,30 @@ bool isClient(int argc, char **argv)
 
 int main(int argc, char **argv)
 {
+  Arch arch;
+  arch.init();
+
+  Log log;
+
+  if (isHelp(argc, argv)) {
+    showHelp();
+    return 0;
+  }
+
+  // Create a shared memory segment with a unique key
+  // This is to prevent a new instance from running if one is already running
+  QSharedMemory sharedMemory("deskflow-core");
+
+  // Attempt to attach first and detach in order to clean up stale shm chunks
+  // This can happen if the previous instance was killed or crashed
+  if (sharedMemory.attach())
+    sharedMemory.detach();
+
+  // If we can create 1 byte of SHM we are the only instance
+  if (!sharedMemory.create(1)) {
+    LOG_WARN("an instance of deskflow core (server or client) is already running");
+    return 0;
+  }
 #if SYSAPI_WIN32
   // HACK to make sure settings gets the correct qApp path
   QCoreApplication m(argc, argv);
@@ -52,10 +86,6 @@ int main(int argc, char **argv)
   ArchMiscWindows::setInstanceWin32(GetModuleHandle(nullptr));
 #endif
 
-  Arch arch;
-  arch.init();
-
-  Log log;
   EventQueue events;
 
   if (isServer(argc, argv)) {
--- a/src/apps/deskflow-core/deskflow-core.cpp
+++ b/src/apps/deskflow-core/deskflow-core.cpp
@@ -9,6 +9,7 @@
 #include "arch/Arch.h"
 #include "base/EventQueue.h"
 #include "base/Log.h"
+#include "common/ExitCodes.h"
 #include "deskflow/ClientApp.h"
 #include "deskflow/ServerApp.h"
 
@@ -61,7 +62,7 @@ int main(int argc, char **argv)
 
   if (isHelp(argc, argv)) {
     showHelp();
-    return 0;
+    return s_exitSuccess;
   }
 
   // Create a shared memory segment with a unique key
@@ -76,7 +77,7 @@ int main(int argc, char **argv)
   // If we can create 1 byte of SHM we are the only instance
   if (!sharedMemory.create(1)) {
     LOG_WARN("an instance of deskflow core (server or client) is already running");
-    return 0;
+    return s_exitDuplicate;
   }
 #if SYSAPI_WIN32
   // HACK to make sure settings gets the correct qApp path
@@ -98,5 +99,5 @@ int main(int argc, char **argv)
     showHelp();
   }
 
-  return 0;
+  return s_exitSuccess;
 }
--- a/src/apps/deskflow-core/deskflow-core.cpp
+++ b/src/apps/deskflow-core/deskflow-core.cpp
@@ -26,6 +26,7 @@ const static auto kHeader = QStringLiteral("%1-core: %2\n").arg(kAppId, kDisplay
 
 void showHelp()
 {
+  std::cout << qPrintable(kHeader) << qPrintable(kAppDescription) << "\n\n";
   std::cout << "Usage: deskflow-core <server | client> [...options]" << std::endl;
   std::cout << "server - start as a server (deskflow-server)" << std::endl;
   std::cout << "client - start as a client (deskflow-client)" << std::endl;
--- a/src/apps/deskflow-daemon/deskflow-daemon.cpp
+++ b/src/apps/deskflow-daemon/deskflow-daemon.cpp
@@ -9,6 +9,7 @@
 #include "arch/Arch.h"
 #include "base/EventQueue.h"
 #include "base/Log.h"
+#include "common/ExitCodes.h"
 #include "common/Settings.h"
 #include "deskflow/DaemonApp.h"
 #include "deskflow/ipc/DaemonIpcServer.h"
--- a/src/apps/deskflow-gui/deskflow-gui.cpp
+++ b/src/apps/deskflow-gui/deskflow-gui.cpp
@@ -8,6 +8,7 @@
 
 #include "VersionInfo.h"
 #include "common/Constants.h"
+#include "common/ExitCodes.h"
 #include "common/UrlConstants.h"
 #include "gui/Diagnostic.h"
 #include "gui/DotEnv.h"
@@ -68,18 +69,18 @@ int main(int argc, char *argv[])
 
   if (!parser.errorText().isEmpty()) {
     qCritical().noquote() << parser.errorText() << "\nUse --help for more information.";
-    return 1;
+    return s_exitArgs;
   }
 
   if (parser.isSet(helpOption)) {
     QTextStream(stdout) << kHeader << QStringLiteral("  %1\n\n").arg(kAppDescription)
                         << parser.helpText().replace(QApplication::applicationFilePath(), kAppId);
-    return 0;
+    return s_exitSuccess;
   }
 
   if (parser.isSet(versionOption)) {
     QTextStream(stdout) << kHeader << kCopyright << Qt::endl;
-    return 0;
+    return s_exitSuccess;
   }
 
   // Create a shared memory segment with a unique key
@@ -102,7 +103,7 @@ int main(int argc, char *argv[])
       QMessageBox::information(nullptr, QObject::tr("Deskflow"), QObject::tr("Deskflow is already running"));
     }
     socket.disconnectFromServer();
-    return 0;
+    return s_exitDuplicate;
   }
 
 #if !defined(Q_OS_MAC)
--- a/src/lib/common/CMakeLists.txt
+++ b/src/lib/common/CMakeLists.txt
@@ -13,6 +13,7 @@ unset(SERVER_BINARY)
 
 add_library(common STATIC
   Common.h
+  ExitCodes.h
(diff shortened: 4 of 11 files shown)


nbolton (commented):
Wouldn't this be better as a function rather than copy/pasted? I understand there are probably two reasons why you did it this way: 1) the client/server bins are going anyway, and 2) bad duplication is better than bad abstraction.

That said, copy/pasting code like this makes my eyes itch. Would you be against a function, at least temporarily, until we only have the core bin? I'm open to your thoughts here as I'm a bit of a chronic over-abstractor.

sithlord48:
I explaned above 

>I would have like todo this in a helper method of some kind but our shared memory has to live thru the live of the app so moving it else where as a helper method to check will not work unless we first create the shared memory and then that leaves our helper as a single line

I had done it that way at first but becase the shared memory was made in the function when a second instance launched teh old shm was out of scope so they do not see one another. And to make a working helper you have to make in main the shm , do theconnect / reconect and that would leave the helper mehod a line or two

nbolton:
> I explaned above

Damn, totally missed that :facepalm:

src/apps/deskflow-client/deskflow-client.cpp
```diff
@@ -11,13 +11,30 @@
...
+  QSharedMemory sharedMemory("deskflow-core");
+
+  // Attempt to attach first and detach in order to clean up stale shm chunks
+  // This can happen if the previous instance was killed or crashed
+  if (sharedMemory.attach())
+    sharedMemory.detach();
+
+  // If we can create 1 byte of SHM we are the only instance
+  if (!sharedMemory.create(1)) {
+    QTextStream(stdout) << "An instance of deskflow core (server or client) is already running";
```
nbolton: Let's make the log output consistent with existing logs, which are all in lower case.

Also, if we move this code down, we can use the existing log function, and we could also make the log message more useful by using the process name from `argv[0]`. The same applies to the other 2 places this code is copied to. Ah, and this log line actually seems like an error to me, so we should output to stderr and return an error code.

I tried this and it works well:
```
[2025-08-19T14:01:30] DEBUG: core process state changed: starting -> started
	/home/nick/Projects/deskflow/src/lib/gui/core/CoreProcess.cpp:598
[2025-08-19T14:01:30] ERROR: the core process is already running: /home/nick/Projects/deskflow/build/bin/deskflow-server
	/home/nick/Projects/deskflow/src/apps/deskflow-server/deskflow-server.cpp:52
```

```patch
diff --git a/src/apps/deskflow-server/deskflow-server.cpp b/src/apps/deskflow-server/deskflow-server.cpp
index 3c7105c71..c634314ff 100644
--- a/src/apps/deskflow-server/deskflow-server.cpp
+++ b/src/apps/deskflow-server/deskflow-server.cpp
@@ -21,21 +21,6 @@
 
 int main(int argc, char **argv)
 {
-  // Create a shared memory segment with a unique key
-  // This is to prevent a new instance from running if one is already running
-  QSharedMemory sharedMemory("deskflow-core");
-
-  // Attempt to attach first and detach in order to clean up stale shm chunks
-  // This can happen if the previous instance was killed or crashed
-  if (sharedMemory.attach())
-    sharedMemory.detach();
-
-  // If we can create 1 byte of SHM we are the only instance
-  if (!sharedMemory.create(1)) {
-    QTextStream(stdout) << "An instance of deskflow core (server or client) is already running" << Qt::endl;
-    return 0;
-  }
-
 #if SYSAPI_WIN32
   // HACK to make sure settings gets the correct qApp path
   QCoreApplication m(argc, argv);
@@ -53,6 +38,21 @@ int main(int argc, char **argv)
   Log log;
   EventQueue events;
 
+  // Create a shared memory segment with a unique key
+  // This is to prevent a new instance from running if one is already running
+  QSharedMemory sharedMemory("deskflow-core");
+
+  // Attempt to attach first and detach in order to clean up stale shm chunks
+  // This can happen if the previous instance was killed or crashed
+  if (sharedMemory.attach())
+    sharedMemory.detach();
+
+  // If we can create 1 byte of SHM we are the only instance
+  if (!sharedMemory.create(1)) {
+    LOG_ERR("the core process is already running: %s", argv[0]);
+    return s_exitFailed;
+  }
+
   ServerApp app(&events);
   return app.run(argc, argv);
 }
```

Or we could just use `kServerBinName`:
```cpp
LOG_ERR("the core process is already running: %s", kServerBinName);
```

nbolton:
I'm unable to build if there is an instance of `deskflow-client` running:
```
help2man: can't get `--help' info from deskflow-client
```

@sithlord48 said:
> I can't repo this do u somehow have stale cmake files ? The whole client / server folder should not be added when building unified and the calls for each bin to help2man is in the cmakelist within those subfolders.

src/apps/deskflow-client/deskflow-client.cpp
sithlord48: Does not there is not a log object when the message is called.

sithlord48: I don't want the EXE to do anything if its running this is why i check first. before anything else

sithlord48: >  Or we could just use kServerBinName:

good idea... however its misleading since you can't run deskflow-core and deskflow-client if the message for client always says its name users will not know to check for a running deskflow-core or deskflow-server

sithlord48: > Ah, and this log line actually seems like an error to me, so we should output to stderr and return an error code.

Its not an error the app exits without issue or fault. that is why i don't stderr or return non 0 ..

sithlord48:
Remember we expect to see this ONLY if the user is running from the CLI. We do not expect this to be an issue when using the gui (unless a cli instance is running then you see it in the log of the app too)

src/apps/deskflow-client/deskflow-client.cpp
```diff
@@ -9,15 +9,33 @@
...
+  QSharedMemory sharedMemory("deskflow-core");
+
+  // Attempt to attach first and detach in order to clean up stale shm chunks
+  // This can happen if the previous instance was killed or crashed
+  if (sharedMemory.attach())
+    sharedMemory.detach();
+
+  // If we can create 1 byte of SHM we are the only instance
+  if (!sharedMemory.create(1)) {
+    LOG_WARN("An instance of deskflow core (server or client) is already running");
```
nbolton: Small nit, log case consistency.

Edit: Also in 2 other places.

```suggestion
    LOG_WARN("an instance of deskflow core (server or client) is already running");
```

src/apps/deskflow-client/deskflow-client.cpp
```diff
@@ -9,15 +9,33 @@
...
+#include <QSharedMemory>
+
 #if SYSAPI_WIN32
 #include "arch/win32/ArchMiscWindows.h"
 #include <QCoreApplication>
 #endif
 
 int main(int argc, char **argv)
 {
+  Log log;
```
nbolton: I seem to remember creating `Log log` instance before `Arch arch` to be an issue on Windows due to the outputters used... hopefully I am misremembering or maybe no longer an issue. I'll do some testing.

sithlord48: This is why part of why i was using QTextStream since arch needs Arch::init() and this makes a bunch of stuff.

nbolton:
> I'm unable to build if there is an instance of `deskflow-client` running:
> 
> ```
> help2man: can't get `--help' info from deskflow-client
> ```
> 
> @sithlord48 said:
> 
> > I can't repo this do u somehow have stale cmake files ? The whole client / server folder should not be added when building unified and the calls for each bin to help2man is in the cmakelist within those subfolders.

Happens because of non-zero return code when it's already running.

```
ps ax | grep deskflow
 599876 ?        Sl     0:06 /home/nick/Projects/deskflow/build/bin/deskflow-server [...]
```

```
[2/4] Linking CXX executable bin/deskflow-client
FAILED: bin/deskflow-client 
: && /usr/bin/c++ -fPIC -g -fprofile-arcs -ftest-coverage -fprofile-update=atomic -g  src/apps/deskflow-client/CMakeFiles/deskflow-client.dir/deskflow-client_autogen/mocs_compilation.cpp.o src/apps/deskflow-client/CMakeFiles/deskflow-client.dir/deskflow-client.cpp.o -o bin/deskflow-client -L/usr/local/lib -Wl,-rpath,/usr/local/lib:  -lgcov  src/lib/arch/libarch.a  src/lib/base/libbase.a  src/lib/client/libclient.a  src/lib/io/libio.a  src/lib/mt/libmt.a  src/lib/net/libnet.a  src/lib/platform/libplatform.a  src/lib/server/libserver.a  src/lib/deskflow/libapp.a  -lpthread  -lXtst  -lX11  -lxkbfile  -lXext  -lXinerama  -lXrandr  -lXi  src/lib/client/libclient.a  src/lib/platform/libplatform.a  src/lib/server/libserver.a  src/lib/deskflow/libapp.a  /usr/lib64/libQt6DBus.so.6.8.2  /usr/lib64/libxkbcommon.so  /usr/lib64/libglib-2.0.so  /usr/lib64/libei.so  /usr/lib64/libportal.so  /usr/lib64/libgio-2.0.so  /usr/lib64/libgobject-2.0.so  /usr/lib64/libglib-2.0.so  /usr/lib64/libei.so  /usr/lib64/libportal.so  /usr/lib64/libgio-2.0.so  /usr/lib64/libgobject-2.0.so  src/lib/net/libnet.a  src/lib/io/libio.a  src/lib/mt/libmt.a  src/lib/base/libbase.a  src/lib/arch/libarch.a  -lpthread  -lXtst  -lX11  -lxkbfile  -lXext  -lXinerama  -lXrandr  -lXi  /usr/lib64/libssl.so  /usr/lib64/libcrypto.so  src/lib/common/libcommon.a  -lgcov  /usr/lib64/libQt6Network.so.6.8.2  /usr/lib64/libtomlplusplus.so  /usr/lib64/libQt6Xml.so.6.8.2  /usr/lib64/libQt6Core.so.6.8.2 && cd /home/nick/Projects/deskflow/build/src/apps/deskflow-client && QT_QPA_PLATFORM=minimal PATH=/home/nick/Projects/deskflow/build/bin: /usr/bin/help2man --name Keyboard\ and\ mouse\ sharing\ utility\ \(Client\) --include /home/nick/Projects/deskflow/src/apps/res/manpage.txt --no-info deskflow-client -o /home/nick/Projects/deskflow/build/bin/deskflow-client.1
help2man: can't get `--help' info from deskflow-client
Try `--no-discard-stderr' if option outputs to stderr
[3/4] Linking CXX executable bin/deskflow-server
FAILED: bin/deskflow-server 
: && /usr/bin/c++ -fPIC -g -fprofile-arcs -ftest-coverage -fprofile-update=atomic -g  src/apps/deskflow-server/CMakeFiles/deskflow-server.dir/deskflow-server_autogen/mocs_compilation.cpp.o src/apps/deskflow-server/CMakeFiles/deskflow-server.dir/deskflow-server.cpp.o -o bin/deskflow-server -L/usr/local/lib -Wl,-rpath,/usr/local/lib:  -lgcov  src/lib/arch/libarch.a  src/lib/base/libbase.a  src/lib/client/libclient.a  src/lib/io/libio.a  src/lib/mt/libmt.a  src/lib/net/libnet.a  src/lib/platform/libplatform.a  src/lib/server/libserver.a  src/lib/deskflow/libapp.a  -lpthread  -lXtst  -lX11  -lxkbfile  -lXext  -lXinerama  -lXrandr  -lXi  src/lib/client/libclient.a  src/lib/platform/libplatform.a  src/lib/server/libserver.a  src/lib/deskflow/libapp.a  /usr/lib64/libQt6DBus.so.6.8.2  /usr/lib64/libxkbcommon.so  /usr/lib64/libglib-2.0.so  /usr/lib64/libei.so  /usr/lib64/libportal.so  /usr/lib64/libgio-2.0.so  /usr/lib64/libgobject-2.0.so  /usr/lib64/libglib-2.0.so  /usr/lib64/libei.so  /usr/lib64/libportal.so  /usr/lib64/libgio-2.0.so  /usr/lib64/libgobject-2.0.so  src/lib/net/libnet.a  src/lib/io/libio.a  src/lib/mt/libmt.a  src/lib/base/libbase.a  src/lib/arch/libarch.a  -lpthread  -lXtst  -lX11  -lxkbfile  -lXext  -lXinerama  -lXrandr  -lXi  /usr/lib64/libssl.so  /usr/lib64/libcrypto.so  src/lib/common/libcommon.a  -lgcov  /usr/lib64/libQt6Network.so.6.8.2  /usr/lib64/libtomlplusplus.so  /usr/lib64/libQt6Xml.so.6.8.2  /usr/lib64/libQt6Core.so.6.8.2 && cd /home/nick/Projects/deskflow/build/src/apps/deskflow-server && QT_QPA_PLATFORM=minimal PATH=/home/nick/Projects/deskflow/build/bin: /usr/bin/help2man --name Keyboard\ and\ mouse\ sharing\ utility\ \(Server\) --include /home/nick/Projects/deskflow/src/apps/res/manpage.txt --no-info deskflow-server -o /home/nick/Projects/deskflow/build/bin/deskflow-server.1
help2man: can't get `--help' info from deskflow-server
Try `--no-discard-stderr' if option outputs to stderr
[4/4] Generating developer documentation
ninja: build stopped: subcommand failed.
```

Edit: Do we have a better way than to invoke `deskflow-[server|client] --help` from `help2man`? If not, I guess we could just say "We don't support building while you have server/client running" but that seems a bit annoying.

sithlord48:
> Edit: Do we have a better way than to invoke deskflow-[server|client] --help from help2man? If not, I guess we could just say "We don't support building while you have server/client running" but that seems a bit annoying. 

no not right now,  and this is happening on CI randomly with this PR.. this will have to get resolved first. Drafting until i do

sithlord48:
@nbolton  should be working proper now

winex:
well, this essentially breaks multi-server and multiseat usage, when you run more than 1 display server on the same host. can't it check `$DISPLAY` env var on linux?

currently, i must use patched build to run server and client, but different `DISPLAY=` for each:
```diff
-  QSharedMemory sharedMemory("deskflow-core");
+  QSharedMemory sharedMemory(isServer(argc, argv) ? "deskflow-core-server" : "deskflow-core-client");
```
but with 3rd X11 server, you are still unable to run more than 1 client! :(

maybe just add an option to suppress this? though, proper checking of `DISPLAY` and `WAYLAND_DISPLAY` vars would be even better

sithlord48:
Why would you want multi server ? 

How many "normal" users are using multi seat ?

winex:
anyone can install second GPU for either:
* encoding: obs, streaming, transcoding/decoding, w/e...
* TV, and you would love to run second X11/Wayland for that
* virtualization, with gpu pass-through

nowadays, more and more users (_when OS allows it, lol_) are starting to use that...

TCO is way less for a single PC with 2 GPUs. moreso, almost every modern CPU (_without **F** suffixes!_) now have video core (because manufacturers put h264/hevc/av1 codecs into silicon) which can be used any way you want

why would any software intentionally disable such support, if that worked before?!

### edit:
btw, `deskflow` is a must for me since this September: i've moved away from 2 X11-screens (:0.0 :0.1) on a single GPU to multiseat with triple GPUs, which works even better, but only with `deskflow` i don't need to touch second mouse at all (_well, when my fixes are applied_)

sithlord48:
I am running a 3rd screen from a second GPU right now and i did not need to start a new X11 session or Wayland Session todo so . The mouse works fine between them and i would suspect most people would use their second gpu in similar to this. 

If you want discuss this further please open up a new discussion.

miranda:
I can no longer update to the latest deskflow... I am stuck on 1.23 forever unless this change is reverted.
Why would you want to restrict the client from running twice on the same machine?
I am also using a multiple GPU setup but I cannot share the mouse between the screens without deskflow, as they are not contiguously arranged.

nbolton:
@sithlord48 Would you be open to adding a `--no-single-instance` CLI arg or `AllowMultipleInstances` setting... or something like that?

sithlord48:
I don't plan for this to be single instance forever just need fix the IPC so we can attach / detach the gui to / from the core. As well as a few other things.. I would be open to an option to allow multi instances  but this comes with some technical issues as the code sits now.

How could this even work on windows ? It is not possible in our current set up.

miranda:
> I would be open to an option to allow multi instances but this comes with some technical issues as the code sits now.

What was wrong with it before that it had to be changed?  Everything seemed to work fine when it was a separate client and server, rather than this "core" thing.

rickybrent:
> How could this even work on windows ? It is not possible in our current set up.

If this was a change made for Windows, perhaps it could be limited to Windows as well?

Issue #7951, opened 2024-12-05
Use unified core in GUI when `BUILD_UNIFIED` is set


nbolton:
### Feature Request

When `BUILD_UNIFIED` config option is set, the GUI won't be able to find the core binary (e.g. `deskflow-server` because it's now `deskflow-core` with the `server` option).

The GUI should run `deskflow-core server` instead of `deskflow-server` and `deskflow-core client` instead of `deskflow-client` when `BUILD_UNIFIED` is set.

PR #8868, opened 2025-08-19
feat: Unified core binary

sithlord48:
fixes: #7951

The gui can use `deskflow-core`
Remove `deskflow-client` and `deskflow-server`
Unify the deskflow-core --help to show all help (for man page ease)
Adjust wix to make rule to allow `deskflow-core` in and out of the firewall.

Tested on 
 - [X] Linux
 - [x] Windows
 - [x] macOS


src/lib/gui/core/CoreProcess.cpp
```diff
@@ -363,16 +365,30 @@ void CoreProcess::start(std::optional<ProcessMode> processModeOption)
 
   QString app;
   QStringList args;
+
+  qInfo() << "MODE IS" << mode() << "ARGS ARE" << args.join(", ");
```
nbolton: Please no all caps rage.

src/lib/gui/core/CoreProcess.cpp
```diff
@@ -363,16 +365,30 @@ void CoreProcess::start(std::optional<ProcessMode> processModeOption)
...
+  if (mode() == Server && !addServerArgs(args, app)) {
     qWarning("failed to add server args for core process, aborting start");
     return;
-  } else if (mode() == Settings::CoreMode::Client && !addClientArgs(args, app)) {
+  } else if (mode() == Client && !addClientArgs(args, app)) {
     qWarning("failed to add client args for core process, aborting start");
     return;
   }
 
+  if (app.endsWith("core") || app.endsWith("core.exe")) {
```
nbolton: This seems like a hack. How about using `kUnifiedCore` constant or something like that, which is based on CMake config?

sithlord48: when you define it at build time its not able to be changed and you can change the client / server binaries by changing the settings file.

src/lib/gui/core/CoreProcess.cpp
sithlord48: oh that is left over and i should remvoe this anyway

src/lib/gui/core/CoreProcess.cpp
sithlord48: This will be resolved when we only ship the one core app anyway.

src/lib/gui/core/CoreProcess.cpp
```diff
@@ -363,16 +365,26 @@ void CoreProcess::start(std::optional<ProcessMode> processModeOption)
...
     qWarning("failed to add client args for core process, aborting start");
     return;
   }
 
+  if (mode() == Server) {
+    args.prepend("server");
+  } else if (mode() == Client) {
+    args.prepend("client");
+  } else {
+    qFatal("Core started without mode");
```
nbolton:
```suggestion
    qFatal("core started without mode");
```

nbolton (changes requested):
Only very minor nit.

nbolton:
> * Linux
> * Windows
> * macOS

Works well. I tested the Windows dev build and dist package, both are good.

src/lib/gui/core/CoreProcess.cpp
sithlord48: done

nbolton (approved):

PR #8873, opened 2025-08-20
MainWindow Slight gui update

sithlord48:
move start and restart inline with mode options
 update labels for the mode buttons actions and menu actions 
 use callstart / callstop for connect and disconnect icons

Server mode (not started)

df-SNC

client mode (connected)
df-CC

client mode (DC)
df-CNC

### DIFF
--- a/src/lib/gui/MainWindow.ui
+++ b/src/lib/gui/MainWindow.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>781</width>
-    <height>483</height>
+    <width>758</width>
+    <height>466</height>
    </rect>
   </property>
   <property name="sizePolicy">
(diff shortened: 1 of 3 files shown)


src/lib/gui/MainWindow.ui
```diff
@@ -293,7 +341,7 @@
                 </sizepolicy>
                </property>
                <property name="text">
-                <string>Server IP address or hostname:</string>
+                <string>Server Ip or hostname:</string>
```
nbolton:
```suggestion
                <string>Server IP or hostname:</string>
```

src/lib/gui/MainWindow.ui
```diff
@@ -385,29 +521,6 @@
            </property>
           </spacer>
          </item>
-         <item>
-          <widget class="QPushButton" name="btnRestartCore">
-           <property name="enabled">
-            <bool>false</bool>
-           </property>
-           <property name="text">
-            <string>Rest&amp;art</string>
```
nbolton, line 394: ~~Just thinking out loud... What was the rationale behind removing the 'Restart' functionality from the stop/start button? I'm fine with this, but I think users might miss it. I know they can access from the File menu or Ctrl+S, but I bet many users will miss this.~~

Edit: Oops, totally missed this lovely new restart button :facepalm: (my powers of observation are clearly suffering right now)

image

nbolton (approved):

PR #8881, opened 2025-08-22
Disable meta mod (fixes alt key bug)

nbolton:
Fixes: #8840

I have no idea why, but removing the [meta key](https://en.wikipedia.org/wiki/Meta_key) mapping fixes the Alt key on my Arch+KDE client.

I do wonder if the [original] meta key is actually used today. It often gets mixed up with the super ("logo") key (which is called meta now by many DE's such as KDE).

My meta/logo/super key works fine with this PR.

image

CC @wismill @whot

### DIFF
--- a/src/lib/platform/EiKeyState.cpp
+++ b/src/lib/platform/EiKeyState.cpp
@@ -159,12 +159,12 @@ std::uint32_t EiKeyState::convertModMask(xkb_mod_mask_t xkbModMaskIn) const
       modMaskOut |= (1 << kKeyModifierBitAltGr);
     else if (strcmp(XKB_VMOD_NAME_LEVEL5, name) == 0)
       modMaskOut |= (1 << kKeyModifierBitLevel5Lock);
-    else if (strcmp(XKB_VMOD_NAME_META, name) == 0)
-      modMaskOut |= (1 << kKeyModifierBitMeta);
     else if (strcmp(XKB_VMOD_NAME_NUM, name) == 0)
       modMaskOut |= (1 << kKeyModifierBitNumLock);
     else if (strcmp(XKB_VMOD_NAME_SCROLL, name) == 0)
       modMaskOut |= (1 << kKeyModifierBitScrollLock);
+    else if (strcmp(XKB_VMOD_NAME_META, name) == 0) // possibly the old meta (not the new meta/super/logo key)
+      LOG_DEBUG2("modifier mask %s ignored", name);
     else if (strcmp(XKB_MOD_NAME_MOD2, name) == 0) // spare, sometimes mapped to num lock.
       LOG_DEBUG2("modifier mask %s ignored", name);
     else if (strcmp(XKB_MOD_NAME_MOD3, name) == 0) // spare, could be mapped to alt_r, caps lock, scroll lock, etc.
--- a/src/lib/platform/EiKeyState.cpp
+++ b/src/lib/platform/EiKeyState.cpp
@@ -163,11 +163,10 @@ std::uint32_t EiKeyState::convertModMask(xkb_mod_mask_t xkbModMaskIn) const
       modMaskOut |= (1 << kKeyModifierBitNumLock);
     else if (strcmp(XKB_VMOD_NAME_SCROLL, name) == 0)
       modMaskOut |= (1 << kKeyModifierBitScrollLock);
-    else if (strcmp(XKB_VMOD_NAME_META, name) == 0) // possibly the old meta (not the new meta/super/logo key)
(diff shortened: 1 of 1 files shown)


nbolton:
Edit: Nope, that also breaks it. How strange.

Hmm, this also fixes the bug:
```patch
diff --git a/src/lib/platform/EiKeyState.cpp b/src/lib/platform/EiKeyState.cpp
index dbaeeb958..56da2cd10 100644
--- a/src/lib/platform/EiKeyState.cpp
+++ b/src/lib/platform/EiKeyState.cpp
@@ -139,6 +139,7 @@ std::uint32_t EiKeyState::convertModMask(xkb_mod_mask_t xkbModMaskIn) const
     static const auto XKB_VMOD_NAME_NUM = "NumLock";
     static const auto XKB_VMOD_NAME_SCROLL = "ScrollLock";
     static const auto XKB_VMOD_NAME_SUPER = "Super";
+    static const auto XKB_VMOD_NAME_META = "Meta";
     static const auto XKB_MOD_NAME_MOD2 = "Mod2";
     static const auto XKB_MOD_NAME_MOD3 = "Mod3";
     static const auto XKB_MOD_NAME_MOD5 = "Mod5";
@@ -152,7 +153,8 @@ std::uint32_t EiKeyState::convertModMask(xkb_mod_mask_t xkbModMaskIn) const
       modMaskOut |= (1 << kKeyModifierBitControl);
     else if (strcmp(XKB_MOD_NAME_ALT, name) == 0 || strcmp(XKB_VMOD_NAME_ALT, name) == 0)
       modMaskOut |= (1 << kKeyModifierBitAlt);
-    else if (strcmp(XKB_MOD_NAME_LOGO, name) == 0 || strcmp(XKB_VMOD_NAME_SUPER, name) == 0)
+    else if (strcmp(XKB_MOD_NAME_LOGO, name) == 0 || strcmp(XKB_VMOD_NAME_SUPER, name) == 0 ||
+             strcmp(XKB_VMOD_NAME_META, name) == 0)
       modMaskOut |= (1 << kKeyModifierBitSuper);
     else if (strcmp(XKB_MOD_NAME_MOD5, name) == 0 || strcmp(XKB_VMOD_NAME_LEVEL3, name) == 0)
       modMaskOut |= (1 << kKeyModifierBitAltGr);
```

sithlord48:
I always though Meta and Super were synonyms for the same key ... (now i know better...)

nbolton:
> Meta and Super

Confuses the hell out of me every single time I encounter it. Now I just see them and think... IDK what they are any more, better look it up... nope, none the wiser.

sithlord48 (approved):
Does not break my "Logo" key and other mods also work for me still !

src/lib/platform/EiKeyState.cpp
```diff
@@ -159,15 +159,14 @@ std::uint32_t EiKeyState::convertModMask(xkb_mod_mask_t xkbModMaskIn) const
...
     else if (strcmp(XKB_VMOD_NAME_NUM, name) == 0)
       modMaskOut |= (1 << kKeyModifierBitNumLock);
     else if (strcmp(XKB_VMOD_NAME_SCROLL, name) == 0)
       modMaskOut |= (1 << kKeyModifierBitScrollLock);
-    else if (strcmp(XKB_MOD_NAME_MOD2, name) == 0) // spare, sometimes mapped to num lock.
-      LOG_DEBUG2("modifier mask %s ignored", name);
-    else if (strcmp(XKB_MOD_NAME_MOD3, name) == 0) // spare, could be mapped to alt_r, caps lock, scroll lock, etc.
+    else if ((strcmp(XKB_VMOD_NAME_META, name) == 0) || // possibly the old meta (not the new meta/super/logo key)
+             (strcmp(XKB_MOD_NAME_MOD2, name) == 0) ||  // spare, sometimes mapped to num lock.
+             (strcmp(XKB_MOD_NAME_MOD3, name) == 0)     // spare, could be mapped to alt_r, caps lock, scroll lock, etc.
```
wismill, line 168: `Mod3` usually encodes `LevelFive`, which is used in some keyboard layouts to access the 5th level. It should probably not be omitted.

wismill (commented):
`Meta` is usually encoded like `Alt`, i.e. to `Mod1`. In that case, both share the same state.

Added to that, if KDE interprets `Meta` as `Super` (the logo key), then it might explain the mess.

src/lib/platform/EiKeyState.cpp
nbolton: >  It should probably not be omitted.

Thanks, I'll raise a PR.

PR #8884, opened 2025-08-23
LogTests are now QTests

sithlord48:
Port the LogTests to QTest.


src/unittests/base/LogTests.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+
+void LogTests::printTestWithArgs()
+{
+  Log log(false);
+
+  std::stringstream buffer;
+  std::streambuf *old = std::cout.rdbuf(buffer.rdbuf());
+
+  log.print(nullptr, 0, LEVEL_INFO "test %d %.2f %s", 1, 1.234, "test arg");
```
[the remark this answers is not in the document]
nbolton: What? Why does it think 1.234 is an int?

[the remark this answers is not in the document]
sithlord48: idk i just told it to ignore that cause its used in a test anyway .

src/unittests/base/LogTests.cpp
nbolton, line 1: Just thinking out loud...

What do you think about the style `{function-to-test}_{scenario}_{expectation}`?

> Be named like so: {function-to-test}_{scenario}_{expectation}
https://github.com/deskflow/deskflow/wiki/Unit-Testing#writing-tests

I find this makes it easier for others to understand the intent:
- What function are you testing? (it's not always clear, especially in complex tests)
- What is the scenario or input you're trying to create? (with lots of setup, this can get confusing, fast)
- What is the expectation you're trying to test against? (expects can be a bit arcane/esoteric)

In other languahes like TypeScript, the `describe` grouping helps to communicate to other developers what function is under test, e.g.
```ts
describe("someFunction", {
it("does something in some certain condition") {
//...
}
it("does something else in some other situation") {
//...
}
}
```

I find that test names like `printTest` can be a bit vague... what are you expecting the output to be? When we use a more structured function name with underscores, it makes the developer more likely to supply conditon and expectatation.

I see that you're starting all your test cases with the function under test, which is good, and I wonder if we could make it clearer by visually separating the function name from the condition and expectation.

When we use camelCase for long sentences, they get quite tricky to read fast and you end up compromising and trying to shorten the function name: e.g. `printWithErrorValidOutput` or `printInfoWithFileAndLine`

nbolton (approved):

src/unittests/base/LogTests.cpp
sithlord48: i fixed that one, but as a side note i also I plan to remove all these pointless tests when we drop this object for Qt Logging..

nbolton: Ah yes, it is reinventing the wheel, isn't it?

Issue #8852, opened 2025-08-12
Replace CLI args with Qt settings file read in Core


nbolton:
Blocks: #8849

The Core* is launched via the GUI, and is controlled through CLI args. This is a problem for a two important reasons:
1. Extra code to maintain, unnecessarily; CLI parsing in the Core and CLI construction in the GUI. The Core was developed a long time before the GUI, which was developed as a wrapper on top of the Core; extra code means more bugs and more expensive maintenance. 
2. When working on the Core directly (i.e., taking the GUI totally out of the equation), we launch the Core directly. Using CLI args is tricky during Core development, since it's fiddly to make changes to the args; much faster to edit a text file, which doesn't break your flow.

The solution is most likely to delete all CLI arg parsing code in the Core, replacing it with simple and easy-to-maintain Qt settings read code. This way, when the GUI writes to the config file (which already happens anyway) we can simply re-read the config from the Core. Initially we can restart the process, but long-term a hot-reload feature would be useful so the Core process doesn't need to be restarted when changing the config.

Once this is done, we can say goodbye to the TOML parser, which becomes redundant.

** "Core" is a collective term for `deskflow-core`/`deskflow-server`/`deskflow-client`

sithlord48:
So how should this work ? 

1. Can we pass args to the core ? if so what ? 
2. If an arg is set it would override or replace what is in the settings (if a settings exists ? )
3. Should the core just attempt to load the settings automatically ? is there a switch -c or something (optional , mandatory ?? )

nbolton:
I suspect the best design is to _remove the args_ altogether, except for `-c` + `--config`. People do not use the Core CLI directly as a day-to-day tool like `ls` or `cp`, and so args become a hindrance rather than a convenience.

Edit: @sithlord48 This is assigned to me, and I plan on working on this. Let me know if you'd rather take it over.

sithlord48:
I might get to it before you. Lmk when you really starting work on it

sithlord48:
I may have started a thing #8886

PR #8886, opened 2025-08-23
Use settings in core objects

sithlord48:
Fixes: #8852 

Remove args from the `IApp` Classes new arg are handled by the new `CoreArgParser` and contained in the new `CoreArgs.h` header. 

The CoreArgParser will be used to parse the arguments for `deskflow-core`. Parsed arguments are then written into the settings file (the normal one unless you provide `-s`.  All of the options shown for deskflow-core will be optional when this is finished. (As in they will use the value in settings if not provided). With the exception of Setting log will write to the log file this is the only arg passed to deskflow-core other than the coremode.

```
❯ ./deskflow-core --help
deskflow-core: 1.24.0.50 (5f404e53)
Usage: deskflow-core coremode [options]
Keyboard and mouse sharing utility

Options:
  -h, --help                       Display Help on the command line
  -v, --version                    Display version information
  -s, --settings <configFile>      override configuration file to use
  -i, --interface <address>        Use a specific interface for the connection.
                                   Instead of any address in client mode or
                                   listening on all addresses in server mode
                                   address.
                                   address must be the IP address or hostname of
                                   an interface on the system.
  -p, --port <port>                Port to use in place of default port
  -n, --name <screen-name>         use screen-name instead the hostname to
                                   identify this screen
  --log-level <level>              filter out log messages with priority below
                                   level.
                                   level may be:
                                   FATAL, ERROR, WARNING, NOTE, INFO, DEBUG,
                                   DEBUG1, DEBUG2.
  -l, --log <file>                 Write messages to file
  --secure <value>                 Enable TLS encryption (default: true)
  --tls-cert <file>                Use file in place of default TLS certificate
                                   path
  --prevent-sleep <value>          When true the machine will be prevented from
                                   sleeping while the program is running
  -r, --restartOnFailure <value>   Set if the core should automatically restart
                                   if it fails
  --display <display>              When in X mode, connect to the X server at
                                   <display>
  --useHooks <value>               Sets if hooks are used for windows desks
  --peerCertCheck <value>          Server Mode: Enable client SSL certificate
                                   checking (default: true)
  --serverConfig <path>            Server Mode: File to load as server config
  --yscroll <value>                Client Mode: Vertical scrolling delta
                                   (default: 120)
  --languageSync <value>           Client Mode: Sync language with client
                                   (default: false)
  --invertScrollDirection <value>  Client Mode: Set scroll direciton to be
                                   inverted (default: false)
  --remoteHost <hostname>          Client Mode: Sets the remote host in the
                                   settings file

Arguments:
  coremode                         The mode to start in either: server or
                                   client
```

Version info prints 
```
❯ ./deskflow-core --version
deskflow-core v1.23.0.152 (88217f30), protocol v1.8
Copyright (C) 2024-2025 Deskflow Devs
Copyright (C) 2012-2025 Symless Ltd.
Copyright (C) 2009-2012 Nick Bolton
Copyright (C) 2002-2009 Chris Schoeneman
```
 
- Remove the old argparser and friends.
- Add Option to SettingsDialog to adjust ScrollSpeed


nbolton (approved):
Unable to test as OOO, but LGTM.

sithlord48:
testing is preferred .

Rebased and resolved conflicts

nbolton (approved):
Looking for what changed since my last approve, guessing more old code deleted?

Same as last time, still OOO so can't test but LGTM.

PR #8914, opened 2025-09-01
Sonar cleanup 15

sithlord48:
Address some more sonar issues.


nbolton (approved):
Interesting, single case switches... must have been vestigial.

PR #8921, opened 2025-09-04
Better log dock

sithlord48:
- Split the LogDock into its own item
 - Allow the Log dock to float or be attached to the bottom  (dock will always start docked on to the bottom)

My help w/ weird sizing on GNOME ?


src/lib/gui/widgets/LogDock.cpp
```diff
@@ -0,0 +1,71 @@
...
+  m_lblTitle->setVisible(!floating);
+  m_btnClose->setVisible(!floating);
+  show();
+}
+
+void LogDock::setupTitleWidget()
+{
+  m_btnFloat->setStyleSheet(deskflow::gui::kStyleFlatButtonHoverable);
+  m_btnFloat->setCheckable(true);
+  m_btnFloat->setIcon(QIcon::fromTheme(QIcon::ThemeIcon::ViewRestore));
```
sithlord48: This icon needs to be checked on
 Windows 
 Mac 
 Gnome

If its missing on any ill have to it in our theme

nbolton:
> My help w/ weird sizing on GNOME ?

Similar issues on GNOME. This is what it looks like when I detach the log.

image

nbolton:
The icons look a bit too big on Windows.

image

Edit: And the macOS icons look OK, but don't match Linux.

Screenshot 2025-09-04 at 09 31 10

nbolton (approved):

sithlord48:
The icons should be platform native. But looking at how windows picks the "refresh" icon.. I ll have to at least pack an icon for windows at least. Do you have "breeze" theme installed on your gnome machine ? if yoru not using breeze theme do you get an icon in gnome?

sithlord48:
@nbolton thoughts on remove the whole logic for fixed window sizing at least then users  can make the window smaller if it does this kinda stuff.

nbolton:
> Do you have "breeze" theme installed on your gnome machine ? if yoru not using breeze theme do you get an icon in gnome?

Breeze detach is always white, so invisible in light mode.

image

---

image

---

image

nbolton:
> remove the whole logic for fixed window sizing

For me personally, I'm happy as long as I can resize it and it remembers what size I made it.

src/lib/gui/widgets/LogDock.cpp
```diff
@@ -0,0 +1,85 @@
+/*
+ * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2025 Chris Rizzitello <sithlord48@gmail.com>
```
nbolton, line 3: Did you want to say Deskflow Developers?

sithlord48:
breeze is the "light" color theme "breeze-dark" would have the icons from dark schemes

src/lib/gui/widgets/LogDock.cpp
sithlord48: no, i use deskflow developers when moving code around not writing from scratch.

nbolton:
> breeze is the "light" color theme "breeze-dark" would have the icons from dark schemes

Indeed, though that doesn't explain why the icon is white when using light mode style (i.e. regular Breeze). The screenshot is confusing as our GUI uses GNOME legacy style (not modern dark mode).

image

sithlord48:
I Tested on Debian 13 with GNOME 48 unable to repo that icon issue.

nbolton:
> I Tested on Debian 13 with GNOME 48 unable to repo that icon issue.

OK, I'll test on a VM. Maybe my "default" GNOME Tweak settings are not actually default.

nbolton:
Yeah, looks fine on stock Debian 13.

Screenshot_debian-13_2025-09-11_10:55:13

> Maybe my "default" GNOME Tweak settings are not actually default.

I'll chalk that up to experience I guess... GNOME.

PR #8930, opened 2025-09-09
refactor: ScreenConfigDialog, make sure to populate the screen model after server config

sithlord48:
This should correct the inability to write to the server config


nbolton (approved):
That fixed it, thanks.

PR #8935, opened 2025-09-10
fix: Store server config window state in memory instead of on disk

### LINKED ISSUE #8931: Accept client prompt not showing on Windows server
### Project

Deskflow

### Deskflow version info

Deskflow: 1.23.0.203 (539ddcaf)
Qt: 6.9.1
System: Windows 11 Version 24H2

### Build type

### Operating systems (OS)

- [x] Windows
- [ ] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

### Signing on macOS

- [ ] I am not using macOS

### Continuous build

### OS versions/distros

Windows 11

### Deskflow configuration

Windows server

1. Reset settings or delete all clients from server config
2. Select server mode (if not already selected)
3. Connect a client

Expect: Accept client dialog shows
Actual:
```
[2025-09-10T09:07:45] WARNING: unrecognised client name "spock-tux", check server config
[2025-09-10T09:07:45] DEBUG: server config dialog visible, skipping new client prompt
```

In reality, the 'server config dialog visible' is not actually visible.

### Log output

```shell

```

### Additional information

_No response_

nbolton:
Fixes #8931


sithlord48:
I forget why (and it could be something we have since fixed) But not storing this as a setting was preventing it from not being shown in some cases. This needs to be tested thoroughly before landing

nbolton:
Ah, is this related to hide on start or something like that?

nbolton:
Ignore my last comment.

> preventing it from not being shown in some cases

Hmm, something like a race condition from the server config being non-blocking once upon a time?

sithlord48:
> Ignore my last comment.
> 
> > preventing it from not being shown in some cases
> 
> Hmm, something like a race condition from the server config being non-blocking once upon a time?

Yes something like that i just don't recall exactly what or why it happened. Just that the solution was at that time to have them both check settings first. Its a hack and it should be done how you are doing it here. So hopefully we have resolved it

Issue #8937, opened 2025-09-10
Remove dead service install/remove code


nbolton:
### Feature Request

I realised that I am the only person using `ArchDaemonWindows::installDaemon` and `ArchDaemonWindows::uninstallDaemon` purely for development; users use WiX to install/uninstall the service. Keeping these doesn't make sense as it causes more maintenance burden (e.g. I just spent the last hour fixing unicode bugs in those 2 functions as part of #8640).

I will remove this dead code and write some simple install/uninstall script in Python under https://github.com/deskflow/scripts (if I need to).

sithlord48:
Is this code used when you enable / disable the option to use the daemon ?

nbolton:
> Is this code used when you enable / disable the option to use the daemon ?

No, it's dead code. It was used before you refactored the WiX installer to use the builtin WiX service management and since then it was only used by me installing/uninstalling from my IDE during dev... which was pointless since we have `sc.exe` on Windows. It's highly likely that `sc.exe` didn't exist in 2003(?) when the original Windows daemon functions were written.

PR #8938, opened 2025-09-10
chore: Remove unused daemon installation and uninstallation methods

nbolton:
Fixes: #8937


nbolton:
> Nice remove some more unneeded code!

Feels good.

PR #8942, opened 2025-09-10
docs: Simplify bug_report.yml to make it easier to use

nbolton:
I think this would be much easier to use while achieving the same result. Not sure how it's going to look formatting-wise... there's no preview.


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -66,60 +66,21 @@ body:
...
-      required: true
-
   - type: textarea
     id: config
     attributes:
       label: Deskflow configuration
       description: |
         Please provide a very brief description of your configuration.
-        Let us know what OS your server and client are running.
+        Let us know what OS your server and client are running, including all OS versions.
```
sithlord48, line 75: ideally they would provide the about info copied from the about dialog, as well as what distro (for linux)

sithlord48: Please copy the about info from the aboutdialog in deskflow

ex 
 ``` 
Deskflow: 1.23.0.206 (66a022a0)
Qt: 6.9.2
System: Arch Linux
Session: KDE (wayland)
```

Comment (nbolton) [reply .github/ISSUE_TEMPLATE/bug_report.yml:75]: We already have that, see L25:
```
  - type: textarea
    id: version
    attributes:
      label: Deskflow version info
      description: You can find the Deskflow version number on the About screen.
      placeholder: |
        Deskflow: 1.2.3.4 (deadc0de)
        Qt: 3.2.1
        System: Hannah Montana Linux (Workstation Edition)
        Session: Unity (Compiz)
    validations:
      required: true
```

The problem is, they're likely to only paste one of their computers where _they_ think the bug is (but that's just a wild guess for most users). Even if we ask them to paste it from every computer they probably won't bother or won't notice we asked.

Hence, I ask people for the OS version of all their computers.

PR #8948, opened 2025-09-12
Make user agent RFC 9110, add popularity contest, improve update check dialog

nbolton:
- Fixes user agent header not being RFC 9110 10.1.5
- Only send extra user agent data in popularity contest mode
- Fixes update check URL settings key (should be GUI, not Core)
- Remove custom tab stops from settings dialog and use default

New design for version/update check dialog:
image

### DIFF
--- a/src/lib/common/Settings.h
+++ b/src/lib/common/Settings.h
@@ -67,6 +67,7 @@ public:
     inline static const auto AutoStartCore = QStringLiteral("gui/startCoreWithGui");
     inline static const auto AutoUpdateCheck = QStringLiteral("gui/enableUpdateCheck");
     inline static const auto UpdateCheckUrl = QStringLiteral("gui/updateCheckUrl");
+    inline static const auto JoinPopularityContest = QStringLiteral("gui/joinPopularityContest");
     inline static const auto CloseReminder = QStringLiteral("gui/closeReminder");
     inline static const auto CloseToTray = QStringLiteral("gui/closeToTray");
     inline static const auto LogExpanded = QStringLiteral("gui/logExpanded");
@@ -207,6 +208,7 @@ private:
     , Settings::Gui::AutoStartCore
     , Settings::Gui::AutoUpdateCheck
     , Settings::Gui::UpdateCheckUrl
+    , Settings::Gui::JoinPopularityContest
     , Settings::Gui::CloseReminder
     , Settings::Gui::CloseToTray
     , Settings::Gui::LogExpanded
@@ -229,6 +231,7 @@ private:
     , Settings::Gui::AutoStartCore
     , Settings::Gui::ShownFirstConnectedMessage
     , Settings::Gui::ShownServerFirstStartMessage
+    , Settings::Gui::JoinPopularityContest
     , Settings::Core::PreventSleep
     , Settings::Core::UseWlClipboard
     , Settings::Server::ExternalConfig
--- a/src/lib/gui/Messages.cpp
+++ b/src/lib/gui/Messages.cpp
@@ -8,14 +8,18 @@
 
 #include "Logger.h"
 
+#include "VersionChecker.h"
+
 #include "common/Enums.h"
 #include "common/Settings.h"
 #include "common/UrlConstants.h"
 #include "common/VersionInfo.h"
 
 #include <QCheckBox>
+#include <QLabel>
 #include <QMessageBox>
 #include <QPushButton>
+#include <QVBoxLayout>
 #include <memory>
 
 namespace deskflow::gui::messages {
@@ -298,22 +302,94 @@ void showWaylandLibraryError(QWidget *parent)
   );
 }
 
-bool showUpdateCheckOption(QWidget *parent)
+OnlineServicesOptions showOnlineServicesOptions(QWidget *parent)
 {
-  QMessageBox message(parent);
-  message.addButton(QObject::tr("No thanks"), QMessageBox::RejectRole);
-  const auto checkButton = message.addButton(QObject::tr("Check for updates"), QMessageBox::AcceptRole);
-  message.setText(
-      QObject::tr(
-          "<p>Would you like to check for updates when %1 starts?</p>"
-          "<p>Checking for updates requires an Internet connection.</p>"
-          "<p>URL: <pre>%2</pre></p>"
-      )
-          .arg(kAppName, Settings::value(Settings::Gui::UpdateCheckUrl).toString())
-  );
+  QDialog dialog(parent);
+  auto layout = new QVBoxLayout(&dialog);
 
-  message.exec();
-  return message.clickedButton() == checkButton;
+  auto about = new QLabel(
+      QObject::tr(
+          "Some optional %1 features require an Internet connection. "
+          "Would you like to enable them?"
+      )
+          .arg(kAppName)
+  );
+  about->setWordWrap(true);
+  layout->addWidget(about);
+
+  auto prompt = new QLabel(QObject::tr("Choose which features to enable:"));
+  layout->addWidget(prompt);
+
+  auto cbCheckForUpdates = new QCheckBox(QObject::tr("Check for new versions when app starts"));
+  layout->addWidget(cbCheckForUpdates);
+
+  auto cbPopularityContest = new QCheckBox(QObject::tr("Participate in popularity contest"));
+  cbPopularityContest->setEnabled(false);
+  cbPopularityContest->setToolTip(
+      QObject::tr("Requires the version check. The popularity contest is based on the user agent.")
+  );
+  layout->addWidget(cbPopularityContest);
+
+  auto url = new QLabel(
+      QObject::tr("<b>URL:</b>&nbsp;&nbsp;<code>%1</code>")
+          .arg(Settings::value(Settings::Gui::UpdateCheckUrl).toString())
+  );
+  layout->addWidget(url);
+
+  auto userAgent = new QLabel();
+  userAgent->setWordWrap(true);
+  userAgent->setVisible(false);
+  layout->addWidget(userAgent);
+
+  auto buttonBox = new QDialogButtonBox(QDialogButtonBox::Ok);
+  auto cancelButton = buttonBox->addButton(QObject::tr("No thanks"), QDialogButtonBox::RejectRole);
+  layout->addWidget(buttonBox);
+
+  QObject::connect(buttonBox, &QDialogButtonBox::accepted, &dialog, &QDialog::accept);
+  QObject::connect(buttonBox, &QDialogButtonBox::rejected, &dialog, &QDialog::reject);
+
+  auto okButton = buttonBox->button(QDialogButtonBox::Ok);
+  okButton->setEnabled(false);
+
+  auto update = [&]() {
+    // Popularity contest requires update check.
+    cbPopularityContest->setEnabled(cbCheckForUpdates->isChecked());
+    if (!cbCheckForUpdates->isChecked()) {
+      cbPopularityContest->setChecked(false);
+    }
+
+    // Disable accept if no checkbox is checked. This is important because the user may not take the
+    // time understand they need to check at least one option.
+    okButton->setEnabled(cbCheckForUpdates->isChecked() || cbPopularityContest->isChecked());
+
+    const auto static userAgentText = QObject::tr("<b>User agent:</b>&nbsp;&nbsp;<code>%1</code>");
+
+    if (cbPopularityContest->isChecked()) {
+      userAgent->setText(userAgentText.arg(VersionChecker::getUserAgent(true)));
+      userAgent->setVisible(true);
+    } else if (cbCheckForUpdates->isChecked()) {
+      userAgent->setText(userAgentText.arg(VersionChecker::getUserAgent(false)));
+      userAgent->setVisible(true);
+    } else {
+      userAgent->setVisible(false);
+    }
+
+    dialog.adjustSize();
+  };
+
+  QObject::connect(cbCheckForUpdates, &QCheckBox::toggled, update);
+  QObject::connect(cbPopularityContest, &QCheckBox::toggled, update);
+
+  dialog.exec();
+
+  if (dialog.result() != QDialog::Accepted) {
+    return {};
+  }
+
+  OnlineServicesOptions options;
+  options.checkForUpdates = cbCheckForUpdates->isChecked();
+  options.joinPopularityContest = cbPopularityContest->isChecked();
+  return options;
 }
 
 bool showDaemonOffline(QWidget *parent)
--- a/src/lib/gui/VersionChecker.cpp
+++ b/src/lib/gui/VersionChecker.cpp
@@ -22,15 +22,27 @@ VersionChecker::VersionChecker(QObject *parent) : QObject(parent), m_network{new
   connect(m_network, &QNetworkAccessManager::finished, this, &VersionChecker::replyFinished, Qt::UniqueConnection);
 }
 
+QString VersionChecker::getUserAgent(bool popularityContestEnabled)
+{
+  if (!popularityContestEnabled) {
+    return QStringLiteral("%1/%2").arg(kAppName, kVersion);
+  }
+
+  const static auto arch = QSysInfo::currentCpuArchitecture();
+  const static auto os = QSysInfo::prettyProductName();
+  const static auto osFamily = QSysInfo::kernelType();
+  const static auto language = QLocale::system().name();
+  return QStringLiteral("%1/%2 (%3; %4; %5; %6)").arg(kAppName, kVersion, os, osFamily, arch, language);
+}
+
 void VersionChecker::checkLatest() const
 {
-  const QString url = Settings::value(Settings::Gui::UpdateCheckUrl).toString();
+  const auto userAgent = getUserAgent(Settings::value(Settings::Gui::JoinPopularityContest).toBool());
+  const auto url = Settings::value(Settings::Gui::UpdateCheckUrl).toString();
   qDebug("checking for updates at: %s", qPrintable(url));
   auto request = QNetworkRequest(url);
-  auto userAgent = QString("%1 %2 on %3").arg(kAppName, kVersion, QSysInfo::prettyProductName());
+  qDebug() << "user agent header:" << userAgent;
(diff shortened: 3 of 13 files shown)


sithlord48 (changes requested):
I am 100% not ok with sending users machine info to our api.

Why do we need to send this data ? 

not to mention any stats you collect would be HIGHLY skewed as it woudl only be those with an update check who report in our popularity contest

nbolton:
The only reason we’d like to collect OS information is so we can have a rough understanding of what platforms our community is actually running on. That could help to decide as a community where to spend our limited time (for example, which OSes to fix bugs for, or which bugs to prioritise if they affect a significant share of users).

You’re right that the numbers will never be perfect, since they only include people with update checks enabled. But even an imperfect signal is better than guessing in the dark.

I 100% agree that privacy matters. One simple compromise would be to add a checkbox so that anyone who doesn’t want to send a user agent can easily opt out. That way, people who are happy to share basic OS info can help with our community priorities, and those who aren’t can keep update checks fully private. 

Personally, I would absolutely opt-in to this, much like I opt-in to Debian's optional popularity contest.

Does that sound like a fair middle ground? 🙂

Edit: Something like this...
image

sithlord48:
It has to be Opt-in and needs to have an option in the preferences to toggle it along with the update (i.e seperate options)

sithlord48:
Also combine those commits around the popularity contest into one commit .. the initial introduction of it should include the oiptons and gui parts.

sithlord48 (changes requested):
Feature needs to be opt-in. The settings dialog shoud allow me to change the option if update check is enabled.

nbolton:
> option in the preferences to toggle it along with the update (i.e seperate options)

Good catch! I forgot about that. Working on it now 👍

src/lib/common/Settings.h
```diff
@@ -62,6 +61,8 @@ class Settings : public QObject
   {
     inline static const auto Autohide = QStringLiteral("gui/autoHide");
     inline static const auto AutoUpdateCheck = QStringLiteral("gui/enableUpdateCheck");
+    inline static const auto UpdateCheckUrl = QStringLiteral("gui/updateCheckUrl");
+    inline static const auto JoinPopularityContest = QStringLiteral("gui/joinPopularityContest");
```
sithlord48, line 70: This needs to be false by default

nbolton: Agreed. Good news: I already designed it this way.

sithlord48: Incorrect you need to add it to `Settings::defaultValue` so when its called with the key false is returned as its default. Right now its "default" is not set

nbolton: Could you elaborate on the behavior you're seeing?

When there is no value set, `Settings::value(Settings::Gui::JoinPopularityContest).toBool()` returns `false` for me, as `false` is the default when there is no value.

Are you saying that you get a `true` value from this code when there is no value set?

Edit: Effectively the code is doing this, so I can't understand how you'd see `true`:
```cpp
QVariant v;
bool b = v.toBool(); // b == false
```

src/lib/gui/Messages.cpp
```diff
@@ -318,11 +318,20 @@ bool showUpdateCheckOption(QWidget *parent)
...
+          "<p><b>URL:</b>&nbsp;&nbsp;<code>%2</code></p>"
+          "<p>Optional: Participating in the popularity contest sends your %1 version, "
+          "language, and OS name to the %1 API which helps us prioritize bug fixes.</p>"
       )
-          .arg(kAppName, Settings::value(Settings::Core::UpdateUrl).toString())
+          .arg(kAppName, Settings::value(Settings::Gui::UpdateCheckUrl).toString())
   );
 
+  auto popularityContest = new QCheckBox(QObject::tr("Participate in popularity contest"));
+  QObject::connect(popularityContest, &QCheckBox::toggled, [&](bool enabled) {
```
sithlord48: You do not need to capture `[&]` here as Settings is not part of this class and we don't read the checkbox at all

src/lib/common/Settings.h
sithlord48: yes you get a bool cause it has to convert the variant into something .. you are getting from settings an empty and invalid QVariant. So when you call Settings::defaultValue(Settings::Gui::Join..)  you get an empty variant this can cause our setValue to remove the key .

sithlord48 (changes requested):

src/lib/gui/VersionChecker.cpp
```diff
@@ -40,6 +40,7 @@ void VersionChecker::replyFinished(QNetworkReply *reply)
   const auto httpStatus = reply->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
   if (reply->error() != QNetworkReply::NoError) {
     qWarning("version check server error: %s", qPrintable(reply->errorString()));
+    qWarning("version check server response: %s", qPrintable(QString(reply->readAll())));
```
sithlord48, line 54: If we are going to do this can you please here note this is added because of the Qtbug .... this way if the bug gets fixed we can remove this or adjust the code and know why we added it

src/lib/gui/dialogs/SettingsDialog.ui
```diff
@@ -642,6 +642,8 @@
   <tabstop>cbLanguageSync</tabstop>
   <tabstop>cbScrollDirection</tabstop>
   <tabstop>cbAutoUpdate</tabstop>
+  <tabstop>cbAutoHide</tabstop>
```
sithlord48: We do not care about the tab order.  This should really be removed completely. 

This overrides the tab order, we want the default that should make the order from top left to bottom right i.e what we are forcing now w/o having to redo it if we move a control's location on the form.

sithlord48:
Small Nitpick: 

How about using a tooltip to describe what the popularity contest does / sends instead of being on the dialog ? 

Edit Edit, 

Come to think of it i would like this to not be tied at all to the update check..

Perhaps we should have this instead be a button you can press on the dialog to report this way its a onetime report not every update check

sithlord48:
How about a message box that has two check boxes saying and we use this in place of our current message box 

```                                                         
   Some descriptive message 
    [ ] Check for updates when deskflow starts 

    [ ] Participate in the in popularity contest,         
         The following information about this install will be uploaded.
                   [ SHOW  Gathered info here ]
          
                                               [ OK  ]
```

nbolton:
> two check boxes

That's an idea. I'll try it out and see what it looks like.

src/lib/gui/VersionChecker.cpp
nbolton: Ah, it wasn't because of the Qt bug per se, that was just an interesting side note about `errorString`'s true purpose. You'd have to do this anyway as `errorString` is just a description of the HTTP code (which was deprecated in HTTP/2).

src/lib/gui/dialogs/SettingsDialog.ui
nbolton: I assume you mean remove `<tabstops>.*</tabstops>`

src/lib/common/Settings.cpp
```diff
@@ -229,6 +229,10 @@ QVariant Settings::defaultValue(const QString &key)
     return 120;
   }
 
+  if (key == Gui::JoinPopularityContest) {
+    return false;
```
sithlord48: there is no need to do this anymore just add the new key to the list `m_defaultFalseValues` (in Settings.h)

src/lib/gui/Messages.cpp
```diff
@@ -318,11 +318,20 @@ bool showUpdateCheckOption(QWidget *parent)
       QObject::tr(
           "<p>Would you like to check for updates when %1 starts?</p>"
           "<p>Checking for updates requires an Internet connection.</p>"
-          "<p>URL: <pre>%2</pre></p>"
+          "<p><b>URL:</b>&nbsp;&nbsp;<code>%2</code></p>"
+          "<p>Optional: Participating in the popularity contest sends your %1 version, "
+          "language, and OS name to the %1 API which helps us prioritize bug fixes.</p>"
```
sithlord48: ```"language, and OS name to the %2 API which helps us prioritize bug fixes.</p>"```

src/lib/common/Settings.cpp
nbolton: Oh that's right, I did think about that but immediately forgot about it for some reason!

src/lib/gui/Messages.cpp
nbolton: Ignore this for now, it's still a WIP

sithlord48:
Please explicty show the data to be captured. 

It woudl be nice to see im gonna send this info 

"dfv1.24.0,linux,wayland,x86_64" or something like that  to show exaactly what is sent.

sithlord48 (commented):
Lots of places to use auto . 

Can you add some initial translations for your new strings with google translate and tag the translation team to look at them .

src/lib/gui/VersionChecker.cpp
```diff
@@ -17,20 +17,40 @@
 #include <QRegularExpression>
 #include <climits>
 
+//
+// Free functions
+//
+
+QString getUserAgent()
+{
+  if (!Settings::value(Settings::Gui::JoinPopularityContest).toBool()) {
```
sithlord48: Use QStringLiteral for the QStings your returning.

nbolton:
> Please explicty show the data to be captured.
> 
> It woudl be nice to see im gonna send this info
> 
> "dfv1.24.0,linux,wayland,x86_64" or something like that to show exaactly what is sent.

Nice idea.

> Lots of places to use auto .
> 
> Can you add some initial translations for your new strings with google translate and tag the translation team to look at them .

I'll take a look.

translations/deskflow_es.ts
```diff
@@ -730,6 +722,34 @@ Nombres válidos:
...
+        <translation type="unfinished">Requiere la comprobación de versión. El concurso de popularidad se basa en el user-agent.</translation>
+    </message>
+    <message>
+        <source>&lt;b&gt;User agent:&lt;/b&gt;&amp;nbsp;&amp;nbsp;&lt;code&gt;%1&lt;/code&gt;</source>
+        <translation type="unfinished">&lt;b&gt;User agent:&lt;/b&gt;&amp;nbsp;&amp;nbsp;&lt;code&gt;%1&lt;/code&gt;</translation>
+    </message>
+    <message>
+        <source>&lt;b&gt;URL:&lt;/b&gt;&amp;nbsp;&amp;nbsp;&lt;code&gt;%1&lt;/code&gt;</source>
+        <translation type="unfinished">&lt;b&gt;URL:&lt;/b&gt;&amp;nbsp;&amp;nbsp;&lt;code&gt;%1&lt;/code&gt;</translation>
+    </message>
```
nbolton, line 772: I dug into whether to use "User agent" and "URL" or some native translation. Apparently in the majority of cases and languages, it's common to use the English terms for technical things like this.

translations/deskflow_ru.ts
```diff
@@ -733,6 +725,34 @@ Valid names:
...
+        <source>Participate in popularity contest</source>
+        <translation type="unfinished">Участвовать в конкурсе популярности</translation>
+    </message>
+    <message>
+        <source>Requires the version check. The popularity contest is based on the user agent.</source>
+        <translation type="unfinished">Требуется проверка версии. Конкурс популярности основан на user-agent.</translation>
+    </message>
+    <message>
+        <source>&lt;b&gt;User agent:&lt;/b&gt;&amp;nbsp;&amp;nbsp;&lt;code&gt;%1&lt;/code&gt;</source>
+        <translation type="unfinished">&lt;b&gt;User-agent:&lt;/b&gt;&amp;nbsp;&amp;nbsp;&lt;code&gt;%1&lt;/code&gt;</translation>
```
nbolton, line 770: ... with the exception of Russian which apparently adds a hyphen? 🤷

translations/deskflow_es.ts
sithlord48: Only translate user facing strings. all others should just be literals

translations/deskflow_ru.ts
sithlord48: ? did you change the string and not rebuild ?

translations/deskflow_es.ts
nbolton: Could you rephrase and add more context? These are indeed user facing strings which can be found on the new dialog.

translations/deskflow_ru.ts
nbolton: What do you mean by that? I rebuild and test all translations if that's what you're wondering. My comment was about the Russian translation being an outlier.

sithlord48: OH Well just adjust the translation...  I thought you ment it was showing it in the source string.

translations/deskflow_es.ts
sithlord48: use the google translate strings let the translators worry about that. They will know better than us.

Just make sure your not sending the translated user string to the server.

translations/deskflow_ru.ts
nbolton: Not quite what I meant. In hindsight, my initial observation needed more context. Let's try that again...

@levpr1c Is "User-agent" (with a hyphen) most common in Russian rather than what is used in English which is "User agent" (with no hyphen)?

Edit: I'll assume this is fine, but let me know otherwise.

translations/deskflow_es.ts
nbolton: Or, are you suggesting that we split up this line?

`const auto static userAgentText = QObject::tr("<b>User agent:</b>&nbsp;&nbsp;<code>%1</code>");`

Perhaps it would make more sense as:

```cpp
const auto static userAgentText =
        QStringLiteral("<b>%1</b>&nbsp;&nbsp;<code>%2</code>").arg(QObject::tr("User agent:"));
```

And same for the URL:
```cpp
  auto url = new QLabel(QStringLiteral("<b>%1</b>&nbsp;&nbsp;<code>%2</code>")
                            .arg(QObject::tr("URL:"))
                            .arg(Settings::value(Settings::Gui::UpdateCheckUrl).toString()));
```

sithlord48:
My first issues here got lost in the long long 

It was this. I see some issue with this
1. I can't be in the popularity contest and not have updates on :frowning_face: 
     - This is not ideal.  These are two separate things and i could see a few reasons users may want to send popularity data without checking for updates. The most common being those who have deskflow via a package manager. 
2. The popularity reporting is reported when you do an update check. 
    - This is going to skew our data towards users who restart the app or reboot their machines being more popular. Just think of how popular my (or your) computer config will be if we were to turn on the popularity contest as is.

So what can we do to fix this....
  - Popularity and Online check should be untangled. 
  - We should only report the install once sending the popularity info when the users presses OK on this dialog
  - We need to remember this somehow or fingerprint the install we could use something like https://doc.qt.io/qt-6/qsysinfo.html#machineUniqueId to keep track upstream if an id has reported its version info and with that version info. To be fair idk how i feel about fingerprinting but the alternative of saving in settings can result in duplicate data if the user resets their settings this is the only approach i can see avoid that issue. However with both we can avoid sending data if we don't need as well as filter existing info from being used in our stats. 
  - We report X11 or Wayland, this can change per boot.. this just begs the question of collecting usage data.. Now are users ok with this... So we end up with more a slider of what can we share from NONE - Popular only - minimal stats - full usage. Cause one we find one data point that is useful you start to think about how others could also help.
     - This brings be back to initial concern of usage tracking and reporting at all. For us to use it we need to attach some kind of identifier for the install / machine. This is just more unnecessary breadcrumb that could be potentially used against the reporter by a bad actors. I am aware of how this sounds but all data points can tell you something.. Im not saying we would not sanitize any data we share, but am just saying if we store it its possible someone steals it from our db. This kinda of collection always give me pause its nothing about this patch at all.

translations/deskflow_es.ts
sithlord48: Generally when translating its best to the full string and insert as needed. This way if the language demands the translator can move things around (i.e languages where Actions come before the Subject) . And with that just be sure you add some comments for the translators (`//:` before the tr call) So they know what is going into each replacement (again incase the order has to change). Even allowing translation of the html tags is good incase we use a color is that is wrong for the context (i.e red for bad in Japanese) this again lets the translators. Localize for their language. Another good reason todo this is if they need to adjust the font size to fit it into some area we can't resize.

nbolton:
> * I can't be in the popularity contest and not have updates on ☹️

Fair point. The original idea was just a case of measuring the data that was being sent anyway. Was meant to be a nice little quick add on. Then the privacy discussion happened and things got complicated.

> * The popularity reporting is reported when you do an update check.

The API only records it once and then ignores all further requests from that IP. Obviously not perfect, but I was just aiming for 'good enough'. I get that it's inefficient to keep sending it, so we could certainly have a ~~setting~~ state entry to record that it was sent. That'd need to be reset each time the version number changes (edit: or we record the version number in a `popularityContestLastVersion` state entry), as the version number is measured too. ~~I can see that kind of server/client state management getting a bit complicated.~~ (edit: maybe not if we use the `popularityContestLastVersion` idea)

translations/deskflow_es.ts
nbolton: Ah OK, so leave it as is?
Edit: I'll assume yes.

sithlord48:
Well you need to also be careful w/ using a `state` entry since those files are expected to just be removed sometimes. This is why I was thinking the route of a machineID then we send this w/ and we can then on server side see oh machine ID is has been using 1.24.0 now new version is 1.25.0 and we just say oh we seen machine:version so lets +1 the new version and on the server our check will now be  machine:1.25.0. this way if they downgrade we can also look to track the numbers of users who have downgraded and from what version to what other one.

nbolton:
Ok, that's interesting. So instead of having the state management burden on the client side, we shift it over to the server-side where it's much easier to fix state bugs. 

The trade-off is that we would be sending the request every time the GUI starts. Before I start making changes, are we on the same page?

sithlord48:
No im saying you need to do both.. we don't want to send messages when we don't need to (client check) and we don't want to add messages from the same machine @ version to teh tally (server check)

nbolton:
May reopen something like this in future but just for RFC 9110 user agent format.

PR #8971, opened 2025-09-21
add wayland clipboard support

lxe:
Updated copy of @Mic92's https://github.com/deskflow/deskflow/pull/8648 

> this is currently based on wl-copy/wl-paste.
> One could probably just implement wl-copy/wl-paste without adding much complexity and better performance.

### DIFF
--- a/src/lib/platform/CMakeLists.txt
+++ b/src/lib/platform/CMakeLists.txt
@@ -126,12 +126,16 @@ elseif(UNIX)
 
   if(LIBEI_FOUND)
     list(APPEND PLATFORM_SOURCES
+      EiClipboard.cpp
+      EiClipboard.h
       EiEventQueueBuffer.cpp
       EiEventQueueBuffer.h
       EiKeyState.cpp
       EiKeyState.h
       EiScreen.cpp
       EiScreen.h
+      WaylandClipboard.cpp
+      WaylandClipboard.h
     )
     # The Portal sources also require EI.
     if(LIBPORTAL_FOUND)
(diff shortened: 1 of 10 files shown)


sithlord48 (changes requested):
This should be redone to not require the wl-clipboard binaries. This as discussed in #8648

lxe:
OK, I was actually thinking of stealing the code directly from WL clipboard

sithlord48:
Many of these files also do not end in an empty new line can you please update the files so all end with an empty new line.

sithlord48:
As i said in the other Mr instead of that this should attempt to use the [xdg-clipboard-portal](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.impl.portal.Clipboard.html) . But working even with the wl-roots protocol is a start.. Once this is moved over to a one of those ill test it again.

KoljaFrahm:
Can we not maybe think about using this as an intermediate solution until https://github.com/flatpak/xdg-desktop-portal/pull/1803 is finished? I've been following the discussion for a over year now and it seems that the progress is very slow. (as you said yourself @sithlord48) 
This seems like a good intermediate solution, and there seems to be interest in that by several people.
Now on some distributions there is no x11 anymore and we urgently need a solution for this.

sithlord48:
There has been no response the the review comments I left so without that and it being updated no i can't see how we could land it

KoljaFrahm:
@lxe You think you can quickly implement the review comments from https://github.com/deskflow/deskflow/pull/8648 and rebase?
How I see it, it's just adding a line add the end of two files and changing one type
Then we can go on and finally use wayland.

sithlord48:
superseeded by #9133

nbolton:
> superseeded by #8648

Are you sure it's not the other way round? To me it appears that the current PR supersedes #8648 (rather than being superseded by it).

Edit: Ah, I see you meant #9133

KoljaFrahm:
I think @sithlord48 meant https://github.com/deskflow/deskflow/pull/9133

PR #9001, opened 2025-10-05
Core process cleanup make quoted args

sithlord48:
- Make the function `makeQuotedArgs` a private member of `CoreProcess` 
 - Simplify the method ` CoreProcess::makeQuotedArgs` by using `QString::replaceInStrings` instead of a for loop.


src/lib/gui/core/CoreProcess.cpp
```diff
@@ -77,22 +77,12 @@ QString processStateToString(CoreProcess::ProcessState state)
...
-    if (arg.contains(' ')) {
-      quoted << QString("\"%1\"").arg(arg);
-    } else {
-      quoted << arg;
-    }
-  }
-
-  return quoted.join(" ");
+  QStringList command = {args};
+  command.replaceInStrings("'", "\"");
```
nbolton: It appears that this new code is doing something totally different.

Before: `C:\Program Files` would be converted to `"C:\Program Files"`

After: `C:\Program Files` remains `C:\Program Files` and Windows won't be able to understand it, as everything after the 1st space will be an arg.

Can you see anywhere that we use `'` char for args?

sithlord48: Ah yes yes .. hold on..

src/lib/gui/core/CoreProcess.h
```diff
@@ -128,6 +128,7 @@ private Q_SLOTS:
   QString correctedAddress() const;
   QString requestDaemonLogPath();
   void persistLogDir() const;
+  QString makeQuotedArgs(const QString &app, const QStringList &args) const;
```
nbolton: In hindsight, we could make this static.
```suggestion
  static QString makeQuotedArgs(const QString &app, const QStringList &args);
```

sithlord48: yeah i was thinking that as well
Edit: done

sithlord48:
Superseeded by #9023

PR #9123, opened 2025-11-05
Adjust OS X MouseMove handling to use live cursor position.

brett19:
I wrote this patch many years ago so the comments / PR title are based on my recollection of why this is necessary.  The demonstration of the original issue consisted of an OS X Server connected to a Windows 11 Client, you could then center your cursor on the remote system (Windows) and start performing perfect(ish) circles physically with the mouse.  What you would observe is that on the remote system, while the circles would be circular-ish, your cursor would slowly drift away towards one side of the screen, demonstrating that the mouse position information was not being collected accurately.

### DIFF
--- a/src/lib/platform/OSXScreen.mm
+++ b/src/lib/platform/OSXScreen.mm
@@ -913,8 +913,17 @@ void OSXScreen::handleSystemEvent(const Event &event)
   }
 }
 
-bool OSXScreen::onMouseMove(CGFloat mx, CGFloat my)
+bool OSXScreen::onMouseMove()
 {
+  // when we receive a mouse-move event, it is possible it was queued for a period
+  // and that the mouse has already moved again since then.  to handle this, we need
+  // to query the current mouse position rather than using the position in the event.
+  CGEventRef event = CGEventCreate(NULL);
+  CGPoint pos = CGEventGetLocation(event);
+  CFRelease(event);
+  CGFloat mx = pos.x;
+  CGFloat my = pos.y;
+
   LOG_DEBUG2("mouse move %+f,%+f", mx, my);
 
   CGFloat x = mx - m_xCursor;
@@ -1613,7 +1622,6 @@ OSXScreen::handleCGInputEventSecondary(CGEventTapProxy proxy, CGEventType type, 
 CGEventRef OSXScreen::handleCGInputEvent(CGEventTapProxy proxy, CGEventType type, CGEventRef event, void *refcon)
 {
   OSXScreen *screen = (OSXScreen *)refcon;
-  CGPoint pos;
 
   switch (type) {
   case kCGEventLeftMouseDown:
@@ -1630,8 +1638,9 @@ CGEventRef OSXScreen::handleCGInputEvent(CGEventTapProxy proxy, CGEventType type
   case kCGEventRightMouseDragged:
   case kCGEventOtherMouseDragged:
   case kCGEventMouseMoved:
-    pos = CGEventGetLocation(event);
-    screen->onMouseMove(pos.x, pos.y);
+    // we intentionally ignore the position in the event here as the events are
+    // queued and will no longer be accurate when we process them.
+    screen->onMouseMove();
 
     // The system ignores our cursor-centering calls if
     // we don't return the event. This should be harmless,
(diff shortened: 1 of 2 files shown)


nbolton:
@brett19 Hey, looks like you have some conflicts. Would you mind resolving those?

brett19:
Hey @nbolton, I've updated the PR.  Is there any chance there is someone around who has some more experience dealing with OS X who might be able to verify the logic of the change?  It certainly fixes the issue I was encountering on my local system, but I'm not entirely convinced that the reasoning behind why it fixes it is sound.

src/lib/platform/OSXScreen.mm
```diff
@@ -915,6 +915,15 @@
 
 bool OSXScreen::onMouseMove(CGFloat mx, CGFloat my)
 {
+  // when we receive a mouse-move event, it is possible it was queued for a period
+  // and that the mouse has already moved again since then.  to handle this, we need
+  // to query the current mouse position rather than using the position in the event.
+  CGEventRef event = CGEventCreate(NULL);
+  CGPoint pos = CGEventGetLocation(event);
```
nbolton, line 922: I wonder if it makes sense to move this up the stack by one? We're already calling `CGEventGetLocation` on L1642 (L1633 on master) though with a different event.

https://github.com/deskflow/deskflow/blob/0a67f63af6bba4d1cfefa7befc6e0e104ec6cdda/src/lib/platform/OSXScreen.mm#L1632-L1634

CC @feschber who probably has these functions paged in.

feschber: The whole relative motion computation always depends on how the cursor is warped. I dont know what deskflow is doing here exactly so I would have to read into this a bit more, first.

This might be relevant:
https://stackoverflow.com/questions/40904830/constrain-mouse-movement-to-region-in-os-x-without-jitter/40922095#40922095

brett19: Ahh.  This does appear to confirm my old theory.  As a side-note though, it's pretty funny how complicated that SO post seems to imply that OS X has made handling the mouse :D

@nbolton: It probably does make sense to omit the `pos = CGEventGetLocation(event);` and passing the event down to onMouseMove if it's just going to be ignored.  It could also more sense to move my little 'event recreation' block into the top level function, and pass its x,y to onMouseMove instead.  It's not clear to me which one would be "cleaner".

brett19: I've adjusted the PR to remove some of the unused code.

feschber: > As a side-note though, it's pretty funny how complicated that SO post seems to imply that OS X has made handling the mouse :D

Yes... it's by far the most difficult API to work with.

brett19:
I've rebased the PR onto the latest master along with some cleanup to the original.  It looks like the CI check will need approval due to coming from a fork.

nbolton (approved):
LGTM, I tried reproducing the drift but not sure if I'm seeing things or whether the mouse cursor was drifting. In any case it doesn't seem to have any negative effects!

nbolton (changes requested):
Oh, actually, could you change your commit message to the [conventional commit format](https://www.conventionalcommits.org/en/v1.0.0/)?

sithlord48 (changes requested):
Other then the lint and commit message should be `fix: ....`  looks go to me

sithlord48:
I've updated the commit message and linting so it should be good now @nbolton

sithlord48 (approved):

nbolton (approved):

brett19:
@nbolton When can I expect this change to make it into an upstream Synergy release?

nbolton:
> upstream

Downstream ;-)

I've added this to the backports plan for the next update. We are aiming for the coming weeks.

brett19:
Hey @nbolton, any updates on this?  I haven't seen anything to imply this is included in the latest changelogs.

nbolton:
> Hey @nbolton, any updates on this? I haven't seen anything to imply this is included in the latest changelogs.

Decided to put it in the next beta: https://github.com/symless/synergy/pull/190

PR #9312, opened 2025-12-31
fix(macos): prevent automatic termination during memory pressure

### LINKED ISSUE #7329: Random reconnects with Windows server and macOS client
### Operating Systems ###

Server: Windows 11 23H2
Client: macOS Sonoma 14.2.1

### Synergy Version ###

1.15.0 on both sides (also tried with 1.14.6.19-stable)

### Steps to Reproduce ###

There is no exact steps to reproduce this. I'm just controlling the client and in random intervals it disconnects, the cursor jumps back to the server, and the client reconnects.

### Log ###

There are no errors or anything special in the logs, it just prints the following:

```
[2024-01-16T09:42:01] INFO: process exited normally
[2024-01-16T09:42:01] INFO: detected process not running, auto restarting
[2024-01-16T09:42:02] INFO: starting client
[2024-01-16T09:42:02] INFO: log level: INFO
[2024-01-16T09:42:02] NOTE: started client
```

### Extra Info ###

I've take a step further and gathered the logs from the Console.app on macOS. I don't know if it helps or not, but this is the whole log which I get when the reconnect happens: https://pastebin.com/U6tqYCen.

johnweldon:
Disable macOS Automatic Termination to prevent the server/client from being killed during memory pressure events. Adds NSSupportsAutomaticTermination to Info.plist and programmatically disables termination in the Cocoa app.

Fixes #7329

🤖 Generated with [Claude Code](https://claude.com/claude-code) (But _not_ AI Trash)

Edit: (sithlord48) 
 - Updated to remove code. We only need plist or code change  
 - Added a commit to put deskflow in the "Utilities" category

### DIFF
--- a/src/apps/res/deskflow.plist.in
+++ b/src/apps/res/deskflow.plist.in
@@ -31,7 +31,9 @@
 	<key>NSHumanReadableCopyright</key>
 	<string>@BUNDLE_COPYRIGHT@</string>
 
-	<!-- Don't Allow Os to kill for memory -->
+	<!-- Prevent macOS from terminating during memory pressure -->
+	<key>NSSupportsAutomaticTermination</key>
+	<false/>
 	<key>NSSupportsSuddenTermination</key>
 	<false/>
 
--- a/src/apps/res/deskflow.plist.in
+++ b/src/apps/res/deskflow.plist.in
@@ -30,6 +30,8 @@
 	<true/>
 	<key>NSHumanReadableCopyright</key>
 	<string>@BUNDLE_COPYRIGHT@</string>
+	<key>LSApplicationCategoryType</key>
+	<string>public.app-category.utilities</string>
 
 	<!-- Prevent macOS from terminating during memory pressure -->
 	<key>NSSupportsAutomaticTermination</key>


sithlord48 (changes requested):
Looks like it should not break anything

Commit message is wrong , should be `fix: macos ...`

src/lib/platform/OSXCocoaApp.m
```diff
@@ -21,6 +21,12 @@ void runCocoaApp()
 
   [NSApplication sharedApplication];
 
+  // Prevent macOS from killing the process during memory pressure.
```
sithlord48: In #7329  you say it add the line to the plist OR make add this other call.
Since we have added the line to the plist we don't need to also add this code change so please remove it

sithlord48 (changes requested):

sithlord48:
> 🤖 Generated with [Claude Code](https://claude.com/claude-code) (But not AI Trash)

I don't really care for little stuff like this that is gonna be a single line in the plist. I have issues when people are submitting patches that are large "AI" generated code they do not understand and also do not fully test.

sithlord48 (approved):
@nbolton  This looks good to me now can you approve / land when ready. (I made a change so i can't anymore)
@johnweldon  Ill keep this updated from here on out.

nbolton (approved):
Worth a shot! 🤷‍♂️

nbolton:
@sithlord48 I'm wondering if this change is actually effective. According to the docs, the detail on [`NSSupportsAutomaticTermination`](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html) mentions memory ("reclaim the memory used by the app") but as the rationale for terminating an already-idle app, not as the trigger.

> NSSupportsAutomaticTermination (Boolean - macOS). This key contains a Boolean value that indicates whether the app supports automatic termination in OS X v10.7 and later. Automatic termination allows an app that is running to be terminated automatically by the system when certain conditions apply. Primarily, the app can be terminated when it is hidden or does not have any visible windows and is not currently being used. The system may terminate such an app in order to reclaim the memory used by the app.
>
> An app may programmatically disable and reenable automatic termination support using the disableAutomaticTermination and enableAutomaticTermination methods of NSProcessInfo. The app might do this to prevent being terminated during a critical operation.

Memory-pressure terminations on macOS are actually handled by jetsam/`memorystatus` in the macOS kernel:
https://developer.apple.com/documentation/xcode/identifying-high-memory-use-with-jetsam-event-reports

AFAIK the kernel doesn't change it's behavior based on `.plist` files.AFAIK the kernel doesn't change it's behavior based on `.plist` files.

johnweldon:
FWIW, the issue has not re-occurred for me since this change.

nbolton:
> FWIW, the issue has not re-occurred for me since this change.

Good to know, thanks for the info. What's memory use like after the Core process has been running a few days?

johnweldon:
@nbolton -- the short answer: it's stable. After ~13d 18h continuous
uptime on Synergy 1.20.0, the Core process is sitting at ~18.7 MB RSS.

Snapshot from Activity Monitor on this host today:

| Process                 | PID  | RSS     | Threads | CPU time |
| ----------------------- | ---- | ------- | ------- | -------- |
| `synergy` (GUI)         | 3623 | 37.3 MB | 6       | 1:22.85  |
| `synergy-server` (Core) | 3710 | 18.7 MB | 11      | 39:01.89 |

Host: macOS 26.4.1 (build 25E253), Darwin 25.4.0, arm64 (Apple Silicon
/ M2 Air), Synergy 1.20.0 from `/Applications/Synergy.app`.

Across multi-week sessions both before and after this PR I've seen the
Core process stay in roughly that 15-20 MB range. CPU is ~1% on Core
during active use, 0% when idle.

Thank you for all your work on this product... I've used it
continuously for what feels like at least ten or twelve years

nbolton:
> Synergy 1.20.0, the Core process is sitting at ~18.7 MB RSS.

What about with Deskflow?

johnweldon:
> What about with Deskflow?

@nbolton -- apologies, I muddled the two. The numbers I posted are from Symless's commercial Synergy 1.20.0 build (`/Applications/Synergy.app`), which is what I actually run on this machine day-to-day; I don't have a Deskflow build installed, so that data isn't apples-to-apples for what you're asking about. Feel free to disregard the earlier snapshot.

nbolton:
> > What about with Deskflow?
> 
> @nbolton -- apologies, I muddled the two. The numbers I posted are from Symless's commercial Synergy 1.20.0 build (`/Applications/Synergy.app`), which is what I actually run on this machine day-to-day; I don't have a Deskflow build installed, so that data isn't apples-to-apples for what you're asking about. Feel free to disregard the earlier snapshot.

I'm the maintainer of the Synergy 1 commercial downstream code. Changes from Deskflow flow down into Synergy over time, so if it's not happening in Synergy but is happening in Deskflow, then this is useful information. Would you be able to try Deskflow? Deskflow and Synergy 1 are protocol compatible so you'd only need to update your Mac.

PR #9350, opened 2026-01-17
Feat: Implement Toggle Screen Action (#9135)

### LINKED ISSUE #9135: Add Toggle Screen option
### Feature Request

This option is similar to the option from inputLeap, as seen in the image below

Image

Basically it's about being able to switch between 2 or more screens (in the exact order specified in the list)

### Voting Instructions

Ashutosh0x:
fixes: #9135 

Implemented the Toggle Screen action as requested in #9135. Allows cycling through a list of screens or all screens.

### DIFF
--- a/src/lib/gui/dialogs/ActionDialog.cpp
+++ b/src/lib/gui/dialogs/ActionDialog.cpp
@@ -28,6 +28,7 @@ ActionDialog::ActionDialog(QWidget *parent, const ServerConfig &config, Hotkey &
       ui->comboActionType, QOverload<int>::of(&QComboBox::currentIndexChanged), this, &ActionDialog::actionTypeChanged
   );
   connect(ui->listScreens, &QListWidget::itemChanged, this, &ActionDialog::itemToggled);
+  connect(ui->listToggleScreenOrder, &QListWidget::itemChanged, this, &ActionDialog::itemToggled);
   connect(ui->buttonBox, &QDialogButtonBox::accepted, this, &ActionDialog::accept);
   connect(ui->buttonBox, &QDialogButtonBox::rejected, this, &ActionDialog::reject);
 
@@ -51,6 +52,13 @@ ActionDialog::ActionDialog(QWidget *parent, const ServerConfig &config, Hotkey &
 
     ui->listScreens->addItem(newListItem);
 
+    auto *toggleListItem = new QListWidgetItem(screen.name());
+    toggleListItem->setCheckState(Qt::Checked);
+    if ((m_action.typeScreenNames().indexOf(screen.name()) == -1) &&
+        (m_action.haveScreens() && !m_action.typeScreenNames().isEmpty()))
+      toggleListItem->setCheckState(Qt::Unchecked);
+    ui->listToggleScreenOrder->addItem(toggleListItem);
+
     ui->comboSwitchToScreen->addItem(tr("Switch to %1").arg(screen.name()));
     if (screen.name() == m_action.switchScreenName())
       ui->comboSwitchToScreen->setCurrentIndex(ui->comboSwitchToScreen->count() - 1);
@@ -58,6 +66,7 @@ ActionDialog::ActionDialog(QWidget *parent, const ServerConfig &config, Hotkey &
 
   ui->keySequenceWidget->setVisible(false);
   ui->groupScreens->setVisible(false);
+  ui->groupToggleScreenOrder->setVisible(false);
   ui->listScreens->setEnabled(!ui->keySequenceWidget->keySequence().isMouseButton());
   ui->comboSwitchToScreen->setVisible(false);
   ui->comboSwitchInDirection->setVisible(false);
@@ -76,10 +85,13 @@ void ActionDialog::accept()
 
   m_action.typeScreenNames().clear();
 
-  int screenCount = ui->listScreens->count();
+  QListWidget *sourceList =
+      isToggleScreenAction(ui->comboActionType->currentIndex()) ? ui->listToggleScreenOrder : ui->listScreens;
 
-  for (int i = 0; i < ui->listScreens->count(); i++) {
-    const auto &item = ui->listScreens->item(i);
+  int screenCount = sourceList->count();
+
+  for (int i = 0; i < sourceList->count(); i++) {
+    const auto &item = sourceList->item(i);
     m_action.typeScreenNames().append(item->text());
     if (item->checkState() == Qt::Unchecked) {
       screenCount--;
@@ -87,7 +99,7 @@ void ActionDialog::accept()
     }
   }
 
-  if (screenCount == ui->listScreens->count())
+  if (screenCount == sourceList->count())
     m_action.typeScreenNames().clear();
 
   m_action.setHaveScreens(screenCount);
@@ -105,7 +117,7 @@ void ActionDialog::updateSize()
 {
   setMaximumSize(QSize(16777215, 1677215));
   adjustSize();
-  if (!isKeyAction(ui->comboActionType->currentIndex()))
+  if (!isKeyAction(ui->comboActionType->currentIndex()) && !isToggleScreenAction(ui->comboActionType->currentIndex()))
     setMaximumSize(size());
 }
 
@@ -125,7 +137,13 @@ void ActionDialog::actionTypeChanged(int index)
   ui->buttonBox->button(QDialogButtonBox::Ok)->setEnabled(canSave());
   ui->keySequenceWidget->setVisible(isKeyAction(index));
   ui->groupScreens->setVisible(isKeyAction(index));
-  ui->listScreens->setEnabled(!ui->keySequenceWidget->keySequence().isMouseButton());
+  ui->groupToggleScreenOrder->setVisible(isToggleScreenAction(index));
+
+  if (isKeyAction(index)) {
+    ui->listScreens->setEnabled(!ui->keySequenceWidget->keySequence().isMouseButton());
+  } else {
+    ui->listScreens->setEnabled(true);
+  }
   ui->comboSwitchToScreen->setVisible(index == ActionTypes::SwitchTo);
   ui->comboSwitchInDirection->setVisible(index == ActionTypes::SwitchInDirection);
   ui->comboLockCursorToScreen->setVisible(index == ActionTypes::ModifyCursorLock);
@@ -135,6 +153,11 @@ void ActionDialog::actionTypeChanged(int index)
 bool ActionDialog::isKeyAction(int index) const
 {
   return ((index == ActionTypes::PressKey) || (index == ActionTypes::ReleaseKey) || (index == ActionTypes::ToggleKey));
+}
+
+bool ActionDialog::isToggleScreenAction(int index) const
+{
+  return (index == ActionTypes::ToggleScreen);
 }
 
 bool ActionDialog::canSave() const
@@ -148,6 +171,17 @@ bool ActionDialog::canSave() const
     }
     return (!ui->keySequenceWidget->keySequence().toString().isEmpty() && (totalChecked > 0));
   }
+
+  if (isToggleScreenAction(ui->comboActionType->currentIndex())) {
+    const QList<QListWidgetItem *> items = ui->listToggleScreenOrder->findItems("*", Qt::MatchWildcard);
+    int totalChecked = 0;
+    for (const auto &item : items) {
+      if (item->checkState() == Qt::Checked)
+        totalChecked++;
+    }
+    return (totalChecked > 0);
+  }
+
   return true;
 }
 
--- a/src/lib/gui/dialogs/ActionDialog.h
+++ b/src/lib/gui/dialogs/ActionDialog.h
@@ -35,12 +35,13 @@ public:
     inline static const auto SwitchToNextScreen = 5;
     inline static const auto ModifyCursorLock = 6;
     inline static const auto RestartServer = 7;
+    inline static const auto ToggleScreen = 8;
   };
 
   ActionDialog(QWidget *parent, const ServerConfig &config, Hotkey &hotkey, Action &action);
   ~ActionDialog() override;
 
-protected Q_SLOTS:
+protected:
   void accept() override;
 
 private:
@@ -49,6 +50,7 @@ private:
   void itemToggled() const;
   void actionTypeChanged(int index);
   bool isKeyAction(int index) const;
+  bool isToggleScreenAction(int index) const;
   bool canSave() const;
 
   std::unique_ptr<Ui::ActionDialog> ui;
--- a/src/lib/gui/dialogs/ActionDialog.ui
+++ b/src/lib/gui/dialogs/ActionDialog.ui
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
+﻿<?xml version="1.0" encoding="UTF-8"?>
 <ui version="4.0">
  <class>ActionDialog</class>
  <widget class="QDialog" name="ActionDialog">
@@ -78,6 +78,11 @@
          <string>Restart the server</string>
         </property>
        </item>
+       <item>
+        <property name="text">
+         <string>Toggle screen</string>
+        </property>
+       </item>
       </widget>
      </item>
     </layout>
@@ -151,7 +156,7 @@
       <sizepolicy hsizetype="Preferred" vsizetype="Preferred">
        <horstretch>0</horstretch>
        <verstretch>0</verstretch>
-      </sizepolicy>
+       </sizepolicy>
      </property>
      <property name="title">
       <string>Computers to receive this event</string>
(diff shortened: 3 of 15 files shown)


sithlord48 (changes requested):
Seems to have alot of unrelated changes. 

Do not fix commit with more commits.

Ashutosh0x:
@sithlord48 You're absolutely right - I apologize for the unrelated commits. I accidentally branched from the wrong base which included my Wayland clipboard work from PR #9348.

I've cleaned up the PR by force-pushing a new branch based on upstream master with only the Toggle Screen Action commit. The PR now contains only the changes relevant to #9135.

Thank you for catching this!

sithlord48 (commented):
Have not built yet .but i see your using some non Qt Types . Use the Qt Types when and where you can.

src/lib/gui/dialogs/ActionDialog.h
```diff
@@ -35,6 +35,7 @@ class ActionDialog : public QDialog
     inline static const auto SwitchToNextScreen = 5;
     inline static const auto ModifyCursorLock = 6;
     inline static const auto RestartServer = 7;
+    inline static const auto ToggleScreen = 11;
```
sithlord48: Why 11 ?

src/lib/server/Config.cpp
```diff
@@ -1072,6 +1072,20 @@ void Config::parseAction(
     action = new InputFilter::SwitchToNextScreenAction(m_events);
   }
 
+  else if (name == "toggleScreen") {
+    std::vector<std::string> screens;
```
sithlord48, line 1076: Where possible use Qt types. we perfer them over the std since they are generally eaiser to use and just as portable (insome cases more so)

sithlord48 (changes requested):

Ashutosh0x:
@sithlord48
1. **ActionDialog.h**: Fixed the magic number 11. It now uses static_cast<int>(Action::Type::toggleScreen).
2. **Config.cpp types**: I kept std::vector/std::string here because InputFilter::ToggleScreenAction (in src/lib/server/InputFilter.h) takes std::vector<std::string> in its constructor. Using Qt types in Config.cpp would require converting them back to std types immediately to pass them to InputFilter classes, which seems inefficient and inconsistent with the existing server library code style.

src/lib/gui/dialogs/ActionDialog.h
sithlord48: > ActionDialog.h: Fixed the magic number 11. It now uses static_cast(Action::Type::toggleScreen).

sithlord48 (approved):
Unable to fully test this Code seams reasonable. 
@nbolton  can you test this please a bit more before we land it.

sithlord48 (changes requested):
Fix the linting please. 
I recommend  using a precommit git hook see our clang format page for an example script near the bottom https://github.com/deskflow/deskflow/wiki/Clang%E2%80%90Format-Tips-%26-Tricks

Ashutosh0x:
@sithlord48 Applied clang-format==20.1.0 to the modified files. Thanks for the tip on versioning!

sithlord48:
yeah is annoying that the version of clang-format matters :(

sithlord48 (changes requested):
Please fix the lint issues.

Ashutosh0x:
@sithlord48 Applied clang-format 20.1.0 to *all* modified files (previously missed Action.cpp, Server.cpp, etc - only formatted Dialog/Config). Should pass now.

Ashutosh0x:
@sithlord48 Enhanced ActionDialog.h to use Action::Type enums for *all* action types, not just ToggleScreen, to be consistent and avoid magic numbers entirely.

src/lib/gui/dialogs/ActionDialog.h
```diff
@@ -25,33 +25,34 @@ class ActionDialog : public QDialog
...
-    inline static const auto SwitchTo = 3;
-    inline static const auto SwitchInDirection = 4;
-    inline static const auto SwitchToNextScreen = 5;
-    inline static const auto ModifyCursorLock = 6;
-    inline static const auto RestartServer = 7;
-  };
-
-  ActionDialog(QWidget *parent, const ServerConfig &config, Hotkey &hotkey, Action &action);
-  ~ActionDialog() override;
+  inline static const auto PressKey = static_cast<int>(Action::Type::keyDown);
```
sithlord48: Great idea but please do this in a ` refactor` commit .

sithlord48 (changes requested):

Ashutosh0x:
@sithlord48 Per your request, I have split the changes into two commits:
1. 
efactor: use Action::Type enums... - Updates existing constants to use enums.
2. eat: implement toggle screen action - Adds the new feature code.

This keeps the refactoring separate from the new logic.

src/lib/gui/dialogs/ActionDialog.h
```diff
@@ -9,10 +9,10 @@
 
 #pragma once
 
+#include "../Action.h"
```
sithlord48: Bad include style 

use `#include "gui/Action.h"` please

src/lib/gui/dialogs/ActionDialog.h
```diff
@@ -25,33 +25,33 @@ class ActionDialog : public QDialog
...
-    inline static const auto SwitchInDirection = 4;
-    inline static const auto SwitchToNextScreen = 5;
-    inline static const auto ModifyCursorLock = 6;
-    inline static const auto RestartServer = 7;
-  };
-
-  ActionDialog(QWidget *parent, const ServerConfig &config, Hotkey &hotkey, Action &action);
-  ~ActionDialog() override;
+  inline static const auto PressKey = static_cast<int>(Action::Type::keyDown);
+  inline static const auto ReleaseKey = static_cast<int>(Action::Type::keyUp);
```
sithlord48: These should be in a namespace of ActionTypes  or put them back into the struct. We have a few similar namespaces help alot here.

sithlord48 (changes requested):
Im just gonna make this a draft until you fix the feature commit.. this is a bunch of trash you added on accident.. when you have fixed it you can mark this ready for review again.

Ashutosh0x:
Future improvements roadmap (Post-Merge):

1. **Refactor Action Class**: Store `m_type` internally as `Action::Type` instead of `int` for type safety.
2. **Unit Tests**: Add test cases for `ToggleScreen` with 0, 1, and N screens.
3. **Tooling**: Add `git pre-commit` hook for `clang-format`.

sithlord48:
Pick one of your draft prs and focus on making that work.. Fix this feature commit please..

src/lib/gui/dialogs/ActionDialog.h
```diff
@@ -33,25 +34,26 @@ class ActionDialog : public QDialog
...
-ActionDialog(QWidget *parent, const ServerConfig &config, Hotkey &hotkey, Action &action);
-~ActionDialog() override;
+  ActionDialog(QWidget *parent, const ServerConfig &config, Hotkey &hotkey, Action &action);
+  ~ActionDialog() override;
 
 protected Q_SLOTS:
-void accept() override;
+  void accept() override;
+
+private Q_SLOTS:
```
sithlord48: there is no reason for these to be Q_SLOTS normal private methods are fine here.

sithlord48 (changes requested):
Small change ill try to do a proper test sometime soon. besure you rebase to the current code. as a few prs have landed.

Ashutosh0x:
@sithlord48 I've refactored `ActionDialog` to remove the unnecessary `Q_SLOTS` as suggested. I also moved the action type constants into a dedicated `ActionTypes` namespace and applied `clang-format==20.1.0` to ensure everything passes the lint checks.

Ashutosh0x:
@sithlord48 I've refactored `ActionDialog` to remove the unnecessary `Q_SLOTS` as suggested. I also moved the action type constants into a dedicated `ActionTypes` namespace and applied `clang-format==20.1.0`. I've also squashed the commit history into a single clean commit.

sithlord48 (approved):
The toggle screen action is not visiable in list for me to pick , Did you forget to add it to the list ?

sithlord48 (changes requested):
The new action is missing from the list of actions. did you forget to add it ? 
Be sure you are checking locally before submiting

Ashutosh0x:
@sithlord48 Good catch! I've added the missing "Toggle screen" entry to the `comboActionType` dropdown in `ActionDialog.ui`. The dropdown was missing entries for the mouse actions and toggle screen action - they should all be visible now.

src/lib/gui/dialogs/ActionDialog.ui
```diff
@@ -78,6 +78,26 @@
          <string>Restart the server</string>
         </property>
        </item>
+       <item>
+        <property name="text">
+         <string>Press mouse button</string>
```
sithlord48: Why are these added? these should just be a keypress.

sithlord48 (changes requested):

sithlord48:
please comment on review issues directly. 

The list needs the title changes from "screens to receive the event" to a list that allows you to change the order of the items..

Ashutosh0x:
You're right - I added the mouse actions because they exist in the `Action::Type` enum (indices 8-10) and were causing the index mismatch. I'll remove them since they're not needed for this PR.

For the toggle screen list UI, I understand now - it needs to be reorderable so users can set the cycling order. I'll update the `groupScreens` to use a list widget with drag-and-drop reordering enabled.

Ashutosh0x:
@sithlord48 I've addressed your feedback:

1. **Removed mouse button actions** - They've been removed from the dropdown as you requested
2. **Updated toggle screen list** - Changed the title to "Toggle screen order (drag to reorder)" and enabled drag-drop reordering (`InternalMove` mode) so users can set the cycling order

The list now supports dragging items to reorder them, which will determine the toggle sequence.

Ashutosh0x:
I'm investigating why the list isn't showing. The toggle screen list should be visible when the action is selected. Let me check if it's a visibility logic issue or if the list needs screens to be configured first. Will fix and push shortly.

Ashutosh0x:
Fixed! The issue was an index mismatch - the dropdown has 9 items (indices 0-8) but `toggleScreen` is at index 11 in the `Action::Type` enum. When you selected "Toggle screen" from the dropdown (index 8), it didn't match `ActionTypes::ToggleScreen` (11).

Changed `isToggleScreenAction()` to check for dropdown index 8 instead of the enum value. The toggle screen list should now be visible when the action is selected.

Force-pushed SHA da5cd56.

Ashutosh0x:
Removed `ActionTypes::ToggleScreen` from the namespace since it was misleading (enum value is 11 but dropdown index is 8). The function now just uses the hardcoded value 8 with a comment explaining why.

Ashutosh0x:
Added checkboxes to the toggle screen list! Now users can:
- **Check/uncheck** screens to select which ones to include in the toggle cycle
- **Drag-drop** to reorder the cycling sequence

This allows for custom toggle sets like:
- F5 = toggle between client 1 + 5 (only those 2 checked)
- F6 = toggle between client 5 + 3 + 9 (only those 3 checked and ordered)

The `accept()` method now saves data from the correct list based on action type (toggle screen uses `listToggleScreenOrder`, other actions use `listScreens`).

Force-pushed SHA 7646cf0.

Ashutosh0x:
Fixed the save bug! The issue was that `accept()` was saving the dropdown index (8) directly as the action type, but `Action::Type::toggleScreen` is at enum index 11. This caused it to be saved as `mouseDown` (enum index 8).

Added mapping logic: when dropdown index is 8, it now correctly maps to `Action::Type::toggleScreen` (enum value 11) before saving.

Force-pushed with the fix.

Ashutosh0x:
Updated `canSave()` to validate toggle screen actions - now requires at least one screen to be checked in the toggle screen list, similar to how key actions are validated. This prevents saving an empty toggle screen action.

Ashutosh0x:
Simplified the implementation by reordering the `Action::Type` enum! Moved mouse actions (mouseDown, mouseUp, mousebutton) to the end, after toggleScreen. 

This makes dropdown indices match enum values perfectly, eliminating all the complex mapping logic:
- Removed index mapping in `accept()`
- Simplified `isToggleScreenAction()` to use `ActionTypes::ToggleScreen`
- Re-added `ActionTypes::ToggleScreen` constant (now it's correct)

Much cleaner solution as suggested by @sithlord48!

Ashutosh0x:
Changed `#include "../Action.h"` to `#include "gui/Action.h"` as requested.

Checked `m_actionTypeNames` in `ActionDialog.cpp` and it uses the existing action list from `Action` class via `names()`. Since we moved the mouse actions to the end of the `Action::Type` enum, the existing code that maps enum to string names should work correctly assuming it iterates over the enum or is updated accordingly. Note: I'll double check `Action.cpp` or wherever `names()` is defined to ensure the string list matches the new enum order.

Ashutosh0x:
Fixed! I've updated `m_actionTypeNames` in `Action.h` to match the new enum order:
- Moved mouse action strings ("mouseDown", "mouseUp", "mousebutton") to the end
- "toggleScreen" is now at index 8, after "restartServer"

This ensures the string mapping is consistent with the enum values and UI indices.

Ashutosh0x:
Reverted the `ActionTypes` namespace change and moved the constants back into the `ActionDialog` class as static members, as requested. This restores the previous structure while keeping the new functionality.

Ashutosh0x:
Refactored `ActionDialog` to use forward declaration for `Action` in the header, removing the `#include "gui/Action.h"` dependency.

- Moved static constant definitions to `ActionDialog.cpp`
- `ActionDialog.h` now only contains declarations: `static const int PressKey;`
- This ensures clean separation and faster compilation.

Ashutosh0x:
Restored `ActionTypes` as a `struct` to group the constants.

- **ActionDialog.h**: Defines `struct ActionTypes` with static member declarations (keeps header lightweight without `Action.h` include).
- **ActionDialog.cpp**: Defines the constants using `Action::Type`.
- **Usage**: Updated code to use `ActionTypes::Constant`.

Also, Regarding translated strings (.ts files):
I don't have the build environment to generate/update the `.ts` files locally. Since these need to be updated via `lupdate` or build target, can you please handle the translation update on your end?

Ashutosh0x:
That sounds perfect! Thank you for handling the translation updates and the final polish. I really appreciate your guidance and improvements on this PR. 

Standing by if you need anything else!

sithlord48 (dismissed):
Locally tested seams to work

sithlord48 (dismissed):
Quick smoke test.
 1. Runs. 
 2. Saves info 
 3. doesn't crash 
 4. need more clients to test better
 @nbolton ill leave this for you to land

src/lib/gui/dialogs/ActionDialog.cpp
```diff
@@ -51,13 +52,22 @@ ActionDialog::ActionDialog(QWidget *parent, const ServerConfig &config, Hotkey &
 
     ui->listScreens->addItem(newListItem);
 
+    // Also add to toggle screen order list with checkbox
```
nbolton: Comments should explain _why_ rather than _how/what_. If explaining how/what was the norm, literally every line of code would have a comment.

If you can rephrase it explain _why_ this is unusual, then keep the comment. If you can refactor the code to self-document, even better.

Otherwise, please remove it...
```suggestion
```

src/lib/gui/dialogs/ActionDialog.cpp
```diff
@@ -76,18 +86,23 @@ void ActionDialog::accept()
 
   m_action.typeScreenNames().clear();
 
-  int screenCount = ui->listScreens->count();
+  // For toggle screen action, use the toggle screen order list
+  // For other actions, use the standard screens list
```
nbolton:
```suggestion
```

nbolton (changes requested):
This is classic AI slop, please remove all the redundant comments.

sithlord48:
Update based and removed the pointless comments.

src/lib/server/InputFilter.h
```diff
@@ -224,6 +224,24 @@ class InputFilter
     IEventQueue *m_events;
   };
 
+  // ToggleScreenAction
```
nbolton, line 227:
```suggestion
```

src/lib/server/Server.h
```diff
@@ -77,6 +77,20 @@ class Server
     std::string m_screen;
   };
 
+  //! Toggle screen data
```
nbolton, line 80: Is this a helpful Doxygen comment?

nbolton (commented):
I'm being slightly picky.

src/lib/server/Server.h
sithlord48: nope is not even the style, But here again all the secions begin with a simlar comment.

src/lib/server/InputFilter.h
sithlord48: Every class in that file has the same coment before the class start.  I would leave that one  we can remove them later if we want all together.

nbolton (changes requested):
Tested, not working. I see this error on my Windows server:
```
[2026-01-22T13:41:25.787] ERROR: cannot read configuration "C:/ProgramData/Deskflow/deskflow-server.conf": read error: line 38: unknown screen name "" in toggleScreen
	C:\Projects\deskflow\src\lib\deskflow\ServerApp.cpp:140
```

@sithlord48 I wonder if this feature should replace or enhance the existing "Switch to next computer" feature by just adding the computer list there instead. The current label in this PR of "Toggle screen" is too vague and doesn't really make sense (it cycles between computers, not screens), and it's confusing why it's a separate action to the "Switch to next computer" action.

Image

sithlord48:
I think we should use this name at least for now we can look to combine them after this is confirmed to be in a working state.

sithlord48:
@Ashutosh0x  do you plan to continue to work on this ?

sithlord48:
> > @Ashutosh0x do you plan to continue to work on this ?
> 
> i can continue but need bounty i worked alot 🙂💰

That is not how Foss works..

PR #9415, opened 2026-02-02
feat: handle input capture session persistence

### LINKED ISSUE #8032: Persist permission dialog in XDG Desktop Portal (Wayland, etc)

whot:
## Description
Use the new libportal API to start a CreateSession2, then set the session persistence and read/write the tokens as required. This allows us to start a server without a permission dialog - provided the portal supports it and the user is ok with that.

See https://github.com/flatpak/xdg-desktop-portal/pull/1898 for a list of components/branches/mrs that are required to get this working, you will need libportal, xdg-desktop-portal-gnome and xdg-desktop-portal.

## Disclosure of AI use
Google Gemini was asked for how to read/write a string to a file with Qt, see the helper functions.

## Related Issue

#8032
fixes: #9599

## How Has This Been Tested?

This assumes a Fedora 43 box, adjust as required
```
> toolbox create deskflow
> toolbox enter deskflow
... you are now in the toolbox container
toolbox> sudo dnf builddep deskflow
toolbox> sudo dnf builddep -y xdg-desktop-portal xdg-desktop-portal-gnome libportal

toolbox> cd xdg-desktop-portal
toolbox> meson setup build -Dprefix=/usr
toolbox> meson compile -C build

... repeat for libportal/xdg-desktop-portal-gnome

toolbox> cd deskflow
... setup and build


toolbox> flatpak-spawn --host ./build/src/xdg-desktop-portal-gnome --replace &
toolbox> flatpak-spawn --host ./build/src/xdg-desktop-portal --replace &
toolbox> ./build/bin/deskflow
```
The updated dialog has a checkbox for rememberin the permission which gets stored in `XDG_STATE_HOME/deskflow/restore-token.txt` and *next* time that token will be re-used (if allowed) and deskflow-server no longer requires the user interaction.

**This is a Draft PR**. I won't have time to finish it and polish it but all the pieces are there so it can be picked up and integrated correctly (esp. the token read/write). So any volunteers, please take this!

It's backwards-compatible to the current portals (I hope, haven't tested :)

#### Edit, Sithlord48
 - Added a check for the new functions in libportal should build now with any version of libportal
 - Remove the temp helpers and write the token info to `Settings::Server::XdpRestoreToken`
 - Added to the `configuration.md` the new setting
 
Waiting on https://github.com/flatpak/libportal/pull/214 to land

### DIFF
--- a/src/lib/platform/PortalInputCapture.cpp
+++ b/src/lib/platform/PortalInputCapture.cpp
@@ -12,6 +12,10 @@
 #include "base/Log.h"
 #include "base/TMethodJob.h"
 
+#ifdef HAVE_LIBPORTAL_INPUTCAPTURE_RESTORE
+#include "common/Settings.h"
+#endif
+
 #include <sys/socket.h> // for EIS fd hack, remove
 #include <sys/un.h>     // for EIS fd hack, remove
 
@@ -20,6 +24,7 @@ namespace deskflow {
 PortalInputCapture::PortalInputCapture(EiScreen *screen, IEventQueue *events)
     : m_screen{screen},
       m_events{events},
+      m_portalVersion(0),
       m_portal{xdp_portal_new()}
 {
   m_glibMainLoop = g_main_loop_new(nullptr, true);
@@ -79,20 +84,9 @@ void PortalInputCapture::handleSessionClosed(XdpSession *session)
   m_signals.at(Signal::SessionClosed) = 0;
 }
 
-void PortalInputCapture::handleInitSession(GObject *object, GAsyncResult *res)
+void PortalInputCapture::setupSession(XdpInputCaptureSession *session)
 {
-  LOG_DEBUG("portal input capture session initialized");
   g_autoptr(GError) error = nullptr;
-
-  auto session = xdp_portal_create_input_capture_session_finish(XDP_PORTAL(object), res, &error);
-  if (!session) {
-    LOG_ERR("failed to initialize input capture session, quitting: %s", error->message);
-    g_main_loop_quit(m_glibMainLoop);
-    m_events->addEvent(Event(EventTypes::Quit));
-    return;
-  }
-
-  m_session = session;
 
   auto fd = xdp_input_capture_session_connect_to_eis(session, &error);
   if (fd < 0) {
@@ -113,6 +107,46 @@ void PortalInputCapture::handleInitSession(GObject *object, GAsyncResult *res)
   m_signals.at(ZonesChanged) = g_signal_connect(G_OBJECT(m_session), "zones-changed", G_CALLBACK(zonesChanged), this);
   m_signals.at(SessionClosed) = g_signal_connect(G_OBJECT(parentSession), "closed", G_CALLBACK(sessionClosed), this);
   handleZonesChanged(m_session, nullptr);
+}
+
+void PortalInputCapture::handleInitSession(GObject *object, GAsyncResult *res)
+{
+  g_autoptr(GError) error = nullptr;
+
+  LOG_DEBUG("portal input capture session initialized");
+
+  auto session = xdp_portal_create_input_capture_session_finish(XDP_PORTAL(object), res, &error);
+  if (!session) {
+    LOG_ERR("failed to initialize input capture session, quitting: %s", error->message);
+    g_main_loop_quit(m_glibMainLoop);
+    m_events->addEvent(Event(EventTypes::Quit));
+    return;
+  }
+
+  m_session = session;
+
+  setupSession(session);
+}
+
+void PortalInputCapture::handleStart(GObject *object, GAsyncResult *res)
+{
+  g_autoptr(GError) error = nullptr;
+
+#ifdef HAVE_LIBPORTAL_INPUTCAPTURE_RESTORE
+  LOG_DEBUG("portal input capture session initialized");
+  if (!xdp_input_capture_session_start_finish(m_session, res, &error)) {
+    LOG_ERR("failed to start input capture session, quitting: %s", error->message);
+    g_main_loop_quit(m_glibMainLoop);
+    m_events->addEvent(Event(EventTypes::Quit));
+    return;
+  }
+
+  auto restoreToken = xdp_input_capture_session_get_restore_token(m_session);
+  if (restoreToken) {
+    Settings::setValue(Settings::Server::XdpRestoreToken, QString(restoreToken));
+  }
+#endif
+  setupSession(m_session);
 }
 
 void PortalInputCapture::handleSetPointerBarriers(const GObject *, GAsyncResult *res)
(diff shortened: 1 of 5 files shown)


sithlord48:
Thanks @whot  I'll get the style stuff fixed and give it a test soon.

whot:
thanks, style is fixed now, if it's just a running of `clang-format` even I can do that ;)

sithlord48:
> It's backwards-compatible to the current portals (I hope, haven't tested :)

This does not seam to be so some methods are not found w/ libportal 0.9.1 `xdp_input_capture_session_start_finish` and some others are not here. Some of this will need to be wrapped in a version check . Any suggestions for how best to keep this compatible with both versions ? I am also ok saying we need libportal xx for the release after this has landed on the portal side

> Google Gemini was asked for how to read/write a string to a file with Qt, see the helper functions.

For future reference we just save things using our `Settings` class

whot:
oh, true. at *runtime* it is backwards-compat but I forgot about building against older libportal. this will need some CMake check for the new function and you can ifdef out bits om `initSession()` and assume `case 1`.

nbolton (changes requested):
Thanks for the PR. Would you mind changing to `camelCase` instead of `snake_case` for the function names and vars? [Of course, does not apply to the lib functions]

sithlord48:
I pushed some updates here 
 - Use a cmake check_symbol_exists to check for the restore ability in libportal's inputcapture support. (guard new code if its not found) 
 - Use camelCase vars for the few added. 
 - Use the settings to save the restore token similarly to how we do it in the RemoteDesktop portal

sithlord48:
@whot  this should now be up to our standards. It also no longer breaks the build when libportal does not have the correct methods. If the portal side is not going to change IMHO we could land this as its "safe enough" if that will help you land the other parts faster.

whot:
Thanks for the fixes, much appreciated. The big ticket item for the portal bits is the clipboard integration which requires more changes on the deskflow side than this. The inputcapture persistence bits are based on those so I recommend not merging until the clipboard is sorted.

Oh, and getting clipboard support via portals into deskflow would go a long way towards validating the portal additions so they can actually be merged. :chicken:, :egg: etc.

sithlord48:
Honestly @whot  if you can help w/ clipboard here that would be a MASSIVE help .  We can add any if checks we need here to align w/ libportal and that way you can test a full stack

src/lib/platform/PortalInputCapture.cpp
```diff
@@ -115,6 +111,46 @@ void PortalInputCapture::handleInitSession(GObject *object, GAsyncResult *res)
...
+#ifdef HAVE_LIBPORTAL_INPUTCAPTURE_RESTORE
+  LOG_DEBUG("portal input capture session initialized");
+  if (!xdp_input_capture_session_start_finish(m_session, res, &error)) {
+    LOG_ERR("failed to start input capture session, quitting: %s", error->message);
+    g_main_loop_quit(m_glibMainLoop);
+    m_events->addEvent(Event(EventTypes::Quit));
+    return;
+  }
+
+  m_sessionRestoreToken = xdp_input_capture_session_get_restore_token(session);
```
whot:
```suggestion
  m_sessionRestoreToken = xdp_input_capture_session_get_restore_token(m_session);
```

src/lib/platform/PortalInputCapture.cpp
```diff
@@ -115,6 +111,46 @@ void PortalInputCapture::handleInitSession(GObject *object, GAsyncResult *res)
...
+  if (m_sessionRestoreToken) {
+    Settings::setValue(Settings::Server::XdpRestoreToken, QString(m_sessionRestoreToken));
+  }
```
whot: `/home/whot/code/deskflow/src/lib/platform/PortalInputCapture.cpp:146:70: error: invalid conversion from ‘const char*’ to ‘char*’ [-fpermissive]`

```suggestion
  auto token = xdp_input_capture_session_get_restore_token(session);
  if (token) {
    Settings::setValue(Settings::Server::XdpRestoreToken, QString(token));
  }
```
There's no real point saving the token as member variable, it's a single-use token. You set it before `_start` and then you get a new token back[^1] and save that to your Settings.

[^1]: it *may* be the same token value but that's an implementation detail.

whot (commented):
fwiw, to build errors I encountered

src/lib/platform/PortalInputCapture.cpp
sithlord48: just shadowing what we do in the other portal

sithlord48:
Should both be fixed now. I no longer store the token as a member var

whot:
Pushed a fixup comment on top for easier review - in my local setup I had CreateSession2 fail, crashing deskflow which could be construed as bad UX.

sithlord48:
Note: We don't "Squash" land or anything so when this is ready for review some one has to fix up that fixup.

nbolton:
Will try to test as well as #9431 while I'm building all those libs.

> Note: We don't "Squash" land or anything so when this is ready for review some one has to fix up that fixup.

Yeah we can worry about getting the commits perfect when it's time to land.

sithlord48:
Edit: wrong pr the comment was intended for #9431

nbolton:
> my libportal is too old

You'll need to build and install it.
- `libportal` (https://github.com/flatpak/libportal/pull/214)
- `xdg-desktop-portal` (https://github.com/flatpak/xdg-desktop-portal/pull/1898)

On GNOME, I'm also having to build latest:
- `xdg-desktop-portal-gnome`
- `gsettings-desktop-schemas`

Let me know if it's easier on KDE and I might test on my Arch laptop as Wayland server.

sithlord48:
On kde the portal backend has a pr somewhere I think.. ( this will likey be used to test it on the kde side also )

sithlord48:
Updated and squashed the 'fixup' commit into the first one.

sithlord48:
@whot  any idea if libportal-next will contain both this and the clipboard or only the clipboard ?  trying to figure out if i this should be rebased on the changes needed for the clipboard or not.

whot:
it'll all depend on when https://github.com/flatpak/xdg-desktop-portal/pull/1898 gets pulled. Once that one is in, libportal will follow likely immediately

sithlord48:
@whot  can you ping me when the methods used here are not going to change so we can land this part.

sithlord48:
libportal #214 has landed this should be safe to include now ( but will need a libportal release for users to use)

sithlord48 (dismissed):
LGTM!!!!

nbolton (approved):
Tested and working great! 🎉 

Used: 
- `libportal` master (`567019d813dccfbb42fbfd3aaf7d6c7a4bad7c48`)
- `xdg-desktop-portal` master (`2473967902b1bc461243e9e6ecb4e17649734a4b`)
- `xdg-desktop-portal-gnome` [wip/inputcapture-persistence](https://gitlab.gnome.org/whot/xdg-desktop-portal-gnome/-/tree/wip/inputcapture-persistence)

I guess normal users will need to wait for your `wip/inputcapture-persistence` PR to land, as well as `libportal` and `xdg-desktop-portal` updates.

Thank you @whot for this wonderful PR! Now I get to archive [accept-portal-dialog](https://github.com/deskflow/accept-portal-dialog) 😄 

Image

Side note, fixed a tiny nit (all our logging is lower case):
```diff
diff --git a/src/lib/platform/PortalInputCapture.cpp b/src/lib/platform/PortalInputCapture.cpp
index 99385af892..0a063cf401 100644
--- a/src/lib/platform/PortalInputCapture.cpp
+++ b/src/lib/platform/PortalInputCapture.cpp
@@ -198,7 +198,7 @@ gboolean PortalInputCapture::initSession()
 
   switch (m_portalVersion) {
   case 0:
-    LOG_WARN("Portal bug: input capture version is %d", m_portalVersion);
+    LOG_WARN("portal bug: input capture version is %d", m_portalVersion);
     // fallthrough
   case 1:
     xdp_portal_create_input_capture_session(
```

krisrok:
Which portal versions include the neccessary changes?

sithlord48:
depends on the backend you use . you will need libporal 0.10.0 also

krisrok:
Thanks for the swift reply. That might be the culprit then since on Ubuntu it's only libportal 0.9.1 (and xdg-desktop 1.21.1, xdg-desktop-gnome 50.0).

sithlord48:
IIRc its gnome 50 and plasma 6.7.1

adrianinsaval:
I'm now on kde 6.7.1 and libportal 0.10.0 on fedora 44 but it's still not working, do I need to do anything else? do I need a specific version of deskflow? I'm running the v1.26.0.294 rpm from the continuous build release

sithlord48:
You should need 
 1. libportal 0.10.0
 2. plasma 6.7.1
 3. xdg-desktop-portal-kde 6.7.1
 4. deskflow's continuous version.

It should work for you if you have these three. You should see a check box when prompted to allow this always if you check that it should allow the capture session to persist

holzgeist:
> IIRc its gnome 50 and plasma 6.7.1

For gnome it looks like we'll have to wait for version 51. The enabling commit is not part of a release yet: https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/commit/50ae755bc1b230af20c931138cd6a0c9c6243bee

Issue #8486, opened 2025-04-12
Support sending right modifier keys instead of just left modifiers


ArsenicBismuth:
### Feature Request

Currently, when you send right modifier keys such as ROption or RControl from macOS, both of them are recognized as LAlt and LCtrl respectively in Windows.

You can check them here: en.key-test.ru (seems like the best, as other online keyboard testers doesn't differentiate between left & right modifiers).

(I'm specifically talking about my setup which is: macOS server, Windows client. I don't know if different OS combination has this behavior or not)

sithlord48:
Please tell me how this is effects your setup ?
 Keyboards tend to have only a L + R control / command / alt so you can type correctly not use them as seperate keys.

ArsenicBismuth:
@sithlord48 Games have L & R differentiated sometimes. And I have various shortcuts/hotkey in my Windows machine that depends on the R modifiers.

>  not use them as seperate keys.

But, why having incomplete implementation when you could make it whole?

Various RDP also have L & R differentiated. Because the implied usage is not what matters, it's the properness of the implementation.

I understand that it's non-trivial. Searching the keyword in Google shows how many different apps have this issue, some managed to solve it. But, I think for an app that's supposed to send any keyword/input to another, it's simply the proper behavior to have them differentiated.

![Image](https://github.com/user-attachments/assets/c9340d9a-6e32-4cf1-8943-cc17ea5c1b62)

jaepil-choi:
I second this. ANSI keyboard Asian users usually have language conversion button in right alt. 
This is different from the left alt.

ArsenicBismuth:
@Seraphli You're essentially asking the same thing. Shift is also part of modifier keys dude.

Seraphli:
Well, I just noticed you hadn’t mentioned it explicitly, so I thought I’d point it out.

gawtDamn:
Just wanted to weigh in that I would use this as well, as MacOS allows assigning different actions to the different modifier keys, and I was using this previously.

tonycody:
> Please tell me how this is effects your setup ? Keyboards tend to have only a L + R control / command / alt so you can type correctly not use them as seperate keys.

This is especially important for Chinese input methods. Typically, Chinese input methods allow you to distinguish between Chinese and English input modes by pressing the right Shift key. Without this, you would need to use keyboard shortcuts to switch input methods.

tonycody:
[...]

PR #9441, opened 2026-02-08
fixes AltGr for clients w/windows+finnish-layout.

aalm:
## Description
fixes this(AltGr/LVL3 sent by X11 server with fi-layout):
`DEBUG1: key ef7e is not on keyboard`
^as seen from the client log, and so ie. pressing '2' with it doesn't work.

## Disclosure of AI use
no AI was involved with this simple diff, just plenty of logs and experience
of already getting this fixed between X11-to-X11, and the great linux tools
nowadays available on windows thanks to WSL(xev, xkbcli interactive-..).

## Related Issue
atleast related to #4411 but nowadays deskflow does send it correctly over,
and the windows clients needs this to properly separate AltGr from Ctrl+Alt,
which isn't working.

## How Has This Been Tested?
rpi5 server:
(Deskflow: 1.25.0.220 (4200de66)
Qt: 6.8.2
System: Debian GNU/Linux 13 (trixie)
Session: LXDE (x11)

laptop(x64) client:
Deskflow: 1.25.0.224 (ee550587)
Qt: 6.7.3
System: Windows 11 Version 25H2

To reproduce bug:
1. Set Finish layout on Wayland server
2. Set Finish layout on Windows client
3. Move mouse to client
4. Press AltGr + 2

On master: `2` is typed
On this PR: `@` is typed (fix works as intended)

# To do

As is, this would definately belong behind a switch, because this doesn't
consider layouts not using AltGr at all - just wanted to keep this minimal.

- [x] Check if replacing `KeyModifierControl | KeyModifierAlt` with `KeyModifierAltGr` causes issues
- [ ] ~~Move fix into `getKeyMap` so that keymap contains the fix~~

### DIFF
--- a/src/lib/platform/MSWindowsKeyState.cpp
+++ b/src/lib/platform/MSWindowsKeyState.cpp
@@ -191,7 +191,7 @@ const KeyID MSWindowsKeyState::s_virtualKey[] = {
     /* 0x0a2 */ {kKeyControl_L},        // VK_LCONTROL
     /* 0x0a3 */ {kKeyControl_R},        // VK_RCONTROL
     /* 0x0a4 */ {kKeyAlt_L},            // VK_LMENU
-    /* 0x0a5 */ {kKeyAlt_R},            // VK_RMENU
+    /* 0x0a5 */ {kKeyAltGr},            // VK_RMENU
     /* 0x0a6 */ {kKeyNone},             // VK_BROWSER_BACK
     /* 0x0a7 */ {kKeyNone},             // VK_BROWSER_FORWARD
     /* 0x0a8 */ {kKeyNone},             // VK_BROWSER_REFRESH
@@ -448,7 +448,7 @@ const KeyID MSWindowsKeyState::s_virtualKey[] = {
     /* 0x1a2 */ {kKeyControl_L},    // VK_LCONTROL
     /* 0x1a3 */ {kKeyControl_R},    // VK_RCONTROL
     /* 0x1a4 */ {kKeyAlt_L},        // VK_LMENU
-    /* 0x1a5 */ {kKeyAlt_R},        // VK_RMENU
+    /* 0x1a5 */ {kKeyAltGr},        // VK_RMENU
     /* 0x1a6 */ {kKeyWWWBack},      // VK_BROWSER_BACK
     /* 0x1a7 */ {kKeyWWWForward},   // VK_BROWSER_FORWARD
     /* 0x1a8 */ {kKeyWWWRefresh},   // VK_BROWSER_REFRESH
@@ -1075,7 +1075,7 @@ void MSWindowsKeyState::getKeyMap(deskflow::KeyMap &keyMap)
           static const Modifier modifiers[] = {
               {VK_SHIFT, VK_SHIFT, 0x80u, KeyModifierShift},
               {VK_CAPITAL, VK_CAPITAL, 0x01u, KeyModifierCapsLock},
-              {VK_CONTROL, VK_MENU, 0x80u, KeyModifierControl | KeyModifierAlt}
+              {VK_CONTROL, VK_MENU, 0x80u, KeyModifierAltGr}
           };
           static const size_t s_numModifiers = sizeof(modifiers) / sizeof(modifiers[0]);
           static const size_t s_numCombinations = 1 << s_numModifiers;


sithlord48 (changes requested):
Looks ok but can you update the commit message to be `fix: ...`

sithlord48:
I fixed the commit message and rebased this upon master

sithlord48 (commented):
Updated the base and fixed up the commit message . I can't test it but the code looks ok.

For now I am going to make this a draft when you have updated it and its ready again please mark it ready for review.

sithlord48 (commented):
@nbolton  ill leave this up to you

nbolton:
> As is, this would definately belong behind a switch, because this doesn't
> consider layouts not using AltGr at all - just wanted to keep this minimal.

Thanks for pointing this out. I did wonder how it might impact non-AltGr users.

Edit: I added a todo to remind me to investigate. Currently trying to install the Suomi language pack but Windows is giving me some trouble.

sithlord48 (commented):
Don't have a good way to test but we should check what does happen if altGR is not there does it just send alt_R ?

nbolton:
Reproduced bug:
1. Set Finish layout on Wayland server
2. Set Finish layout on Windows client
3. Move mouse to client
4. Press AltGr + 2

On master: `2` is typed
On this PR: `@` is typed (fix works as intended)

sithlord48:
Do you anyway to test w/o a altgr ? i can't even use it casue in the vm that is the "host" key

nbolton:
I'm just wondering if `KeyModifierControl | KeyModifierAlt` evaluates to the same value as `KeyModifierAltGr`. Pretty sure they do.

The effective change seems to be this:
```cpp
  if (key == kKeyAltGr) {
    return VK_RMENU;
  }
```

Apparently right alt (e.g. US keyboard) is supposed to produce `VK_RMENU` and AltGr (e.g. European keyboard) should produce a sequence of `VK_LCONTROL` then `VK_RMENU` but I need to confirm that, and if it's even necessary (since the fix in this PR seems to work fine). I just wonder what the "proper" way of doing this is.

And, do we need a case for right alt? Is that a separate PR maybe?

----

Edit: I'm also wondering if the proper fix here is actually not to modify `MSWindowsKeyState::mapKeyToVirtualKey`:
https://github.com/deskflow/deskflow/blob/b33eb0135dbab76bd79c21078d192482b27255f3/src/lib/platform/MSWindowsKeyState.cpp#L701-L712

... but instead maybe we should modify the key map code in `MSWindowsKeyState::getKeyMap` which populates `m_keyToVKMap`:
https://github.com/deskflow/deskflow/blob/b33eb0135dbab76bd79c21078d192482b27255f3/src/lib/platform/MSWindowsKeyState.cpp#L916-L918

sithlord48:
IIRC, At this time we just send the keypress and for our modifier internal is just ALT no distinction between L / R.. 

There is a feature request to support both https://github.com/deskflow/deskflow/issues/8486

nbolton:
> Edit: I'm also wondering if the proper fix here is actually not to modify `MSWindowsKeyState::mapKeyToVirtualKey`:
> ... but instead maybe we should modify the key map code in `MSWindowsKeyState::getKeyMap` which populates `m_keyToVKMap`:

@aalm Would you mind taking a look to see if it's possible to do it this way?
Edit: Setting to draft until this is discussed.

amagus:
Tried reproducing the issue using windows as server and I couldn't.
On all cases, I set the keyboard to be Finnish (QWERTY), using latest continuous release.

Windows (server), Windows (client) => altgr works as expected
macOS (server) (with alt=altgr on configs), Windows (client) => altgr works as expected

I was hoping to try Nick's suggestion to see if it works, but I think this issue is very specific to wayland (or linux) as a server and I don't have that setup available to test.

nbolton:
I paged the details out now but I was definitely able to repro it on master and this PR fixed it. You have to have Finnish layout on both server and client. The server OS is not particularly relevant, as it's a Windows client fix. I plan to have a stab at my earlier suggestion (moving the logic to `getKeyMap`) if nobody else does. Bit swamped at the moment but I should have some time soon to get stuck in.

amagus:
@nbolton have you set the OS language as Finnish as well? Because I kept the language as english and tried setting the keyboard layout to finnish on both server (windows and mac) and client (windows) and I couldn't reproduce it on master (altgr + 2 produces "@" as expected)

nbolton:
Keyboard layout must be Finnish on both computers. OS language is not relevant.

nbolton:
@aalm I found a slightly cleaner way of applying your fix; I moved the logic to `getKeyMap` so that the actual keymap contains the fix.

```patch
diff --git a/src/lib/platform/MSWindowsKeyState.cpp b/src/lib/platform/MSWindowsKeyState.cpp
index e2d178f682..d9719fe122 100644
--- a/src/lib/platform/MSWindowsKeyState.cpp
+++ b/src/lib/platform/MSWindowsKeyState.cpp
@@ -703,9 +703,6 @@ UINT MSWindowsKeyState::mapKeyToVirtualKey(KeyID key) const
   if (key == kKeyNone) {
     return 0;
   }
-  if (key == kKeyAltGr) {
-    return VK_RMENU;
-  }
   KeyToVKMap::const_iterator i = m_keyToVKMap.find(key);
   if (i == m_keyToVKMap.end()) {
     return 0;
@@ -1159,6 +1156,8 @@ void MSWindowsKeyState::getKeyMap(deskflow::KeyMap &keyMap)
     }
   }
 
+  m_keyToVKMap[kKeyAltGr] = VK_RMENU;
+
   // restore keyboard layout
   ActivateKeyboardLayout(activeLayout, 0);
 }
```

@sithlord48 I'll squash this into one commit before I ready the PR.

Edit: Actually, I'm pulling my last commit out and just working on my branch instead (`fix/altgr-win`). I'll modify this PR once I'm happy with my branch.

Edit 2: OK, so it seems that `if` might actually be redundant (so moving the logic to `getKeyMap` might be pointless). I've requested a change to the PR. I didn't want to go ahead and just delete it before asking @aalm why it's there.

src/lib/platform/MSWindowsKeyState.cpp
```diff
@@ -703,6 +703,9 @@ UINT MSWindowsKeyState::mapKeyToVirtualKey(KeyID key) const
   if (key == kKeyNone) {
     return 0;
   }
+  if (key == kKeyAltGr) {
+    return VK_RMENU;
+  }
```
nbolton: @aalm This seems to be a redundant check. When I remove it, the Windows client fix still works (I can type `@` with AltGr+2). Does it cover some other scenario or is it just left over from debugging? 

The actual fix appears to be the combination of the change to `s_virtualKey` and the change to `modifiers` (in `getKeyMap`).

If you concur, I'm happy to remove this `if` for you (or you can do it).

Side note: I tested this with US layout on the Windows client and there doesn't seem to be any negative side effects, at least not with a Wayland server. On the Windows client, both alt keys are pressed when right-alt is pressed with US layout.

nbolton (changes requested):

src/lib/platform/MSWindowsKeyState.cpp
nbolton: @aalm I'm going to assume that it's vestigial from early debugging attempts:
```cpp
  if (key == kKeyAltGr) {
    return VK_RMENU;
  }
```

After removing it, the bug fix still works so it appears to be redundant.

I'll edit your commit to remove the lines.

Let me know if this is a mistake.

nbolton (approved):

sithlord48 (approved):

PR #9480, opened 2026-02-17
i18n(ja): update Japanese(ja) translation

ykasap:
## Description

Update unfinished ja translation, and modify some existing texts for consistency.

## Disclosure of AI use

None used.

## How Has This Been Tested?

Published the translation by Qt linguist and tested locally with deskflow 1.26.0.


sithlord48 (approved):
Thanks

PR #9530, opened 2026-02-27
fix: properly handling infinite timeout for macOS on EventQueue

amagus:
## Description
Fixes an unhandled situation of a negative timeout for the `OSXEventQueueBuffer` variant of the Event Queue.

-1 should be handled as an infinite timeout, but it was leading to a busy loop on macOS.

Taking a closer look on libc++ implementation of condition_variable::wait_for:
```cpp
template <class _Rep, class _Period, class _Predicate>
inline bool
condition_variable::wait_for(unique_lock<mutex>& __lk, const chrono::duration<_Rep, _Period>& __d, _Predicate __pred) {
  return wait_until(__lk, chrono::steady_clock::now() + __d, std::move(__pred));
}
```
The implementation already uses `wait_until` + `now`, hence using it one level up will bring no real harm.

## Disclosure of AI use
No AI Used for this PR

## Related Issue
fixes: #9529

## How Has This Been Tested?
Tested on macOS Tahoe 26.3, closed the lid on the client and observed the CPU usage over ssh.

### DIFF
--- a/src/lib/platform/OSXEventQueueBuffer.cpp
+++ b/src/lib/platform/OSXEventQueueBuffer.cpp
@@ -31,9 +31,10 @@ void OSXEventQueueBuffer::waitForEvent(double timeout)
 {
   std::unique_lock lock(m_mutex);
   if (m_dataQueue.empty()) {
-    auto duration = std::chrono::duration<double>(timeout);
     LOG_DEBUG2("waiting for event, timeout: %f seconds", timeout);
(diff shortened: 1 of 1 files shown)


sithlord48 (approved):
Tested on mac os looks to at least not be any worse than it was before. I will land this so the users with issues can test it out for us.

trzecieu:
Thanks! I can confirm this issue is fixed.

PR #9539, opened 2026-03-02
fix(win): handle dead-key overlap correctly and add integration regre…

i-santos:
## Description
This change rewrites the Windows dead-key handling in `MSWindowsHook` so relay mode and local/server mode no longer share the same dead-key flow. 

The fix added explicit tracking for the pending dead key and its transition state, and changed the relay path to decide when that dead key should be preserved, composed, emitted as a standalone character, or cleared before processing the next key. 

This was needed because the two paths do not interact with Windows keyboard composition in the same way, and using one shared flow caused dead keys to be applied to the wrong character, applied twice, or lost during fast overlapping input.

## Disclosure of AI use
This PR was AI-assisted for code and test authoring (implementation and test-writing support).  
Final behavior was validated through local build/test execution and manual verification by the author.

## Related Issue
fixes: [#9254](https://github.com/deskflow/deskflow/issues/9254)

## How Has This Been Tested?

**Regression verification strategy**

1. Add integration tests that reproduce the dead-key overlap sequences that were causing the bug.
2. Check out `master`.
3. Cherry-pick the commit that adds the new tests, without applying the fix.
4. Run the new integration tests against the pre-fix implementation.
5. Confirm that the tests fail on `master`, showing that they reproduce the original bug.
6. Check out `fix/windows-deadkey`.
7. Run the same integration tests again.
8. Confirm that the tests pass on `fix/windows-deadkey`, showing that the fix resolves the reproduced cases.

**Manual validation in real keyboard/layout conditions**

Keyboard Layout: `Portuguese (Brazil ABNT)`

- Client: `~ down, ~ up (dead key pending), a down, o down, a up, o up` -> expected: `não`; before the fix: `nãõ` (the dead key leaked into the next character and was consumed twice).
- Client: `´ down, ´ up (dead key pending), a down, e down, a up, e up` -> expected: `áe`; before the fix: `áé` (the dead key was consumed twice instead of only composing with `a`).
- Server: `t down, ´ down (dead key pending), t up, ´ up, a down, a up` -> expected: `tá`; before the fix: `ta` (the dead key was lost and never applied).
- Client, same sequence as the server scenario above: `t down, ´ down, t up, ´ up, a down, a up` -> expected: `tá`; before the fix: `t´a` (the same overlap produced a different failure on the client side, where the dead key was emitted as a standalone character instead of composing with `a`).

`down` means the key was pressed, and `up` means the key was released. 

When two `down` events appear in the sequence, that explicitly shows an overlap: the second key was pressed before the first key was fully released, which is what happens during fast typing.

### DIFF
--- a/src/lib/platform/MSWindowsDeadKeyUtils.h
+++ b/src/lib/platform/MSWindowsDeadKeyUtils.h
@@ -1,0 +1,183 @@
+/*
+ * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
+ * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
+ */
+
+#pragma once
+
+#include <Windows.h>
+
+namespace deskflow {
+namespace win32 {
+
+struct DeadKeyState
+{
+  WPARAM m_deadVirtKey = 0;
+  LPARAM m_deadLParam = 0;
+  WCHAR m_deadChar = 0;
+  BYTE m_deadKeyState[256] = {0};
+};
+
+enum class DeadKeyEventType
+{
+  KeyDown,
+  KeyUp
+};
+
+enum class DeadKeyAction
+{
+  StoreDead,
+  KeepPending,
+  ConsumeWithComposedChar,
+  ConsumeWithDeadFallback,
+  ConsumeBeforeNewCandidate,
+  NoDeadInteraction
+};
+
+struct DeadKeyTransitionInput
+{
+  DeadKeyEventType m_eventType = DeadKeyEventType::KeyDown;
+  int m_toUnicodeResult = 0;
+  UINT m_vk = 0;
+  UINT m_button = 0;
+  bool m_isModifierOrToggle = false;
+  bool m_hasPendingDead = false;
+  UINT m_deadOwnerButton = 0;
+};
+
+struct DeadKeyTransition
+{
+  DeadKeyAction m_action = DeadKeyAction::NoDeadInteraction;
+  bool m_setOwner = false;
+  bool m_clearOwner = false;
+  bool m_clearPending = false;
+  bool m_emitDeadFallback = false;
+  bool m_recomputeWithoutDead = false;
+};
+
+inline bool isModifierOrToggleVK(UINT vk)
+{
+  switch (vk) {
+  case VK_SHIFT:
+  case VK_LSHIFT:
+  case VK_RSHIFT:
+  case VK_CONTROL:
+  case VK_LCONTROL:
+  case VK_RCONTROL:
+  case VK_MENU:
+  case VK_LMENU:
+  case VK_RMENU:
+  case VK_LWIN:
+  case VK_RWIN:
+  case VK_CAPITAL:
+  case VK_NUMLOCK:
+  case VK_SCROLL:
+    return true;
+
+  default:
+    return false;
+  }
+}
+
+inline void sanitizeDeadKeyState(BYTE keys[256], UINT deadVK)
+{
+  BYTE sanitized[256] = {0};
+  for (UINT i = 0; i < 256; ++i) {
+    if (i == deadVK || isModifierOrToggleVK(i)) {
+      sanitized[i] = keys[i];
+    }
+  }
+
+  for (size_t i = 0; i < 256; ++i) {
+    keys[i] = sanitized[i];
+  }
+}
+
+inline bool shouldClearDeadKeyState(int toUnicodeResult, bool isKeyUpEvent)
+{
+  return (toUnicodeResult == 1 || toUnicodeResult == 2) && !isKeyUpEvent;
+}
+
+inline DeadKeyTransition decideDeadKeyTransition(const DeadKeyTransitionInput &input)
+{
+  const bool isKeyDown = (input.m_eventType == DeadKeyEventType::KeyDown);
+  const bool hasPendingDead = input.m_hasPendingDead;
+  const bool isModifierOrToggle = input.m_isModifierOrToggle;
+  const bool isOwnerButton = (input.m_deadOwnerButton != 0 && input.m_button == input.m_deadOwnerButton);
+
+  // Priority 1: a fresh dead key candidate was pressed.
+  // Store it immediately so subsequent key events can consume or fallback it.
+  // NOTE: order matters; this check must run before "no pending dead".
+  if (isKeyDown && input.m_toUnicodeResult < 0) {
+    return {DeadKeyAction::StoreDead, false, true, false, false, false};
+  }
+
+  // Priority 2: no dead key is pending, so this event does not interact with dead-key state.
+  if (!hasPendingDead) {
+    return {DeadKeyAction::NoDeadInteraction, false, false, false, false, false};
+  }
+
+  // Priority 3: key-up handling while a dead key is pending.
+  // If this key-up belongs to the owner button, force fallback and clear;
+  // otherwise keep pending for the next key-down.
+  if (!isKeyDown) {
+    if (isOwnerButton) {
+      return {DeadKeyAction::ConsumeWithDeadFallback, false, true, true, true, false};
+    }
+    return {DeadKeyAction::KeepPending, false, false, false, false, false};
+  }
+
+  // Priority 4: modifiers/toggles do not consume pending dead keys.
+  if (isModifierOrToggle) {
+    return {DeadKeyAction::KeepPending, false, false, false, false, false};
+  }
+
+  // Priority 5: key-down composition outcomes.
+  // result == 1: composed char produced (consume pending dead key).
+  if (input.m_toUnicodeResult == 1) {
+    return {DeadKeyAction::ConsumeWithComposedChar, false, true, true, false, false};
+  }
+
+  // result == 2: pending dead key did not compose with current key.
+  // Emit fallback for dead key and continue with current key.
+  if (input.m_toUnicodeResult == 2) {
+    return {DeadKeyAction::ConsumeWithDeadFallback, false, true, true, true, false};
+  }
+
+  // Priority 6: track ownership for overlap scenarios.
+  if (input.m_deadOwnerButton == 0) {
+    return {DeadKeyAction::KeepPending, true, false, false, false, false};
+  }
+
+  if (isOwnerButton) {
+    return {DeadKeyAction::KeepPending, false, false, false, false, false};
+  }
+
+  // Priority 7: overlap with a different key/button while dead is pending.
+  // Consume pending dead fallback first, clear dead state, then recompute.
+  return {DeadKeyAction::ConsumeBeforeNewCandidate, false, true, true, true, true};
+}
+
+inline void captureDeadKeyState(
+    DeadKeyState &state, WPARAM deadVirtKey, LPARAM deadLParam, const BYTE keys[256], WCHAR deadChar = 0
+)
+{
+  state.m_deadVirtKey = deadVirtKey;
+  state.m_deadLParam = deadLParam;
+  state.m_deadChar = deadChar;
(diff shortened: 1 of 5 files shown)


src/lib/platform/MSWindowsHook.cpp
```diff
@@ -188,26 +197,276 @@ static WPARAM makeKeyMsg(UINT virtKey, WCHAR wc, bool noAltGr)
...
+
+static void clearToUnicodeDeadKeyState(UINT flags)
+{
+  BYTE emptyState[256] = {};
+  WCHAR unicode[2] = {0, 0};
+  const UINT scanCode = MapVirtualKey(VK_SPACE, MAPVK_VK_TO_VSC);
+
+  // ToUnicode maintains an internal dead-key compose state.
+  // Flush it defensively when a dead key has just been consumed.
+  while (ToUnicode(VK_SPACE, scanCode, emptyState, unicode, 2, flags) < 0) {
```
amagus: Does this really need to be a while loop?

src/lib/platform/MSWindowsHook.cpp
```diff
@@ -188,26 +197,276 @@ static WPARAM makeKeyMsg(UINT virtKey, WCHAR wc, bool noAltGr)
...
+      g_dead.m_releaseVirtKey = g_dead.m_virtKey;
+    }
+  }
+}
+
+static void clearToUnicodeDeadKeyState(UINT flags)
+{
+  BYTE emptyState[256] = {};
+  WCHAR unicode[2] = {0, 0};
+  const UINT scanCode = MapVirtualKey(VK_SPACE, MAPVK_VK_TO_VSC);
```
amagus: any particular reason to hardcode VK_SPACE?

src/lib/platform/MSWindowsHook.cpp
```diff
@@ -188,26 +197,276 @@ static WPARAM makeKeyMsg(UINT virtKey, WCHAR wc, bool noAltGr)
...
       // incorrectly restore the dead key state.
-      g_deadRelease = g_deadVirtKey;
+      g_dead.m_releaseVirtKey = g_dead.m_virtKey;
+    }
+  }
+}
+
+static void clearToUnicodeDeadKeyState(UINT flags)
+{
+  BYTE emptyState[256] = {};
```
amagus: Maybe, for consistency, this should be initialized with {0}

src/lib/platform/MSWindowsHook.cpp
```diff
@@ -188,26 +197,276 @@ static WPARAM makeKeyMsg(UINT virtKey, WCHAR wc, bool noAltGr)
...
+      // We need to keep track of this because g_dead.m_virtKey will be
       // cleared later on; this would cause the dead key release to
       // incorrectly restore the dead key state.
-      g_deadRelease = g_deadVirtKey;
+      g_dead.m_releaseVirtKey = g_dead.m_virtKey;
+    }
+  }
+}
+
+static void clearToUnicodeDeadKeyState(UINT flags)
```
amagus: This static function is used just once, can't it be merged to clearTrackedDeadKeyState?

src/lib/platform/MSWindowsHook.cpp
```diff
@@ -610,8 +872,11 @@ EHookResult MSWindowsHook::install()
   }
 
   // discard old dead keys
-  g_deadVirtKey = 0;
-  g_deadLParam = 0;
+  g_dead.m_virtKey = 0;
```
amagus: can't you create a reset() function on the DeadRuntimeState instead?

amagus (changes requested):
It's really hard to understand what is going on because there's several refactors along with fixes; Several parts of the code got moved to adjacent static functions.

Could you please split your PR in at least two different commits?
The first one should be a `refactor:`, without trying to fix anything.
The second one should be the real `fix:` commit.

Anyway, the commit message prefix is not following our standards, could you please check https://github.com/deskflow/deskflow/wiki/Hacking-Guide and adjust your commit messages accordingly?

src/lib/platform/MSWindowsHook.cpp
```diff
@@ -255,138 +514,141 @@ static bool keyboardHookHandler(WPARAM wParam, LPARAM lParam)
...
+    const WCHAR effectiveDeadChar = (deadChar != 0) ? deadChar : g_dead.m_char;
+    WPARAM deadCharAndVirtKey = makeKeyMsg((UINT)g_dead.m_virtKey, effectiveDeadChar, noAltGrFallback);
+    PostThreadMessage(g_threadID, DESKFLOW_MSG_KEY, deadCharAndVirtKey, g_dead.m_lParam & 0x7fffffffu);
+    PostThreadMessage(g_threadID, DESKFLOW_MSG_KEY, deadCharAndVirtKey, g_dead.m_lParam | 0x80000000u);
+  };
+
+  WPARAM charAndVirtKey = 0;
+  bool clearDeadAfterRestore = transition.m_clearPending;
+  switch (transition.m_action) {
+  case deskflow::win32::DeadKeyAction::store_dead:
```
amagus: add `using deskflow::win32::DeadKeyAction` inside the switch and avoid namespaces on each case

src/lib/platform/MSWindowsDeadKeyUtils.h
```diff
@@ -0,0 +1,165 @@
...
+
+enum class DeadKeyEventType
+{
+  KeyDown,
+  KeyUp
+};
+
+enum class DeadKeyAction
+{
+  store_dead,
```
amagus: enums must follow pascal case

sithlord48 (changes requested):
What @amagus  said

nbolton (changes requested):
Your PR description reads like a predictive text matrix that lost its context window. We require manually written, carbon-based PR descriptions, not generated token mush. Please read the following resources:
- [Our hacking guide](https://github.com/deskflow/deskflow/wiki/Hacking-Guide)
- [RFC 406i (RAGS)](https://406.fail)

src/lib/platform/MSWindowsHook.cpp
i-santos: I refactored it to use a Max Attempts approach instead of a while with no stop condition

src/lib/platform/MSWindowsHook.cpp
i-santos: Yes. `VK_SPACE` is used as a reliable dead-key flush key. it exists in virtually every layout and commonly resolves a pending dead key into its standalone accent form.

src/lib/platform/MSWindowsHook.cpp
i-santos: yes. done

src/lib/platform/MSWindowsHook.cpp
i-santos: yes. merged

src/lib/platform/MSWindowsHook.cpp
i-santos: yes, done

src/lib/platform/MSWindowsHook.cpp
i-santos: agreed. done

src/lib/platform/MSWindowsDeadKeyUtils.h
i-santos: fixed it

i-santos:
> It's really hard to understand what is going on because there's several refactors along with fixes; Several parts of the code got moved to adjacent static functions.
> 
> Could you please split your PR in at least two different commits? The first one should be a `refactor:`, without trying to fix anything. The second one should be the real `fix:` commit.
> 
> Anyway, the commit message prefix is not following our standards, could you please check https://github.com/deskflow/deskflow/wiki/Hacking-Guide and adjust your commit messages accordingly?

I reorganized the commits into a more coherent series with separate `refactor`, `test`, and `fix` commits, and updated the commit messages to follow the expected convention.

sithlord48:
This is out of base with our master branch please rebase this upon current master , When you are updated and ready to have it reviewed be sure to click the ready for review button .

i-santos:
> Your PR description reads like a predictive text matrix that lost its context window. We require manually written, carbon-based PR descriptions, not generated token mush. Please read the following resources:
> 
>     * [Our hacking guide](https://github.com/deskflow/deskflow/wiki/Hacking-Guide)
> 
>     * [RFC 406i (RAGS)](https://406.fail)

English is not my first language, so I used AI help both to improve the wording and to assist with the dead-key state transition logic.

To be transparent, the transition/state-machine part of this fix was beyond what I could confidently derive on my own. The bug itself was introduced by human-written code, and the existing implementation already showed signs that this area was difficult to reason about. Looking at the diff before this fix, there are comments such as 

> This handles the obscure situation, 

which suggests the behavior was already not straightforward.

Because of that, I did not want to rely on AI output blindly. I added a substantial set of integration tests to validate the behavior across multiple keyboard layouts and to make sure the result was grounded in observable behavior, not just plausible code generation.

I am not a maintainer, so I am not in a position to decide whether this approach is acceptable for the project. I completely understand if the PR is not something you want to merge.

That said, I really like Deskflow and had almost stopped using it because this bug made Portuguese input impractical in daily use. I eventually decided to fix it and contribute it back as my first open-source contribution. I am genuinely happy that I can now use Deskflow daily again.

Thank you for the work on the project, and I will fully respect whatever decision you make about the PR.

sithlord48:
Looks like there is a small issue w/ code lint.

i-santos:
> Looks like there is a small issue w/ code lint.

working on that now.

nbolton:
> English is not my first language, so I used AI help both to improve the wording and to assist with the dead-key state transition logic.

Thanks for being honest. In future, just keep it concise and include distilled insight. See other PRs for examples.

nbolton (changes requested):
Please reword the PR description:
1. What does "Manual validation on dead-key fast-typing scenarios (including overlap ordering) in real keyboard/layout conditions." mean -- please replace everything in "How Has This Been Tested?" with what you actually did to manually test it so that we can do the same. Specific steps would be more useful.
2. Distill everything in "Description" into a simple plain english description of what code you changed and why. One simple paragraph is sufficient unless the change is truly very complex (which I doubt).

i-santos:
Updated the PR description.

Manual verification may be harder to reproduce on keyboards using a US layout, but the `How Has This Been Tested?` section now includes four concrete bug reproduction scenarios that were used to validate the behavior.

nbolton:
> Updated the PR description.
> 
> Manual verification may be harder to reproduce on keyboards using a US layout, but the `How Has This Been Tested?` section now includes four concrete bug reproduction scenarios that were used to validate the behavior.

This is like 1000% times better, thank you! I'll test it tomorrow.

nbolton:
> Keyboard Layout: `Portuguese (Brazil ABNT)`

What OS and keyboard layout are you using on the server and client? 

On macOS is it Brazilian or Brazilian ABNT2? 

On Windows I see "Portuguese (Brazil ABNT)"

src/lib/platform/MSWindowsDeadKeyUtils.h
nbolton, line 1: Testing with Windows server, this PR introduces 2 new bugs:
1. Completely breaks dead keys on macOS client - when ´ + a is typed, only a is produced.
1. On Linux (Wayland) clients, it causes keys to be repeated and space chars to be missed (happens regardless of keyboard layouts chosen).

nbolton (changes requested):

i-santos:
On macOS it is Brazilian.

I don't have macOS. 
I'll figure out how I can test it before marking as ready to review again.

nbolton:
> I'll figure out how I can test it before marking as ready to review again.

OK, thanks, yes we must test thoroughly before marking ready for review.

sithlord48:
@i-santos  are you still working on this or should we take it over ? 
@amagus   could you look at why this is failing for mac os please ?

PR #9544, opened 2026-03-04
Window settings

sithlord48:
## Description

Before: 
Screenshot_20260304_170645

After: 
Screenshot_20260304_202032Screenshot_20260304_201953

## Disclosure of AI use

None
## Related Issue

None
## How Has This Been Tested?

Run

### DIFF
--- a/src/lib/gui/dialogs/SettingsDialog.ui
+++ b/src/lib/gui/dialogs/SettingsDialog.ui
@@ -22,6 +22,9 @@
    </property>
    <item>
     <widget class="QTabWidget" name="tabWidget">
+     <property name="currentIndex">
+      <number>0</number>
+     </property>
      <widget class="QWidget" name="tabGeneral">
       <property name="sizePolicy">
        <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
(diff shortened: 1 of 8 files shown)


src/lib/gui/dialogs/SettingsDialog.ui
```diff
@@ -143,6 +135,125 @@
...
+              </property>
+             </widget>
+            </item>
+           </layout>
+          </widget>
+         </item>
+         <item row="1" column="0">
+          <widget class="QLabel" name="label">
+           <property name="text">
+            <string>When the mainwindow is closed</string>
```
amagus: maybe should be "main window" instead of "mainwindow"?

src/lib/gui/dialogs/SettingsDialog.ui
```diff
@@ -143,6 +135,125 @@
        </item>
       </layout>
      </widget>
+     <widget class="QWidget" name="tabWindow">
+      <attribute name="title">
+       <string>&amp;Window Behavior</string>
```
nbolton: Maybe just 'Window' -- all settings control behavior so the word is implied.
```suggestion
       <string>&amp;Window</string>
```

nbolton (approved):

PR #9545, opened 2026-03-05
build: ensuring macOS codesign is triggered for changes in core

amagus:
## Description
This ensures that changes on deskflow-core binary also trigger resigning in macOS.

Please note that it creates a stamp file to keep track of the last time the signature happen. 
Using the binary itself was not possible given that codesign itself touches the binary files.
POST_BUILD scripting triggered resigning every time, which is not intended.

## Disclosure of AI use
Claude Code guidance on how to properly create a stamp file

## Related Issue
N/A

## How Has This Been Tested?
Tested building with codesign and without codesign.
When it should, changes on the gui or in core triggered `codesign`

### DIFF
--- a/cmake/MacCodesign.cmake
+++ b/cmake/MacCodesign.cmake
@@ -1,0 +1,49 @@
+# SPDX-FileCopyrightText: (C) 2025-2026 Deskflow Contributors
+# SPDX-License-Identifier: MIT
+
+# Warning: Do not use for CI/production, as the `entitlements-dev.plist` file adds special
+# entitlements that are only appropriate for local development.
+#
+# macOS made TCC stricter so that if you don't sign your local dev builds properly, macOS will
+# nag you to remove and re-approve the app every time you make a change to the binary which is
+# extremely annoying during development.
+#
+# If you were to use ad-hoc signing (i.e. not specify a certificate), TCC would still nag you
+# because the binary identity is anchored not on the app ID, but on the CD hash (which changes
+# based on the binary contents).
+#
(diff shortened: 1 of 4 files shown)


cmake/MacCodesign.cmake
```diff
@@ -0,0 +1,50 @@
...
+
+  # Use a stamp file because codesign modifies the binaries it signs.
+  add_custom_command(
+    OUTPUT ${stamp_file}
+    COMMAND /usr/bin/codesign
+            --force
+            --options runtime
+            --entitlements "${CMAKE_SOURCE_DIR}/src/apps/res/entitlements-dev.plist"
+            --sign "${APPLE_CODESIGN_DEV}"
+            "${bundle_dir}"
```
sithlord48: Are  you not able to do this ? 

`$<TARGET_BUNDLE_DIR:${CMAKE_PROJECT_PROPER_NAME}>`

amagus: I tried to use the directory as stamp, it didn't work, cmake requires a target file...
And I didn't wanted to add any file... codesign updates the timestamps on all the binaries, which make it to be resigned every build (not the actual goal)

sithlord48 (approved):

nbolton:
It appears this PR has made it harder to develop locally, as each time you launch a new build, you need to re-grant permissions. Unfortunately this is very tedious and slows down testing for me. I will raise an issue.

Edit: #9547
> Solved by blowing `build` dir away 🎉

Issue #9552, opened 2026-03-08
Restrict deskflow client to using specific network


BenBergman:
### Feature Request

I only want to use deskflow when I have my laptop connected to a dock with a wired network connection, not over wifi or VPN. Right now that means I have to manually start/stop deskflow whenever I move around or else I might accidentally connect to the server at inappropriate times.

Today I noticed the "Network IP" option in the advanced settings and hoped it would be what I wanted, but when I disconnect from the dock deskflow automatically switches to my wifi network. Adding a toggle here to restrict deskflow to the specified IP might be a sufficient solution.

### Voting Instructions

React with :+1: or :-1: to vote on this request

sithlord48:
Agree this can be done a bit better.  Ill assign myself and see how to improve this further.

nbolton:
> automatically switches to my wifi network

Agreed, anything auto needs a setting.

sithlord48:
Can you try the above PR .. this adds the old ip to the list preventing automatic from being used if the ip no longer exists when you open the dialog.

When the selected network is not connected you should see 
`WARNING: cannot bind address: Cannot assign requested address`

It does not stop trying in case that ip shows back up

nbolton:
@BenBergman See: https://github.com/deskflow/deskflow/pull/9553#issuecomment-4030177393

sithlord48:
Re opened as this issue for for client mode.

alizane:
Hi @sithlord48  I see #9553 landed to fix the server-side UI state for this, but the client is still falling back to 0.0.0.0 when the specific interface drops (triggering the operating system will select network interface automatically warning). I'm currently digging into the socket binding logic in net/ to prevent this fall back if a strict IP is set. Let me know if you're okay with me putting up a PR for the client side!

sithlord48:
you don't need my permission to  propose a code change. You only need to follow the contribution guidelines

PR #9553, opened 2026-03-08
fix: do not reset network if old interface is missing

sithlord48:
## Description

Add the current interface to the selection combo it its not added. This prevents reset when the interface is not longer valid. 

## Disclosure of AI use

None
## Related Issue

Server side fix for: #9552

## How Has This Been Tested?

set a bogus 10.0.9.1 address for interface in config, run check the set interface that is used and set in the settings and shows in the combo of interfaces.

### DIFF
--- a/src/lib/gui/dialogs/SettingsDialog.cpp
+++ b/src/lib/gui/dialogs/SettingsDialog.cpp
@@ -63,6 +63,11 @@ SettingsDialog::SettingsDialog(QWidget *parent, const IServerConfig &serverConfi
     }
   }
 
+  if (const auto interface = Settings::value(Settings::Core::Interface).toString();
+      !interface.isEmpty() && (ui->comboInterface->findData(interface) == -1)) {
+    ui->comboInterface->addItem(interface, interface);
+  }
+
   loadFromConfig();
 
   adjustSize();


nbolton:
I'll wait until @BenBergman has tested. 

@BenBergman To get the packages, go here and scroll down: https://github.com/deskflow/deskflow/actions/runs/22830644042

BenBergman:
Thanks for taking a crack at this so quickly! (and linking the builds!)

Sadly, this does not resolve the issue. I tested on a macOS and Linux client. On both, I connected to my dock, set the network to the IP address assigned to the dock (set static by the router) and then disconnecting the dock Deskflow still fails over to the wifi network. On the Linux client I also tried setting the network IP to that of my wifi connection, then turned off wifi. The connection definitely drops, but it deskflow reconnects after about a second.

BenBergman:
I suppose these messages from the log might be helpful:

```
[2026-03-10T11:44:57.820] NOTE: server is dead
[2026-03-10T11:44:57.820] INFO: leaving screen
[2026-03-10T11:44:57.820] WARNING: failed to connect to server: server is not responding
[2026-03-10T11:44:58.821] IPC: connecting to '192.168.1.190': 192.168.1.190:24800
[2026-03-10T11:44:58.822] WARNING: cannot bind address: Cannot assign requested address
[2026-03-10T11:44:58.822] WARNING: operating system will select network interface automatically
```

sithlord48:
When connected to the dock you want to use ETH and when not you want to use WIFI ? 

That is not at all what this is doing.. This makes it so if you set the Connecting IP to Eth  you can only connect with that IP address. 

> [2026-03-10T11:44:58.822] WARNING: operating system will select network interface automatically 

Yes i guess we need to not let the os pick a fallback now

BenBergman:
> When connected to the dock you want to use ETH and when not you want to use WIFI ?

No, that is the existing behaviour when I specify the IP of ETH (I think). What I want is for Deskflow to only work with ETH and if that interface disappears to not have a Deskflow connection. I just did a test with a connection on WIFI to confirm the behaviour was an issue that way as well where the interface still existed but was turned off.

> This makes it so if you set the Connecting IP to Eth you can only connect with that IP address.

From my testing, this isn't working.

sithlord48:
yes because as client there is that fallback we need to prevent still . ill see what i can do 
Edit: not seeing why this happens i don't see us really asking for a new ip or anything. @nbolton  can you look at this im sure its gonna be something in arch that im not aware of.

nbolton:
> can you look at this

Sure, I'll give my head a scratch. I guess this is a draft until we figure it out?

sithlord48:
Draft or land as is and make a pr to fix it for client mode. This will fix the case of setting the server to an IP then that IP goes away . We now at least continue to keep the users choice, so up to you on land now follow up to fix client or add to this one.

nbolton:
> Draft or land as is

OK, I will test.

Edit: Looks good. Still not sure about using "link visited" for error... seems super jank to me.

image

nbolton (approved):

sithlord48:
For the idk 5th or 6th time that is one of the palette colors we have acces to there is not Error color and assuming colors meanins is a bad idea (red is not a global negative color) Its worse to force a random no palette color as that can have to low contrast with the background

nbolton:
> that is one of the palette colors we have acces to there is not Error color and assuming colors meanins is a bad idea (red is not a global negative color)

Is there one that isn't purple?

sithlord48:
Its up to your color theme

PR #9563, opened 2026-03-12
fix: retry OpenClipboard() on Windows

nbolton:
OpenClipboard() can fail transiently when another process holds the clipboard mutex. Add a retry loop (5 attempts, 5ms delay) so that Deskflow handles brief contention gracefully instead of immediately failing.

## Description

Cherry pick of retry loop from #9474

## Disclosure of AI use

See #9474

## Related Issue

Unknown (#9474 lists tentatively related Windows clipboard issues, but none specifically show logs with the clipboard failing to open)

## How Has This Been Tested?

Uncertain (#9474 explains some testing but the accuracy of the contributor’s claim could not be verified)


sithlord48 (approved):
ok with me

PR #9566, opened 2026-03-13
feat: support link-local connections in Preferences dialog

eean:
most link-local IPs are just noise, but if an entire interface is link-local only, maybe someone is hooking up their computers directly by ethernet cable

## Description
Currently all link-local addresses are filtered out of the preferences dialog. This makes sense, given that ~every interface has a fe80 address. But I do actually use Deskflow with link-local address for two laptops on (otherwise) different networks connected by Ethernet, so I added logic to allow link-local addresses from interfaces that don't have other IPs.

## Disclosure of AI use
Used Gemini to double check the std::ranges::any_of api usage

## How Has This Been Tested?
Built, and tested the dialog on Linux. I'm sure this code works fine on other platforms but I do wonder if Windows is more aggressive at handing out link-local IP addresses to disconnected interfaces.

### DIFF
--- a/src/lib/gui/core/NetworkMonitor.cpp
+++ b/src/lib/gui/core/NetworkMonitor.cpp
@@ -71,11 +71,15 @@ QStringList NetworkMonitor::validAddresses()
     const bool isVirtualType = interface.type() == QNetworkInterface::Virtual;
     const bool isVirtual = isVirtualInterface(interface.humanReadableName()) || isP2P || isVirtualType;
     const auto addressEntries = interface.addressEntries();
+    // exclude link-local IP addresses on interfaces with better options
+    const bool hasNonLinkLocal = std::ranges::any_of(addressEntries, [](const QNetworkAddressEntry &entry) {
+      return !entry.ip().isLoopback() && !entry.ip().isLinkLocal();
+    });
 
     for (const auto &entry : addressEntries) {
       const QHostAddress address = entry.ip();
 
-      if (address.isLinkLocal() || address.isLoopback() || uniqueAddresses.contains(address)) {
+      if (address.isLoopback() || (hasNonLinkLocal && address.isLinkLocal()) || uniqueAddresses.contains(address)) {
         continue;
       }
 
@@ -98,6 +102,8 @@ QStringList NetworkMonitor::validAddresses()
   std::ranges::sort(physicalIP4, [](const QHostAddress &a, const QHostAddress &b) {
     if (a.isPrivateUse() != b.isPrivateUse())
       return a.isPrivateUse();
+    if (a.isLinkLocal() != b.isLinkLocal())
+      return a.isLinkLocal();
     return a.toIPv4Address() < b.toIPv4Address();
   });
 
@@ -108,6 +114,8 @@ QStringList NetworkMonitor::validAddresses()
   std::ranges::sort(physicalIP6, [](const QHostAddress &a, const QHostAddress &b) {
     if (a.isPrivateUse() != b.isPrivateUse())
       return a.isPrivateUse();
+    if (a.isLinkLocal() != b.isLinkLocal())
+      return a.isLinkLocal();
     return a.toString() < b.toString();
   });
 
--- a/src/lib/gui/core/NetworkMonitor.cpp
+++ b/src/lib/gui/core/NetworkMonitor.cpp
@@ -71,11 +71,16 @@ QStringList NetworkMonitor::validAddresses()
     const bool isVirtualType = interface.type() == QNetworkInterface::Virtual;
     const bool isVirtual = isVirtualInterface(interface.humanReadableName()) || isP2P || isVirtualType;
     const auto addressEntries = interface.addressEntries();
+    // link-local is only worth offering when it's a physical interface's only address, e.g. a direct ethernet cable
+    const bool hasNonLinkLocal = std::ranges::any_of(addressEntries, [](const QNetworkAddressEntry &entry) {
+      return !entry.ip().isLoopback() && !entry.ip().isLinkLocal();
+    });
+    const bool skipLinkLocal = isVirtual || hasNonLinkLocal;
 
     for (const auto &entry : addressEntries) {
       const QHostAddress address = entry.ip();
 
-      if (address.isLinkLocal() || address.isLoopback() || uniqueAddresses.contains(address)) {
+      if (address.isLoopback() || (skipLinkLocal && address.isLinkLocal()) || uniqueAddresses.contains(address)) {
         continue;
       }
 
@@ -98,6 +103,8 @@ QStringList NetworkMonitor::validAddresses()
   std::ranges::sort(physicalIP4, [](const QHostAddress &a, const QHostAddress &b) {
     if (a.isPrivateUse() != b.isPrivateUse())
       return a.isPrivateUse();
+    if (a.isLinkLocal() != b.isLinkLocal())
+      return a.isLinkLocal();
     return a.toIPv4Address() < b.toIPv4Address();
   });
 
@@ -108,6 +115,8 @@ QStringList NetworkMonitor::validAddresses()
   std::ranges::sort(physicalIP6, [](const QHostAddress &a, const QHostAddress &b) {
     if (a.isPrivateUse() != b.isPrivateUse())
       return a.isPrivateUse();
+    if (a.isLinkLocal() != b.isLinkLocal())
+      return a.isLinkLocal();
     return a.toString() < b.toString();
   });
 
--- a/src/lib/gui/core/NetworkMonitor.h
+++ b/src/lib/gui/core/NetworkMonitor.h
@@ -47,7 +47,7 @@ public:
   void stopMonitoring();
 
   /**
-   * @brief Get list of all available IP addresses (excluding local and link-local addresses)
+   * @brief Get list of all available IP addresses (excluding loopback addresses)
    * @return IP address list
    */
   static QStringList validAddresses();


src/lib/gui/core/NetworkMonitor.cpp
```diff
@@ -98,6 +102,8 @@ QStringList NetworkMonitor::validAddresses()
   std::ranges::sort(physicalIP4, [](const QHostAddress &a, const QHostAddress &b) {
     if (a.isPrivateUse() != b.isPrivateUse())
       return a.isPrivateUse();
+    if (a.isLinkLocal() != b.isLinkLocal())
```
sithlord48, line 106: why is this added ?

amagus:
I personally think this is only solving your unique problem. 
A general solution, in my opinion, would be way better. 
We need something like a "Show link-local" option near the ip combo box, and if that is turned ticked, we show all the ips.

src/lib/gui/core/NetworkMonitor.cpp
eean: to keep the public IP addresses at the end

eean:
> I personally think this is only solving your unique problem. A general solution, in my opinion, would be way better. We need something like a "Show link-local" option near the ip combo box, and if that is turned ticked, we show all the ips.

I don't think it's that unique, at least not for deskflow users. I'd rather have a "show local link-only interfaces" option if we want to go that way. I'd be happy to implement this, though I think it's a bit messy.

Really what users want to control is what interface deskflow runs on. But we need a IP address string to bind to, so I wonder if there's a clever way to make configuration dialog more interface-oriented. Like instead of a simple combo box maybe a selection widget with IP addresses organized by interface.

amagus:
> I don't think it's that unique, at least not for deskflow users. 

What I meant by uniqueness is that you are defining that interfaces with link-local only should be listed differently from the others.
I understand the use case, I consider it valid, but I also consider valid the case of using link locals in interfaces with other ip addresses... If all the machines are local, restricting to link local may still be effective.
I understand that you may want to listen on specific interfaces but unfortunately not how binding works. 
Listing all link local may be messy, indeed, but it may be even messier to select an interface...

sithlord48:
Yes it can be more interface oriented but we need to fix up some back end code to use QNetwork before hand.  It won't make the situation any cleaner since now your gonna want to not only restrict via interface but also ips bound to that interface. doing it by ip is just generally eaiser for users to understand

I agree that if we are to show local link addresses then showing them should be an option that is off by default. Personally I don't even see the point because If you are an advanced enough user to want to do restrictions to a local link interface then you are also a skilled enough user to just put the desired ip in your config file to set the interface avoiding the issue since deskflow will now add the address to the list (but it will show always show not connected coloring on the server's using IP label)

eean:
> What I meant by uniqueness is that you are defining that interfaces with link-local only should be listed differently from the others.

Ah yeah I misunderstood you. Your right it's actually not hard to think of a reason you'd want to use the link-local address of an interface that provided 'better' IP addresses.

> since deskflow will now add the address to the list 

It doesn't for me, I'm pretty sure the drop down just uses the function I modified here. So you're right in that I did get my system working, but it feels broken, thus this PR. There's actually no reason to set the IP address you want to set the server to listen on that isn't a bit advanced or esoteric.

I'll tinker on this some more, either come back with that a checkbox for filtering link-local addresses, or a screenshot of a QTreeView mock up.

sithlord48:
Make sure you are using the continuous version i just added that recently with https://github.com/deskflow/deskflow/pull/9553

nbolton:
> I do actually use Deskflow with link-local address for two laptops on (otherwise) different networks connected by Ethernet

Good point.

> Personally I don't even see the point because If you are an advanced enough user to want to do restrictions to a local link interface then you are also a skilled enough user to just put the desired ip

Also a good point...

Idea: What if we allow the user to manually enter an IP for these edge cases so they don't need to manually edit the config file.

> I'll tinker on this some more

I guess it's a draft then?

sithlord48:
> Idea: What if we allow the user to manually enter an IP for these edge cases so they don't need to manually edit the config file.

Thought about this but then we also have to let them remove their custom items, I am hesitant to let users type in things to that box due to how breakable it could be to put the wrong character.

sithlord48:
@eean  will you be updating this PR ? Or should we close it as abandonded ?

eean:
> @eean will you be updating this PR ? Or should we close it as abandonded ?

I'll get back to it soon

eean:
@nbolton  @sithlord48 was thinking of implementing something like this. I think including the interface names as headers that are unselectable makes it easier to find the interface you want.  I mocked it up in Designer, but I would implement it in plain C++, using a QListWidget, and a style sheet to implement the indenting.

One idea is to also have a Other option, then we can exclude the virtual interfaces and deal with interfaces that have disappeared.
image
image
 
I had to rename it to .xml for github to accept the .ui file:
[some proposals.xml](https://github.com/user-attachments/files/26281542/some.proposals.xml)

sithlord48:
To do this correctly you want to make a QAbstractListModel class that we then can use in lists / comboboxes etc.. Idk if this is worth that amount of effort.

Take the route your taking make a fully realized version and we will look at it. No matter what we should be able to use some if not all the code in the future too.

I've not really thought about the ux here but I would prefer a simple list with custom items that can be added by typing in the combobox itself. if they are not valid you can just delete the text in the editable combo. 

> and a style sheet to implement the indenting.

Don't use stylesheets if you can avoid it.

eean:
QListWidget is good enough for a widget for interfaces and ip addresses. It's just a different way of looking what's in the combobox already.

sithlord48:
No QListWidget is not backed by models. 

What you want to do is make a QAbstractListView based model then we can use that model in a QListView or a QComboBox or any of the other model backed objects.

eean:
There's really no reason to use a model. We're talking about ~15 strings that will all be visible at once. Well that's my opinion, but I guess not really related to this PR, you could use a model or not for any UI discussed, including the current one.

sithlord48:
> I guess not really related to this PR, you could use a model or not for any UI discussed, including the current one. 

Totally just continue what you were doing and we will check it out when you think its ready

eean:
I don't mind requests for code refactoring afterward, but I want to get the UI design nailed down more before I implement it.

sithlord48:
@eean Any updates here is this something you are still working on ?

sithlord48 (dismissed):
Tested this locally and now i can see the local links if that is all that is on that interface. I think this is a pretty good improvement. Sorry it took this long to review. 

If you want to further improve the UI for this it can be done w/ a future PR

src/lib/gui/core/NetworkMonitor.cpp
```diff
@@ -71,11 +71,15 @@ QStringList NetworkMonitor::validAddresses()
     const bool isVirtualType = interface.type() == QNetworkInterface::Virtual;
     const bool isVirtual = isVirtualInterface(interface.humanReadableName()) || isP2P || isVirtualType;
     const auto addressEntries = interface.addressEntries();
+    // exclude link-local IP addresses on interfaces with better options
+    const bool hasNonLinkLocal = std::ranges::any_of(addressEntries, [](const QNetworkAddressEntry &entry) {
```
nbolton, line 75: Was it intentional to allow link-local on virtual interfaces too? On my box this adds fe80 entries for tailscale0 and every docker veth.

Image

nbolton: This is what it looks like on `master`:

Image

sithlord48: well the idea is that if the interface only has local link .. and those sure do .. 

we could hide the virtual ones if you prefer this

nbolton: @eean @amagus Thoughts? Should we exclude things like Docker and Tailscale?

sithlord48: @nbolton  i have updated this again it should now filter out the virtual without a real ip (like it does on our current master branch)

nbolton: Still there.

image

Skip link-local addresses on virtual interfaces too, not just on interfaces that have a better address.

Example:
```patch
diff --git a/src/lib/gui/core/NetworkMonitor.cpp b/src/lib/gui/core/NetworkMonitor.cpp
index 5cc9bae475..6c61650462 100644
--- a/src/lib/gui/core/NetworkMonitor.cpp
+++ b/src/lib/gui/core/NetworkMonitor.cpp
@@ -71,27 +71,28 @@ QStringList NetworkMonitor::validAddresses()
     const bool isVirtualType = interface.type() == QNetworkInterface::Virtual;
     const bool isVirtual = isVirtualInterface(interface.humanReadableName()) || isP2P || isVirtualType;
     const auto addressEntries = interface.addressEntries();
-    // exclude link-local IP addresses on interfaces with better options
+    // link-local is only worth offering when it's a physical interface's only address, e.g. a direct ethernet cable
     const bool hasNonLinkLocal = std::ranges::any_of(addressEntries, [](const QNetworkAddressEntry &entry) {
       return !entry.ip().isLoopback() && !entry.ip().isLinkLocal();
     });
+    const bool skipLinkLocal = isVirtual || hasNonLinkLocal;
 
     for (const auto &entry : addressEntries) {
       const QHostAddress address = entry.ip();
 
-      if (address.isLoopback() || (hasNonLinkLocal && address.isLinkLocal()) || uniqueAddresses.contains(address)) {
+      if (address.isLoopback() || (skipLinkLocal && address.isLinkLocal()) || uniqueAddresses.contains(address)) {
         continue;
       }
 
       uniqueAddresses.insert(address);
 
       if (address.protocol() == QHostAddress::IPv6Protocol) {
-        if (isVirtual && hasNonLinkLocal)
+        if (isVirtual)
           virtualIP6.append(address);
         else
           physicalIP6.append(address);
       } else {
-        if (isVirtual && hasNonLinkLocal)
+        if (isVirtual)
           virtualIP4.append(address);
         else
           physicalIP4.append(address);

```

image

nbolton (approved):

PR #9582, opened 2026-03-16
fix: kill core orphan on gui death (linux)

nbolton:
## Description

On Linux, child processes are not automatically killed when the parent unexpectedly dies.

> The prctl() system call provides a rather useful PR_SET_PDEATHSIG option to allow a signal to be sent to child processes when the parent unexpectedly dies. 
> [Source](https://smackerelofopinion.blogspot.com/2015/11/using-prsetpdeathsig-to-reap-child.html)

## Disclosure of AI use

Used Claude to find these links:
https://doc.qt.io/qt-6/qprocess.html#setChildProcessModifier
https://man7.org/linux/man-pages/man2/pr_set_pdeathsig.2const.html

## Related Issue

Fixes #9581

## How Has This Been Tested?

Run: `killall deskflow`
Expect: `ps ax | grep deskflow-core` has no results

### DIFF
--- a/src/lib/gui/core/CoreProcess.cpp
+++ b/src/lib/gui/core/CoreProcess.cpp
@@ -14,6 +14,11 @@
 #include "OSXHelpers.h"
 #endif
 
+#ifdef Q_OS_LINUX
+#include <signal.h>
+#include <sys/prctl.h>
+#endif
+
 #include <QCoreApplication>
 #include <QDir>
 #include <QFile>
@@ -202,6 +207,14 @@ void CoreProcess::startForegroundProcess(const QStringList &args)
   // core command can be easily copy/pasted to the terminal for testing.
   const auto quoted = makeQuotedArgs(m_appPath, args);
   qInfo("running command: %s", qPrintable(quoted));
+
+#ifdef Q_OS_LINUX
+  m_process->setChildProcessModifier([] {
+    // the core process becomes orphaned when the gui process exits abruptly (e.g. with kill -9),
+    // so ensure the os also kills the core when that happens to the gui.
+    prctl(PR_SET_PDEATHSIG, SIGTERM);
+  });
+#endif
 
   m_process->start(m_appPath, args);
 


sithlord48 (approved):
Seams to work as expected, I think we can do better w/ maybe a QProcess solution that works on all platforms..

nbolton:
> Seams to work as expected, I think we can do better w/ maybe a QProcess solution that works on all platforms..

Yeah I guess it is a workaround for a bigger problem.

MultiKoopa:
Hmm. Doesn't seem to work on steam Deck. If I switch into desktop mode multiple times without rebooting, it says it's still running if I try to use again. I'm guessing something is different with whatever steam deck does when switching to desktop mode?

sithlord48:
Are you running the continuous version ?

MultiKoopa:
> Are you running the continuous version ?

I'm using the flatpak; can you use the continuous version on an immutable OS? can you install a non-Discover store flatpak on SteamOS?

sithlord48:
yes

MultiKoopa:
> yes

well now this happens on an infinite loop: 

deskflow problem

the log just shows the mouse entering the screen, no other errors.

Even more annoying this popup happens infinitely if I attach the usb dongle for the mouse directly to the steam deck. And the popup is definitely coming from Deskflow; quitting it stops the popup.

Deskflow is unusable like this.

sithlord48:
what version of kde plasma is on there ?

MultiKoopa:
> what version of kde plasma is on there ?

6.4.3. Can't be updated.

sithlord48:
This maybe fixed in kde 6.6.6  sounds like it was a bug in the portals https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/commit/cdb8e76a962d5557262653b59d5d1cce6e379d5a

MultiKoopa:
> This maybe fixed in kde 6.6.6 sounds like it was a bug in the portals https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/commit/cdb8e76a962d5557262653b59d5d1cce6e379d5a

Welp. Hopefully SteamOS upgrades their plasma version soon.

sithlord48:
IIRc you should be using (by default) X11 for the desktop mode that should not involve any portals.

MultiKoopa:
> IIRc you should be using (by default) X11 for the desktop mode that should not involve any portals.

Unfortunately, the latest major SteamOS update switched to Wayland.

https://store.steampowered.com/news/app/1675200/view/697641379212298072

sithlord48:
I've updated the flatpak to use kde 6.11 this may somehow resolve your issue 
https://github.com/deskflow/deskflow/pull/10012 is the pr (not landed yet) 
https://github.com/deskflow/deskflow/actions/runs/30387515183/artifacts/8699859219 is the flatpak made from the last action run on that branch https://github.com/deskflow/deskflow/actions/runs/30387515183

MultiKoopa:
> I've updated the flatpak to use kde 6.11 this may somehow resolve your issue #10012 is the pr (not landed yet) https://github.com/deskflow/deskflow/actions/runs/30387515183/artifacts/8699859219 is the flatpak made from the last action run on that branch https://github.com/deskflow/deskflow/actions/runs/30387515183

appreciate it. I tried something else before I saw this: leaving DeskFlow on my Windows PC, and putting the old Barrier back on the Steam Deck. Seems to work just fine. I originally switched to DeskFlow because Barrier to Barrier was having a lot of problems with the new SteamOS update. But this seems to work perfectly.

MultiKoopa:
> I've updated the flatpak to use kde 6.11 this may somehow resolve your issue #10012 is the pr (not landed yet) https://github.com/deskflow/deskflow/actions/runs/30387515183/artifacts/8699859219 is the flatpak made from the last action run on that branch https://github.com/deskflow/deskflow/actions/runs/30387515183

ok Barrier stopped working correctly, so I tried again with Deskflow. Looked in the Discover store and realized there's a more stable version within Flathub. Didn't know you could do that. The problem seems fixed now. thank you!

PR #9587, opened 2026-03-17
fix: high-resolution scrolling on macOS clients

dadealus:
## Description
This change addresses an issue where micro-updates from high-resolution mice (like the Logitech G502 X) were being truncated to zero on macOS clients, causing the scroll wheel to appear non-functional. 

The fix implements scroll accumulation and switches the event unit to pixels (). This ensures that fractional scroll deltas are preserved and eventually result in smooth on-screen movement.

## Disclosure of AI use
I used Gemini to help navigate the codebase, identify the truncation point in , and help draft the accumulation logic to ensure it correctly handled the client's scaling and inversion settings.

## How Has This Been Tested?
- **Server**: Arch Linux KDE Wayland
- **Client**: macOS (OSX 26.3.1)
- **Hardware**: Logitech G502 X
- **Results**: Verified that 'micro-ticks' now result in smooth scrolling on the Mac client, and the scroll speed can be further tuned using the client-side ratio settings.

### DIFF
--- a/src/lib/deskflow/ISecondaryScreen.h
+++ b/src/lib/deskflow/ISecondaryScreen.h
@@ -70,6 +70,26 @@ public:
     return delta;
   }
 
+  double getYScrollScale() const
+  {
+    return m_yScrollScale;
+  }
+
+  double getXScrollScale() const
+  {
+    return m_xScrollScale;
+  }
+
+  bool getInvertYScroll() const
+  {
+    return m_invertYScroll;
+  }
+
+  bool getInvertXScroll() const
+  {
+    return m_invertXScroll;
+  }
+
 private:
   /**
    * @brief this member is used to modify the verical scroll direction.
--- a/src/lib/platform/OSXScreen.h
+++ b/src/lib/platform/OSXScreen.h
@@ -240,6 +240,8 @@ private:
      Evil, and this should be moved to a place where it need not
      be mutable as soon as possible. */
   mutable MouseButtonState m_buttonState;
+  mutable double m_scrollRemainderX = 0.0;
+  mutable double m_scrollRemainderY = 0.0;
   using MouseButtonEventMapType = std::map<uint16_t, CGEventType>;
   std::vector<MouseButtonEventMapType> MouseButtonEventMap;
 
--- a/src/lib/platform/OSXScreen.mm
+++ b/src/lib/platform/OSXScreen.mm
@@ -588,21 +588,55 @@ void OSXScreen::fakeMouseRelativeMove(int32_t dx, int32_t dy) const
 void OSXScreen::fakeMouseWheel(ScrollDelta delta) const
 {
   if (delta.x != 0 || delta.y != 0) {
-    // use server's acceleration with a little boost since other platforms
-    // take one wheel step as a larger step than the mac does.
-    delta = applyScrollModifier(
-        {static_cast<int32_t>(3.0 * delta.x / s_scrollDelta), static_cast<int32_t>(3.0 * delta.y / s_scrollDelta)}
-    );
-    // create a scroll event, post it and release it.  not sure if kCGScrollEventUnitLine
-    // is the right choice here over kCGScrollEventUnitPixel
-    CGEventRef scrollEvent = CGEventCreateScrollWheelEvent(nullptr, kCGScrollEventUnitLine, 2, delta.y, delta.x);
+    // Multiplier for lines per scroll tick (3.0 is a legacy boost).
+    const double linesPerTick = 3.0;
+    // Multiplier for pixels per line (common on macOS).
+    const double pixelsPerLine = 15.0;
 
-    // Fix for sticky keys
-    CGEventFlags modifiers = m_keyState->getModifierStateAsOSXFlags();
-    CGEventSetFlags(scrollEvent, modifiers);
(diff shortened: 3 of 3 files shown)


sithlord48 (commented):
You may want to do something similar to what we do in the XWindowScreen Impl of fakeScrollWheel.

dadealus:
Thank you for the feedback. I reviewed the `XWindowsScreen` implementation as suggested. 

The X11 implementation ensures movement by rounding any non-zero delta up to a full 120-unit 'click' (one line). My implementation for macOS takes a similar 'always move' approach but with higher precision to **preserve the high-resolution smooth scrolling features** of mice like the G502 X, MX series, and more:

1. **Accumulation**: Instead of just rounding up to a full line, it uses `double` precision to accumulate remainders. This ensures that high-resolution data is preserved over time rather than being discarded or bloated into full notches.
2. **Pixel Resolution**: I switched the event unit to `kCGScrollEventUnitPixel` (approximately 15x finer than a line). My 'round up' logic ensures that any non-zero micro-update results in at least a **1-pixel** movement immediately.

 This provides the same functional guarantee as the X11 logic, but with the smooth, sub-line precision that matches the native macOS scrolling experience. I found this sub-pixel accumulation to be significantly smoother during testing with high-DPI sensors. Let me know if you would still prefer I align the logic even more strictly with the X11 implementation.

sithlord48:
> The X11 implementation ensures movement by rounding any non-zero delta up to a full 120-unit 'click' (one line).  
 
This seams like the right thing todo.

> High def scrolling... 

But Why ? (just don't even get why this is a thing on hardware) 

Tldr:  I think for a solution for this to be "correct" it should work the same on all screen types . And maybe better to calculate this in the base class. Since that would affect all the subclasses

sithlord48 (changes requested):
I think the correct way to support this is to do it in the base class so that all subclasses have this same ability

sithlord48:
Going to draft until this gets updated.

nbolton:
> I think the correct way to support this is to do it in the base class so that all subclasses have this same ability

Just a hunch, I haven't groked the fix yet but isn't it an OS-specific fix? If so, I'm not sure how that would work. What leads you to think the fix can be moved up into the parent class?

sithlord48:
The "fix" is about using doubles and keeping a remainder for the unused scroll ..  its not really os specific .. I think we used to do this on xwindows.

dadealus:
Thanks for the feedback, everyone. I appreciate the focus on getting the architecture right. As a contributor looking to improve the macOS experience, I’d like to provide some technical context to help bridge the discussion:

**1. Modern Hardware Standards (HID++ & Apple Native)**
High-resolution scrolling is the native standard for modern macOS. Apple’s own Magic Mouse and Trackpad use high-precision, pixel-based scrolling to provide the "fluid" feel macOS is known for. High-performance mice (like the Logitech G502 X via HID++) use similar "micro-updates"—they don't wait for a full "notch" to send a signal; they report fine-grained physical movement as it happens.

**2. Fixing the "Dead Zone"**
By forcing these micro-updates into integer "lines" (`kCGScrollEventUnitLine`), Deskflow was truncating any movement smaller than a full line to zero. This created a "dead zone" where slow physical scrolling resulted in zero movement on the screen. My fix switches to `kCGScrollEventUnitPixel` and uses `double` precision to match how macOS handles this natively.

**3. Modern Standards vs. Legacy X11**
Regarding the X11 comparisons—I'm running a Wayland-native environment and avoid X11 as much as possible, as it's legacy. X11’s traditional "discrete button" model (Button 4/5) is why scrolling can feel "chunky" on older systems. Both macOS and Wayland have moved past this to support high-precision axes. I believe Deskflow should align with these modern standards rather than trying to fit them into a legacy X11-style logic.

**4. Regarding the Base Class Refactor**
I see the value in `sithlord48`’s suggestion to move the accumulation logic to a base class, but I also share `nbolton`’s concern about whether this can truly be generic. 

While "accumulating a remainder" is a generic concept, the *implementation* of how those remainders are dispatched is deeply platform-specific. On macOS, the "correct" fix is switching the unit to pixels (`kCGScrollEventUnitPixel`), which is a CoreGraphics-specific enum. On X11 or Windows, "scrolling" is still largely tied to fixed 120-unit "notches" or virtual button clicks.

Moving this to a base class would require a significant architectural refactor to unify how all platforms report fractional deltas—converting everything to a shared unit before dispatching. Given that this PR is a surgical fix for a confirmed bug that makes high-res mice nearly unusable on macOS clients today, I’d propose we land this platform-specific fix first. We can then use it as a reference for a larger, multi-platform refactor of the scroll architecture in a future PR.

sithlord48:
> 3. Modern Standards vs. Legacy X11
Regarding the X11 comparisons—I'm running a Wayland-native environment and avoid X11 as much as possible, as it's legacy. X11’s traditional "discrete button" model (Button 4/5) is why scrolling can feel "chunky" on older systems. Both macOS and Wayland have moved past this to support high-precision axes. I believe Deskflow should align with these modern standards rather than trying to fit them into a legacy X11-style logic.

The same input methods are used on both wayland and X11....,  this maybe only because ei abstracts it for us. that you see the changes in Deskflow that you do.  

What i am saying is we take care of the movement / scaling in the base class we can calculate the leftover in teh base the platforms have todo whatever

sithlord48:
I think this should be more like #9975

PR #9661, opened 2026-04-14
# feat(macos): add native MDM support via NSUserDefaults

jdipanfilo:
## Description

Lets an administrator enforce Deskflow settings on macOS with a configuration profile,
delivered through MDM such as Jamf Pro, Mosyle, Kandji or Intune.

Any setting key can be enforced. The preference key is the settings key exactly as
written, for example `security/tlsEnabled`, `core/port` or `server/enableClipboard`, so
there is no list of manageable settings to maintain.

Enforcement happens in `Settings::value()` and `Settings::setValue()`, so it applies to
the GUI and `deskflow-core` alike and edits to the config file are overridden.
`Settings::setValue()` now returns `false` and warns when it refuses a write rather than
returning silently. `Settings::isManaged()` is declared in `Settings` and defined per
platform: `ManagedSettings.mm` reads forced values through `NSUserDefaults` on macOS,
`ManagedSettings.cpp` reports nothing managed everywhere else.

In the GUI, each control names the setting it edits in a `managedSetting` property and a
shared helper disables the ones that are enforced, showing a note explaining why.

Documented in `docs/user/configuration.md`.

## AI tools used for this PR

Claude Opus 4.8 (claude.ai) was used for implementation and for working through review
feedback. 
Claude Sonnet 4.6 (Claude Code) was used for code review. Translations were filled in with Google Translate and left marked unfinished.

## Related Issue

Proposed in #9659

## How Has This Been Tested?

Tested on macOS 15.7.7, 26.5.1 and 26.5.2 on an MDM enrolled Mac as well as on an unenrolled device.

On the enrolled Mac, the settings were delivered as a Jamf Pro configuration profile
using an Application & Custom Settings payload with the preferred domain
`org.deskflow.deskflow` and this plist:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>security/tlsEnabled</key>
    <false/>
    <key>core/port</key>
    <integer>12345</integer>
    <key>server/enableClipboard</key>
    <false/>
</dict>
</plist>
```

On the unenrolled Mac the same payload was written directly to
`/Library/Managed Preferences/org.deskflow.deskflow.plist` with PlistBuddy.

In both cases: enforced controls are disabled and show the enforced value, settings
without a policy stay editable, and edits to the config file are overridden. A value for example
written with `defaults write org.deskflow.deskflow log/level -int 2` is correctly not
treated as enforced, since it is not in a forced domain. Removing the profile makes the affected controls editable again.

### DIFF
--- a/src/lib/common/CMakeLists.txt
+++ b/src/lib/common/CMakeLists.txt
@@ -9,6 +9,12 @@ endif()
 configure_file(Constants.h.in Constants.h @ONLY)
 configure_file(VersionInfo.h.in VersionInfo.h @ONLY)
 
+if(APPLE)
+  set(managed_sources ManagedSettings.mm)
+else()
+  set(managed_sources ManagedSettings.cpp)
+endif()
+
 add_library(common STATIC
   Coordinate.h
   Enums.h
@@ -16,6 +22,7 @@ add_library(common STATIC
   I18N.h
   I18N.cpp
   LogLevel.h
+  ${managed_sources}
   NetworkProtocol.h
   PlatformInfo.h
   Settings.h
@@ -34,3 +41,7 @@ target_include_directories(common
     $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/src/lib>
     $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/src/lib>
 )
+
+if(APPLE)
+  target_link_libraries(common PRIVATE "-framework Foundation")
+endif()
--- a/src/lib/common/ManagedSettings.cpp
+++ b/src/lib/common/ManagedSettings.cpp
@@ -1,0 +1,19 @@
+/*
+ * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
+ * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
+ */
+
+#include "Settings.h"
+
+bool Settings::isManaged(const QString &key)
+{
+  Q_UNUSED(key)
+  return false;
+}
+
+QVariant Settings::managedValue(const QString &key)
+{
+  Q_UNUSED(key)
+  return QVariant();
+}
--- a/src/lib/common/ManagedSettings.mm
+++ b/src/lib/common/ManagedSettings.mm
@@ -1,0 +1,48 @@
+/*
+ * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
+ * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
+ */
+
+#include "Settings.h"
+
+#import <Foundation/NSString.h>
+#import <Foundation/NSUserDefaults.h>
+#import <Foundation/NSValue.h>
+
+// Qt headers must come after Foundation to avoid symbol conflicts
+#include <QString>
+#include <QVariant>
+
+#include <cstring>
+
+bool Settings::isManaged(const QString &key)
+{
+  // objectIsForcedForKey: reports only values coming from a forced domain, so a
+  // value the user wrote themselves is never mistaken for an enforced policy
+  return [[NSUserDefaults standardUserDefaults] objectIsForcedForKey:key.toNSString()];
+}
+
+QVariant Settings::managedValue(const QString &key)
+{
+  if (!isManaged(key))
+    return QVariant();
+
+  id value = [[NSUserDefaults standardUserDefaults] objectForKey:key.toNSString()];
+  if (!value)
+    return QVariant();
+
+  if ([value isKindOfClass:[NSNumber class]]) {
+    // NSNumber wraps BOOL as __NSCFBoolean, so the encoded type is the only way
+    // to tell a policy bool from a policy int
+    const char *encodedType = [value objCType];
+    if (strcmp(encodedType, @encode(BOOL)) == 0 || strcmp(encodedType, @encode(bool)) == 0)
+      return QVariant([value boolValue]);
+    return QVariant([value intValue]);
+  }
+
+  if ([value isKindOfClass:[NSString class]])
+    return QVariant(QString::fromNSString(value));
+
+  return QVariant();
+}
--- a/src/lib/common/Settings.cpp
+++ b/src/lib/common/Settings.cpp
@@ -6,6 +6,8 @@
 
 #include "Settings.h"
 
+#include <QDebug>
+
 #include "LogLevel.h"
 #include "NetworkProtocol.h"
 #include "UrlConstants.h"
@@ -315,13 +317,18 @@ QString Settings::logLevelText()
   return Settings::value(Log::Level).toString();
 }
 
-void Settings::setValue(const QString &key, const QVariant &value)
+bool Settings::setValue(const QString &key, const QVariant &value)
 {
+  if (isManaged(key)) {
+    qWarning() << "not changing" << key << "as it is enforced by an administrator policy";
+    return false;
+  }
+
   const bool useState = Settings::m_stateKeys.contains(key) && !instance()->isPortableMode();
   auto settings = useState ? instance()->m_stateSettings : instance()->m_settings;
 
   if (settings->value(key) == value)
-    return;
+    return true;
 
   if (!value.isValid())
     settings->remove(key);
@@ -334,10 +341,14 @@ void Settings::setValue(const QString &key, const QVariant &value)
 
   settings->sync();
   Q_EMIT instance()->settingsChanged(key);
+  return true;
 }
 
 QVariant Settings::value(const QString &key)
 {
+  if (isManaged(key))
+    return managedValue(key);
+
   const bool useState = Settings::m_stateKeys.contains(key) && !instance()->isPortableMode();
   auto settings = useState ? instance()->m_stateSettings : instance()->m_settings;
   return settings->value(key, defaultValue(key));
--- a/src/lib/common/Settings.h
+++ b/src/lib/common/Settings.h
@@ -174,7 +174,7 @@ public:
   static Settings *instance();
   static void setSettingsFile(const QString &settingsFile = QString());
   static void setStateFile(const QString &stateFile = QString());
-  static void setValue(const QString &key = QString(), const QVariant &value = QVariant());
+  static bool setValue(const QString &key = QString(), const QVariant &value = QVariant());
   static QVariant value(const QString &key = QString());
   static void restoreDefaultSettings();
   static QVariant defaultValue(const QString &key);
@@ -192,6 +192,14 @@ public:
   static NetworkProtocol networkProtocol();
   static void save(bool emitSaving = true);
   static QStringList validKeys();
+
+  /**
+   * @brief Checks if a setting is enforced by an administrator policy.
+   * Defined per platform; only platforms with a policy backend can return true.
+   * @param key The settings key to check.
+   * @return true if the setting is enforced and may not be changed.
+   */
+  static bool isManaged(const QString &key);
   static QStringList validGroups();
   static QString portableSettingsFile();
   static void removeUnknownScreens(const QStringList &knownScreens);
@@ -226,6 +234,14 @@ private:
    * @return a valid computerName
    */
   static QString cleanComputerName(const QString &name);
+
+  /**
+   * @brief Gets the value an administrator policy enforces for a setting.
+   * Defined per platform alongside isManaged().
+   * @param key The settings key to look up.
+   * @return The enforced value, or an invalid QVariant if there is none.
+   */
+  static QVariant managedValue(const QString &key);
 
   QSettings *m_settings = nullptr;
   QSettings *m_stateSettings = nullptr;
--- a/src/lib/gui/dialogs/ServerConfigDialog.cpp
+++ b/src/lib/gui/dialogs/ServerConfigDialog.cpp
@@ -16,6 +16,7 @@
 #include "dialogs/ActionDialog.h"
 #include "dialogs/HotkeyDialog.h"
 #include "dialogs/ScreenSettingsDialog.h"
+#include "gui/StyleUtils.h"
 
 #include <QFileDialog>
 #include <QMessageBox>
@@ -433,6 +434,9 @@ void ServerConfigDialog::loadFromConfig()
   m_clipboardSize = Settings::value(Settings::Server::ClipboardSize).toUInt();
   ui->sbClipboardSizeLimit->setValue(m_clipboardSize);
 
+  // Runs last so a managed control cannot be re-enabled by the code above
+  applyManagedLocks();
+
   ui->listHotkeys->clear();
   for (const Hotkey &hotkey : std::as_const(serverConfig().hotkeys()))
     ui->listHotkeys->addItem(hotkey.text());
@@ -451,6 +455,29 @@ void ServerConfigDialog::loadFromConfig()
   } else {
     server->markAsServer();
   }
+}
+
+void ServerConfigDialog::applyManagedLocks()
+{
+  // Each control records the setting it edits so the lock check stays generic.
+  ui->rbProtocolSynergy->setProperty("managedSetting", Settings::Server::Protocol);
+  ui->rbProtocolBarrier->setProperty("managedSetting", Settings::Server::Protocol);
+  ui->cbHeartbeat->setProperty("managedSetting", Settings::Server::EnableHeatbeat);
+  ui->sbHeartbeat->setProperty("managedSetting", Settings::Server::Heartbeat);
+  ui->cbRelativeMouseMoves->setProperty("managedSetting", Settings::Server::RelativeMouseMoves);
+  ui->cbWin32KeepForeground->setProperty("managedSetting", Settings::Server::Win32KeepForeground);
+  ui->cbSwitchDelay->setProperty("managedSetting", Settings::Server::EnableSwitchDelay);
+  ui->sbSwitchDelay->setProperty("managedSetting", Settings::Server::SwitchDelay);
+  ui->cbSwitchDoubleTap->setProperty("managedSetting", Settings::Server::EnableSwitchDoubleTap);
+  ui->sbSwitchDoubleTap->setProperty("managedSetting", Settings::Server::SwitchDoubleTap);
+  ui->groupExternalConfig->setProperty("managedSetting", Settings::Server::ExternalConfig);
+  ui->lineConfigFile->setProperty("managedSetting", Settings::Server::ExternalConfigFile);
+  ui->cbDefaultLockToComputerState->setProperty("managedSetting", Settings::Server::DefaultLockToComputerState);
+  ui->cbDisableLockToComputer->setProperty("managedSetting", Settings::Server::DisableLockToComputer);
+  ui->cbEnableClipboard->setProperty("managedSetting", Settings::Server::EnableClipboard);
+  ui->sbClipboardSizeLimit->setProperty("managedSetting", Settings::Server::ClipboardSize);
+
+  deskflow::gui::applyManagedLocks(this, tr("Managed by your organization"));
 }
 
 void ServerConfigDialog::initConnections() const
--- a/src/lib/gui/dialogs/SettingsDialog.cpp
+++ b/src/lib/gui/dialogs/SettingsDialog.cpp
@@ -14,6 +14,7 @@
 #include "common/I18N.h"
 #include "common/Settings.h"
 #include "gui/Messages.h"
+#include "gui/StyleUtils.h"
 #include "gui/TlsUtility.h"
 #include "gui/core/NetworkMonitor.h"
 
@@ -402,6 +403,41 @@ void SettingsDialog::updateControls()
   ui->widgetLogFilename->setEnabled(writable && logToFile);
 
   updateTlsControls();
+
+  // Runs last so a managed control cannot be re-enabled by the code above
+  applyManagedLocks();
+}
+void SettingsDialog::applyManagedLocks()
+{
+  // Each control records the setting it edits so the lock check stays generic.
+  // Paired radio buttons share a key so both halves lock together.
+  ui->sbPort->setProperty("managedSetting", Settings::Core::Port);
+  ui->comboInterface->setProperty("managedSetting", Settings::Core::Interface);
+  ui->comboLogLevel->setProperty("managedSetting", Settings::Log::Level);
+  ui->groupLogToFile->setProperty("managedSetting", Settings::Log::ToFile);
+  ui->lineLogFilename->setProperty("managedSetting", Settings::Log::File);
+  ui->cbGuiDebug->setProperty("managedSetting", Settings::Log::GuiDebug);
+  ui->cbElevateDaemon->setProperty("managedSetting", Settings::Daemon::Elevate);
+  ui->rbAutoHide->setProperty("managedSetting", Settings::Gui::Autohide);
+  ui->rbShowOnStart->setProperty("managedSetting", Settings::Gui::Autohide);
+  ui->rbCloseToTray->setProperty("managedSetting", Settings::Gui::CloseToTray);
+  ui->rbExitOnClose->setProperty("managedSetting", Settings::Gui::CloseToTray);
+  ui->rbIconMono->setProperty("managedSetting", Settings::Gui::SymbolicTrayIcon);
+  ui->rbIconColorful->setProperty("managedSetting", Settings::Gui::SymbolicTrayIcon);
+  ui->cbAutoUpdate->setProperty("managedSetting", Settings::Gui::AutoUpdateCheck);
+  ui->cbShowVersion->setProperty("managedSetting", Settings::Gui::ShowVersionInTitle);
+  ui->cbPreventSleep->setProperty("managedSetting", Settings::Core::PreventSleep);
+  ui->comboLanguage->setProperty("managedSetting", Settings::Core::Language);
+  ui->cbRunEnterCommand->setProperty("managedSetting", Settings::Core::EnableEnterCommand);
+  ui->cbRunExitCommand->setProperty("managedSetting", Settings::Core::EnableExitCommand);
+  ui->lineCommandEnter->setProperty("managedSetting", Settings::Core::ScreenEnterCommand);
+  ui->lineCommandExit->setProperty("managedSetting", Settings::Core::ScreenExitCommand);
+  ui->groupSecurity->setProperty("managedSetting", Settings::Security::TlsEnabled);
+  ui->lineTlsCertPath->setProperty("managedSetting", Settings::Security::Certificate);
+  ui->comboTlsKeyLength->setProperty("managedSetting", Settings::Security::KeySize);
+  ui->cbRequireClientCert->setProperty("managedSetting", Settings::Security::CheckPeers);
+
+  deskflow::gui::applyManagedLocks(this, tr("Managed by your organization"));
 }
 
 void SettingsDialog::updateRequestedKeySize() const
--- a/translations/deskflow_ru.ts
+++ b/translations/deskflow_ru.ts
@@ -1088,6 +1088,10 @@ Enabling this setting will disable the server config GUI.</source>
         <translation>Выбор файла конфигурации</translation>
     </message>
     <message>
+        <source>Managed by your organization</source>
+        <translation type="unfinished">Управляется вашей организацией</translation>
+    </message>
+    <message>
         <source>Enable lock to computer at startup</source>
         <translation>Включать привязку к компьютеру при запуске</translation>
     </message>
@@ -1229,6 +1233,10 @@ Enabling this setting will disable the server config GUI.</source>
     <message>
         <source>Reset to default values</source>
         <translation type="unfinished">Сбросить до значений по умолчанию</translation>
+    </message>
+    <message>
+        <source>Managed by your organization</source>
+        <translation type="unfinished">Управляется вашей организацией</translation>
     </message>
     <message>
         <source>Automatic</source>
--- a/translations/deskflow_ru.ts
+++ b/translations/deskflow_ru.ts
@@ -182,6 +182,10 @@ p, li { white-space: pre-wrap; }
         <translation type="unfinished">Сбросить до значений по умолчанию</translation>
     </message>
     <message>
+        <source>Managed by your organization</source>
+        <translation type="unfinished">Управляется вашей организацией</translation>
+    </message>
+    <message>
         <source>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Allow the client to slow the rate it attempts to reconnect to the server when connections attempts are failing. The delay between connection attempts will  start at 1 second intervals and can be a maxium of 5 minutes between connection attempts.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</source>
         <translation type="unfinished">&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Позволить клиенту замедлять частоту попыток повторного подключения к серверу в случае неудачных попыток соединения. Интервал между попытками подключения будет начинаться с 1 секунды и может достигать максимума в 5 минут.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</translation>
     </message>
--- a/docs/user/configuration.md
+++ b/docs/user/configuration.md
@@ -283,6 +283,32 @@ screens\size=15
 ```
 
 
+## Administrator managed settings
+
+On macOS an administrator can enforce settings with a configuration profile delivered
+through MDM, such as Jamf Pro, Mosyle, Kandji or Intune. Deskflow reads managed
+preferences for its own bundle identifier, `org.deskflow.deskflow`. Platforms without a
+policy backend never report a setting as managed.
+
+Any key documented above can be enforced. The preference key is the settings key exactly
+as written, for example `security/tlsEnabled`, `core/port` or `server/enableClipboard`.
+
+A managed setting always reports the enforced value, writes to it are rejected, and the
+control that edits it is disabled in the GUI with a note explaining why. Removing the
+policy restores the value the user had before.
+
+To try this without enrolling a device, write the file directly:
+
+```bash
+sudo mkdir -p "/Library/Managed Preferences"
+sudo /usr/libexec/PlistBuddy \
+  -c "Add :security/tlsEnabled bool false" \
+  -c "Add :core/port integer 12345" \
+  "/Library/Managed Preferences/org.deskflow.deskflow.plist"
+```
+
+Delete that file to return to normal behaviour.
+
 # Server Config
 
 The `deskflow-server` command accepts the `-c` or `--config` option, which takes one argument,
--- a/src/lib/gui/StyleUtils.h
+++ b/src/lib/gui/StyleUtils.h
@@ -12,8 +12,10 @@
 #include <QIcon>
 #include <QPalette>
 #include <QStyleHints>
+#include <QWidget>
 
 #include "common/Constants.h"
+#include "common/Settings.h"
 
 namespace deskflow::gui {
 
@@ -48,6 +50,28 @@ inline void updateIconTheme()
   else
     QIcon::setFallbackThemeName(themeName);
   QIcon::setFallbackSearchPaths({QStringLiteral(":/icons/%1").arg(themeName)});
+}
+
+/**
+ * @brief Disables any child widget of @p root bound to an administrator managed setting.
+ *
+ * A widget declares which setting it edits through a dynamic property named
+ * "managedSetting" holding the settings key. Widgets whose key returns true from
+ * Settings::isManaged() are disabled and given @p tooltip to explain why.
+ *
+ * @param root The widget whose children should be checked.
+ * @param tooltip Text shown on any widget that gets locked.
+ */
+inline void applyManagedLocks(QWidget *root, const QString &tooltip)
+{
+  const auto widgets = root->findChildren<QWidget *>();
+  for (QWidget *widget : widgets) {
+    const auto key = widget->property("managedSetting").toString();
+    if (!key.isEmpty() && Settings::isManaged(key)) {
+      widget->setEnabled(false);
+      widget->setToolTip(tooltip);
+    }
+  }
 }
 } // namespace deskflow::gui
 
--- a/src/lib/gui/dialogs/ClientConfigDialog.cpp
+++ b/src/lib/gui/dialogs/ClientConfigDialog.cpp
@@ -8,6 +8,7 @@
 #include "ui_ClientConfigDialog.h"
 
 #include "common/Settings.h"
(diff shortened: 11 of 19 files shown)


src/apps/deskflow-gui/CMakeLists.txt
```diff
@@ -98,9 +98,11 @@ elseif(APPLE)
     if (APPLE_CODESIGN_DEV)
       configure_mac_codesign(${target})
     endif()
+
```
sithlord48: Remove these un needed white space changes .

src/lib/gui/config/MacOSMDMSettings.h
```diff
@@ -0,0 +1,34 @@
+/*
+ * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2024 Deskflow Contributors
```
sithlord48: Is it 2024 ?

src/lib/gui/config/MacOSMDMSettings.h
```diff
@@ -0,0 +1,34 @@
...
+#pragma once
+
+#ifdef Q_OS_MAC
+
+#include <QString>
+#include <QVariant>
+
+namespace deskflow::gui::mdm {
+
+/// Returns true if the given key has been set by MDM policy.
```
sithlord48: if your gonna describe the methods use doxygen style comments so the dev docs are filled out correctly .

src/lib/gui/config/ServerConfig.cpp
```diff
@@ -14,6 +14,10 @@
 #include <QAbstractButton>
 #include <QPushButton>
 
+#ifdef Q_OS_MAC
```
sithlord48: please use #if defined () here and in other places where you are guarding

sithlord48 (changes requested):
Dates are wrong for the copyright info .. 
very ugly w/ all the os checks needed.  but idk much can be done about that.

Who is going to maintain this for new settings...

@amagus  please look at this.

src/lib/gui/config/MacOSMDMSettings.mm
```diff
@@ -0,0 +1,126 @@
...
+#import <Foundation/Foundation.h>
+#import <CoreFoundation/CoreFoundation.h>
+#include "MacOSMDMSettings.h"
+
+// Qt headers — must come after Foundation to avoid conflicts
+#include <QString>
+#include <QVariant>
+#include <QStringList>
+
+static NSString *const kBundleID = @"org.deskflow.deskflow";
```
amagus: can't you retrieve this value programatically from the .app?

src/lib/gui/config/MacOSMDMSettings.mm
```diff
@@ -0,0 +1,126 @@
...
+///         full MDM enrollment.
+static id managedValue(NSString *key)
+{
+  // Tier 1: NSUserDefaults (works on real MDM-enrolled devices)
+  id val = [[NSUserDefaults standardUserDefaults] objectForKey:key];
+  if (val) {
+    return val;
+  }
+
+  // Tier 2: CFPreferences machine-wide (AnyUser + CurrentHost)
```
amagus: I am not sure Tier 2 and 3 are desirable, it will just cause confusion to know exactly where the settings are coming from. Stick with NSUserDefaults.

src/lib/gui/config/MacOSMDMSettings.mm
```diff
@@ -0,0 +1,126 @@
...
+{
+  return managedValue(toNSString(key)) != nil;
+}
+
+QVariant value(const QString &key)
+{
+  return toQVariant(managedValue(toNSString(key)));
+}
+
+bool boolValue(const QString &key, bool defaultValue)
```
amagus: These functions are adding yet another level to get values falling back to default values... Can't you just fetch the QVariant from Settings and inherit their default values from there? Why do we need default MDM values? Shouldn't these be the ones from the Settings? (MDM would just override these)

src/lib/gui/config/ServerConfig.cpp
```diff
@@ -435,6 +439,24 @@ size_t ServerConfig::setClipboardSharingSize(size_t size)
   swap(size, m_ClipboardSharingSize);
   return size;
 }
+bool ServerConfig::clipboardSharing() const
+{
+#ifdef Q_OS_MAC
+  if (deskflow::gui::mdm::isManaged("ClipboardSharingEnabled")) {
```
amagus: Can't we do this in a better generic way? Like a map with the allowable MDM settings and possibly a function that takes the setting name and tell if the setting can be changed or not?

src/lib/gui/dialogs/SettingsDialog.cpp
```diff
@@ -216,9 +220,21 @@ void SettingsDialog::updateText()
 
 void SettingsDialog::accept()
 {
+#ifdef Q_OS_MAC
```
amagus: We need a better way to check that, these if guards do not look good, this will become even messier if we want to allow firther MDM settings in the future.

amagus (changes requested):
I think these changes are only blocking the GUI from doing the changes. IMHO the loading of the settings should be blocking reading these settings from the config files as well. 
We also should think on this change in a more general way, it may also be useful to apply MDM settings when using Windows... Guarding all the MDM under if guards for macOS look too restrictive to me.

src/lib/gui/config/MacOSMDMSettings.h
jdipanfilo: Will do!

src/lib/gui/config/ServerConfig.cpp
jdipanfilo: Affirmative and thank you!

jdipanfilo:
Thank you very much for the detailed and insightful feedback! I've worked on the requested changes in an updated commit. I'd love to submit for another review should the team be open, and I am of course happy to make further adjustments and refinements.  

@sithlord48:
Fixed copyright year to 2026
Removed unnecessary whitespace changes in CMakeLists.txt
Changed all #ifdef Q_OS_MAC to #if defined(Q_OS_MAC)
Added Doxygen-style comments to the header

@amagus:
Dropped Tiers 2 and 3 — implementation now uses NSUserDefaults standardUserDefaults with addSuiteNamed: only
Bundle ID retrieved programmatically via [[NSBundle mainBundle] bundleIdentifier]
Replaced typed accessors with a generic isManaged() / value() API backed by a QHash key map
Moved MacOSMDMSettings from gui to common so enforcement applies at Settings::value() and Settings::setValue() — MDM values are now respected by both the GUI and deskflow-core, not just UI controls
Removed all per-setting #if defined() guards from dialogs — Settings handles enforcement automatically for any key in the map

Additionally, I've added docs/user/mdm-macos.md, docs/dev/mdm-macos.md, and deploy/macos/deskflow-mdm.mobileconfig.

Thank you again!

docs/dev/mdm-macos.md
```diff
@@ -0,0 +1,76 @@
+# macOS MDM Support — Developer Guide {#macos_mdm_dev}
```
sithlord48: Remove this useless document.

deploy/macos/deskflow-mdm.mobileconfig
```diff
@@ -0,0 +1,62 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sithlord48: What is this file for ? its in deploy but not deployed. Remove it.

docs/user/mdm-macos.md
```diff
@@ -0,0 +1,60 @@
+# macOS MDM Support
```
sithlord48: This document is also pointless. You should add a note to the configuration doc to say 

1. MDM is supported 
2. Bundle idenditifer 
3. Make a note for each item that is covered by MDM

src/apps/deskflow-gui/CMakeLists.txt
```diff
@@ -101,6 +101,7 @@ elseif(APPLE)
   else()
     set_target_properties(${target} PROPERTIES MACOSX_BUNDLE FALSE)
   endif()
+
```
sithlord48: Unrelated whitespace change remove

src/lib/common/CMakeLists.txt
```diff
@@ -9,7 +9,15 @@ endif()
 configure_file(Constants.h.in Constants.h @ONLY)
 configure_file(VersionInfo.h.in VersionInfo.h @ONLY)
 
+if(APPLE)
```
sithlord48, line 12: I feel like this would work better in the platform library that is where platform specific support for other things are.

src/lib/gui/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ set(target gui)
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
 if(APPLE)
-  set (platform_sources
+  set(platform_sources
```
sithlord48: Revert these whitespace changes

src/lib/gui/CMakeLists.txt
```diff
@@ -136,3 +136,5 @@ if(WIN32)
     net
   )
 endif()
+
```
sithlord48: Revert these whitespace changes

sithlord48 (commented):
Find a way todo this w/o all the #If definted checks. 

This should be done in a more generic so the support can work across platforms using the proper backends. 

You must use our pr template , PRS without the correct template will be closed.

sithlord48 (changes requested):
Find a way todo this w/o all the #If definted checks. 

This should be done in a more generic so the support can work across platforms using the proper backends. 

You must use our pr template, PRS without the correct template will be closed.

sithlord48:
Please mark this ready for review after you fixed the issues, rebased on current master and fixing the code format issues.

jdipanfilo:
Thank you for the tremendously helpful feedback! I've addressed the requested changes:
Changes and adjustments made:

Moved MacOSMDMSettings from common to the platform library where all other platform-specific macOS code lives (OSXScreen, OSXClipboard, etc.)
Removed docs/dev/mdm-macos.md
Removed deploy/macos/deskflow-mdm.mobileconfig
Added MDM documentation to the existing docs/user/configuration.md — covers MDM support, bundle identifier, manageable settings table, and testing instructions
Reverted all whitespace-only changes in CMakeLists.txt files
Applied clang-format v20.1.0 formatting fixes
Rebased on current master

Regarding the #if defined() guards — these are now minimal. The only guards remaining are in Settings.cpp (two) and the UI lock code in the dialogs. The guards in accept() and loadFromConfig() were removed when we moved enforcement to Settings::value() and Settings::setValue().
Regarding cross-platform — the Settings interception points are already in place for any platform. Adding Windows or Linux MDM support would only require a new platform-specific implementation file with no changes to the existing guards.
I've also verified that MDM enforcement applies continuously at the application level — not just the GUI. Settings::value() always returns the MDM value when a key is managed, and any read of a managed setting returns the MDM value regardless of what is stored in the config file. Direct edits to Deskflow.conf are overridden and corrected back to the MDM-enforced value on next launch.

Ready for re-review when you have the opportunity!

src/apps/deskflow-gui/CMakeLists.txt
```diff
@@ -98,9 +98,11 @@ elseif(APPLE)
     if (APPLE_CODESIGN_DEV)
       configure_mac_codesign(${target})
     endif()
+
   else()
     set_target_properties(${target} PROPERTIES MACOSX_BUNDLE FALSE)
   endif()
+
```
sithlord48: Please revert this whitespace change.

src/lib/common/CMakeLists.txt
```diff
@@ -32,3 +32,4 @@ target_include_directories(common
     $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/src/lib>
     $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/src/lib>
 )
+
```
sithlord48, line 44: Please revert this whitespace change.

src/lib/platform/MacOSMDMSettings.mm
```diff
@@ -0,0 +1,89 @@
+/*
+ * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2026 Deskflow Contributors
+ * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
+ */
+
+#include "MacOSMDMSettings.h"
```
sithlord48: perhaps just include this in settings (or common) since this seams to be needed from settings.. maybe need to have one platform based item there now..

src/lib/platform/MacOSMDMSettings.h
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+#pragma once
+
+#if defined(Q_OS_MAC)
+
+#include <QString>
+#include <QVariant>
+
+namespace deskflow::gui::mdm {
```
sithlord48: not a good namespace deskflow::settings::adminManagement or something as this name space would be used on linux / windows also

sithlord48 (changes requested):
Still some whitespace changes that don't need to be here.

Maybe a good idea is to make a class

common/managedSettings.h this would have the items we need and hide all the #ifDef  mac os stuff

then just have settings check 
deskflow::settings::isManaged on all oses where it ca only return true ever on mac os if mdm

jdipanfilo:
Thank you, and I greatly appreciate the feedback! I'll get to work on the needed items.

amagus:
I think that #if defined() guards are not needed at all. 
The checks will just always tell that a given configuration may be changed in platforms without MDM implemented.

jdipanfilo:
Thank you for your continued guidance and wisdom! Here is what has been addressed in the updated commit:
@sithlord48:

Created common/ManagedSettings.h and ManagedSettings.cpp — a cross-platform facade that hides all platform-specific code from callers. Settings.cpp and the dialogs now call deskflow::settings::admin::isManaged() and deskflow::settings::admin::value() with no #if defined() guards
Renamed namespace from deskflow::gui::mdm to deskflow::settings::admin — appropriate for cross-platform use
MacOSMDMSettings remains in platform/ as the internal macOS implementation, included only from ManagedSettings.cpp
Reverted all remaining whitespace changes in CMakeLists.txt files

@amagus:

#if defined() guards are now eliminated from all callers — ManagedSettings.cpp contains the only platform guards, hidden from the rest of the codebase. On non-macOS platforms isManaged() always returns false and value() always returns an invalid QVariant
Added ServerConfig::recall() MDM override — clipboard sharing is now enforced at multiple layers. When the application loads, recall() reads the stored value from Deskflow.conf and then immediately overrides it with the MDM value if managed. At read time, clipboardSharing() always checks MDM before returning the stored value. For TLS, Port, and LogLevel, Settings::value() always returns the MDM value when managed, and Settings::setValue() silently blocks any writes. This establishes a clear priority chain: MDM policy > config file > application default. Verified that direct edits to Deskflow.conf are corrected back to the MDM-enforced value on next launch — the GUI locks are purely visual indicators of this underlying enforcement

Ready for re-review when you have the opportunity and thank you again!

src/lib/platform/MacOSMDMSettings.h
```diff
@@ -0,0 +1,31 @@
...
+ * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2026 Deskflow Contributors
+ * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
+ */
+
+#pragma once
+
+#if defined(Q_OS_MAC)
+
+#include <QString>
```
sithlord48: This happens in common now.. perhaps even in `ManagedSettings` itself for mac os.. no need for this file we can do all our if defs in that class.

sithlord48 (changes requested):
Well that is much much better.

Somethings I would consider
 - Common can not link platform so move all the stuff in platform to common. (yes i know i had you move it, sorry)
 - Combine your `ManagedSettings` with the platform code. Only include the `mm` for mac os otherwise just use a header that returns false for the check. 
 - The whole only some settings are managed thing is sloppy. checks all over to see if things are in managed mode etc, This should be completely transparent to callers. And ideally not have these checks in the dialogs but provided by the ManagedSettings class

jdipanfilo:
Thank you and I once again gravely appreciate your wisdom. I will dive in and get to work!

src/lib/platform/MacOSMDMSettings.h
jdipanfilo: Follow up question before I get too far down the road regarding removing MDM checks from the dialogs. One approach would be to add a lockIfManaged(settingsKey, widget, tooltip) helper to ManagedSettings, but that would add a Qt Widgets dependency to common. 
May you advise on the preferred approach?

sithlord48: common should not link Qt Widgets.

jdipanfilo: Understood and hopefully you are not tired of me thanking you.  I have addressed points 1 and 2 locally — all MDM code will live in `common/ManagedSettings.mm` with no `platform` dependency, and `platform/MacOSMDMSettings.h/.mm` will be removed entirely.

Before pushing, I wanted to check in on point 3 — since `common` cannot link Qt Widgets, `ManagedSettings` cannot lock UI controls directly. The current enforcement at `Settings::value()` and `Settings::setValue()` is fully transparent to callers with no checks needed anywhere.

The remaining `isManaged()` calls in the dialogs are purely for visual locking (graying out controls and showing tooltips). To consolidate these, I'm considering adding a single `applyManagedLocks()` method to each dialog that maps settings keys to their corresponding widgets in one place:

```cpp
void SettingsDialog::applyManagedLocks()
{
    static const QMap<QString, QWidget *> managedWidgets = {
        {Settings::Core::Port, ui->sbPort},
        {Settings::Security::TlsEnabled, ui->groupSecurity},
        {Settings::Log::Level, ui->comboLogLevel},
    };
    for (auto it = managedWidgets.begin(); it != managedWidgets.end(); ++it) {
        if (deskflow::settings::admin::isManaged(it.key())) {
            it.value()->setEnabled(false);
            it.value()->setToolTip(tr("Managed by your organization"));
        }
    }
}
```

This keeps `ManagedSettings` widget-free, consolidates all locking logic in one place per dialog, and means adding a new managed setting only requires one map entry. Would this approach be acceptable?

jdipanfilo:
Thank you for all the feedback and guidance! Here is what has been addressed in this updated commit:

@sithlord48:

- Moved all MDM code back to `common` — `common` no longer links `platform`
- Consolidated `MacOSMDMSettings.h/.mm` and `ManagedSettings.cpp` into a single `common/ManagedSettings.mm` — on macOS this contains the full NSUserDefaults implementation with `dispatch_once` suite registration. On other platforms a `.cpp` stub returns `false`/invalid. `platform/MacOSMDMSettings.h` and `platform/MacOSMDMSettings.mm` have been removed entirely
- Replaced scattered `isManaged()` checks in the dialogs with a single `applyManagedLocks()` method per dialog — a `QList<QPair>` maps settings keys to their corresponding widgets, keeping all UI locking logic consolidated in one place. Since `common` cannot link Qt Widgets, this approach keeps `ManagedSettings` widget-free while still eliminating the scattered checks from the rest of the dialog code
- Adding a new managed setting now only requires adding one entry to the `QHash` key map in `ManagedSettings.mm` and one entry to the `applyManagedLocks()` map in the relevant dialog

The architecture remains fully extensible — adding Windows or Linux MDM support would only require providing a new platform-specific implementation file with no changes to `Settings.cpp`, the dialogs, or `ServerConfig`.

Ready for re-review when you have a chance , and I again appreciate the feedback!

src/lib/gui/config/ServerConfig.cpp
```diff
@@ -173,7 +174,10 @@ void ServerConfig::recall()
       settings().value("clipboardSharingSize", (int)ServerConfig::defaultClipboardSharingSize()).toULongLong()
   );
   setClipboardSharing(settings().value("clipboardSharing", true).toBool());
-
+  // Apply admin policy override if managed
```
sithlord48: This comment is redundant.

src/lib/common/ManagedSettings.mm
```diff
@@ -0,0 +1,111 @@
...
+  static const QHash<QString, QString> map = {
+      {Settings::Security::TlsEnabled, QStringLiteral("TLSEnabled")},
+      {Settings::Core::Port, QStringLiteral("ServerPort")},
+      {Settings::Log::Level, QStringLiteral("LogLevel")},
+      {QStringLiteral("ClipboardSharingEnabled"), QStringLiteral("ClipboardSharingEnabled")},
+  };
+  return map;
+}
+
+/// Looks up the MDM preference value for a given MDM key using NSUserDefaults.
```
sithlord48: do these comment really add any value ?

sithlord48 (changes requested):
Redundant comments. but i also notice added strings w/o new translation. This makes me suspect you are not even building the code locally...

jdipanfilo:
I am indeed building locally, but something is certainly off. I'll pull back to draft.

src/lib/common/ManagedSettings.mm
jdipanfilo: Good point, too verbose on my part. Amending :)

jdipanfilo:
Identified and resolved the missing .cpp stub for Linux/Windows/FreeBSD that was missed. Also the redundant comments have been removed. Continuing with further verification prior to a requesting review.

sithlord48:
You at least need to rebase on current master

And your commit message should not be a novel...

jdipanfilo:
The following has been addressed:

- **Rebased** on current master
- **Commit message** shortened to a concise summary

Also fixed in this commit:
- Added missing `ManagedSettings.cpp` stub for non-Apple platforms — this was the cause of the Linux/Windows/FreeBSD build failures
- Removed redundant comments identified during review.

src/lib/common/ManagedSettings.cpp
```diff
@@ -0,0 +1,23 @@
+/*
+ * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2026 Deskflow Contributors
+ * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
+ */
+
+#include "ManagedSettings.h"
+
+namespace deskflow::settings::admin {
```
amagus: Why do we need this file with classless functions? Mainly, I think this namespace is useless
Can't we declare the function in Settings class and just define it differently here or in ManagedSettings.mm? (the second one is to be used in macOS builds, and the first on all the other ones...)

src/lib/gui/config/ServerConfig.cpp
```diff
@@ -172,7 +173,9 @@ void ServerConfig::recall()
       settings().value("clipboardSharingSize", (int)ServerConfig::defaultClipboardSharingSize()).toULongLong()
   );
   setClipboardSharing(settings().value("clipboardSharing", true).toBool());
-
+  if (deskflow::settings::admin::isManaged(QStringLiteral("ClipboardSharingEnabled"))) {
```
amagus: there's no need to add code here... simply make it always fetch the managed value on calls to Settings::value

src/lib/gui/config/ServerConfig.cpp
```diff
@@ -434,6 +437,20 @@ size_t ServerConfig::setClipboardSharingSize(size_t size)
   swap(size, m_ClipboardSharingSize);
   return size;
 }
+bool ServerConfig::clipboardSharing() const
```
amagus: No need for this as well... You can add a function like Settings::isManaged that takes the config name that you want...
You will need to keep a list of manageable settings somewhere inside Settings class and nowhere else...

src/lib/gui/dialogs/SettingsDialog.cpp
```diff
@@ -535,3 +537,19 @@ void SettingsDialog::setButtonBoxEnabledButtons() const
 }
 
 SettingsDialog::~SettingsDialog() = default;
+
+void SettingsDialog::applyManagedLocks()
+{
+  const auto managed = tr("Managed by your organization");
+  const QList<QPair<QString, QWidget *>> managedWidgets = {
```
amagus: apply these locks more widely...

src/lib/gui/config/ServerConfig.cpp
jdipanfilo: Awesome feedback and thank you!

sithlord48:
thank you @amagus  for handing this review for me !

jdipanfilo:
Thank you @amagus  and @sithlord48!

Checking in on the feedback to make sure I understand. 

1. Add `Settings::isManaged()` as a static method inside the `Settings` class — the list of manageable settings lives only there
2. `Settings::value()` already returns the managed value transparently, so the extra MDM checks in `ServerConfig::recall()` and `ServerConfig::clipboardSharing()` can be removed entirely
3. `ManagedSettings.mm` (macOS) and `ManagedSettings.cpp` (other platforms) remain as platform-specific implementation files but feed into `Settings` class methods rather than a separate namespace
4. `applyManagedLocks()` should be applied more widely across the dialogs. Should applyManagedLocks() lock all writable controls that correspond to managed settings or is there a more preferred approach?

jdipanfilo:
@amagus and @sithlord48 I am leaning to keeping `Settings::isManaged("ClipboardSharingEnabled")` available for `ServerConfig` to call explicitly to avoid regressions a larger refactor could introduce.
I'd love any feedback here if the team would like this approached differently.

jdipanfilo:
- Added `Settings::isManaged(const QString &key)` as a static method — the list of manageable settings (`m_managedKeys`) now lives only inside the `Settings` class
- Removed the `deskflow::settings::admin` namespace entirely — `ManagedSettings.mm` (macOS) and `ManagedSettings.cpp` (other platforms) now expose two internal free functions (`hasManagedValue()` / `managedSettingValue()`) that `Settings` calls into
- `Settings::value()` transparently returns the managed value, and `Settings::setValue()` blocks writes for managed keys
- `ServerConfig` now calls `Settings::isManaged()` / `Settings::value()` directly

For `ClipboardSharingEnabled` — since it is read/written via `ServerConfig` using `QSettingsProxy` rather than `Settings::value()`, it can't be made fully transparent at the `Settings` level without a larger `ServerConfig` refactor. For now `ServerConfig` calls `Settings::isManaged("ClipboardSharingEnabled")` explicitly. Happy to move it fully into `Settings` keys if that's preferred.

amagus:
I still don't think you need special functions to any of the settings. Access them directly.

src/lib/gui/config/ServerConfig.cpp
amagus: This needs to be removed.

src/lib/gui/config/ServerConfig.cpp
```diff
@@ -434,6 +436,20 @@ size_t ServerConfig::setClipboardSharingSize(size_t size)
...
+{
+  static const QString kKey = QStringLiteral("ClipboardSharingEnabled");
+  if (Settings::isManaged(kKey)) {
+    return Settings::value(kKey).toBool();
+  }
+  return m_ClipboardSharing;
+}
+
+bool ServerConfig::isClipboardSharingManaged() const
```
amagus: Please remove, access the setting directly using Settings::isManaged wherever this is needed.

src/lib/gui/dialogs/ServerConfigDialog.cpp
```diff
@@ -329,6 +331,9 @@ void ServerConfigDialog::removeAction()
 
 void ServerConfigDialog::toggleClipboard(bool enabled)
 {
+  if (serverConfig().isClipboardSharingManaged()) {
```
amagus: Check the setting directly.

src/lib/gui/dialogs/ServerConfigDialog.cpp
```diff
@@ -516,3 +521,18 @@ void ServerConfigDialog::onChange()
   ui->buttonBox->button(QDialogButtonBox::Ok)
       ->setEnabled(!isAppConfigDataEqual || !(m_originalServerConfig == m_serverConfig));
 }
+
+void ServerConfigDialog::applyManagedLocks()
+{
+  const auto managed = tr("Managed by your organization");
```
amagus: You are missing the translations for this string.

jdipanfilo: Crud! Thank you.

src/lib/gui/dialogs/SettingsDialog.cpp
amagus: not addressed. Each widget can potentially check its own status instead of applying it externally.

src/lib/gui/dialogs/ServerConfigDialog.cpp
```diff
@@ -516,3 +521,18 @@ void ServerConfigDialog::onChange()
...
+  const QList<QPair<QString, QWidget *>> managedWidgets = {
+      {QStringLiteral("ClipboardSharingEnabled"), ui->cbEnableClipboard},
```
amagus: Each widget can potentially check its own status instead of applying it externally.

amagus (changes requested):
Please review the changes and ensure the manageable settings are defined in a single place. The widgets that need to be able to know the setting they are related to and fetch if it's allowed to be changed or not.

jdipanfilo:
Thank you! Here is what I am thinking for the changes.
The manageable settings list stays in a single place (`Settings::m_managedKeys`). 
For the widgets knowing their related setting, my plan is to assign each managed control a dynamic property naming its settings key. Then a single shared helper iterates the dialog's child widgets, reads that property, and locks any widget whose key returns true from `Settings::isManaged()`
Would this align with the preferred approach?

sithlord48:
seems reasonable, But i will look over the code when you have written it.

jdipanfilo:
Implemented the widget-property approach:

- `isClipboardSharingManaged()` removed — `ServerConfig` and `toggleClipboard()` now call `Settings::isManaged()` directly
- Each managed control declares its setting via a `managedSetting` dynamic property. A shared helper in `StyleUtils.h` iterates the dialog's child widgets, reads that property, and locks any whose key returns true from `Settings::isManaged()` — so the widget→setting association lives on the widget, and the managed list stays only in `Settings::m_managedKeys`
- Added the missing translation entry for "Managed by your organization" (regenerated the `.ts` files)

Please share any additional feedback when able!

sithlord48:
Please when your ready to have review done be sure to undraft / mark ready for review the PR again this will ping us that its state has changed. And we will review it again.

nbolton:
> Please when your ready to have review done be sure to undraft / mark ready for review the PR again this will ping us that its state has changed. And we will review it again.

I'm guessing you meant to leave this as draft?

jdipanfilo:
Hi @nbolton! I should have taken it out of draft so that a review could be requested as @sithlord48 called out. My apologies for the confusion I caused.
Thank you!

src/lib/common/ManagedSettings.mm
```diff
@@ -0,0 +1,99 @@
...
+// Qt headers -- must come after Foundation to avoid symbol conflicts
+#include <QHash>
+#include <QString>
+#include <QVariant>
+
+#include "Settings.h"
+
+static const QHash<QString, QString> &settingsToMDMKeyMap()
+{
+  static const QHash<QString, QString> map = {
```
sithlord48: So we can still only have these settings "managed" ? Seems very restrictive for admins trying to do such a thing.

amagus: will this list be different between macOS and other OSs? Why the list is only on the macOS file? Shouldn't this be a general setting for all OSs?

src/lib/common/ManagedSettings.mm
```diff
@@ -0,0 +1,99 @@
...
+      {Settings::Security::TlsEnabled, QStringLiteral("TLSEnabled")},
+      {Settings::Core::Port, QStringLiteral("ServerPort")},
+      {Settings::Log::Level, QStringLiteral("LogLevel")},
+      {QStringLiteral("ClipboardSharingEnabled"), QStringLiteral("ClipboardSharingEnabled")},
```
amagus: why this is different and not on Settings?

translations/deskflow_ru.ts
```diff
@@ -1262,6 +1266,10 @@ Enabling this setting will disable the server config GUI.</source>
         <source>Reset to default values</source>
         <translation type="unfinished">Сбросить до значений по умолчанию</translation>
     </message>
+    <message>
+        <source>Managed by your organization</source>
+        <translation type="unfinished"></translation>
```
amagus: add missing translations, use google translate but still keep them as unfinished.

src/lib/gui/config/ServerConfig.cpp
amagus: Still not addressed.

src/lib/gui/config/ServerConfig.cpp
```diff
@@ -172,7 +172,9 @@ void ServerConfig::recall()
       settings().value("clipboardSharingSize", (int)ServerConfig::defaultClipboardSharingSize()).toULongLong()
   );
   setClipboardSharing(settings().value("clipboardSharing", true).toBool());
-
+  if (Settings::isManaged(QStringLiteral("ClipboardSharingEnabled"))) {
```
amagus: check for all the managed settings instead of focusing on just some...

src/lib/common/Settings.h
```diff
@@ -188,6 +195,12 @@ class Settings : public QObject
   QSettings *m_stateSettings = nullptr;
   std::shared_ptr<QSettingsProxy> m_settingsProxy;
 
+  // Settings that can be enforced by an admin policy (e.g. macOS MDM).
+  inline static const QStringList m_managedKeys = {
```
amagus: this is also defined in ManagedSettings.mm... If this list is really needed, please store it in a single place

src/lib/common/Settings.cpp
```diff
@@ -289,8 +290,16 @@ QString Settings::logLevelText()
...
 
+bool Settings::isManaged(const QString &key)
+{
+  return m_managedKeys.contains(key) && hasManagedValue(key);
+}
+
 void Settings::setValue(const QString &key, const QVariant &value)
 {
+  if (Settings::isManaged(key)) {
+    return;
```
amagus: shouldn't this be an error? otherwise the user (or even our own code) may think the value got updated...

src/lib/common/ManagedSettings.mm
```diff
@@ -0,0 +1,99 @@
...
+  return managedValue(toNSString(map.value(settingsKey))) != nil;
+#else
+  Q_UNUSED(settingsKey)
+  return false;
+#endif
+}
+
+QVariant managedSettingValue(const QString &settingsKey)
+{
+#if defined(Q_OS_MAC)
```
amagus: is this really needed? can't just the linker find the correct definition from the classes?

src/lib/common/ManagedSettings.mm
```diff
@@ -0,0 +1,99 @@
...
+    return QVariant(QString::fromNSString(val));
+  }
+  return QVariant();
+}
+
+#endif // Q_OS_MAC
+
+bool hasManagedValue(const QString &settingsKey)
+{
+#if defined(Q_OS_MAC)
```
amagus: this file is MAC exclusive no need for guards...

src/lib/common/ManagedSettings.mm
```diff
@@ -0,0 +1,99 @@
+/*
+ * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2026 Deskflow Contributors
+ * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
+ */
+
+#include "ManagedSettings.h"
+
+#if defined(Q_OS_MAC)
```
amagus: this file is exclusive for macs, this guard is useless

src/lib/common/ManagedSettings.mm
```diff
@@ -0,0 +1,99 @@
...
+
+#import <Foundation/Foundation.h>
```
amagus: do you really need the entire foundation headers for your changes?

src/lib/common/ManagedSettings.mm
jdipanfilo: I'll work to expand beyond the initial settings focused on.

jdipanfilo:
Work is in progress. Pardon the delay....

jdipanfilo:
Rebased and reworked:

- The list is gone. Any setting can be enforced, and the preference key is the settings key verbatim, so there is nothing to keep in sync between `Settings` and the macOS file.
- `Settings::isManaged()` and `managedValue()` are declared in `Settings` and defined per platform in `ManagedSettings.mm` / `.cpp`, per @amagus's earlier suggestion. `ManagedSettings.h` is deleted. Tell me if I read that wrong.
- `ClipboardSharingEnabled` is just `Settings::Server::EnableClipboard` now, so the `ServerConfig` workaround is deleted.
- `setValue()` returns `false` and warns instead of silently dropping the write.
- Guards and the umbrella Foundation import are gone from the `.mm`.
- Translations filled in from Google Translate, left `unfinished`.
- Also locked `ClientConfigDialog`, which edits six client settings and had no locks — found while checking whether anything outside the two dialogs binds to a setting.

Removing the guards turned up a bug: `addSuiteNamed:` was a no-op, and plain `objectForKey:` reads any domain, so a value set with `defaults write` was treated as an enforced policy. Now uses `objectIsForcedForKey:`. Verified a profile value locks the control and a `defaults write` value does not.

sithlord48:
failing build....

jdipanfilo:
The CI failure was a missing ClientConfigDialog translation context — the "Managed by your organization" string had entries for ServerConfigDialog and SettingsDialog but not the client dialog, so the build regenerated it and tripped the "unexpected repo changes" gate. 
Regenerated the six .ts files to include it.

sithlord48:
Just an FYI we are looking at this for after 1.27.0

sithlord48:
@jdipanfilo  Drafting until we get 1.27.0 out. Once that happens I will undraft this one. It looks like it has some conflicts also.

jdipanfilo:
@sithlord48 Rebased on current master, conflicts resolved. Also fixed a fractional-value bug in the scroll-scale handling while addressing.

amagus:
> Also fixed a fractional-value bug in the scroll-scale handling while addressing.

Can we have it amended to the initial commit instead? 
I don't consider it as a bugfix worth a new commit (given it is fixing bugs from the initial commit)

src/lib/common/Settings.cpp
```diff
@@ -374,8 +374,12 @@ bool Settings::setValue(const QString &key, const QVariant &value)
 
 QVariant Settings::value(const QString &key)
 {
-  if (isManaged(key))
-    return managedValue(key);
+  // A policy holding a type with no QVariant equivalent falls through to the
```
amagus, line 377: comment not very usefull (the code itself is self documenting)

jdipanfilo: Thank you! I've amended the fractional-value fix into the initial commit and removed the comment.

PR #9707, opened 2026-04-30
fix: ensure correct screen name is set for switchToScreen

ts-kang:
## Description
Fixes a bug where incorrect screen name could be set for switchToScreen in ActionDialog

Korean:
deskflow_bug

Japanese:
deskflow_bug_2

previous implementation got the screen name by removing translated "Switch to " string from the combo box text:
```
m_action.setSwitchScreenName(ui->comboSwitchToScreen->currentText().remove(tr("Switch to ")));
```
if the translation is not as expected, it behaves weirdly as seen above.

this change stores the original screen name in user data (`Qt::UserRole`) for each combo box item and reads it back directly:
```
ui->comboSwitchToScreen->addItem(tr("Switch to %1").arg(screen.name()), screen.name());
...
m_action.setSwitchScreenName(ui->comboSwitchToScreen->currentData().toString());
```
https://doc.qt.io/qt-6/qcombobox.html#addItem

## AI tools used for this PR
none

## Related Issue
I could not find any related issue.

## How Has This Been Tested?
built using github workflow after fork
and tested on Windows 11 x64

### DIFF
--- a/src/lib/gui/dialogs/ActionDialog.cpp
+++ b/src/lib/gui/dialogs/ActionDialog.cpp
@@ -51,7 +51,7 @@ ActionDialog::ActionDialog(QWidget *parent, const ServerConfig &config, Hotkey &
 
     ui->listScreens->addItem(newListItem);
 
-    ui->comboSwitchToScreen->addItem(tr("Switch to %1").arg(screen.name()));
+    ui->comboSwitchToScreen->addItem(tr("Switch to %1").arg(screen.name()), screen.name());
     if (screen.name() == m_action.switchScreenName())
       ui->comboSwitchToScreen->setCurrentIndex(ui->comboSwitchToScreen->count() - 1);
   }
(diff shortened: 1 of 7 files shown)


src/lib/gui/dialogs/ActionDialog.cpp
```diff
@@ -51,7 +51,7 @@ ActionDialog::ActionDialog(QWidget *parent, const ServerConfig &config, Hotkey &
 
     ui->listScreens->addItem(newListItem);
 
-    ui->comboSwitchToScreen->addItem(tr("Switch to %1").arg(screen.name()));
+    ui->comboSwitchToScreen->addItem(tr("Switch to %1").arg(screen.name()), screen.name());
```
nbolton, line 54: What the...?

ts-kang: second argument of QComboBox::addItem stores custom user data
https://doc.qt.io/qt-6/qcombobox.html#addItem
https://doc.qt.io/qt-6/qcombobox.html#currentData-prop

sithlord48 (approved):
This is likely safer than the translation removal

src/lib/gui/dialogs/ActionDialog.cpp
nbolton: Oh! I totally missed that.

Edit: I wonder if it might be worth pulling a const out for readability, just a thought. Eh... probably fine.

nbolton (approved):
LGTM

PR #9695, opened 2026-04-27
[WIP] Long running Core process with file log tail

nbolton:
:warning:  WIP: Unfinished, commits missing, not ready for review.

## Description

When there is an existing Core process (e.g. GUI crashes and restarts), we do not handle this in the GUI.

This PR changes the GUI to reuse an existing Core process. It depends on #9645 which kills bad version Core processes for when GUI crashes and user attempts to fix it by updating to a newer version (we can't take over an existing process if it has a bad version).

## Disclosure of AI use

`claude-opus-4-7` to draft specific changes that I proposed, which I edited.

## Related Issue

Depends on #9645
Fixes #8879

## How Has This Been Tested?

1. Start Core from CLI
2. Run GUI

## To do

- [x] Get handover working when we see existing
- [ ] Fix merge conflicts
- [ ] Send current state back to Core IPC client
- [ ] Wire up Core process stop/start controls
- [ ] Apply config without restarting Core

PR #9716, opened 2026-05-03
Do not use Qt::DirectConnection across threads

sithlord48:
## Description

Remove the `Qt::DirectConnection` for the daemon and ipc quit connections. We will let Qt decided what to do , It will pick Qt::QueuedConnection across threads as using DirectConnection is only for objects within the same thread. 

## AI tools used for this PR

None

## Related Issue

None ? 

## How Has This Been Tested?

Needs testing on windows but i have run it locally to see if the ipc server still closes.

### DIFF
--- a/src/apps/deskflow-core/deskflow-core.cpp
+++ b/src/apps/deskflow-core/deskflow-core.cpp
@@ -131,9 +131,7 @@ int main(int argc, char **argv)
   QApplication::setApplicationName(QStringLiteral("%1 Core").arg(kAppName));
 
   const auto ipcServer = new deskflow::core::ipc::CoreIpcServer(&app); // NOSONAR - Qt managed
-  QObject::connect(
-      ipcServer, &deskflow::core::ipc::IpcServer::stopProcessRequested, coreApp, &App::quit, Qt::DirectConnection
-  );
+  QObject::connect(ipcServer, &deskflow::core::ipc::IpcServer::stopProcessRequested, coreApp, &App::quit);
   ipcServer->listen();
 
   QThread coreThread;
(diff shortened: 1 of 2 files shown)


nbolton (commented):
Getting some intermittent race bugs with this change, where the IPC messages sometimes don't arrive in the right order. I'll do some more testing over the next few days to understand why.

I remember I had to use direct connection to make IPC more stable/reliable, but I've forgotten why.

sithlord48:
Try forcing QueuedConnection but that should be what "Auto" is picking . If this needs to block you can do BlockingQueuedConnection

nbolton:
> Try forcing QueuedConnection but that should be what "Auto" is picking . If this needs to block you can do BlockingQueuedConnection

Sure, let's try that.

Pretty consistently, when I start the GUI on Windows, I see this error on the `qtDecidedConnection` branch (doesn't happen on `master`):

image

Though this is perhaps because of a deeper root cause: when you start the GUI it sends the start command to the Daemon even if the Core is already running, so we should probably fix that first before trying to paper over the problem by fiddling with the Qt event connection type in this PR, if that makes sense. Instead, we can either try to land #9695 first, which is designed to re-use an existing running Core, or just stop the GUI telling the Daemon to run the Core when it starts (that'll probably be fine, as long as the Core tells the GUI it's current client/server connection states on IPC connect, but I can't remember if I did that yet... maybe not?)

sithlord48:
Sure go for fixing the known bad first. Like i said i changed this as this is generally the wrong way to connect two objects on seperate threads.

sithlord48:
For now i will close this as we need to check a few other things first.

nbolton:
> For now i will close this as we need to check a few other things first.

Sounds good, I'll avoid `DirectConnection` in future and try `BlockingQueuedConnection`.

sithlord48:
Try to not force a connection type at all if you can.

nbolton:
> Try to not force a connection type at all if you can.

Normally I don't but in this case it "fixed" (or masked) some race bugs, which is usually the only reason I ever block.

Issue #3744, opened 2014-10-13
Screen edge scaling is not intuitive


nbolton:
... uses largest of multiple monitor dimensions, instead of "nearest" monitor's dimensions.

**Imported issue:**
- Author: Brian Porter
- Date: 2013-10-31 20:10:28
- Legacy ID: 3814
## Steps to reproduce
1. Configure a Synergy server with a single monitor and a client with multiple monitors each having different resolutions, or where the multiple monitors are offset from one another so the tops and bottoms don't "line up" perfectly.
2. _(Different resolutions or having the client's monitors offset is key to this issue!)_
3. _(Operating systems do not seem to matter in my tests, but my server is Win7 x64 and my client is OS X 10.8.5.)_
4. Configure the client's monitors so that the smaller of the two is in the middle, with the larger client monitor on one side and the server's monitor on the other.
5. Configure Synergy so that the virtual screen arrangement mirrors the physical arrangement described above. For example: 
   
   `Large client monitor <-> Small client monitor <-> Server monitor`
6. In this configuration, switching the cursor between Synergy's screens will cause it to be positioned relative to the height of the client's larger outside monitor instead of relative to the client's smaller middle monitor.
7. [This diagram](https://www.dropbox.com/s/rcg6j7ebcokgrez/synergy_1.4.15_screen_edge_scaling_issue.png) will help illustrate the issue further.
## Expected

Given the above 3(+) monitor configuration; the cursor should be positioned on the destination monitor using the percentage scaled position of the cursor when leaving the originating machine, as compared to the dimension (width or height) of the **exiting monitor**.
## Actual

The cursor is positioned on the destination monitor using the percentage-scaled position of the cursor when leaving the originating machine, as compared to the _combined maximum dimension (width or height) of **all** monitors_ on the client.

The effect of this bug looks a lot like setting fractional and expanded transition dimensions in your config. **I realize the following config is invalid, but imagine the consequences of it:**

```
section: links
    Client:
        right(0,100) = Server(40,90)
    Server:
        left(0,100) = Client(-40,110)  # Invalid, but use your imagination and remember values are in percentages.
end
```
## Example

Given the above 3 monitor configuration:
- Cursor starts on the client.
- Cursor travels right off the edge of the machine and should transition to the server. 
[...]

PR #9723, opened 2026-05-07
Fix Windows bottom-edge switching on uneven multi-monitor layouts

tdw46:
## Description
Fix bottom-edge switching when the desktop bounds extend below a visible monitor edge.

The edge math is now shared in `base` and used by Windows, macOS, X11, EI, and GUI config generation.

## AI tools used for this PR
OpenAI Codex, GPT-5, was used to investigate the bug, edit code, run local checks, and update this draft PR. Estimated manual coding: 70%, AI coding: 30%. I needed to clean up some of the files. 

## Related Issue
None filed.

## How Has This Been Tested?
Windows 11, Qt 6.7.3, MSVC 2022:

- `clang-format` 20.1.0 over `src/`
- `git diff --check`
- Release build with `CMAKE_COMPILE_WARNING_AS_ERROR=ON`
- `ScreenEdgesTests.exe -txt`
- `ServerConfigTests.exe -txt`

`reuse lint` was run on a clean Windows worktree. It still reports existing extensionless icon alias files under `src/apps/res/icons/...`; this PR does not touch those files.

macOS and Linux builds were not run locally.

### DIFF
--- a/src/lib/base/CMakeLists.txt
+++ b/src/lib/base/CMakeLists.txt
@@ -1,3 +1,4 @@
+# SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
 # SPDX-FileCopyrightText: (C) 2024 Chris Rizzitello <sithlord48@gmail.com>
 # SPDX-FileCopyrightText: (C) 2012 - 2024 Symless Ltd
 # SPDX-FileCopyrightText: (C) 2009 - 2012 Nick Bolton
@@ -27,6 +28,8 @@ add_library(base STATIC
   Log.h
   LogLevel.h
   PriorityQueue.h
+  ScreenEdges.cpp
+  ScreenEdges.h
   SimpleEventQueueBuffer.cpp
   SimpleEventQueueBuffer.h
   Stopwatch.cpp
--- a/src/lib/base/ScreenEdges.cpp
+++ b/src/lib/base/ScreenEdges.cpp
@@ -1,0 +1,551 @@
+/*
+ * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
+ * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
+ */
+
+#include "base/ScreenEdges.h"
+
+#include "base/DirectionTypes.h"
+
+#include <algorithm>
+
+namespace deskflow {
+
+namespace {
+
+struct EdgeSegment
+{
+  int32_t edge = 0;
+  int32_t start = 0;
+  int32_t end = 0;
+};
+
+int32_t right(const ScreenRect &rect)
+{
+  return rect.x + rect.w;
+}
+
+int32_t bottom(const ScreenRect &rect)
+{
+  return rect.y + rect.h;
+}
+
+bool contains(const ScreenRect &rect, int32_t x, int32_t y)
+{
+  return x >= rect.x && x < right(rect) && y >= rect.y && y < bottom(rect);
+}
+
+bool containsAny(const std::vector<ScreenRect> &screens, int32_t x, int32_t y)
+{
+  for (const auto &screen : screens) {
+    if (contains(screen, x, y)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+uint32_t sideMask(Direction side)
+{
+  switch (side) {
+    using enum Direction;
+  case Left:
+    return static_cast<uint32_t>(DirectionMask::LeftMask);
+  case Right:
+    return static_cast<uint32_t>(DirectionMask::RightMask);
+  case Top:
+    return static_cast<uint32_t>(DirectionMask::TopMask);
+  case Bottom:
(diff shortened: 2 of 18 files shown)


sithlord48:
This needs to follow our [contribution guidelines](https://github.com/deskflow/deskflow/wiki/Hacking-Guide) before we will review it. 

This also fixes a cross platform problem for one platform.

tdw46:
Thanks! I’ll take a look

tdw46:
Updated for the hacking guide: rewrote the branch to one typed `fix:` commit, replaced the PR body with the template, added copyright/license updates, moved the edge math into shared base code, and wired it into Windows/macOS/X11/EI plus GUI config generation. Local Windows build/tests pass; macOS/Linux builds need CI.

src/lib/platform/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+# SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
```
sithlord48: No changes were made to this file in this PR please revert this change

sithlord48 (commented):
What about other edges ?  
@nbolton  iirc the current way for this is to add up(x,y) or down(x,y) to the link section ?

tdw46:
Good point. I only handled the bottom-edge case that reproduced in this layout.

My read is that the existing link syntax can already describe constrained ranges with `up(x,y)` / `down(x,y)`, and this patch is using that path for the below-screen case. I'm not sure whether the right answer is to generalize this same projection to left/right/top in this PR or keep this scoped to the observed bottom-edge bug, so I'll leave that for @nbolton to confirm before I expand it.

I also reverted the accidental copyright-only change in `src/lib/platform/CMakeLists.txt`; I removed the same no-op touch from `src/unittests/platform/CMakeLists.txt` too.

pLum0:
Great PR — the per-monitor rect tracking and `ScreenEdges` helper are exactly the infrastructure needed for a related long-standing issue.

**Related Issue:** #3744 ("Screen edge scaling is not intuitive") — open since 2013. The core problem: when a client has multiple monitors with different vertical positions (e.g., an internal laptop display offset lower than the external display), cursor transition coordinates are mapped against the combined bounding box instead of the individual monitor at the edge. This causes the cursor to appear at the wrong Y position when transitioning left/right between machines.

**How it relates to this PR:** PR #9723 already introduces `m_monitorRects` on every platform and the `ScreenRect`/`ScreenEdges` helpers. That's most of the plumbing needed to fix #3744. The missing piece is a client-side coordinate remapping:

1. **`remapEntryCoordinates(bounds, monitors, x, y)`** in `ScreenEdges` — recovers the fraction the server computed against the bounding box and remaps it to the edge monitor's coordinate space
2. **`getMonitorRects()`** virtual method on `PlatformScreen` — exposes per-monitor rects to the remapping logic
3. **Call it from `Client::enter()`** before `mouseMove()` — the client already knows its bounds and individual monitors; it can correct the server's bounding-box-based coordinates locally
4. **Fix EiScreen bugs** — `m_x/m_y/m_w/m_h` are `uint32_t` (should be `int32_t`; negative offsets are silently lost), and `updateShape()` stores max edge positions instead of actual width/height (`max_x - min_x`)

This approach requires **no protocol changes** — the server sends coordinates as before, and the client remaps them using its local monitor geometry.

**Question:** Would you be open to including this in the same PR, or should I open a separate PR that builds on top of this one? The changes are closely related (same infrastructure) but address different symptoms (bottom-edge holes vs. general left/right/top/bottom coordinate mapping).

sithlord48:
@pLum0  please dont use LLMs to make comments. 

This pr should cover all screen directions if it's gonna fix one of them .

tdw46:
Thanks all for the input here. I expanded this to cover all four edges now, and added client-side remapping so entry coordinates are corrected against the client monitor geometry without a protocol change.

I tested the Windows/server build locally, including the top-edge hole case. I’m keeping this as draft for now while I rebuild/test the Mac client side for the coordinate mapping behavior.

Checks run locally:
- Release build: `deskflow`, `deskflow-core`, `ScreenEdgesTests`, `ServerConfigTests`
- `ScreenEdgesTests`: 16 passed
- `ServerConfigTests`: 9 passed
- `git diff --check`

tdw46:
﻿Quick update: I tested the previous client-side coordinate remap on macOS and it broke return paths through the hole: exiting upward landed on the wrong Windows monitor, and side exits did not cross as expected.

I rewrote that part to keep the mapping server/config-side instead: remove the client enter remap, emit extra partial links for the hole's bordering edges, and leave the platform visible-edge projection to do the coordinate work. Keeping this draft while I retest the Mac build.

(edit: this still has some issues, so i'll do further testing in yet another branch until i get it working then merge into this PR)

sithlord48:
You can test here if you like but if you do so on your fork you can run CI there and also check the linting etc will pass.

sithlord48:
this has been stale for a while @tdw46  i will close this for now you can re open it when you have had time to do the updates discussed

Issue #8492, opened 2025-04-15
Specific key (vk1Dsc7B) not recognized


superuser7777:
### Project

Deskflow

### Deskflow version info

**server
Deskflow: 1.21.2.0 (daa09546)
Qt: 6.9.0
System: Windows 10 Version 21H2(japanese)

**client
Deskflow: 1.21.2.0 (daa09546)
Qt: 6.9.0
System: Windows 10 Version 22H2(japanese)

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [x] Windows
- [ ] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [ ] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [ ] I am not using Wayland on Linux

### Signing on macOS

- [ ] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### Continuous build

- [ ] I have tried the latest continuous build and the issue persists
- [x] I am unable to try the latest continuous build

### OS versions/distros

**server
System: Windows 10 Version 21H2(japanese)

**client
System: Windows 10 Version 22H2(japanese)

### Deskflow configuration

**server
Deskflow: 1.21.2.0 (daa09546)
Qt: 6.9.0
System: Windows 10 Version 21H2(japanese)

**client
Deskflow: 1.21.2.0 (daa09546)
Qt: 6.9.0
System: Windows 10 Version 22H2(japanese)

### What steps will reproduce the problem?

Detect "vk1Dsc7B" key when keyhooked via autohotkey on client side. Not detected.

vk1Dsc7B: 無変換(japanese key name)

### Log output

```shell

```

superuser7777:
[update]

The problem continues even with the latest build.
Probably, the maintainers are using keyboards that do not have the vk1Dsc7B key physically. I understand that it is a low priority. The "vk1Dsc7B" key is an important key for converting IME in a multibyte environment.

I only have python, shell, and ahk coding skills, but is there anything I can help you when debugging?

```
**server
Deskflow: 1.22.0.0 (00962837)
Qt: 6.9.0
System: Windows 10 Version 21H2(japanese)

**client
Deskflow: 1.22.0.0 (00962837)
Qt: 6.9.0
System: Windows 11 Version 24H2(japanese)
```

[...]

PR #9736, opened 2026-05-10
fix: add Muhenkan (Non-Convert) key support for Japanese keyboard

takatu-kazusige:
## Problem

Currently, Deskflow does not support the "Muhenkan" (Non-Convert) key, which is essential for Japanese keyboard layouts. While a bug report exists (#8492), it remains unresolved. The key is currently ignored.

## Solution

This PR adds full support for the Muhenkan key by:

1. Defining a new `KeyID` for Muhenkan based on its KeySym value.
2. Adding the key to the global key name map.
3. Updating the Windows virtual key (VK) table to recognize `VK_NONCONVERT`.
4. Ensuring proper mapping for both Windows and Linux through existing conversion logic.

## Changes

- **KeyTypes.h**: Defined `kKeyMuhenkan` as `0xEF22`.
- **KeyTypes.cpp**: Added `"Muhenkan"` to `KeyNameMapEntry`.
- **MSWindowsKeyState.cpp**: Updated `s_virtualKey` table to map `0x01d` and `0x11d` (`VK_NONCONVERT`) to `kKeyMuhenkan` instead of `kKeyNone`.

### Testing Environment

* **Server OS**: Windows 11 (Japanese Keyboard)
* **Client OS**: Linux (Ubuntu 26.04/Wayland)
* **Verification**: Built from source with the proposed changes.

### Test Scenarios

1. **Client-side Event Verification (Linux)**:
* Used `xev` on the Linux client to monitor incoming key events from the Windows server.
* **Result**: Confirmed that pressing the "Muhenkan" key on the server correctly generates the following event on the client:
`keycode 102 (keysym 0xff22, Muhenkan)`

2. **Basic Regression**:
* Verified that the **Henkan** (Convert) key and other Japanese-specific keys continue to function normally without any interference from the new mapping.

## Technical Details (Review Notes)

* **KeyID Selection**: `kKeyMuhenkan` is set to `0xEF22`. This aligns with `XK_Muhenkan` (`0xff22`) because Deskflow's internal KeyID for special keys is derived by the bitwise operation `(KeySym & 0x0FFF) | [cite_start]0xE000` (or similar logic used in `XDGKeyUtil`).

* **Cross-Platform Impact**:
  * **Windows**: The `s_virtualKey` update allows the server to capture the key and the client to replay it via the automatically generated `m_keyToVKMap`.
  * **Linux**: `XDGKeyUtil::mapKeySymToKeyID()` will handle the conversion automatically once the KeyID is defined.
  * **macOS**: No changes made as the key does not exist on standard Mac layouts, but compatibility is maintained.

### DIFF
--- a/src/lib/deskflow/KeyTypes.cpp
+++ b/src/lib/deskflow/KeyTypes.cpp
@@ -73,6 +73,7 @@ const KeyNameMapEntry kKeyNameMap[] = {
     {"F35", kKeyF35},
     {"Find", kKeyFind},
     {"Help", kKeyHelp},
+    {"Muhenkan", kKeyMuhenkan},
     {"Henkan", kKeyHenkan},
     {"Home", kKeyHome},
     {"Hyper_L", kKeyHyper_L},
--- a/src/lib/deskflow/KeyTypes.h
+++ b/src/lib/deskflow/KeyTypes.h
@@ -114,6 +114,7 @@ static const KeyID kKeyPause = 0xEF13;  /* Pause, hold */
 static const KeyID kKeyScrollLock = 0xEF14;
 static const KeyID kKeySysReq = 0xEF15;
 static const KeyID kKeyEscape = 0xEF1B;
+static const KeyID kKeyMuhenkan = 0xEF22;         /* Cancel Conversion */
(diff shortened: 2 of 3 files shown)


sithlord48 (changes requested):

Issue #9739, opened 2026-05-11
Client running in service mode cannot connect to the server


redsigma:
### Sanity checks

- [x] I have done the sanity checks, and my issue persists
- [x] These sanity checks are not relevant to the bug

### Deskflow version info

Deskflow: 1.26.0.177 (c3ecd0b8)
Qt: 6.10.2
System: Windows 10 Version 22H2

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [x] Windows
- [ ] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Deskflow configuration

- Windows 10 22H2, server
- Windows 11 23H2, client
- Debian 13 Trixie, client

### What steps will reproduce the problem?

1. Start deskflow client  in service mode
2. Try to connect to server
3. Logs are constantly saying `failed to verify server certificate fingerprint`

### Log output

```shell
[2026-05-11T11:47:40.122] INFO: starting client process (service mode)
[2026-05-11T11:47:40.122] INFO: sending start to daemon (config file: C:/apps/common/deskflow/settings/Deskflow.conf)
[2026-05-11T11:47:40.240] WARNING: QCoreApplication::applicationDirPath: Please instantiate the QApplication object first
[2026-05-11T11:47:40.241] INFO: initial settings file: C:/Users/cristian.cumparatu/AppData/Roaming/Deskflow/Deskflow.conf
[2026-05-11T11:47:40.241] INFO: settings file changed: C:/apps/common/deskflow/settings/Deskflow.conf
[2026-05-11T11:47:40.246] WARNING: SetProcessDpiAwarenessContext() failed: Access is denied.
Qt's default DPI awareness context is DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2. If you know what you are doing, you can overwrite this default using qt.conf (https://doc.qt.io/qt-6/highdpi.html#configuring-windows).
[2026-05-11T11:47:40.254] NOTE: starting core
[2026-05-11T11:47:40.271] NOTE: configured 1 server address(es)
[2026-05-11T11:48:04.708] INFO: stopping core process (service mode)
[2026-05-11T11:48:04.712] INFO: starting client process (service mode)
[2026-05-11T11:48:04.712] INFO: sending start to daemon (config file: C:/apps/common/deskflow/settings/Deskflow.conf)
[...]

Issue #9738, opened 2026-05-11
Portable mode not working in server mode


redsigma:
### Sanity checks

- [x] I have done the sanity checks, and my issue persists
- [x] These sanity checks are not relevant to the bug

### Deskflow version info

Deskflow: 1.26.0.177 (c3ecd0b8)
Qt: 6.10.2
System: Windows 10 Version 22H2

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [x] Windows
- [ ] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Deskflow configuration

- Windows 10 22H2  , server
- Windows 11 23H2 , client
- Debian 13 Trixie, client

### What steps will reproduce the problem?

1. Put all the Deskflow.conf and deskflow-server.conf in your portable ./settings/ folder
2. Start deskflow in server  mode on windows with processMode=1 (no service)
3. Press Start to start the server mode
4. Will crash with exit code 4 saying that config isnt found

### Log output

```shell
[2026-05-11T11:06:26.560] INFO: starting server process (desktop mode)
[2026-05-11T11:06:26.561] INFO: core config file: F:/Apps/common/deskflow/settings/deskflow-server.conf
[2026-05-11T11:06:26.561] INFO: running command: F:/Apps/common/deskflow3/deskflow-core.exe server
[2026-05-11T11:06:26.592] WARNING: QCoreApplication::applicationDirPath: Please instantiate the QApplication object first
[2026-05-11T11:06:26.597] WARNING: SetProcessDpiAwarenessContext() failed: Access is denied.
Qt's default DPI awareness context is DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2. If you know what you are doing, you can overwrite this default using qt.conf (https://doc.qt.io/qt-6/highdpi.html#configuring-windows).
[2026-05-11T11:06:26.620] ERROR: cannot open configuration "C:/ProgramData/Deskflow/deskflow-server.conf"
[2026-05-11T11:06:26.620] FATAL: deskflow-core.exe: failed to load config: C:/ProgramData/Deskflow/deskflow-server.conf
[2026-05-11T11:06:26.620] ERROR: core stopped with error code: 4
[2026-05-11T11:06:26.593] INFO: initial settings file: C:/Users/Red/AppData/Roaming/Deskflow/Deskflow.conf
[...]

PR #9742, opened 2026-05-11
Use AbsoluteFilePath for all settings returned filepaths

sithlord48:
## Description

make sure to use absoluteFilePath for returned settings files 

## AI tools used for this PR

None

## Related Issue

hopefully path issues on windows (#9739. #9738)

## How Has This Been Tested?

locally run on linux 
@redsigma can you see if this helps your issue on windows

### DIFF
--- a/src/lib/common/Settings.cpp
+++ b/src/lib/common/Settings.cpp
@@ -263,7 +263,7 @@ bool Settings::isPortableMode()
 
 QString Settings::settingsFile()
 {
-  return instance()->m_settings->fileName();
+  return QFileInfo(instance()->m_settings->fileName()).absoluteFilePath();
 }
 
 QString Settings::settingsPath()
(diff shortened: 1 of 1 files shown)


redsigma:
im building it now on my side to see if it fixes it

redsigma:
just tested and it didnt fix the issue.

nbolton (approved):

nbolton:
> just tested and it didnt fix the issue.

Ah... just saw this.

@sithlord48 Is there any point landing if it doesn't fix the bug?

sithlord48:
Yes its the "proper" way we should be returning these paths in some cases it will help like if we pass a path to a program like `--settings` . Tldr it won't break anything current but may help in some very niche cases since we hardly pass paths around anyway .

redsigma:
just for consistency, i tried the latest changes of this PR, but it still didnt fix the issue on windows. I'll update the other PR when i get the minimum changes there.

PR #9744, opened 2026-05-12
feat: add search evdev.xml path for NetBSD/OpenBSD

jg1uaa:
## Description
Currently, searching evdev.xml code in src/lib/deskflow/unix/AppUtilUnix.cpp supports FreeBSD and Linux only.
To support other *BSDs, such as NetBSD and OpenBSD, adding suitable path is needed.

(I am now struggling to run deskflow on OpenBSD-7.8/amd64.)

## AI tools used for this PR
None

## Related Issue
None

## How Has This Been Tested?
run `deskflow-core server` on FreeBSD and Linux(Debian-13), and no error message related evdev.xml.

### DIFF
--- a/src/lib/deskflow/unix/AppUtilUnix.cpp
+++ b/src/lib/deskflow/unix/AppUtilUnix.cpp
@@ -46,9 +46,15 @@ std::vector<std::string> AppUtilUnix::getKeyboardLayoutList()
 
 #if WINAPI_XWINDOWS
   // Check /usr/local first used on bsd and some systems
-  m_evdev = "/usr/local/share/X11/xkb/rules/evdev.xml";
-  if (!std::filesystem::exists(m_evdev))
-    m_evdev = "/usr/share/X11/xkb/rules/evdev.xml";
+  std::vector<std::string> evdev_candidate = {
+      "/usr/share/X11/xkb/rules/evdev.xml",       // Linux
+      "/usr/local/share/X11/xkb/rules/evdev.xml", // FreeBSD, DragonFlyBSD
+      "/usr/X11R7/lib/X11/xkb/rules/evdev.xml",   // NetBSD
+      "/usr/X11R6/share/X11/xkb/rules/evdev.xml", // OpenBSD
+  };
+
+  for (auto it = evdev_candidate.begin(); it != evdev_candidate.end() && !std::filesystem::exists(m_evdev = *it); it++)
+    ;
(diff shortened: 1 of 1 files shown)


sithlord48 (changes requested):
I thought we did this .. maybe its just a single folder we checked 

swap the list order so that linux is checked first. Its the most common case so we should check for it first.

+1 for reading the guidelines !

sithlord48:
Drafted until its updated. make sure when you do update to rebase upon the master branch. If you do not plan to update this let us know and we will update it and 'fix' the review issues.

jg1uaa:
list order has changed, Linux first. Please review with newer code. Thanks.

sithlord48 (changes requested):
Please squash your commits into one . After that i think this is good to go !

sithlord48 (changes requested):
Looks like the code format need a bit of update also be sure to rebase upon master (do not add a new commit to "fix" the format append your current one)

src/lib/deskflow/unix/AppUtilUnix.cpp
```diff
@@ -45,9 +45,15 @@ std::vector<std::string> AppUtilUnix::getKeyboardLayoutList()
...
-  if (!std::filesystem::exists(m_evdev))
-    m_evdev = "/usr/share/X11/xkb/rules/evdev.xml";
+  std::vector<std::string> evdev_candidate = {
+      "/usr/share/X11/xkb/rules/evdev.xml",       // Linux
+      "/usr/local/share/X11/xkb/rules/evdev.xml", // FreeBSD, DragonFlyBSD
+      "/usr/X11R7/lib/X11/xkb/rules/evdev.xml",   // NetBSD
+      "/usr/X11R6/share/X11/xkb/rules/evdev.xml", // OpenBSD
+  };
+
+  for (auto it = evdev_candidate.begin(); it != evdev_candidate.end() && !std::filesystem::exists(m_evdev = *it); it++)
```
sithlord48, line 56: @nbolton well that is just ugly formatting ..

sithlord48 (approved):

nbolton (approved):
We really need to use `xkbcommon` for this.

src/lib/deskflow/unix/AppUtilUnix.cpp
nbolton: Oof

PR #9789, opened 2026-05-26
fix(client): restore cursor positions locally In relative mouse mode

### LINKED ISSUE #9788: relativeMouseMoves resets client cursor to top-left (0,0) when switching screens

### Deskflow version info

**Server**
Deskflow: 1.26.0.207 (https://github.com/deskflow/deskflow/commit/62a8ee749f0ec36b60fae3282143446f1b872d05)
Qt: 6.10.2
System: macOS Sequoia (15.7.5)

**Client**
Deskflow: 1.26.0.207 (https://github.com/deskflow/deskflow/commit/62a8ee749f0ec36b60fae3282143446f1b872d05)
Qt: 6.10.2
System: Windows 10 Version 21H2

### Build types

Official release (installer/package from GitHub releases)

### Deskflow configuration

* macOS server (primary), Windows client (secondary)
* relativeMouseMoves =  true
* defaultLockToScreenState = true
* Two keystrokes to switch in direction between two devices

1. Set up macOS as the Deskflow server and Windows as the client.
2. Enable relative mouse moves (relativeMouseMoves = true) and lock to screen (defaultLockToScreenState = true).
3. Trigger a hotkey to switch to win and move the cursor to any position on the Windows client (e.g., center of the screen).
4. Trigger a hotkey to switch focus back to the macOS server.
5. Switch back to the Windows client.

**Expected**: The Windows cursor should restore to the last position where it was left, identical to the behavior when `relativeMouseMoves = false`.
**Actual**: The Windows cursor is restored to the top-left corner of the screen (0, 0).

### Log output

_No response_

### Additional information

When `relativeMouseMoves = true` and the cursor is locked, `Server::onMouseMoveSecondary() `intentionally returns early without updating the server's absolute coordinate model (m_x/m_y). This is done because relative movement in games can warp the cursor, making the server's absolute tracking unreliable.

However, when switching away from the client via `jumpToScreen()`, the server saves the active client's position using these stale coordinates (setJumpCursorPos(m_x, m_y)). Since `BaseClientProxy` initializes the saved jump position to `(0, 0)`, the server saves `(0, 0)` and subsequently sends it via `CEnter` when the user switches back.

MomoyamaSawa:
## Description
**Root Issue:**

In relative mouse mode, the server intentionally stops updating its absolute coordinate model (m_x/m_y) to avoid drift from in-game cursor warping. When switching away, it saves this stale coordinate as the jump position. Since BaseClientProxy initializes the jump position to (0, 0), the next enter sends (0, 0) to the client, resetting the cursor to the top-left.

This affects any use of relative mouse moves, not just games — switching back to the client on a plain desktop also causes the cursor to jump to (0, 0), forcing users to toggle the option off for non-game sessions.

**Solution:**

server-side tracking is unreliable due to in-game cursor warping In relative mode, whereas the client can always get its exact physical position. So this fix delegates position-restoration memory to the client. The client caches its actual local cursor position on leave and restores it on enter. On Windows, coordinates are saved per-desktop to prevent UAC and secure desktop transitions from polluting the position.

**Scope:** relativeMouseMoves enabled path only. Normal absolute movement paths are unaffected.

## AI tools used for this PR
**Codex GPT-5.5 High:** help identify root cause & check change impact.
**Claude Code Sonnet 4.6 Adaptive:** Assist with cross-validating change impact; help me understand the project architecture.

## Fixed/related issues
fixes: #9788 

## How has this been tested?
**Environment & Config:**

* macOS 15.7.5, Apple Silicon — server (official release v1.26.0.207)
* Windows 10 LTSC (build 19044) — client (debug and release build from this branch, commit dbd5f5fd)
* relativeMouseMoves = true, defaultLockToScreenState = true, hotkey: switchInDirection
* Windows display: 1920×1080, 100% scaling

**Checks:**

1. Ran clang-format 20.1.0 on modified files.
2. Built on Windows: CMake 4.3.2, Ninja 1.13.2, MSVC 19.44.35207, Qt 6.10.2 — build succeeded, unit tests 23/23 passed.
3. macOS and Linux compilation not verified locally; relying on CI. MSWindowsDesks.cpp is Windows-only; Client.cpp changes are minimal conditional logic.

**Manual validation:**

1. Start and connect the server and the client.
2. Use hotkey to switch between mac and win.
3. repeat and observe.

**Before fix:** cursor restored to (0, 0) on each switch from macOS to Windows.
**After fix:** cursor restored to the last known position on the Windows desktop. Verified on both debug and release builds, including both on the UAC secure desktop itself and within UAC-elevated application windows.
**Regression check (relativeMouseMoves = false):** behavior unchanged.

### DIFF
--- a/src/lib/client/Client.cpp
+++ b/src/lib/client/Client.cpp
@@ -1,6 +1,6 @@
 /*
  * Deskflow -- mouse and keyboard sharing utility
- * SPDX-FileCopyrightText: (C) 2025 Deskflow Developers
+ * SPDX-FileCopyrightText: (C) 2025 - 2026 Deskflow Developers
  * SPDX-FileCopyrightText: (C) 2012 - 2016, 2026 Synergy App Ltd
  * SPDX-FileCopyrightText: (C) 2002 Chris Schoeneman
  * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
@@ -154,6 +154,9 @@ void Client::handshakeComplete()
 {
   m_ready = true;
   m_screen->enable();
+  if (m_relativeMouseMoves && !m_hasRelativeRestorePosition) {
+    saveRelativeRestorePosition();
+  }
   sendEvent(EventTypes::ClientConnected);
 }
 
@@ -195,12 +198,20 @@ void Client::getCursorPos(int32_t &x, int32_t &y) const
 void Client::enter(int32_t xAbs, int32_t yAbs, uint32_t, KeyModifierMask mask, bool)
 {
   m_active = true;
+  if (m_relativeMouseMoves && m_hasRelativeRestorePosition) {
+    xAbs = m_relativeRestoreX;
+    yAbs = m_relativeRestoreY;
+    LOG_VERBOSE("using relative restore position: %d,%d", xAbs, yAbs);
+  }
   m_screen->mouseMove(xAbs, yAbs);
   m_screen->enter(mask);
 }
 
 bool Client::leave()
 {
+  if (m_relativeMouseMoves) {
+    saveRelativeRestorePosition();
+  }
   m_active = false;
 
   m_screen->leave();
@@ -283,6 +294,8 @@ void Client::screensaver(bool activate)
 
 void Client::resetOptions()
 {
+  m_relativeMouseMoves = false;
+  m_hasRelativeRestorePosition = false;
   m_screen->resetOptions();
 }
 
@@ -303,6 +316,14 @@ void Client::setOptions(const OptionsList &options)
       if (index != options.end()) {
         m_maximumClipboardSize = *index;
       }
+    } else if (id == kOptionRelativeMouseMoves) {
+      index++;
+      if (index != options.end()) {
+        m_relativeMouseMoves = (*index != 0);
+        if (m_relativeMouseMoves && m_ready && !m_hasRelativeRestorePosition) {
+          saveRelativeRestorePosition();
+        }
+      }
     }
   }
 
@@ -312,6 +333,13 @@ void Client::setOptions(const OptionsList &options)
   }
 
   m_screen->setOptions(options);
+}
+
+void Client::saveRelativeRestorePosition()
+{
+  m_screen->getCursorPos(m_relativeRestoreX, m_relativeRestoreY);
+  m_hasRelativeRestorePosition = true;
+  LOG_VERBOSE("saved relative restore position: %d,%d", m_relativeRestoreX, m_relativeRestoreY);
 }
 
 std::string Client::getName() const
--- a/src/lib/client/Client.h
+++ b/src/lib/client/Client.h
@@ -1,6 +1,6 @@
 /*
  * Deskflow -- mouse and keyboard sharing utility
- * SPDX-FileCopyrightText: (C) 2025 Deskflow Developers
+ * SPDX-FileCopyrightText: (C) 2025 - 2026 Deskflow Developers
  * SPDX-FileCopyrightText: (C) 2012 - 2016, 2026 Synergy App Ltd
  * SPDX-FileCopyrightText: (C) 2002 Chris Schoeneman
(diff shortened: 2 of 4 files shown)


MomoyamaSawa:
Note: The root cause is that CEnter in relative mode carries a stale (0, 0) — the server can't and is no need to track the true position in this mode. A cleaner fix would be a protocol-level "don't warp" signal, but that requires a protocol version bump. This PR fixes it at the client layer where the correct position is always available.

src/lib/client/Client.cpp
```diff
@@ -195,12 +195,24 @@ void Client::getCursorPos(int32_t &x, int32_t &y) const
 void Client::enter(int32_t xAbs, int32_t yAbs, uint32_t, KeyModifierMask mask, bool)
 {
   m_active = true;
+  if (m_relativeMouseMoves) {
+    if (!m_hasRelativeRestorePosition) {
+      saveRelativeRestorePosition();
```
amagus: why do we save when entering?

MomoyamaSawa:
@amagus Before I mark this ready for review, could you please take a quick look at the approach?

amagus:
@MomoyamaSawa I left a comment.
Can you please let me know why do we need to save on enter? Shouldn't we only be saving on leaving?

MomoyamaSawa:
> @MomoyamaSawa I left a comment. Can you please let me know why do we need to save on enter? Shouldn't we only be saving on leaving?

It only saves on the first enter when no saved position exists yet. Subsequent enters restore the position saved on leave.

amagus:
> > @MomoyamaSawa I left a comment. Can you please let me know why do we need to save on enter? Shouldn't we only be saving on leaving?
> 
> It only saves on the first enter when no saved position exists yet. Subsequent enters restore the position saved on leave.

Why do we need that? It should be saved on the first leave of the screen... saving on enter looks wrong from my perspective.

MomoyamaSawa:
> > > @MomoyamaSawa I left a comment. Can you please let me know why do we need to save on enter? Shouldn't we only be saving on leaving?
> > 
> > 
> > It only saves on the first enter when no saved position exists yet. Subsequent enters restore the position saved on leave.
> 
> Why do we need that? It should be saved on the first leave of the screen... saving on enter looks wrong from my perspective.

Without it, the first `enter` to client in relative mode still uses the stale `CEnter` position from the server, which is `(0, 0)` in my reproduction, so the cursor jumps to the top-left before any client-side `leave()` has happened.

The save in `enter()` was intended only to initialize a local position for that first-enter case. After that, it is saved on leave and restored on enter.

If the initial enter should keep using the server-provided position instead, I can remove this fallback.

amagus:
Maybe the first enter should go to the center of the screen instead?

You should not save it anyway... You just need to default to some position...

MomoyamaSawa:
Makes sense, I'll change the initial position to the screen center and only save positions on leave.

Out of curiosity, is there a particular design concern with using the client's current cursor position as the initial fallback on first enter?

amagus:
My concern here is related to saving on enter...
We could be using the initial position of the client, but it should have been saved somewhere else (perhaps on the client constructor? or in some client initialization procedure?)

MomoyamaSawa:
It now happens after `m_screen->enable()` in `handshakeComplete()`, since the platform screen is ready there. `enter()` only uses an already saved restore position now.

I kept the `setOptions()` path only for live config reloads on an already-connected client.

src/lib/client/Client.cpp
amagus: resolved

MomoyamaSawa:
@amagus

amagus (approved):

sithlord48 (dismissed):
I would consider this OK to land.  @nbolton  I will leave landing to you so get a chance to review as well

sithlord48:
@MomoyamaSawa  can you please update this to the newest base .  I will try todo so later today if you have not by then 

Edit: i have updated the base.

nbolton (approved):
Verified with a macOS server and Windows client.

Issue #9791, opened 2026-05-27
Punctuation keys fail on Windows client when macOS server uses Chinese Pinyin IME


MomoyamaSawa:
### Before reporting a bug, please complete the sanity checks above.

- [x] I have done the sanity checks, and my issue persists.

### Deskflow version info

Server
Deskflow: 1.26.0.207 (https://github.com/deskflow/deskflow/commit/62a8ee749f0ec36b60fae3282143446f1b872d05)
Qt: 6.10.2
System: macOS Sequoia (15.7.5)

Client
Deskflow: 1.26.0.207 (https://github.com/deskflow/deskflow/commit/62a8ee749f0ec36b60fae3282143446f1b872d05)
Qt: 6.10.2
System: Windows 10 Version 21H2

### Build types

Official release (installer/package from GitHub releases)

### Deskflow configuration

* macOS server (primary), Windows client (secondary)
* Disable "Use server's keyboard language on this computer" on the client.

### What steps will reproduce the problem?

* The Windows client is using an English US keyboard layout. The issue also reproduces when additional input methods (e.g. Chinese, Japanese) are installed and active on Windows.

1. On the macOS server, switch the active input source to Simplified Chinese Pinyin  (com.apple.inputmethod.SCIM.ITABC) .
2. Switch to the Windows client
3. Press each of the following keys:
`,` `.` `;` `[` `]` `\`
`Shift+1` `Shift+4` `Shift+6` `Shift+9` `Shift+0` `Shift+-`
4. Switch the macOS input source back to ABC and press the same keys again.

**Expected result**
Step 3 and step 4 should produce the same output: standard ASCII punctuation matching the Windows English US layout — `! $ ^ ( ) _` and `, . ; [ ] \`.

**Actual result**
In step 3, keys produce no output on the Windows client. In step 4, all the same keys work correctly. Switching macOS back to ABC immediately restores normal behavior.

### Log output

<details><summary><h3> Diagnostic: TIS/UCKeyTranslate output on macOS server with Simplified Chinese Pinyin active </h3></summary>

The following was captured using a local Swift tool that reads the macOS Text Input Source (TIS) properties and runs UCKeyTranslate() against both the current keyboard layout source and the ASCII-capable layout source.

Active keyboard layout source while Pinyin is active:

* ID: com.apple.keylayout.PinyinKeyboard
* Has Unicode key layout data: yes
* User-selectable: no (bundled with the input method, not available in System Settings)

[...]

Issue #9465, opened 2026-02-12
Mac server using Simplified Chinese Pinyin input source cannot input comma/period on Windows 10 client


notevery:
### Feature Request

When using Deskflow with:

* MacBook Pro (server)
* Windows 10 (client)

I found that punctuation keys (`,` and `.`) cannot be typed on the Windows client **only when the Mac server input source is set to “Simplified Chinese – Pinyin”**.

If I switch the Mac input source to **ABC**, the issue immediately disappears and punctuation works normally on Windows.

This seems related to how macOS input methods process key events before Deskflow forwards them.

---

## Environment

**Server (Mac)**

* Device: MacBook Pro 15" 2019
* macOS: 15.7.3
* Deskflow: 1.25.0 (Server)
* Input Source:

  * Works: ABC
  * Not working: Simplified Chinese – Pinyin

**Client (Windows)**

* OS: Windows 10
* Deskflow: 1.25.0 (Client)
* Keyboard layout: English (US)

---

## Steps to Reproduce

1. Start Deskflow (Mac as server, Windows as client)
2. On Mac, switch input source to **Simplified Chinese – Pinyin**
3. Move cursor to Windows client screen
4. Try typing:

   * `,`
   * `.`
   * `/`
   * `[`
   * `]`
   * `\`

---

## Actual Behavior

* Comma and period do not input correctly on Windows.
* Sometimes no character is produced.
* Other keys (letters, numbers) work normally.
* If switching Mac input source to **ABC**, punctuation works immediately without restarting Deskflow.

---

## Expected Behavior

Punctuation keys (`,` and `.`) should work regardless of the Mac input source.

Deskflow should ideally forward raw key codes instead of processed input method results.

---

## Additional Notes

* The issue did not occur when using ABC input source.
* This appears to be related to macOS input method processing intercepting or modifying key events before Deskflow forwards them.
* Restarting Deskflow does not resolve the issue unless switching the input source to ABC.

---

### Voting Instructions

React with :+1: or :-1: to vote on this request

sithlord48:
Please try the [continuous](https://github.com/deskflow/deskflow/releases/tag/continuous) version to see if your issue persists.

sithlord48:
Should have been fixed by #9435 
This should be fixed in 1.26.0 Please try 1.26.0 and if this is still an issue we can reopen this.

UrsineBear:
Server (Mac)

Device: Mac mini M2

macOS: 26.4

[...]

PR #9733, opened 2026-05-10
fix(mac): protect TIS API calls with a mutex (they are not thread safe)

amagus:
## Description
Prevents crashes when there's too many keyboard events being processed on the server.
For example, the issue can easily be triggered by using a barcode reader on a long QR Code.

We fetch the keyboard layout information using TIS APIs and they are not thread safe. 
This information can be potentially cached but there's a risk of quick changes or app specific layouts. The minimal safest solution is simply to protect TIS calls with a shared global mutex.

## AI tools used for this PR
Claude Code (Sonnet 4.6) to help find all the places using TIS in the code base and protect them.

## Related Issue
fixes: #9731

## How Has This Been Tested?
Tested on my local setup with macOS Tahoe 26.4.1 as server + macOS Tahoe 26.4.1 as client + 2 Windows 11 clients

PR #9792, opened 2026-05-27
fix(macos): avoid non-selectable IME layouts for remote keys

MomoyamaSawa:
## Description
macOS input methods can expose non-user-selectable keyboard layouts bundled with the input method rather than the user's real selectable keyboard layout. Translating remote key events through those layouts can send IME punctuation semantics to clients instead of the expected ASCII punctuation.

Use the current ASCII-capable keyboard layout when the active input source has no Unicode key layout data and the reported keyboard layout is not directly selectable. Normal selectable keyboard layouts continue to use the existing translation path.

This affects macOS servers forwarding keyboard input to remote clients while an input-method-bundled layout is active.

## AI tools used for this PR
* Codex GPT-5.5 High: Write reproduction scripts, assist with log analysis, help identify root cause & check change impact.
* Claude Code Sonnet 4.6 Adaptive: Assist with cross-validating change impact; help me understand the overall project architecture.

## Fixed/related issues
#9465 
#9791

## How has this been tested?
**Environment:**

- macOS 15.7.5, Apple Silicon
- Debug build
- CMake 4.3.2, Ninja 1.13.2, Qt 6.11.1

**Checks:**

- Ran `clang-format 20.1.0 --dry-run --Werror src/lib/platform/OSXKeyState.cpp`.
- Built `Deskflow` and `deskflow-core` successfully.
- Verified `deskflow-core --version` starts successfully:
  `deskflow-core v1.26.0.210 (bf65080b), protocol v1.8`
- Ran the unit tests: 23/24 passed.

Note: `OSXKeyStateTests::fakePollShift`, `OSXKeyStateTests::fakePollChar`, and `OSXKeyStateTests::fakePollCharWithModifier` fail in this local environment. These are local macOS HID key injection/polling checks and are unrelated to the changed `mapKeyFromEvent` translation path.

Manual validation:

- macOS server (build), Windows client (1.26.0.207).
- With macOS Simplified Chinese Pinyin active, verified that affected punctuation keys now produce ASCII punctuation on the Windows client.
- Verified that switching macOS back to ABC still produces the expected punctuation.
- Verified macOS official Japanese Hiragana/Romaji input still produces ASCII punctuation on the Windows client.
- Verified the client option "Use server's keyboard language on this computer" remains disabled during the test.

### DIFF
--- a/src/lib/platform/OSXKeyState.cpp
+++ b/src/lib/platform/OSXKeyState.cpp
@@ -173,6 +173,33 @@ bool isModifier(uint8_t virtualKey)
   return (modifiers.find(virtualKey) != modifiers.end());
 }
 
+AutoTISInputSourceRef copyKeyboardLayoutForKeyTranslation()
+{
+  std::lock_guard<std::mutex> lock(g_tisMutex);
+
+  AutoTISInputSourceRef keyboardLayout(TISCopyCurrentKeyboardLayoutInputSource(), CFRelease);
+  AutoTISInputSourceRef inputSource(TISCopyCurrentKeyboardInputSource(), CFRelease);
+
+  const bool inputSourceHasLayout =
+      inputSource && TISGetInputSourceProperty(inputSource.get(), kTISPropertyUnicodeKeyLayoutData) != nullptr;
+
+  CFBooleanRef isSelectCapable = nullptr;
+  if (keyboardLayout) {
+    isSelectCapable =
+        (CFBooleanRef)TISGetInputSourceProperty(keyboardLayout.get(), kTISPropertyInputSourceIsSelectCapable);
+  }
+  const bool keyboardLayoutIsSelectCapable = isSelectCapable && CFBooleanGetValue(isSelectCapable);
+
+  if (inputSource && keyboardLayout && !inputSourceHasLayout && !keyboardLayoutIsSelectCapable) {
+    AutoTISInputSourceRef asciiKeyboardLayout(TISCopyCurrentASCIICapableKeyboardLayoutInputSource(), CFRelease);
+    if (asciiKeyboardLayout) {
+      return asciiKeyboardLayout;
+    }
+  }
+
+  return keyboardLayout;
+}
+
 } // namespace
 
 //
@@ -293,11 +320,10 @@ KeyButton OSXKeyState::mapKeyFromEvent(KeyIDs &ids, KeyModifierMask *maskOut, CG
   }
 
   // get keyboard info
-  AutoTISInputSourceRef currentKeyboardLayout(nullptr, CFRelease);
+  AutoTISInputSourceRef currentKeyboardLayout = copyKeyboardLayoutForKeyTranslation();
   CFDataRef ref = nullptr;
   {
     std::lock_guard<std::mutex> lock(g_tisMutex);
-    currentKeyboardLayout = AutoTISInputSourceRef(TISCopyCurrentKeyboardLayoutInputSource(), CFRelease);
     if (currentKeyboardLayout)
       ref = (CFDataRef)TISGetInputSourceProperty(currentKeyboardLayout.get(), kTISPropertyUnicodeKeyLayoutData);
   }


src/lib/platform/OSXKeyState.cpp
```diff
@@ -173,6 +173,43 @@ bool isModifier(uint8_t virtualKey)
   return (modifiers.find(virtualKey) != modifiers.end());
 }
 
+bool sourceHasUnicodeKeyLayoutData(TISInputSourceRef source)
+{
+  return source && TISGetInputSourceProperty(source, kTISPropertyUnicodeKeyLayoutData) != nullptr;
```
amagus: This uses TIS and is not protected. see #9733

amagus (changes requested):
TIS API is not thread safe, needs protected accesses.

src/lib/platform/OSXKeyState.cpp
```diff
@@ -173,6 +173,43 @@ bool isModifier(uint8_t virtualKey)
...
+}
+
+bool sourceIsSelectCapable(TISInputSourceRef source)
+{
+  if (!source) {
+    return false;
+  }
+
+  auto isSelectCapable = (CFBooleanRef)TISGetInputSourceProperty(source, kTISPropertyInputSourceIsSelectCapable);
```
amagus: also uses TIS

src/lib/platform/OSXKeyState.cpp
```diff
@@ -173,6 +173,43 @@ bool isModifier(uint8_t virtualKey)
...
+{
+  return inputSource && keyboardLayout && !sourceHasUnicodeKeyLayoutData(inputSource) &&
+         !sourceIsSelectCapable(keyboardLayout);
+}
+
+AutoTISInputSourceRef copyKeyboardLayoutForKeyTranslation()
+{
+  std::lock_guard<std::mutex> lock(g_tisMutex);
+
+  AutoTISInputSourceRef keyboardLayout(TISCopyCurrentKeyboardLayoutInputSource(), CFRelease);
```
amagus, line 180: TIS as well...

MomoyamaSawa:
Thanks, updated.

Renamed to copyKeyboardLayoutForKeyTranslationLocked() and moved the call inside the existing std::lock_guard block in mapKeyFromEvent(), so the locking requirement is now explicit at the call site rather than hidden inside the helper.

amagus:
I need to try this with a bit more of time as I think it may cause issues if the user keyboard layout itself is not ascii.
I should be able to try it over the weekend.

src/lib/platform/OSXKeyState.cpp
```diff
@@ -173,6 +173,32 @@ bool isModifier(uint8_t virtualKey)
   return (modifiers.find(virtualKey) != modifiers.end());
 }
 
+// Must be called with g_tisMutex held.
```
amagus: this looks to me like something waiting to be badly used... perhaps you should lock and unlock the mutex for each time one TIS API is called instead of locking it once for all the calls

amagus: Like, you can lock it here, inside the function and move the lock guard from the called to just after this call

MomoyamaSawa:
@amagus Gentle ping on this draft PR.

I noticed #9809 is another approach for the same CJK punctuation issue, but on the Windows client side. My PR tries to avoid generating the IME-bundled layout punctuation on the macOS server side, while #9809 normalizes those characters on the Windows client side.

amagus:
@MomoyamaSawa I really haven't got time to test these stuff... I am really sorry. I will try ASAP. 
@nbolton, I think you are busy as I am, but if you have a chance, please try it on your side and check...

nbolton:
> @amagus Gentle ping on this draft PR.

Is it ready for review now? If so please set to open, thanks.

nbolton:
I'm trying to verify that I can reproduce the bug on master to confirm this PR fixes it but I'm having some difficulty. 

Can you help me reproduce the bug on master?

Steps:
1. Set to Chinese Pinyin on macOS server
2. Move to Windows client and type '.' char

Expected: The bug which is a '。' char
Actual: ...but instead I see '.' on the Widnows client

MomoyamaSawa:
The symptom is actually that the key is **swallowed** (nothing appears on Windows), not that `。` shows up — the Windows client can't synthesize that KeyID so it drops it.

I tested with [Continuous v1.26.0.250](https://github.com/deskflow/deskflow/releases/tag/continuous) on both sides and can reproduce. Not sure why it's not triggering on your end — could you run this while Chinese Pinyin is active and share the output?

`defaults read com.apple.HIToolbox AppleCurrentKeyboardLayoutInputSourceID`

On my side it prints `com.apple.keylayout.PinyinKeyboard`, which is what causes the wrong translation.

The core issue is that macOS server sends KeyID `U+3002` (`。`) because `PinyinKeyboard`'s `UCKeyTranslate` maps `.` there, and the Windows client (English US layout) has no way to synthesize that character, so it drops it.

nbolton:
> could you run this while Chinese Pinyin is active and share the output?
> 
> `defaults read com.apple.HIToolbox AppleCurrentKeyboardLayoutInputSourceID`

Sure.

@amagus Do you have time to try and repro this bug on master? I'm wary of approving fixes for bugs that I can't repro myself but if someone else can confirm it, then I can LGTM.

Edit: Just saw...

> I really haven't got time to test these stuff... I am really sorry. I will try ASAP.

No problem, I'll give it another crack to try and reproduce.

sithlord48 (approved):

sithlord48:
we have user confirmation of this fixing the issue on #10039

PR #9802, opened 2026-05-29
feat: show remote peer name or IP when accepting a connection

anoopmsivadas:
## Description
Currently neither the UI nor logs shows the host name or IP of the client which is trying to connect to the server. The PR adds functions to fetch the remote address and uses the available functions to fetch the host name as well. If the host name lookup fails it will use the IP instead. And the local device name will be fetched from the `ComputerName` settings parameter. The fetched information will be shown in FingerprintPreview widget as suggested in #8888.

## AI tools used for this PR
AI was used initially to understand the code base and control flow.
Code is manually written.

## Fixed/related issues
fixes: #8862 

## How has this been tested?
Tested the BSD socket changes and Qt UI changes on Debian 13 gcc 14.2 Qt version 6.8.2
WinSock changes are partially tested on Win 11 with Visual Studio 2026 CE, MSVC v143, Qt 6.7
The changes are not compiled and tested on Mac

### DIFF
--- a/src/lib/gui/MainWindow.cpp
+++ b/src/lib/gui/MainWindow.cpp
@@ -841,9 +841,13 @@ void MainWindow::handleMissingKeyboardLayouts(const QString &layouts)
   }
 }
 
-void MainWindow::handlePeerFingerprint(const QString &fingerprint)
+void MainWindow::handlePeerFingerprint(const QString &payload)
 {
-  const auto sha256Text = QString(fingerprint).remove(':');
+  const auto payloadList = payload.split('|');
+  bool hasPeerInfo = payloadList.size() == 3;
+  const auto sha256Text = QString(payloadList.value(0)).remove(':');
+  const auto localPeer = hasPeerInfo ? payloadList.value(1) : QString{};
+  const auto remotePeer = hasPeerInfo ? payloadList.value(2) : QString{};
   const Fingerprint sha256 = {QCryptographicHash::Sha256, QByteArray::fromHex(sha256Text.toLatin1())};
 
   const bool isClient = m_coreProcess.mode() == CoreMode::Client;
@@ -868,7 +872,7 @@ void MainWindow::handlePeerFingerprint(const QString &fingerprint)
   }
 
   auto mode = isClient ? FingerprintDialogMode::Client : FingerprintDialogMode::Server;
-  FingerprintDialog fingerprintDialog(this, m_fingerprint, mode, sha256);
+  FingerprintDialog fingerprintDialog(this, m_fingerprint, mode, sha256, localPeer, remotePeer);
 
   if (fingerprintDialog.exec() == QDialog::Accepted) {
     db.addTrusted(sha256);
--- a/src/lib/net/SecureSocket.cpp
+++ b/src/lib/net/SecureSocket.cpp
@@ -414,8 +414,8 @@ int SecureSocket::secureAccept(int socket)
 
   // set connection socket to SSL state
   SSL_set_fd(m_ssl->m_ssl, socket);
-
-  LOG_VERBOSE("accepting secure socket");
+  const std::string &remotePeer = remotePeerNameOrAddress();
+  LOG_VERBOSE("accepting secure socket from '%s'", remotePeer.c_str());
   int r = SSL_accept(m_ssl->m_ssl);
 
   static int retry;
@@ -433,13 +433,15 @@ int SecureSocket::secureAccept(int socket)
 
   // If not fatal and no retry, state is good
   if (retry == 0) {
-    if (m_securityLevel == SecurityLevel::PeerAuth && !verifyCertFingerprint(Settings::tlsTrustedClientsDb())) {
+    if (const auto &localPeer = Settings::value(Settings::Core::ComputerName).toString().toStdString();
+        m_securityLevel == SecurityLevel::PeerAuth &&
+        !verifyCertFingerprint(Settings::tlsTrustedClientsDb(), localPeer, remotePeer)) {
       retry = 0;
       disconnect();
       return -1; // Fail
     }
     m_secureReady = true;
-    LOG_INFO("accepted secure socket");
+    LOG_INFO("accepted secure socket from '%s'", remotePeer.c_str());
     SslLogger::logSecureCipherInfo(m_ssl->m_ssl);
     SslLogger::logSecureConnectInfo(m_ssl->m_ssl);
     return 1;
@@ -499,7 +501,7 @@ int SecureSocket::secureConnect(int socket)
   retry = 0;
   // No error, set ready, process and return ok
   m_secureReady = true;
-  if (verifyCertFingerprint(Settings::tlsTrustedServersDb())) {
+  if (verifyCertFingerprint(Settings::tlsTrustedServersDb(), name, remotePeerNameOrAddress())) {
     LOG_INFO("connected to secure socket");
     if (!showCertificate()) {
       disconnect();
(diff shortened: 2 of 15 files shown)


src/lib/net/TCPSocket.cpp
```diff
@@ -552,3 +553,22 @@ ISocketMultiplexerJob *TCPSocket::serviceConnected(ISocketMultiplexerJob *job, b
...
+  }
+  try {
+    m_remotePeerNameOrAddress = ARCH->addrToName(addr);
+  } catch (ArchNetworkException e) {
+    LOG_DEBUG("error fetching peer name: '%s'", e.what());
+    m_remotePeerNameOrAddress = ARCH->addrToString(addr);
+  }
+  ARCH->closeAddr(addr);
+  return m_remotePeerNameOrAddress;
+}
```
sithlord48: File needs to end with empty new line.

src/lib/gui/MainWindow.cpp
```diff
@@ -1156,12 +1160,11 @@ void MainWindow::setHostName()
     if (existingScreen) {
       body = tr("Screen name already exists");
     } else {
-      body =
-          tr("The name you have chosen is invalid.\n\n"
-             "Valid names:\n"
-             "• Use letters and numbers\n"
-             "• May also use _ or -\n"
-             "• Are between 1 and 255 characters");
+      body = tr("The name you have chosen is invalid.\n\n"
```
sithlord48: Why did this change ?

sithlord48 (changes requested):
Have not done a full review just noticed these

src/lib/gui/MainWindow.cpp
anoopmsivadas: clang format automatically did that. I will revert that if required

sithlord48: Weird; As this file has been formatted. You may need to revert this if the code format check fails. Make sure your using clang-format `20.1.0` or the formatting applied maybe incorrect . 

We can let the workflow run and check if its gonna complain

anoopmsivadas: ah there were two versions present and IDE was using a different version of clang-format. Fixed it now : )

src/lib/net/SecureSocket.cpp
```diff
@@ -431,15 +431,17 @@ int SecureSocket::secureAccept(int socket)
     return -1; // Fail
   }
 
+  const auto serverName = Settings::value(Settings::Core::ComputerName).toString().toStdString();
```
sithlord48: The scope on this is to high. i would suggest one of the two

```
if (retry == 0) {
  const auto serverName = Settings::value(Settings::Core::ComputerName).toString().toStdString();
  if ( ....)  {
    ...
  }
}
```

or (what sonar will suggest) 

```
if (retry == 0 ) {
    if (const auto serverName = Settings::value(Settings::Core::ComputerName).toString().toStdString(); m_securityLevel == SecurityLevel::PeerAuth &&
        !verifyCertFingerprint(Settings::tlsTrustedClientsDb(), serverName, remotePeer)) {
    ...
    }
}
   
```

sithlord48 (changes requested):
✅ Tested with a 1.26.0 peer 

One thing i noticed when running this is that for my remote machine the name includes the domain (.lan / .local for example). So when viewing the dialog i see that  `peer.lan` is the remote name while the screen's name everywhere else is `peer` can we strip the domain peer name in the dialog?

Side note: We can do this with `QHostInfo` to avoid the need for any platform specific code (preferred). However I am unsure how we turn our Address Type into a string. If we could do that name look up is `QHostInfo::lookupHost(ip);`  or make a QHostname set its ip and then ask for the hostName() (as lookup i think includes the domain too)

sithlord48:
Make sure also to rebase this on current master and after you have "fixed" the review issue you can mark it ready for review

anoopmsivadas:
> Side note: We can do this with `QHostInfo` to avoid the need for any platform specific code (preferred). However I am unsure how we turn our Address Type into a string. If we could do that name look up is `QHostInfo::lookupHost(ip);` or make a QHostname set its ip and then ask for the hostName() (as lookup i think includes the domain too)

correct me if I'm wrong, fetching the remote IP still requires getpeername call right? and the addrToString function also uses platform dependent  APIs to convert it to string. So is there any advantage of using QHostInfo over getnameinfo for the name lookup? 
With QHostInfo we can use fromName() or lookupHost(). And the result of both will have domain as well.  We can strip the domain with QString::section or std::string::substr with either approach.

sithlord48:
> correct me if I'm wrong, fetching the remote IP still requires getpeername call right? and the addrToString function also uses platform dependent APIs to convert it to string. So is there any advantage of using QHostInfo over getnameinfo for the name lookup?
With QHostInfo we can use fromName() or lookupHost(). And the result of both will have domain as well. We can strip the domain with QString::section or std::string::substr with either approach.

I was not suggesting this be done with this PR. Sorry if it sounded that way. I was just saying how much easier this will be when our network stack is also using Q Objects.

anoopmsivadas:
> I was not suggesting this be done with this PR. Sorry if it sounded that way. I was just saying how much easier this will be when our network stack is also using Q Objects.

oh got it, that make sense

sithlord48 (commented):
I see a potential  problem here.

Our screen names are not always the hostname. The user can change it to whatever and in some cases this may happen from our parsing too replacing ` ` with `_` for example. I think we need to send the peer name with the request to do this "correctly" in all cases. As is this does work well and it think for most people they use the default host name but some will be confused by this name descritpency. 

@nbolton  with our current setup does the peer sends its name when connecting or only when connected.

anoopmsivadas:
> I think we need to send the peer name with the request to do this "correctly" in all cases.

even if the client send the peer name it will be read only after the TLS handshake right? and we are showing the fingerprint with peer name during the handshake.
Or we could use a custom openSSL extension to send the configured device name with client hello

nbolton:
> @nbolton with our current setup does the peer sends its name when connecting or only when connected.

Client sends name in handshake hello back:

https://github.com/deskflow/deskflow/blob/e8a474ba7d277d8c72cf2085dd94414eb485f7c4/src/lib/client/Client.cpp#L605

On server side we log the client name at verbose level upon handling the hello back message:

https://github.com/deskflow/deskflow/blob/e8a474ba7d277d8c72cf2085dd94414eb485f7c4/src/lib/server/ClientProxyUnknown.cpp#L225

sithlord48:
@anoopmsivadas  seems like we may have it from the initial handshake. Can you adjust this to use the client provided name please. 

Edit: I will put this in draft mode , Undraft after you have updated (helps keeps track of what i need to review)

anoopmsivadas:
> @anoopmsivadas seems like we may have it from the initial handshake. Can you adjust this to use the client provided name please.

Deskflow client is sending the name after the handshake is success. So we won't be able to use that during the handshake to show in the Fingerprint verification dialog. As I mentioned earlier it might be possible to send the name during the TLS client hello with a custom openSSL extension.

sithlord48:
I am hesitant to land this knowning the names can be incorrect... @nbolton  what do u think about tackling this w/ a proto version bump ?  We can land it as is and the names could be sometimes incorrect ( if users renamed the screen)

nbolton:
I like this PR in principle, showing the remote IP especially has been on my mind for a while. I can see why showing the client name in the fingerprint dialog would be helpful, too.

> @nbolton what do u think about tackling this w/ a proto version bump ?
> Deskflow client is sending the name after the handshake is success.

I need to dig into the code and start poking around first (most of this I don't have paged in ATM), but taking what @anoopmsivadas said at face value, changing proto won't help because we show the dialog before the handshake (but like I said, I need to dig into that, so take it with a pinch of salty goodness).

PR #9809, opened 2026-06-01
feat(windows): Map fullwidth CJK punctuation to halfwidth for IME handling

aabbccgg:
## Description
On a secondary (client) Windows machine, punctuation typed from a primary (server) running a CJK input method was silently dropped.

When the server runs a CJK IME (e.g. macOS Pinyin), punctuation arrives as a finished fullwidth glyph — e.g. ， (U+FF0C), 。 (U+3002), 《》 (U+300A/B). The Windows client has no physical key that produces those glyphs, so KeyMap::mapKey() fails to find a button, the keystroke is dropped, and nothing is typed. Latin letters were unaffected (they arrive as ASCII), so only punctuation went missing.

This change adds MSWindowsKeyState::normalizeFullwidthKey(), applied in fakeKeyDown/fakeKeyRepeat before mapping. When an incoming glyph has no local key, it is translated to the ASCII character that lives on the same physical key:

Fullwidth ASCII variants (U+FF01–U+FF5E) → halfwidth (U+0021–U+007E) by the fixed −0xFEE0 offset.
A small table for CJK punctuation that has no fullwidth-form block entry (。、《》【】「」——……""''·¥), each mapped to the ASCII char on its key position.
The client then injects that key, letting the local IME decide the final width according to its own punctuation mode — so the width decision happens on the target side, matching how a directly-attached keyboard behaves. Substitution only happens when the original glyph has no local key, so input that already works is untouched (zero regression).

## AI tools used for this PR
Claude opus 4.8 max (Anthropic) — used for codebase investigation (tracing the key-event pipeline across OSXKeyState → protocol → MSWindowsKeyState/KeyMap), root-cause analysis from server/client verbose logs, and drafting the normalizeFullwidthKey implementation.

## Fixed/related issues
Null

## How has this been tested?
Tested manually with a macOS server (Pinyin input method) and a Windows client:
- Environment: server = macOS, Pinyin IME; client = Windows, Microsoft Pinyin IME.
- Before: typing ，。；：？！《》【】"" … from the server produced nothing on the Windows client (verbose log showed key XXXX is not on keyboard).
- After: the punctuation appears correctly; the verbose log shows normalized fullwidth key XXXX to YYYY.
- With the Windows IME switched to English/halfwidth mode, the same keys produce the halfwidth forms — confirming the width is decided by the local IME.

### DIFF
--- a/src/lib/platform/MSWindowsKeyState.cpp
+++ b/src/lib/platform/MSWindowsKeyState.cpp
@@ -737,16 +737,68 @@ void MSWindowsKeyState::sendKeyEvent(
   }
 }
 
+KeyID MSWindowsKeyState::normalizeFullwidthKey(KeyID id) const
+{
+  KeyID ascii = kKeyNone;
+
+  if (id >= 0xff01u && id <= 0xff5eu) {
+    ascii = id - 0xfee0u;
+  } else {
+    // CJK punctuations that have no fullwidth-form
+    struct FullwidthKey
+    {
+      KeyID m_fullwidth;
+      KeyID m_ascii;
+    };
+    static const FullwidthKey s_fullwidthKeys[] = {
+        {0x3002, '.'},  // 。
+        {0x3001, '\\'}, // 、
+        {0x300a, '<'},  // 《
+        {0x300b, '>'},  // 》
+        {0x3010, '['},  // 【
+        {0x3011, ']'},  // 】
+        {0x2014, '_'},  // —
+        {0x2026, '^'},  // …
+        {0x201c, '"'},  // “
+        {0x201d, '"'},  // ”
+        {0x2018, '\''}, // ‘
+        {0x2019, '\''}, // ’
+        {0x00b7, '`'},  // ·
+        {0x300c, '{'},  // 「
+        {0x300d, '}'},  // 」
+        {0x00a5, '$'},  // ¥
+        {0xffe5, '$'},  // ￥
+    };
+    for (const auto &entry : s_fullwidthKeys) {
+      if (entry.m_fullwidth == id) {
+        ascii = entry.m_ascii;
+        break;
+      }
+    }
+  }
+
+  if (ascii == kKeyNone) {
+    return id;
+  }
+
+  if (getButton(id, pollActiveGroup()) != 0) {
+    return id;
+  }
+
+  LOG_DEBUG("normalized fullwidth key %04x to %04x", id, ascii);
+  return ascii;
+}
+
 void MSWindowsKeyState::fakeKeyDown(KeyID id, KeyModifierMask mask, KeyButton button, const std::string &lang)
 {
-  KeyState::fakeKeyDown(id, mask, button, lang);
+  KeyState::fakeKeyDown(normalizeFullwidthKey(id), mask, button, lang);
 }
 
 bool MSWindowsKeyState::fakeKeyRepeat(
     KeyID id, KeyModifierMask mask, int32_t count, KeyButton button, const std::string &lang
 )
 {
-  return KeyState::fakeKeyRepeat(id, mask, count, button, lang);
+  return KeyState::fakeKeyRepeat(normalizeFullwidthKey(id), mask, count, button, lang);
 }
 
 // We must use SendSAS (Secure Attention Sequence) to simulate Ctrl+Alt+Del (since Windows Vista).
--- a/src/lib/platform/MSWindowsKeyState.h
+++ b/src/lib/platform/MSWindowsKeyState.h
@@ -178,6 +178,12 @@ private:
 
   void addKeyEntry(deskflow::KeyMap &keyMap, deskflow::KeyMap::KeyItem &item);
 
+  // Map a fullwidth CJK punctuation KeyID to the ASCII character on the same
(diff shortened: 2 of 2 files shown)


src/lib/platform/MSWindowsKeyState.cpp
```diff
@@ -737,16 +737,78 @@ void MSWindowsKeyState::sendKeyEvent(
...
+        {0x3010, '['},  // left black lenticular bracket
+        {0x3011, ']'},  // right black lenticular bracket
+        {0x2014, '_'},  // em dash (chinese dash uses two of these)
+        {0x2026, '^'},  // horizontal ellipsis
+        {0x201c, '"'},  // left double quotation mark
+        {0x201d, '"'},  // right double quotation mark
+        {0x2018, '\''}, // left single quotation mark
+        {0x2019, '\''}, // right single quotation mark
+        {0x00b7, '`'},  // middle dot
+        {0x300c, '{'},  // 「
```
sithlord48, line 767: We do no "fix" our commits by adding new commts so please fix your Yuan issue in the first commit.

sithlord48 (changes requested):
I am not able to test (no windows) @amagus  / @nbolton  / @KoljaFrahm  can any of you check ?

nbolton:
> @amagus / @nbolton / @KoljaFrahm can any of you check ?

I plan to test this week.

src/lib/platform/MSWindowsKeyState.cpp
aabbccgg: Done. Squashed into a single commit as requested.

sithlord48 (changes requested):
Please remove the AI comment vomit .. Only comment  were needed not every line needs a comment....

aabbccgg:
> Please remove the AI comment vomit .. Only comment were needed not every line needs a comment....

Done, if you think the comments should be more compact, please feel free to let me know.

src/lib/platform/MSWindowsKeyState.cpp
```diff
@@ -737,16 +737,78 @@ void MSWindowsKeyState::sendKeyEvent(
   }
 }
 
+KeyID MSWindowsKeyState::normalizeFullwidthKey(KeyID id) const
+{
+  // Some IMEs (e.g. Microsoft Pinyin) hand us a finished fullwidth glyph
```
sithlord48: This is not needed comment block. We already have this in the header as the method description, this is repeating it with sightly different words. If you feel this is a better description than it should replace the one in the header.

src/lib/platform/MSWindowsKeyState.cpp
```diff
@@ -737,16 +737,78 @@ void MSWindowsKeyState::sendKeyEvent(
...
+  // that has no physical key on the local keyboard, so the key would be
+  // dropped.  We translate such glyphs to the ASCII character that lives on
+  // the same physical key, then let the local IME re-apply the fullwidth
+  // form according to its own punctuation mode.  This keeps the width
+  // decision on the target side instead of the source.
+  KeyID ascii = kKeyNone;
+
+  // fullwidth ASCII variants (U+FF01 - U+FF5E) are a fixed offset from their
```
sithlord48: This comment is unneeded we can tell these are offset by the if statement and what it does to the value.

src/lib/platform/MSWindowsKeyState.cpp
```diff
@@ -737,16 +737,78 @@ void MSWindowsKeyState::sendKeyEvent(
...
+        break;
+      }
+    }
+  }
+
+  if (ascii == kKeyNone) {
+    return id;
+  }
+
+  // only substitute when the original glyph isn't on the local keyboard
```
sithlord48: Un needed comment

src/lib/platform/MSWindowsKeyState.cpp
```diff
@@ -737,16 +737,78 @@ void MSWindowsKeyState::sendKeyEvent(
...
+  // form according to its own punctuation mode.  This keeps the width
+  // decision on the target side instead of the source.
+  KeyID ascii = kKeyNone;
+
+  // fullwidth ASCII variants (U+FF01 - U+FF5E) are a fixed offset from their
+  // halfwidth counterparts (U+0021 - U+007E)
+  if (id >= 0xff01u && id <= 0xff5eu) {
+    ascii = id - 0xfee0u;
+  } else {
+    // CJK punctuation that has no fullwidth-form block entry.  Each glyph is
```
sithlord48: `// CJK punctuations that have no fullwidth-form. `

explains this well enough

sithlord48 (changes requested):
Most of the comments are redundant so lets remove them.

aabbccgg:
Thank you for your patient, I had removed all redundant comments.
I found my PR will fix the issue #9465
And my PR may conflict with this PR #9792

sithlord48:
@amagus  or @nbolton  are either of you able to test this

nbolton:
> @amagus or @nbolton are either of you able to test this

On my list! Will probably be Monday though.

amagus:
Need to check if it doesn't conflict with #9792

PR #9814, opened 2026-06-01
fix: crash on windows if fixed font is missing

sithlord48:
## Description

use font names on windows for the log
## AI tools used for this PR

None

## Fixed/related issues

fixed: 9765

## How has this been tested?

run on mac os


nbolton (approved):
Tested. Looks good.

PR #9816, opened 2026-06-01
Better local fingerprint preview

sithlord48:
## Description

 - Remove un needed extra layout from the fingerprint preview
 - Add `deskflow::gui::fixedFont()` to get the fixed size font where needed. 
 - Adjust width of the fingerprint preview to fix the hash nicely

 - Center the title on the fingerprint dialog
## AI tools used for this PR

None
## Fixed/related issues

Hash can be cutoff in the fingerprint preview. 
## How has this been tested?

Locally run on linux 
Someone should check this on windows also.

### DIFF
--- a/src/lib/gui/widgets/FingerprintPreview.cpp
+++ b/src/lib/gui/widgets/FingerprintPreview.cpp
@@ -70,16 +70,11 @@ QLayout *FingerprintPreview::sha256Layout(const Fingerprint &fingerprint, const 
   f.setStyleHint(QFont::Monospace);
   m_lblArt->setFont(f);
 
-  auto innersha256Layout = new QHBoxLayout();
-  innersha256Layout->setContentsMargins(0, 0, 0, 0);
-  innersha256Layout->addWidget(m_lblHash);
-  innersha256Layout->addWidget(m_lblArt);
-
   auto sha256Layout = new QVBoxLayout();
   sha256Layout->setContentsMargins(0, 0, 0, 0);
   sha256Layout->addWidget(labelTitle);
-  sha256Layout->addLayout(innersha256Layout);
-  sha256Layout->addSpacerItem(new QSpacerItem(0, 0, QSizePolicy::Preferred, QSizePolicy::Expanding));
+  sha256Layout->addWidget(m_lblHash);
+  sha256Layout->addWidget(m_lblArt);
 
   auto frameSha256 = new QFrame(this);
   frameSha256->setFrameShape(QFrame::StyledPanel);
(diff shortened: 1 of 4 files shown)


src/lib/gui/widgets/FingerprintPreview.cpp
```diff
@@ -59,27 +60,23 @@ QLayout *FingerprintPreview::sha256Layout(const Fingerprint &fingerprint, const
   m_lblHash->setAlignment(Qt::AlignVCenter | Qt::AlignHCenter);
   m_lblHash->setTextInteractionFlags(Qt::TextSelectableByMouse);
   m_lblHash->setVisible(hashMode);
+  m_lblHash->setFont(fixedFont());
+  m_lblHash->setFixedSize(
+      m_lblHash->fontMetrics().horizontalAdvance(QChar('M')) * m_lblHash->text().length() / 4,
+      m_lblHash->fontMetrics().height() * 4
+  );
```
nbolton: Sonar warning:
> implicit conversion loses integer precision: 'qsizetype' (aka 'long long') to 'int'

This would solve the warning, but it's quite chewy:
```suggestion
  m_lblHash->setFixedSize(
      static_cast<int>(m_lblHash->fontMetrics().horizontalAdvance(QChar('M')) * m_lblHash->text().length() / 4),
      m_lblHash->fontMetrics().height() * 4
  );
```

Or, rasier for other devs to read and understand:
```suggestion
  const int rows = 4;
  const auto metrics = m_lblHash->fontMetrics();
  const auto charWidth = metrics.horizontalAdvance(QChar('M'));
  const auto charsPerRow = static_cast<int>(m_lblHash->text().length()) / rows;
  m_lblHash->setFixedSize(charWidth * charsPerRow, metrics.height() * rows);
```

nbolton (commented):
Visually, appears the same. Is that expected?

Edit: Ah, yes, commit says refactor.

Image

Image

src/lib/gui/widgets/FingerprintPreview.cpp
sithlord48: I did the second one but mac os needs at least 2 px for width extra so using 

`fm->maxWidth() +2` for charWidth.

nbolton (commented):
Hmm, now it's ultra wide on server.

Image

And on the client the text is aligned to the center right somehow?

image

p.s. Copy pasted that screenshot from Windows client to Wayland server :heart_eyes:

sithlord48:
Ah layout padding is different on the comparison dialog . Should be better now. We do not need to set a fixed size for  the hash label just make the box a little wider than it thinks it should be.

nbolton (approved):
Very nice! Code looks much cleaner this way. 
Tested both dialog modes (art + hash only) on all 4 desktops and looks great.

PR #9820, opened 2026-06-03
[codex] Harden clipboard and screen transition handling

RuruBros:
## Summary
- Add physical screen layout mapping support for server transitions
- Hide and restore the Windows cursor when moving off the primary screen
- Add a development restart helper for Deskflow processes
- Harden clipboard ownership handling so stale or mismatched clipboard updates are ignored instead of asserting
- Restore Windows cursor visibility during the development restart flow

## Root Cause
Rapid screen transitions can reorder clipboard owner and clipboard data events. A delayed clipboard update from a previous owner could reach the server after ownership had already changed, which hit an ownership assertion. Entering secondary screens at the reciprocal edge could also immediately trigger a switch back at the boundary.

## Impact
The server now accepts clipboard data only from the current clipboard owner, ignores stale sequence numbers, and avoids immediate edge bounce when entering secondary screens. The restart helper also recovers Windows cursor visibility after forced termination.

## Validation
- `git diff --check HEAD~1..HEAD`
- `cmake --build build/windows-debug-vcpkg --target server`
- `tools/restart-deskflow-dev.ps1 -NoGuiBuild`
- Checked recent daemon logs for assertion, crash, and clipboard ignore patterns

## Notes
The current local build cache has `BUILD_TESTS=OFF`, so unit tests were not available in this build directory.

### DIFF
--- a/docs/user/configuration.md
+++ b/docs/user/configuration.md
@@ -384,6 +384,26 @@ This indicates that computer ''larry'' is to the right of computer ''moe'' (so m
 
 Note that links do not have to be symmetrical; for instance, here the edge between ''moe'' and ''curly'' maps to different ranges depending on if you're going up or down. In fact links don't have to be bidirectional. You can configure the right of ''moe'' to go to ''larry'' without a link from the left of ''larry'' to ''moe''. It's possible to configure a computer with no outgoing links; the cursor will get stuck on that computer unless you have a hot key configured to switch off of that computer.
 
+### physical-layout section
+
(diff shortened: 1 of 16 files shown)


thepythonguy-sm:
Feel like this all is vibecoded. Never mind, it is.

sithlord48:
Please go read the [contributing](https://github.com/deskflow/deskflow/wiki/Contributing) guide and then after you have fixed this PR you can reopen it.

Issue #9639, opened 2026-04-06
After modifying the configuration and restarting the service on the Windows server, the client fails to reconnect automatically.


MCsonger:
### Sanity checks

- [x] I have done the sanity checks, and my issue persists
- [x] These sanity checks are not relevant to the bug

### Deskflow version info

```
Deskflow: 1.26.0.0 (760e3b99)
Qt: 6.10.1
System: macOS Tahoe (26.4.1)
```

After modifying the configuration and restarting the service on the Windows server, the Mac client stops working and fails to reconnect automatically. An extra mouse operation is required to restart the Mac client before it can reconnect successfully.

<img width="566" height="546" alt="Image" src="https://github.com/user-attachments/assets/3aa57efa-f2d1-43ba-8463-7eba91687618" />

<img width="1250" height="1098" alt="Image" src="https://github.com/user-attachments/assets/48e941ef-6ffb-4c2a-a114-3ce4ee098c47" />

### Operating systems (OS)

- [x] Windows
- [x] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Deskflow configuration

<img width="400" height="508" alt="Image" src="https://github.com/user-attachments/assets/6281dc80-42e6-4d46-9dc7-26b714fd5e74" />

<img width="400" height="508" alt="Image" src="https://github.com/user-attachments/assets/fbb28c69-e669-40d8-98e3-79787f8d8a80" />

### Log output

```shell

```

sithlord48:
English please

MCsonger:
> English please

After modifying the configuration and restarting the service on the Windows server, the Mac client stops working and fails to reconnect automatically. An extra mouse operation is required to restart the Mac client before it can reconnect successfully.

sithlord48:
You also need to provide version info for both your client and server please use the copy button in the about dialog to add this to your description above.

I have seen this with 1.26.0 on my client (X11 flatpak) but its not consistent as to what causes the client to look as it its "connecting" but not try  I have not looked for a duplicate of this issue but i suspect one exists.

MCsonger:
> 1.26.0
It's version 1.26.0.

<img width="541" height="639" alt="Image" src="https://github.com/user-attachments/assets/d1fc2653-587f-491b-ad0b-dee75cdad31f" />

sithlord48:
[...]

Issue #8263, opened 2025-02-28
Server/Client on mac disconnects after either goes to sleep


Toni-Chan:
### Project

Deskflow

### Deskflow version info

Deskflow: 1.19.0.0 (9365091c)
Qt: 6.7.3
System: macOS Sequoia (15.3) (Server), (15.3.1) (Client)

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [ ] Windows
- [x] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [ ] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [ ] I am not using Wayland on Linux

### Signing on macOS

- [x] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### Continuous build

- [x] I have tried the latest continuous build and the issue persists
- [ ] I am unable to try the latest continuous build

### OS versions/distros

System: macOS Sequoia (15.3) (Server), (15.3.1) (Client)

### Deskflow configuration

System: macOS Sequoia (15.3) (Server), (15.3.1) (Client)
Config: Server on left, client on right, but middle screen may belong to either computer

### What steps will reproduce the problem?

When either client or server goes to sleep with Deskflow open. Log outputs will show when the server log when server goes to sleep. From client side, the client will keep connecting, but when the server wakes up, connection will still fail by `Connection Refused`.

This seems to be a new bug introduced by the latest version of how server is controlled. Didn't see this problem in previous versions (but previous version's disconnect every half hour bug is fixed by this).

### Log output

```shell
Server:
[2025-02-27T20:14:04] INFO: switch from "Toni-Personal-M1" to "Toni-Work-M1" at 1419,826
[2025-02-27T20:14:04] INFO: entering screen
[2025-02-27T20:14:05] INFO: switch from "Toni-Work-M1" to "Toni-Personal-M1" at -3840,834
[2025-02-27T20:14:05] INFO: leaving screen
[2025-02-27T20:14:05] WARNING: cursor may not be visible
[2025-02-27T20:42:39] INFO: switch from "Toni-Personal-M1" to "Toni-Work-M1" at 1415,757
[2025-02-27T20:42:39] INFO: entering screen
[...]

PR #9826, opened 2026-06-06
Replace stale same-named client on reconnect instead of rejecting with EBusy

valentt:
## Description

When a client connects with a screen name that is already present in `m_clients`, `Server::adoptClient()` rejects the **new** connection with `kMsgEBusy` ("a client with name … is already connected") and keeps the existing one.

The problem is that the existing connection is very often a **stale / half-open** one: the previous client dropped *ungracefully* — the client machine's display went to sleep, a network blip, an OOM stall, etc. — so no clean disconnect (FIN/RST) ever reached the server. There is no application-level keepalive that reaps such a dead client, and the TCP socket lingers for the system keepalive default (hours). As a result the dead client **holds the screen name indefinitely**, and the real client can never reconnect — every attempt is refused with EBusy — until the **server is restarted**.

This is a frequent, very confusing failure: the client just reports "connection failed" forever after the laptop/monitor sleeps overnight, even though the network and TLS are fine.

### Change

In `adoptClient()`, when `addClient()` fails because the name is taken, disconnect the existing client through the normal `closeClient()` teardown (which removes it from `m_clients`, installs the close-timeout timer, and jumps off it if it was the active screen) and then let the new client take over. Safeguards:

- The **primary screen is never replaced** (`stale != m_primaryClient`).
- If the name still cannot be freed, the original `kMsgEBusy` rejection is kept as a fallback.

This is the standard "new connection wins" behaviour. The only behaviour change for a *live* duplicate is that the newcomer now takes over instead of being rejected — but two live clients sharing one screen name is already a misconfiguration (names must be unique in the server config), whereas the stale-reconnect case is the common real-world one.

## Disclosure of AI use

This change was written with AI assistance (Claude / Claude Code): the root cause was diagnosed and the patch drafted with the tool, then reviewed against the surrounding `Server` teardown logic (`closeClient`, `removeClient`, `handleClientCloseTimeout`) by a human before submitting.

## Related Issue

Related to the "cannot reconnect after the client dropped/slept" class of reports, e.g. #9639 and #8263. Not marked as `fixes` because those reports cover broader scenarios; this PR addresses the specific stale-same-name-blocks-reconnect mechanism.

## How Has This Been Tested?

- Built `deskflow-core` (tag v1.26.0, the same change applies cleanly to `master`) on Fedora 42 (GCC, CMake, Qt6) and ran it as the server.
- Real-world reproduction on an X11 server (Linux) with a Windows client: after the client dropped ungracefully (monitor sleep), `ss` showed a stale `ESTAB` to the client and every reconnect was refused with "a client with name … is already connected" until the server was restarted. With this change the reconnecting client takes over the stale slot and the session re-establishes without a server restart.
- No automated test added yet — happy to add a `ServerTests` case for the duplicate-name takeover if maintainers prefer.

### DIFF
--- a/src/lib/server/Server.cpp
+++ b/src/lib/server/Server.cpp
@@ -242,10 +242,27 @@ void Server::adoptClient(BaseClientProxy *client)
 
   // add client to client list
   if (!addClient(client)) {
-    // can only have one screen with a given name at any given time
-    LOG_WARN("a client with name \"%s\" is already connected", getName(client).c_str());
-    closeClient(client, kMsgEBusy);
-    return;
+    // A client with this name is already connected. This is almost always a
+    // stale/half-open connection left behind when the previous client dropped
+    // ungracefully (display sleep, network blip, OOM): the dead client keeps
+    // holding the screen name and there is no keepalive to reap it, so the real
+    // client could never reconnect until the server was restarted. Instead,
+    // disconnect the existing client and let the new one take over.
+    const std::string name = getName(client);
+    BaseClientProxy *stale = nullptr;
+    if (auto it = m_clients.find(name); it != m_clients.end()) {
+      stale = it->second;
+    }
+    if (stale != nullptr && stale != client && stale != m_primaryClient) {
+      LOG_WARN("client \"%s\" already connected; replacing stale connection", name.c_str());
+      closeClient(stale, kMsgCClose);
+    }
+    if (!addClient(client)) {
+      // could not free the name (e.g. it is the primary screen) -- give up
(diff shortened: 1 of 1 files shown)


sithlord48 (changes requested):

I do not like this solution. Its basically oh someone with your name is connected kick them and replace them with this client. This could possibly be abused by a malicious network device.  I would much prefer if we could fix the root cause of the half connection / stale connection issue.

valentt:
Thanks, I now see that this PR was just a hack, I'll digg deeper to find root cause. Closing this one.

PR #9833, opened 2026-06-08
refactor: replace qFatal with qCritical

nbolton:
## Description
Over-use of qFatal (terminate the program) means that debugging is harder for the end-user and most of our fatals are not legitimate reasons to crash the GUI.

## AI tools used for this PR
~~None~~ Claude Opus

## Fixed/related issues
None?

## How has this been tested?
Run GUI, happy path smoke test.

### DIFF
--- a/src/lib/gui/ScreenSetupModel.cpp
+++ b/src/lib/gui/ScreenSetupModel.cpp
@@ -8,11 +8,12 @@
 
 #include "ScreenSetupModel.h"
 
+#include "common/Constants.h"
+#include "gui/config/Screen.h"
+
 #include <QIODevice>
 #include <QIcon>
 #include <QMimeData>
-
-#include "gui/config/Screen.h"
 
 const QString ScreenSetupModel::m_MimeType = "application/x-deskflow-screen";
 
@@ -22,17 +23,19 @@ ScreenSetupModel::ScreenSetupModel(ScreenList &screens, int numColumns, int numR
       m_NumColumns(numColumns),
       m_NumRows(numRows)
 {
-
-  // bound rows and columns to prevent multiply overflow.
-  // this is unlikely to happen, as the grid size is only 3x9.
-  if (m_NumColumns > 100 || m_NumRows > 100) {
-    qFatal("grid size out of bounds: %d columns x %d rows", m_NumColumns, m_NumRows);
-    return;
+  // bound the grid so that multiplying columns by rows cannot overflow.
+  if (m_NumColumns < 1 || m_NumColumns > kMaxGridSize || m_NumRows < 1 || m_NumRows > kMaxGridSize) {
+    qCritical("grid size out of bounds: %d columns x %d rows", m_NumColumns, m_NumRows);
+    m_NumColumns = kServerGridWidth;
+    m_NumRows = kServerGridHeight;
   }
 
-  const long span = static_cast<long>(m_NumColumns) * m_NumRows;
-  if (span > screens.size()) {
-    qFatal("scrren list (%lld) too small for %d columns x %d rows", screens.size(), m_NumColumns, m_NumRows);
+  const int span = m_NumColumns * m_NumRows;
+  if (span > m_Screens.size()) {
+    qCritical(
+        "screen list too small for grid, screens: %lld, cells: %d", static_cast<long long>(m_Screens.size()), span
+    );
+    m_Screens.resize(span);
   }
 }
 
--- a/src/lib/gui/core/CoreProcess.cpp
+++ b/src/lib/gui/core/CoreProcess.cpp
@@ -105,8 +105,7 @@ CoreProcess::CoreProcess(const ServerConfig &serverConfig)
 {
   m_appPath = QStringLiteral("%1/%2").arg(QCoreApplication::applicationDirPath(), kCoreBinName);
   if (!QFile::exists(m_appPath)) {
-    qFatal("core server binary does not exist");
-    return;
+    qCritical("core server binary does not exist");
   }
 
   connect(m_daemonIpcClient, &ipc::DaemonIpcClient::connected, this, &CoreProcess::daemonIpcClientConnected);
@@ -216,7 +215,8 @@ void CoreProcess::startForegroundProcess(const QStringList &args)
   using enum ProcessState;
 
   if (m_processState != Starting) {
-    qFatal("core process must be in starting state");
+    qCritical("not starting core desktop process, unexpected process state");
+    return;
   }
 
   // only make quoted args for printing the command for convenience; so that the
@@ -245,7 +245,8 @@ void CoreProcess::startForegroundProcess(const QStringList &args)
 void CoreProcess::startProcessFromDaemon()
 {
   if (m_processState != ProcessState::Starting) {
-    qFatal("core process must be in starting state");
+    qCritical("not starting core process from daemon, unexpected process state");
+    return;
   }
 
   const auto configFile = Settings::settingsFile();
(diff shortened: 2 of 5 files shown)


src/lib/gui/config/ServerConfig.cpp
```diff
@@ -258,7 +258,7 @@ QTextStream &operator<<(QTextStream &outStream, const ServerConfig &config)
     outStream << "\t" << "heartbeat = " << config.heartbeat() << Qt::endl;
 
   if (config.protocol() == NetworkProtocol::Unknown)
-    qFatal("unrecognized protocol when writing config");
+    qCritical("unrecognized protocol when writing config");
```
sithlord48: We don't want to write the stream if we do not reconize the values.

src/lib/gui/config/ServerConfig.cpp
```diff
@@ -258,7 +258,7 @@ QTextStream &operator<<(QTextStream &outStream, const ServerConfig &config)
...
   outStream << "\t" << "protocol = " << networkProtocolToOption(config.protocol()) << Qt::endl;
```
sithlord48: I thought i remove this from the server config...

sithlord48 (commented):
qFatal causes the applicaton to end. Just make sure any places where this is changed to critical we are not continuing with bad input.

src/lib/gui/ScreenSetupModel.cpp
```diff
@@ -26,13 +26,13 @@ ScreenSetupModel::ScreenSetupModel(ScreenList &screens, int numColumns, int numR
...
   if (m_NumColumns > 100 || m_NumRows > 100) {
-    qFatal("grid size out of bounds: %d columns x %d rows", m_NumColumns, m_NumRows);
+    qCritical("grid size out of bounds: %d columns x %d rows", m_NumColumns, m_NumRows);
     return;
   }
 
   const long span = static_cast<long>(m_NumColumns) * m_NumRows;
   if (span > screens.size()) {
-    qFatal("scrren list (%lld) too small for %d columns x %d rows", screens.size(), m_NumColumns, m_NumRows);
+    qCritical("scrren list (%lld) too small for %d columns x %d rows", screens.size(), m_NumColumns, m_NumRows);
```
sithlord48: Typo here `scrren` should be `screen`

nbolton:
@sithlord48 Thanks for review, will fix.

src/lib/gui/config/ServerConfig.cpp
sithlord48: Removed by #9838

src/lib/gui/config/ServerConfig.cpp
sithlord48: removed by #9838

src/lib/gui/core/CoreProcess.cpp
```diff
@@ -353,7 +353,7 @@ void CoreProcess::start(std::optional<ProcessMode> processModeOption)
   }
 
   if (m_mode == Settings::CoreMode::None) {
-    qFatal("set core mode before starting");
+    qCritical("set core mode before starting");
```
sithlord48: This maybe could even be an Info message users should see that one and its very unlikey to even be hit.

src/lib/gui/core/CoreProcess.cpp
```diff
@@ -654,7 +654,7 @@ void CoreProcess::clearSettings()
   }
 
   if (processMode != ProcessMode::Service) {
-    qFatal("invalid process mode");
+    qCritical("invalid process mode");
```
sithlord48: return after this critical.

src/lib/gui/validators/LineEditValidator.cpp
```diff
@@ -21,7 +21,7 @@ LineEditValidator::LineEditValidator(QLineEdit *lineEdit, ValidationError *error
 {
 
   if (!m_pLineEdit) {
-    qFatal("validator line edit not set");
+    qCritical("validator line edit not set");
```
sithlord48, line 24: IMHO this should remain Fatal. this will not work w/ a nullptr as its lineEdit.

src/lib/gui/ScreenSetupModel.cpp
```diff
@@ -26,13 +26,13 @@ ScreenSetupModel::ScreenSetupModel(ScreenList &screens, int numColumns, int numR
   // bound rows and columns to prevent multiply overflow.
   // this is unlikely to happen, as the grid size is only 3x9.
   if (m_NumColumns > 100 || m_NumRows > 100) {
-    qFatal("grid size out of bounds: %d columns x %d rows", m_NumColumns, m_NumRows);
+    qCritical("grid size out of bounds: %d columns x %d rows", m_NumColumns, m_NumRows);
```
sithlord48, line 28: Both of these are critical. You can not build a screen model, screen layout without grid being the correct size. maybe here we set it  back to the default size (3x5) instead ?

src/lib/gui/ScreenSetupModel.cpp
sithlord48: Also here maybe we should show the layout is full (i.e too many screens on the layout)

src/lib/gui/core/CoreProcess.cpp
```diff
@@ -271,11 +271,11 @@ void CoreProcess::startProcessFromDaemon()
 void CoreProcess::stopForegroundProcess() const
 {
   if (m_processState != ProcessState::Stopping) {
-    qFatal("core process must be in stopping state");
+    qCritical("core process must be in stopping state");
   }
 
   if (!m_process) {
-    qFatal("process not set, cannot stop");
+    qCritical("process not set, cannot stop");
```
sithlord48: This is fatal if you do not return .

src/lib/gui/core/CoreProcess.cpp
```diff
@@ -216,7 +216,7 @@ void CoreProcess::startForegroundProcess(const QStringList &args)
   using enum ProcessState;
 
   if (m_processState != Starting) {
-    qFatal("core process must be in starting state");
+    qCritical("core process must be in starting state");
```
sithlord48: Should we put the core in that state and move on ?

Edit:
I think this can be logged and we can return. I think putting it in the state may cause un expected issues

sithlord48 (commented):
Some notes on a few of these , Some of these need to return after the log message.

src/lib/gui/ScreenSetupModel.cpp
sithlord48: We do not have a way for users to change the grid size anyway (at this time)

sithlord48:
@nbolton reminder to update this PR

nbolton:
> @nbolton reminder to update this PR

Thanks for the bump, I had totally forgotten about this.

src/lib/gui/core/CoreProcess.cpp
nbolton: Decided to return here. Albeit unlikely now, we do want to know if it happens as it may happen in future due to a new bug we introduce.

src/lib/gui/validators/LineEditValidator.cpp
nbolton: I handled null inside `LineEditValidator::validate` and did `return Intermediate` -- what do you think? I know this is a bit riskier, but I'm really trying not to use `qFatal` for situations we can actually recover from (i.e. by handling the null).

Edit: Why avoid fatal? Kind of the point of this PR is to avoid fatal so users can grab the logs when something breaks.

Actually, IIRC, `Q_ASSERT` in the ctor would actually be idiomatic Qt rather than `qCritical` -- we could do that instead and rely on the null check later on in execution?

I'm easy on this so if you have a strong preference, just say.

src/lib/gui/ScreenSetupModel.cpp
nbolton: Resets to the default grid now.

src/lib/gui/validators/LineEditValidator.cpp
sithlord48: `Q_ASSERT` is only a debug check
we should return if this is null we can not make this item it has to have a line to work upon. This case should also literally never happen  in a real build.

sithlord48 (approved):
Looks alright

PR #9834, opened 2026-06-08
fix(macos): honor locally-held modifier keys on synthetic mouse events

### LINKED ISSUE #8682: "Command+Click multi-selection fails when using mixed input sources (local keyboard + remote mouse)"
### Project

Deskflow

### Deskflow version info

Deskflow version: 1.22.0
Qt version: 6.6.2
Server OS: Windows 11 Pro 23H2 (Deskflow host)
Client OS: macOS 15.5 (Apple Silicon, Mac mini M4)
Connected via: Wi-Fi (same network)

Reproducible: Always

### Build type

### Operating systems (OS)

- [x] Windows
- [x] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [x] I am not using Wayland on Linux

### Signing on macOS

- [x] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### Continuous build

### OS versions/distros

-Windows 11
-macOS 15.5

### Deskflow configuration

- Windows 11 server, macOS 15 client
- each computer has a single monitor
- Windows is on the left, macOS is on the right

1. On my Mac mini, when using both a Windows mouse and Windows keyboard via Deskflow, Command + click allows multiple selection in Finder.

2. On my Mac mini, when using both a Mac mouse and Mac keyboard directly connected, Command + click also works correctly for multiple selection.

3. On my Mac mini, when using a Windows mouse via Deskflow and a Mac keyboard locally, Command + click does not work — only the last clicked item remains selected.

### Log output

```shell

```

### Additional information

_No response_

derhally:
## Description

When a Mac runs as a client and only the mouse is shared from the server (e.g. a Windows server + a local Mac keyboard), modifier keys held on the local Mac keyboard are ignored on clicks coming from the shared mouse. Shift/Command-click multi-selection (Finder, lists, text) only work if the modifier was held on the server's keyboard.

The fix is to OR in the live, session-wide modifier state alongside the server-tracked flags.

## AI tools used for this PR
Claude Code using Opus 4.8

## Fixed/related issues
Fixes: #8682 

## How has this been tested?
I did a local MacOS build and tested that this change works as expected.  I tested both using the PC keyboard and Mac keyboard to make sure existing behavior is working.   Holding down shift or command on the local Mac Keyboard while using the mouse results in the expected results.

### DIFF
--- a/src/lib/platform/OSXScreen.mm
+++ b/src/lib/platform/OSXScreen.mm
@@ -466,9 +466,7 @@ void OSXScreen::postMouseEvent(CGPoint &pos) const
   // Dragging events also need the click state
   CGEventSetIntegerValueField(event, kCGMouseEventClickState, m_clickState);
 
-  // Fix for sticky keys
-  CGEventFlags modifiers = m_keyState->getModifierStateAsOSXFlags();
-  CGEventSetFlags(event, modifiers);
+  CGEventSetFlags(event, getModifiers());
 
   // Set movement deltas to fix issues with certain 3D programs
   SInt64 deltaX = pos.x;
@@ -489,6 +487,14 @@ void OSXScreen::postMouseEvent(CGPoint &pos) const
   CGEventPost(kCGHIDEventTap, event);
 
   CFRelease(event);
+}
+
+CGEventFlags OSXScreen::getModifiers() const
+{
+  // Fix for sticky keys
+  CGEventFlags modifiers = m_keyState->getModifierStateAsOSXFlags();
+  modifiers |= CGEventSourceFlagsState(kCGEventSourceStateHIDSystemState);
+  return modifiers;
 }
 
 void OSXScreen::fakeMouseButton(ButtonID id, bool press)
@@ -549,9 +555,7 @@ void OSXScreen::fakeMouseButton(ButtonID id, bool press)
 
   CGEventSetIntegerValueField(event, kCGMouseEventClickState, m_clickState);
 
-  // Fix for sticky keys
-  CGEventFlags modifiers = m_keyState->getModifierStateAsOSXFlags();
-  CGEventSetFlags(event, modifiers);
+  CGEventSetFlags(event, getModifiers());
 
   m_buttonState.set(index, state);
   CGEventPost(kCGHIDEventTap, event);
@@ -609,9 +613,7 @@ void OSXScreen::fakeMouseWheel(ScrollDelta delta) const
     // is the right choice here over kCGScrollEventUnitPixel
     CGEventRef scrollEvent = CGEventCreateScrollWheelEvent(nullptr, kCGScrollEventUnitLine, 2, delta.y, delta.x);
 
-    // Fix for sticky keys
-    CGEventFlags modifiers = m_keyState->getModifierStateAsOSXFlags();
-    CGEventSetFlags(scrollEvent, modifiers);
+    CGEventSetFlags(scrollEvent, getModifiers());
 
     CGEventPost(kCGHIDEventTap, scrollEvent);
     CFRelease(scrollEvent);
--- a/src/lib/platform/OSXScreen.h
+++ b/src/lib/platform/OSXScreen.h
@@ -107,6 +107,14 @@ private:
   bool updateScreenShape(const CGDirectDisplayID, const CGDisplayChangeSummaryFlags);
   void postMouseEvent(CGPoint &) const;
 
+  /**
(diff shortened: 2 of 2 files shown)


sithlord48:

derhally:
> Hello, Thanks for taking the time to contribute. It seems that this pull request does not conform to our standards. Please review our [Hacking Guide](https://github.com/deskflow/deskflow/wiki/Hacking-Guide).

Is that better?

sithlord48:
Not really its still not using our template.. You have only added some of the sections

sithlord48 (changes requested):
Maybe a method that returns the modifiers would be better here since you would only need to comment why its done once 

something like 

```
// Ensure the correct modifiers  in case of being pressed on local keyboard
CGEventFlags getModifiers()
{
  CGEventFlags modifiers = m_keyState->getModifierStateAsOSXFlags();
  modifiers |= CGEventSourceFlagsState(kCGEventSourceStateCombinedSessionState);
  return modifiers;
}
```

and call `CGEventSetFlags(event, getModifiers());` where needed

sithlord48:
I also think you should remove all but the first line from your commit message

amagus:
Is this exclusive to macOS? 
Does it work fine when using a windows or linux client?
I wonder if the expected results are really to OR the modifiers or not...

I am tending to think that this should not be on by default... (maybe be an option?)
It should be an option and it should align with the behavior in all operating systems...

derhally:
I have tested with Deskflow both running as a Server and Client on MacOS and Windows.

The Windows installation is using the current version 1.26.0
The Mac installation is using my build 

* using the server's mouse (Mac)
* Holding down Shift or Ctrl on PC Keyboard
* Holding down Shift or Ctrl on Mac Keyboard 

In issue #8682, @sithlord48 states that he can't reproduce this with linux, so it seems this may only be an issue with the Mac client.

sithlord48:
I will have to double check with "Mixed" sources I'll try to check that today

src/lib/platform/OSXScreen.mm
```diff
@@ -491,6 +489,18 @@
   CFRelease(event);
 }
 
+CGEventFlags OSXScreen::getModifiers() const
+{
+  // Combine the server-tracked modifier state with the live session-wide
```
sithlord48: These comments are redundant  . Coments on the function should be doxygen style in the header.

sithlord48 (changes requested):

amagus:
@derhally The question here: does it work without changes when you use windows as server? I understand you are trying to fix it when using mac as server, but we need to see if we need a general fix for this feature or not...

sithlord48:
Just tested from wayland server to x11 client and i can hold shift , ctrl etc on the local keyboard (x11) and it works as expected i get the selection method i expect.

sithlord48 (dismissed):
Looks ok @amagus

sithlord48:
@derhally  note for future contributions, We will place your pr into draft after we have reviewed it with issues for you to correct after you do please make the pr "ready for review" again so we know its ready for us to review again ( i did it for you this time)

amagus:
I see, then the pr should be fine...

amagus (approved):

derhally:
> @derhally The question here: does it work without changes when you use windows as server? I understand you are trying to fix it when using mac as server, but we need to see if we need a general fix for this feature or not...

The original issue is with windows as server and the Mac as the client.   I wanted to use a mac keyboard but share the mouse on Windows PC with the mac.

I tested both scenarios with the Mac as the server with my changes and the PC as the server with the current release.

src/lib/platform/OSXScreen.mm
```diff
@@ -491,6 +489,17 @@
   CFRelease(event);
 }
 
+CGEventFlags OSXScreen::getModifiers() const
+{
+  // Server-tracked modifier state (modifiers held on the remote/server side).
```
sithlord48: The comments that describe what the method does should only be in the header not in the middle of the method. 
Nothing complex is going on to where the three lines in the method need comments

sithlord48 (changes requested):

sithlord48 (approved):
@nbolton  ill leave landing to you .

derhally:
@sithlord48 I'm sometimes seeing the modifier keys stick.  I'm going to pull this PR for the time being and do more testing and try to figure it out and come back when I'm confident my change is correct.  Thank you for being patient as I get used to your contribution process.

PR #9843, opened 2026-06-10
fix: remove ClipboardSending handler in proxy destructors (use-after-free)

### LINKED ISSUE #9842: Use-after-free: ServerProxy destructor does not remove ClipboardSending handler (SIGSEGV in ClipboardChunk::send on image copy + reconnect)
## Summary

`ServerProxy::~ServerProxy()` removes its `StreamInputReady` event handler but **not** the `ClipboardSending` handler it registered in the constructor. If a `ClipboardSending` event is still queued when the `ServerProxy` is destroyed (e.g. a reconnect/teardown races an in-flight clipboard transfer), the event loop later dispatches the handler against the freed `ServerProxy`, producing a use-after-free and a SIGSEGV in the clipboard send path.

This reproduces in the field most reliably when **copying a large image** (multi-MB NSPasteboard bitmap on macOS) while the client connection is unstable, because the large transfer is split into multiple `ClipboardSending` chunk events that sit in the queue long enough for a reconnect to occur mid-transfer.

## Affected code

`src/lib/client/ServerProxy.cpp` — constructor registers the handler:

```cpp
m_events->addHandler(EventTypes::ClipboardSending, this, [this](const auto &e) {
    ClipboardChunk::send(m_stream, e.getDataObject());
});
```

…but the destructor only removes `StreamInputReady`:

```cpp
ServerProxy::~ServerProxy()
{
    setKeepAliveRate(-1.0);
    m_events->removeHandler(EventTypes::StreamInputReady, m_stream->getEventTarget());
    // missing: m_events->removeHandler(EventTypes::ClipboardSending, this);
}
```

The same registration pattern (and missing removal) is present in `ClientProxy1_6`.

Verified present on `master` as of 2026-06-09.

## Crash signature (observed in the field)

Platform: macOS 15.7.8 (Darwin 24.6.0), Apple Silicon (arm64). Observed on the synergy-core 1.20.1-dev fork (Synergy 3.6.1), whose `ServerProxy` is identical to current deskflow `master`.

```
Exception: EXC_BAD_ACCESS (SIGSEGV)
Subtype:   KERN_INVALID_ADDRESS at 0x0000000000000020

Faulting thread (App::runEventsLoop):
  ProtocolUtil::vwritef(IStream*, const char*, unsigned int, char*)   +56
  ProtocolUtil::writef(IStream*, const char*, ...)                    +100
  ClipboardChunk::send(IStream*, void*)                               +352
  EventQueue::dispatchEvent(Event const&)                            +116
  EventQueue::loop()                                                 +340
  App::runEventsLoop(void*)                                           +24
  Thread::threadFunc(void*)                                          +132
```

Crash PC disassembly (the virtual `IStream::write` dispatch):

```asm
ldr  x8, [x20]        ; x20 = IStream* this = 0x20  <-- load vtable from 0x20 → fault
ldr  x8, [x8, #0x20]  ; vtable + 0x20 → write()
mov  x0, x20
blr  x8
```

`x20` (the `IStream* stream` passed to `ClipboardChunk::send`) is `0x20` — the value read from `m_stream` (offset 8) of the freed `ServerProxy`, i.e. reused/garbage heap memory rather than a valid pointer. `FAR = 0x20` confirms the dereference of that bogus pointer.

## Sequence

1. Image copied on the client → `OSXClipboard` reads NSPasteboard → `StreamChunker::sendClipboard()` enqueues multiple `ClipboardSending` chunk events.
2. Connection drops / reconnects mid-transfer → `Client` tears down and destroys its `ServerProxy`.
3. `ClipboardSending` handler (bound to the freed `ServerProxy`) is **not** removed from the `EventQueue`.
4. Event-loop thread dispatches the next `ClipboardSending` event → lambda dereferences the dangling `this` → reads garbage `m_stream` → virtual call through a `0x20` vtable → SIGSEGV.

## Proposed fix

Remove the handler in the destructor (mirror the constructor registration):

```cpp
ServerProxy::~ServerProxy()
{
    setKeepAliveRate(-1.0);
    m_events->removeHandler(EventTypes::StreamInputReady, m_stream->getEventTarget());
    m_events->removeHandler(EventTypes::ClipboardSending, this);   // add this
}
```

Apply the equivalent removal wherever `ClientProxy1_6` registers `ClipboardSending`.

## Notes

- This is distinct from #8625 (clipboard-too-large buffer overrun) and the PR #9663 unmarshal size-check work — those are on the *receive/unmarshal* side; this is a lifetime/UAF bug on the *send* side.
- Workaround for affected users: set `clipboardSharing = false` (or a very small `clipboardSharingSize`) on the server, or stabilize the link (wired) to avoid reconnects during transfers.

dreamingbinary:
## Description

`ServerProxy` (client side) and `ClientProxy1_6` (server side) each register an
`EventTypes::ClipboardSending` handler bound to `this` in their constructors, but
neither destructor removes it:

- `ServerProxy::~ServerProxy()` removed only `StreamInputReady`.
- `ClientProxy1_6` used a defaulted destructor (`~ClientProxy1_6() override = default;`) that removed nothing.

If a `ClipboardSending` event is still queued when the proxy is destroyed — e.g. a
reconnect/teardown racing an in-flight clipboard transfer — the event loop later
dispatches the handler against the freed proxy. The captured `this` is dangling;
reading `m_stream` off freed memory yields garbage and the subsequent virtual
`IStream::write` call faults.

Observed in the field on macOS (Apple Silicon) as `EXC_BAD_ACCESS` /
`KERN_INVALID_ADDRESS` in `ClipboardChunk::send` → `ProtocolUtil::vwritef`, most
reproducibly when copying a large image (multi-MB pasteboard data is split into
several `ClipboardSending` chunk events, widening the race window) over an
unstable link. The faulting `IStream*` was the small value `0x20`, i.e. a field
read from the freed proxy rather than a valid stream pointer.

The fix removes the `ClipboardSending` handler in both destructors so it can no
longer fire after the proxy is gone, mirroring each constructor's registration.

## AI tools used for this PR

Claude (Anthropic, Opus 4.8) was used, operated by the repository owner, to:
symbolicate the crash reports, locate the constructor/destructor mismatch in the
source, author the one-line-per-class fix, and draft this PR and the linked issue.
All changes were reviewed by a human before submission.

## Fixed/related issues

fixes: #9842

## How has this been tested?

- Environment: macOS 15.7.8 (Darwin 24.6.0), Apple Silicon (M3), Qt 6.11.1, OpenSSL 3.6.2, CMake/Ninja Release build.
- `deskflow-core` builds cleanly with the change; both modified translation units (`ServerProxy.cpp`, `ClientProxy1_6.cpp`) compile, binary links and runs (`deskflow-core version` reports the patched commit).
- Runtime smoke: client mode boots, reads config, starts, retries connection and shuts down cleanly with no crash / no crash report.
- The fix itself is verified by inspection: each constructor's `addHandler(EventTypes::ClipboardSending, this, …)` now has a matching `removeHandler(EventTypes::ClipboardSending, this)` in the destructor. The original SIGSEGV is a teardown race (a `ClipboardSending` event dispatched after the proxy is freed), which is inherently non-deterministic to reproduce on demand; removing the handler at destruction makes the dangling dispatch impossible.

Maintainers: please confirm the `removeHandler` signature/target matches the registration in your event-queue API — happy to adjust if a different overload is preferred.

## Notes

This is distinct from the clipboard-too-large / unmarshal size-check work (receive
side); this is a handler-lifetime use-after-free on the send side, present in both
the client and server clipboard paths.

### DIFF
--- a/src/lib/client/ServerProxy.cpp
+++ b/src/lib/client/ServerProxy.cpp
@@ -55,6 +55,7 @@ ServerProxy::~ServerProxy()
 {
   setKeepAliveRate(-1.0);
   m_events->removeHandler(EventTypes::StreamInputReady, m_stream->getEventTarget());
+  m_events->removeHandler(EventTypes::ClipboardSending, this);
 }
 
 void ServerProxy::resetKeepAliveAlarm()
--- a/src/lib/server/ClientProxy1_6.cpp
+++ b/src/lib/server/ClientProxy1_6.cpp
@@ -26,6 +26,11 @@ ClientProxy1_6::ClientProxy1_6(const std::string &name, deskflow::IStream *strea
   });
 }
 
+ClientProxy1_6::~ClientProxy1_6()
+{
+  m_events->removeHandler(EventTypes::ClipboardSending, this);
+}
+
 void ClientProxy1_6::setClipboard(ClipboardID id, const IClipboard *clipboard)
 {
   // ignore if this clipboard is already clean
--- a/src/lib/server/ClientProxy1_6.h
+++ b/src/lib/server/ClientProxy1_6.h
@@ -16,7 +16,7 @@ class ClientProxy1_6 : public ClientProxy1_5
 {
 public:
   ClientProxy1_6(const std::string &name, deskflow::IStream *adoptedStream, Server *server, IEventQueue *events);
-  ~ClientProxy1_6() override = default;
+  ~ClientProxy1_6() override;
 
   void setClipboard(ClipboardID id, const IClipboard *clipboard) override;
   bool recvClipboard() override;


sithlord48 (changes requested):
Tons of redundant comments here.. the commit message is WAY To long (the first line is fine)

nbolton (changes requested):
Please reduce comments only to ones that explain why the line is needed in exceptional circumstances where the code is not self-documenting. A one-liner comment is often far more useful, as it's more likely to be read. Also helps to have a space above the comment for readability if the comment is truly needed.

sithlord48:
Not a single line needs a comment for these changes every single line is pretty self documenting in this case.

sithlord48:
Pretty sure it should be removing the handle from `m_stream->getEventTarget()` .

dreamingbinary:
Thanks for the reviews — I've pushed an update that drops all the comments (agreed, the lines are self-documenting) and trims the commit message to the subject line.

On removing from `m_stream->getEventTarget()`: I don't think that's right here. The `ClipboardSending` handler is registered against `this`, not the stream's event target — see `ServerProxy.cpp:46` and `ClientProxy1_6.cpp:24` (`addHandler(EventTypes::ClipboardSending, this, …)`). Since `removeHandler` has to match the target the handler was added with, removing against `this` is the correct pairing.

The handler that *does* use `m_stream->getEventTarget()` is the adjacent `StreamInputReady` one (`ServerProxy.cpp:43`), which is already removed with that same target — I suspect that's the source of the mix-up. Happy to switch both the registration and removal to `getEventTarget()` if you'd prefer that convention, but as written the add/remove pair is consistent.

sithlord48 (approved):
Ok, @nbolton I'll leave landing to you so you can review.

nbolton (approved):

PR #9846, opened 2026-06-10
feat: allow a screen to span multiple columns in the layout grid

mikhailvs:
### Problem

The server layout grid allows one neighbor per direction, so a physically large screen (e.g. an ultrawide spanning the width of three laptops below it, or a portrait monitor next to two stacked screens) can't be represented: moving the cursor toward it only ever reaches one of the screens.

### Solution

A screen can now be resized to span multiple grid cells by dragging its right or bottom edge, directly in the layout grid (no dialog). A screen spanning N cells sits next to several neighbors at once, and the generated config maps the overlap onto the link range syntax the server already supports:

```
section: links
    wide-desktop:
            down(0,33) = laptop-1
            down(33,67) = laptop-2
            down(67,100) = laptop-3
    laptop-2:
            right = laptop-3
            left = laptop-1
            up = wide-desktop(33,67)
```

No core/server changes needed — the parser and cursor position mapping (CellEdge intervals) handle ranged links already; only the GUI couldn't express them.

### Implementation notes

- The screen stays anchored in its top-left cell; covered cells stay empty in the ScreenList and screenIndexAt() resolves which screen covers a cell
- The grid renders spans with QTableView::setSpan; drag-resize handles appear on the right/bottom edges (cursor changes to a resize arrow)
- A resize that would push the span out of the grid or over an occupied cell is rejected; drops and new clients respect spans too
- Spans round-trip through settings and the drag-and-drop mime stream, and are clamped on load if the grid no longer fits them
- Width and height are symmetric — same handling, same link generation

### Screenshots

`wide-desktop` spans three columns, `tall-display` spans two rows:

## AI tools used for this PR

Claude Code (Claude Fable 5, `claude-fable-5`) was used to implement the feature, write the unit test, and generate the screenshots. All changes were reviewed and tested manually.

## Fixed/related issues

None

## How has this been tested?

- New `ScreenTests::span()` unit test covering default values, clamping, equality, and settings round-trip for both width and height; full unit test suite passes
- Built and tested on Fedora 43, Qt 6.10.3, GNOME Wayland
- In daily use on a real 4-computer layout: a Linux server spanning 3 columns above three macOS clients
- Generated config verified against the core parser (ranged links such as `down(0,33)` / `up = name(33,67)` parse and route the cursor correctly)

### DIFF
--- a/src/lib/gui/ScreenSetupModel.h
+++ b/src/lib/gui/ScreenSetupModel.h
@@ -1,5 +1,6 @@
 /*
  * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2026 Mikhail Slyusarev <slyusarevmikhail@gmail.com>
  * SPDX-FileCopyrightText: (C) 2012 Synergy App Ltd
  * SPDX-FileCopyrightText: (C) 2008 Volker Lanz <vl@fidra.de>
  * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
@@ -25,6 +26,11 @@ class ScreenSetupModel : public QAbstractTableModel
   friend class ServerConfigDialog;
 
 public:
+  // QSize(width, height), in grid cells, of the screen at the index -- so a
+  // delegate can distinguish a 1x1 screen from a spanning one, and a wide span
+  // from a tall one, without reaching into the model
+  static constexpr int SpanRole = Qt::UserRole + 1;
+
   ScreenSetupModel(ScreenList &screens, int numColumns, int numRows);
 
   static const QString &mimeType()
@@ -54,6 +60,10 @@ public:
   QMimeData *mimeData(const QModelIndexList &indexes) const override;
   bool isFull() const;
 
+  //! Resize the screen anchored at the given cell, rejecting spans that
+  //! would leave the grid or cover an occupied cell
+  bool trySetSpan(int column, int row, int width, int height);
+
 Q_SIGNALS:
   void screensChanged();
 
@@ -82,6 +92,10 @@ private:
   ScreenList &m_Screens;
   const int m_NumColumns;
   const int m_NumRows;
+
+  // set by dropMimeData when it moves a screen within the grid, so startDrag
+  // knows the source was already handled and only a trash drop must clear it
+  bool m_dropHandled = false;
 
   static const QString m_MimeType;
 };
--- a/translations/deskflow_it.ts
+++ b/translations/deskflow_it.ts
@@ -907,8 +907,8 @@ Inoltre, verifica di poter %1 il file di configurazione del server: %2</translat
 <context>
     <name>ScreenSetupModel</name>
     <message>
-        <source>&lt;center&gt;Screen: &lt;b&gt;%1&lt;/b&gt;&lt;/center&gt;&lt;br&gt;Double click to edit settings&lt;br&gt;Drag screen to the trashcan to remove it</source>
-        <translation>&lt;center&gt;Schermo: &lt;b&gt;%1&lt;/b&gt;&lt;/center&gt;&lt;br&gt;Fai doppio clic per modificare le impostazioni&lt;br&gt;Trascina lo schermo nel cestino per rimuoverlo</translation>
+        <source>&lt;center&gt;Screen: &lt;b&gt;%1&lt;/b&gt;&lt;/center&gt;&lt;br&gt;Double click to edit settings&lt;br&gt;Drag screen to the trashcan to remove it&lt;br&gt;Drag the right or bottom edge to span more cells</source>
+        <translation>&lt;center&gt;Schermo: &lt;b&gt;%1&lt;/b&gt;&lt;/center&gt;&lt;br&gt;Fai doppio clic per modificare le impostazioni&lt;br&gt;Trascina lo schermo nel cestino per rimuoverlo&lt;br&gt;Trascina il bordo destro o inferiore per estendersi su più celle</translation>
     </message>
 </context>
 <context>
--- a/src/lib/gui/ScreenSetupModel.cpp
+++ b/src/lib/gui/ScreenSetupModel.cpp
@@ -1,5 +1,6 @@
 /*
  * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2026 Mikhail Slyusarev <slyusarevmikhail@gmail.com>
  * SPDX-FileCopyrightText: (C) 2025 Chris Rizzitello <sithlord48@gmail.com>
  * SPDX-FileCopyrightText: (C) 2012 Synergy App Ltd
  * SPDX-FileCopyrightText: (C) 2008 Volker Lanz <vl@fidra.de>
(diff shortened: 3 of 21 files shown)


sithlord48:

mikhailvs:
> Hello, Thanks for taking the time to contribute. It seems that this pull request does not conform to our standards. Please review our [Hacking Guide](https://github.com/deskflow/deskflow/wiki/Hacking-Guide).

Thanks for pointing me to that, I've updated the pr body and status to match the doc.

sithlord48 (changes requested):
All the layout related stuff needs to happen in the layout tab not in a dialog. The solution is not to allow computers to span columns in the grid layout but to better communicate the whole edges can be split and the splits linked. This could be an issue w/ a few of these you will fill the allowed screens very quickly since your making each machine take up N columns. Each machine is one cell in our grid and should remain this way.

Lets see if we do this without spanning the grid sections and without needing open the dialog to enter this info

amagus:
Why only wide screens? Can't we have long vertical screens?

sithlord48 (commented):
Other things to note. 

Transaltions need to have translations added using google translate is fine for this
~~What if i my monitor is tall ?~~ (i see amagus asked first)

mikhailvs:
@sithlord48 @amagus Since the feedback is largely design related, if you don't mind before I go into a direction on the basis of my interpretation, can you help me understand a few things?

1. Layout not configured in dialog: I agree, it's better to configure layout directly in the layout ui that already exists. My thought is to use a drag-resize handle to change the span of the monitor. Does that work for you, or did you have something else in mind?
2. One cell per screen/expose edge-splitting: I think exposing edge splitting, while more general, is going to be pretty confusing. It'll require a mental model to match the abstraction in code, which will be difficult to make usable. The mental model I'm working with is roughly "my monitor is wide" not "which behaviors must I define around the edges of the screens I interact with to model a wide monitor". wrt the "fills grid quickly" worry, the span consumes cells because the cells represent physical space, which a large monitor occupies, not to mention the grid size is already configurable (server/gridWidth/gridHeight).
3. Tall screens is a good point, and I think it fits with the drag-handle idea.

I'm happy to implement the feature however is best, but I do want to make sure the direction I go in is acceptable to you. Thanks again for reviewing.

sithlord48:
The grid size is NOT configurable. For all intents and purposes it is always 5x3. Yes i know its in the configuration but its not really something users can change.(or we know works if they do) We have 15 "slots" for a computers to be located at. 

Being able to make the screens wider or taller by dragging is a good start. But mixing the edge spiting w/ spanning is asking for trouble. I think we ultimately need to move away from showing the layout the way we do all together. 

You also should not be adding yourself to any of the copyright stuff , Since you didn't write any of the code

> Claude Code (Claude Fable 5, claude-fable-5) was used to implement the feature, write the unit test, and generate the screenshots. All changes were reviewed and tested manually.

mikhailvs:
Ok, here's where I landed after rereading your comments:

1. Direction: I'll take "wider or taller by dragging is a good start" as the plan, drag-resize handles on the screen right in the layout grid (width and height), spinbox and dialog gone. If you had something different in mind, let me know before I start.
2. wrt mixing edge splitting with spanning, I think we might be describing the same thing from different ends. The span just compiles down to split links in the generated config (down(0,33) = a, down(33,67) = b, etc), there's no second mechanism in the PR. If there's a specific scenario you're worried about can you describe it? I don't want to miss a real footgun.
3. wrt filling the grid, a spanning screen does take extra cells, but cells aren't computers, you still get 15 machines unless the layout actually has that much physical width in it. Seemed like the right trade to me, but open to changing it if there's a constraint I'm not seeing.
4. Copyright: I don't care about having my name in there. Per the hacking guide the options are my name or (C) Deskflow Developers, fwiw Anthropic's terms assign output rights to the user and claim no ownership, so attributing to them isn't a thing. Tell me which you want and I'll amend.

Bottom line, I use this app every day and this change makes a real difference on my setup, so I'm happy to do the rework once the direction is confirmed

sithlord48:
Sounds good lets get this improved for everyone !

translations/deskflow_it.ts
```diff
@@ -908,7 +908,11 @@ Inoltre, verifica di poter %1 il file di configurazione del server: %2</translat
     <name>ScreenSetupModel</name>
     <message>
         <source>&lt;center&gt;Screen: &lt;b&gt;%1&lt;/b&gt;&lt;/center&gt;&lt;br&gt;Double click to edit settings&lt;br&gt;Drag screen to the trashcan to remove it</source>
-        <translation>&lt;center&gt;Schermo: &lt;b&gt;%1&lt;/b&gt;&lt;/center&gt;&lt;br&gt;Fai doppio clic per modificare le impostazioni&lt;br&gt;Trascina lo schermo nel cestino per rimuoverlo</translation>
+        <translation type="vanished">&lt;center&gt;Schermo: &lt;b&gt;%1&lt;/b&gt;&lt;/center&gt;&lt;br&gt;Fai doppio clic per modificare le impostazioni&lt;br&gt;Trascina lo schermo nel cestino per rimuoverlo</translation>
```
sithlord48: You should build this locally with the option to "CLEAN_TR" files this will remove those vanished (no longer used) strings.

sithlord48 (changes requested):
I will try this out later if i get a chance . This looks like too much code for this but i've not really looked into this , its possible we can do this in a more simple way. I have new review later but for now this one item can be done

sithlord48:
@mikhailvs  You will need to remove your merge we can not land any code with merges (per repo rules) Always rebase .

sithlord48:
Quick tested this (mostly for the gui) i do like the spanning may want to be able to resize from any corner. I do not like how it works on the grid with the wider items its  bit hard to see maybe we should consider outlining them. Another thing i noticed is that when we have a tall monitor its label is not vertically centered.  

The config edits look good to .. 

found a weird gui bug too tried moving the wide monitor up then to the middle cell then back to the lower row and when doing that this happened (for some reason they were both selected also) 
Screenshot_20260613_205508

upon saving and reentering they are not selected but sill not it a good place

Screenshot_20260613_205848

src/lib/gui/ScreenSetupModel.h
```diff
@@ -25,6 +26,9 @@ class ScreenSetupModel : public QAbstractTableModel
   friend class ServerConfigDialog;
 
 public:
+  // true when the screen at the index spans more than one cell
+  static constexpr int SpanRole = Qt::UserRole + 1;
```
sithlord48, line 32: If my screen is 2x2 this single role will not work better to add a `WidthRole` and `HeightRole` OR a single role with both as data.

sithlord48 (changes requested):
Do remember to rebase upon master.

Will have time to test later on.

sithlord48:
Still have the same problem w/ being able to place them in a corner . they are also impossible to move 1 cell towards the expansion. The coloring and border outline is too much they look selected always.

mikhailvs:
Ok, here's where I landed after the last couple rounds:

1. wrt the placement/movement bugs, `QTableView::setSpan` was collapsing drop hit-testing anywhere inside a span back to the anchor cell, so the target kept resolving to the screen's current position. I switched the drop target to `columnAt`/`rowAt`, and the move now picks up the screen's whole footprint and, if the destination's occupied, swaps that one screen back to the source anchor in one shot. So a span can move by a single cell or into a corner without the layout ending up half-moved or overlapping. Added tests for those cases plus rejecting moves past the edge.
2. The border was using the selection highlight color, which is why it always looked selected. It's a thin neutral outline now, no fill.
3. `SpanRole` now returns `QSize(width, height)`, which covers the `WidthRole`/`HeightRole` comment and gives the delegate both dimensions for cases like 2x2 without reaching back into the model.
4. Tall label is vertically centered now.
5. On translations, the new strings are translated in all six languages, and I checked with `-no-obsolete` (CLEAN_TRS) that there are no obsolete strings left hanging around.
6. Rebased onto master, single commit, no merge.

The one thing I held back is resize from any corner. The top/left grips have to migrate the anchor cell during the resize, which pulls in a fair bit more resize/cursor handling, so my read was it makes more sense as a follow-up. If you'd rather have it here, let me know and I'll do it.

sithlord48:
Still able to over lap them quite easily

mikhailvs:
You're right, I missed this case. The overlap check was looking at each span on its own, so a spanning screen only checked that its own footprint sat on empty cells in storage. A wide screen and a tall screen can cross on a cell that's covered by both, which reads as empty in storage to each of them, so neither check catches it. I changed it to track occupancy across all the screens at once and reject any cell that's claimed twice, and added a test for the wide+tall cross that failed before and passes now. Pushed.

sithlord48:
@mikhailvs please rebase upon master and mark ready for review when its ready for us to review again

PR #9431, opened 2026-02-06
Clipboard support for Wayland (input capture and remote desktop portals)

whot:
## Description
This adds (a very much in draft state) integration for clipboards via the inputcapture portal, making copy/paste from server to client work OOTB on wayland.

It's based on top of #9415 and will require xdg-desktop-portal, xdg-desktop-portal-gnome, and libportal to run (see https://github.com/flatpak/xdg-desktop-portal/pull/1898). Additionally in the gnome case you will need mutter, other compositors will need similar changes.

It removes the current WlClipboard bits which is probably not ideal but that was yet another layer on top to figure out.

## Disclosure of AI use
The second commit was purely written by Claude after I stared at this for too long without seeing any hope of getting started. Turns out Claude wrote something that at least looks correct and it does work. It needs review but also needs knowledge of deskflow's internals that I don't have...

## How Has This Been Tested?

I've tested this against a Ubuntu 24.04LTS Xorg GNOME session (vanilla with only deskflow installed as extra package). And whoooah, I can copy on the server and paste on the client. It's like the future!

As of 2026-02-06 the other direction (client->server) does not yet work.

Anyway, I'm filing this so it's at least visible, backed up, and maybe someone really wants to take this and run with it so I don't have to :)

Edit: sithlord48
 - Updated Style and rebase upon updated #9415 
 - Add A Check in cmake for `libportal/clipboard.h` to define HAVE_LIBPORTAL_CLIPBOARD 
 - Guard the libportal clipboard use on `HAVE_LIBPORTAL_CLIPBOARD`
 - fixes: #9600  
 - Removes wl-clipboard parts.

## To do:
- [X] Client -> Server Clipboard  (remote desktop) support is not included
- [x] Fix potential race bugs (copy/paste to client not working on first attempt)
- [x] Check type support
- [x] Make sure we are using the proper clipboard max size in this impl
- [x] Squash @nbolton and @KoljaFrahm commits before landing (preferably after review from @whot)
- [x] Copy/paste text and images to/from Wayland client (within scope of this PR?)
- [x] De-dupe code new clipboard between input capture and remote desktop
- [x] Wait for upstream libs to release before landing

## Upstream

- [x]  New release of [xdg-desktop-portal](https://github.com/flatpak/xdg-desktop-portal)
- [x]  New release of [libportal](https://github.com/flatpak/libportal)
- Gnome (probably with Gnome 51):
[...]

Issue #8165, opened 2025-02-02
Clipboard sharing breaks intermittently on X11


ShaunRoselt:
### Project

Deskflow

### Deskflow version number

1.19.0.0

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [x] Windows
- [ ] macOS
- [x] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [ ] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [x] I am not using Wayland on Linux

### Signing on macOS

- [ ] I have authorized the app to run on my Mac
- [x] I am not using macOS

### Continuous build

- [ ] I have tried the latest continuous build and the issue persists
- [x] I am unable to try the latest continuous build

### OS versions/distros

- Windows 11
- SteamOS (Steam Deck)

### Deskflow configuration

- Windows PC has two monitors on the left
- SteamOS Steam Deck has four monitors on the right

### What steps will reproduce the problem?

Just try to copy some text from Windows PC to Steam Deck or vice versa. It doesn't work always. More than half of the time it doesn't copy the text over to the other computer. I have to keep trying and trying and then occasionally it works.  

It's not a big issue since it does work sometimes. It's more so just an irritation that it doesn't consistently work every time.

### Log output

```shell
[2025-02-02T12:54:35] INFO: connected to secure socket
[2025-02-02T12:54:35] INFO: server tls certificate info: /CN=Deskflow
[2025-02-02T12:54:35] INFO: network encryption protocol: TLSv1.3
[2025-02-02T12:54:35] INFO: local languages: en
[2025-02-02T12:54:35] INFO: remote languages: en, en
[2025-02-02T12:54:35] NOTE: connected to server
[2025-02-02T12:54:35] INFO: entering screen
[2025-02-02T12:54:35] INFO: clipboard was updated
[2025-02-02T12:54:35] INFO: clipboard was updated
[2025-02-02T12:54:35] INFO: leaving screen
[2025-02-02T12:54:35] INFO: entering screen
[2025-02-02T12:54:38] INFO: leaving screen
[2025-02-02T12:54:40] INFO: entering screen
[2025-02-02T12:55:02] INFO: leaving screen
[2025-02-02T12:55:03] INFO: entering screen
[2025-02-02T12:55:03] INFO: leaving screen
[...]

Issue #7600, opened 2024-10-01
Clipboard sharing not working on Wayland


rayrayrayraydog:
### Project

Deskflow

### Deskflow version number

1.17.0

### Build type

Local developer build (built it yourself)

### Operating systems (OS)

- [ ] Windows
- [X] macOS
- [X] Linux
- [ ] BSD-derived
- [ ] Other

### OS versions/distros

Server: Fedora 40 UBlue - bazzite (Wayland)
Client: macOS 15.0

### Deskflow configuration

- bazzite Server on the left (Wayland + KDE Plasma 6.1.5), macOS 15.0 client on the right.  Used the GUI with defaults including encryption

### What steps will reproduce the problem?

Start deskflow with above configuration and connect the macOS client.  I cannot share any clipboard contents between the two computers.  It doesn't seem to matter if I copy from the server or the client.  The logs on both computers indicate that the clipboard is updated but I can still only paste from each machine's local clipboard contents.

### Log output

```shell
** From the server ** 

[2024-10-01T19:19:11] INFO: screen "raymac" grabbed clipboard 1 from "raymac"
[2024-10-01T19:19:14] INFO: switch from "raymac" to "k6" at 2185,824
[2024-10-01T19:19:14] INFO: entering screen
[2024-10-01T19:19:14] INFO: screen "raymac" updated clipboard 0
[2024-10-01T19:19:14] INFO: screen "raymac" updated clipboard 1
[2024-10-01T19:21:26] INFO: switch from "k6" to "raymac" at 110,448
[2024-10-01T19:21:26] INFO: leaving screen
[2024-10-01T19:21:33] INFO: switch from "raymac" to "k6" at 2164,390
[2024-10-01T19:21:33] INFO: entering screen
[2024-10-01T19:23:26] INFO: switch from "k6" to "raymac" at 84,26
[2024-10-01T19:23:26] INFO: leaving screen
[2024-10-01T19:23:30] INFO: switch from "raymac" to "k6" at 2187,695
[2024-10-01T19:23:30] INFO: entering screen
```

sithlord48:
From my understanding this is an xdg-desktop-portal implimention problem not a direct deskflow issue.

nbolton:
I made the title a bit more general as clipboard support for both client and server still needs to be implemented for non-X.

nbolton:
[...]

PR #9850, opened 2026-06-11
fix: make Wayland clipboard sharing work reliably

mikhailvs:
## Description

The wl-clipboard backend (`core/wlClipboard`, opt-in) had a few bugs that made it effectively non-functional, plus one that made it actively harmful:

- change detection compared the list of offered mime types between polls, so copying two texts in a row was never detected. Detection now fingerprints the content (capped at 256KB) as well as the types.
- detection ran on a polling thread. On compositors without a data-control protocol (gnome), every wl-paste invocation briefly steals keyboard focus, so background polling drops keystrokes while you type. The poll thread is gone. `checkClipboards()` now checks on demand when the cursor leaves the screen, which is the one moment a focus blip can't interrupt typing and exactly when the clipboard needs syncing.
- clipboard data was passed to wl-copy as a process argument (binary data impossible), and `empty()` plus every `add()` spawned a separate wl-copy, each stealing the selection from the previous one. Writes are now accumulated during the open/close transaction and committed as one wl-copy fed via stdin, picking the most valuable format (wl-copy offers a single mime type per invocation).
- our own writes were detected as external changes, so the server would re-broadcast its own clipboard. The committed content hash is now remembered and adopted silently.
- wl-paste reads had no timeout and could hang the app; all reads are bounded now.
- EiScreen fired `ClipboardChanged` for every clipboard ID when any changed; it's now per-ID.

HTML is no longer dropped on `add()`. Multi-mime simultaneous offers still need data-control or the clipboard portal (#9431), this just makes the existing backend work in the meantime.

## AI tools used for this PR

Claude Code (Claude Fable 5, `claude-fable-5`) was used to implement the changes and write the tests. Reviewed and tested manually on a real setup.

## Fixed/related issues

Related: #7600, #8165. Complements #8031 / #9431 (portal clipboard) rather than competing with it.

## How has this been tested?

- WlClipboardTests extended with regression tests for same-mime-type change detection, binary round-trip (NUL bytes), commit-on-close format priority, and own-write suppression: 19/19 pass on a live GNOME 49 Wayland session (Fedora 43, Qt 6.10.3); full unit suite passes (25/25)
- In daily use as a Wayland server with macOS clients; the keystroke-dropping regression from background polling was caught and fixed during that testing

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -340,16 +340,11 @@ void EiScreen::fakeKey(uint32_t keycode, bool isDown) const
 void EiScreen::enable()
 {
   // Nothing really to be done here
-  if (m_clipboard && m_clipboard->isAvailable()) {
-    m_clipboard->startMonitoring();
-  }
 }
 
 void EiScreen::disable()
 {
-  if (m_clipboard && m_clipboard->isAvailable()) {
-    m_clipboard->stopMonitoring();
-  }
+  // Nothing really to be done here
 }
 
 void EiScreen::enter()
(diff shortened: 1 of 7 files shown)


sithlord48:
Thanks but hopefully very soon we will be removing the ability to use wl-clipboard. compositors will have to provide access via the portal see [#9431](https://github.com/deskflow/deskflow/pull/9431)  for more details

mikhailvs:
fair enough. a note for anyone finding this: the fixes live on my fork until the portal path ships

PR #9858, opened 2026-06-16
Portal Globalshortcuts implementation

### LINKED ISSUE #7596: Wayland: Hotkeys don't work
### Project

Deskflow

### Deskflow version number

1.17.0

### Build type

Local developer build (built it yourself)

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [X] Linux
- [ ] BSD-derived
- [ ] Other

### OS versions/distros

- OpenSUSE Tumbleweed 20240929
- KDE Plasma 6.1.5 running Wayland session

### Deskflow configuration

OpenSUSE server, macOS 15 client. OpenSUSE on the right, macOS on the left.

1. In Deskflow click configure on the server, browse to the "Hotkeys" tab
2. Create any hotkey with any action (for example, "-" for lockCursorToScreen(toggle))
3. Save and optionally restart, observe that the hotkey does nothing, even when focusing the Deskflow gui

### Log output

_No response_

### Additional information

I think this is a Wayland issue, probably more 'security.' Wanted to report this as I don't see an existing issue for it and it doesn't appear to be in the known issues list. Apologies if it is already reported somewhere!

Aside from this everything is working fine (except for the already known issues)

KoljaFrahm:
## Description
Requires at least libportal v0.11.0

Currently only works if you start deskflow from the desktop file, otherwise you get 
`ERROR: failed to initialize global shortcuts session, quitting: GDBus.Error:org.freedesktop.portal.Error.NotAllowed: An app id is required
`

## AI tools used for this PR
I used Microsoft Copilot to get an inspiration on how to implement this. The code in the end is written by me.

## Fixed/related issues

needs: #10117 
needs: https://github.com/flatpak/libportal/pull/226
needs: https://github.com/flatpak/libportal/pull/233
fixes: #7596 

## How has this been tested?
Locally on Wayland, test if lock to screen/restart server works

## Problems with the portal implementation:
- If you change the shortcuts in the compositor or the dialogue, there is not a good way for deskflow to recognise this
-  The one mentioned in https://github.com/deskflow/deskflow/issues/7596#issuecomment-5010813949:
> the `switch to computer` hotkeys appear to fail to start an input capture session; my reading of the XDG Desktop Portal InputCapture implementation, there is only one mechanism for applications to be granted an input capture: pointer barriers. I believe the specification allows compositors to implement their own triggers in addition to pointer barriers, but that sounds like a prime spot for integration fragmentation.
> 
> Keyboard-shortcut invoked capture sessions were briefly mentioned here, [flatpak/xdg-desktop-portal#714](https://github.com/flatpak/xdg-desktop-portal/pull/714), but not explicitly shot down.

See also https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.InputCapture.html: 
> The compositor is always in control of input capturing and may filter events or stop capturing at any time. There is currently no way for an application to activate immediate input capture.
Screenshot From 2026-07-07 18-36-37

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -57,8 +57,9 @@ EiScreen::EiScreen(bool isPrimary, IEventQueue *events, bool usePortal)
       handleConnectedToEisEvent(e);
     });
     if (isPrimary) {
-      m_portalInputCapture = new PortalInputCapture(this, m_events);
       // Portal input capture manages its own clipboard
+      m_portalInputCapture = new PortalInputCapture(this, m_events);
+      m_portalGlobalShortcuts = new PortalGlobalShortcuts(this, m_events);
     } else {
       m_events->addHandler(EventTypes::EISessionClosed, getEventTarget(), [this](const auto &) {
         handlePortalSessionClosed();
@@ -93,6 +94,8 @@ EiScreen::~EiScreen()
   delete m_keyState;
   delete m_clipboard;
 
+  delete m_portalGlobalShortcuts;
+  delete m_portalInputCapture;
   delete m_portalRemoteDesktop;
 }
 
@@ -221,7 +224,10 @@ void EiScreen::warpCursor(int32_t x, int32_t y)
 std::uint32_t EiScreen::registerHotKey(KeyID key, KeyModifierMask mask)
 {
   static std::uint32_t next_id;
-  std::uint32_t id = std::min(++next_id, 1u);
+  std::uint32_t id = ++next_id;
+  if (id == 0) {
+    id = ++next_id;
+  }
 
   // Bug: id rollover means duplicate hotkey ids. Oh well.
 
@@ -232,6 +238,8 @@ std::uint32_t EiScreen::registerHotKey(KeyID key, KeyModifierMask mask)
   }
   set->second.addItem(HotKeyItem(mask, id));
 
+  updatePortalGlobalShortcuts();
+
   return id;
 }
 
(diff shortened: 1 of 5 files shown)


KoljaFrahm:
Other than waiting for the libportal PR to be merged, this is now ready for review.

src/lib/platform/EiScreen.cpp
```diff
@@ -694,6 +725,7 @@ ButtonID EiScreen::mapButtonFromEvdev(ei_event *event) const
 
 bool EiScreen::onHotkey(KeyID keyid, bool isPressed, KeyModifierMask mask)
 {
+#ifndef HAVE_LIBPORTAL_SHORTCUTS
```
sithlord48: this looks wrong,  shouldn't we be doing this IF we have the portal ?

KoljaFrahm: We don't need the function onHotKey anymore if we use the portal global shortcuts (I actually don't know why it is there at all).
`sendEvent(type, HotKeyInfo::alloc(id));` is now send in PortalGlobalShortcuts.cpp.
I did this so we can remove this function when most distros updated libportal.

sithlord48: Ah, thanks for the explanation.

KoljaFrahm: Actually my statement was not correct. We need onHotKey when we're connected to a remote. Then, the InputCapture takes priority over the Shortcut session and we will use this function. I changed and documented this in the code now.

sithlord48:
libportal 0.11 was released i am waiting for a proper package so i can fully test this !

sithlord48 (approved):
Built my own arch package to test this and its working,  Great Work @KoljaFrahm !!
@nbolton please land this after you have had a chance to review.

nbolton:
> Built my own arch package to test this and its working, Great Work @KoljaFrahm !!

Excited about this. Missed hotkeys on Wayland.

sithlord48 (changes requested):
@KoljaFrahm  I just noticed this causes the core to crash on exit. If i have a hotkey (i.e the server has the lock to screen button enabled)

KoljaFrahm:
Can't reproduce this on gnome

sithlord48 (approved):
I think i've found the issue it works when i run it fully installed it was having an issue when running from my IDE only

nbolton:
Interesting. I installed `libportal` from source so maybe I did something wrong?

image

nbolton (approved):
Tested and working (tried a basic hotkey from Wayland server to Windows client) apart from the `libportal` render bug (I guess not our bug or some misconfiguration on my end).

KoljaFrahm:
> Interesting. I installed `libportal` from source so maybe I did something wrong?

This is fixed in gnome 51

PR #9860, opened 2026-06-16
feat: remember previously used server addresses

### LINKED ISSUE #8114: Remember previously used server address (hostname/IP)
### Feature Request

When switching between servers, re-typing different hostnames is a little tedious, especially in development where I frequently change which computer is the server.

It would be handy if the client address (hostname/IP) field was a combo textbox dropdown which I could type new addresses into but also select previously used ones.

Image

Zish19:
## Description

Implemented support for remembering previously used server addresses in client mode using an editable `QComboBox`.

* Replaced server address input with an editable dropdown
* Persisted recent server address history using `QSettings`
* Limited history to 10 entries
* Prevented duplicates and moved reused addresses to the top

## AI tools used for this PR

None

## Fixed/related issues

fixes: #8114

## How has this been tested?

Tested locally on Windows by:

* Adding new server addresses
* Restarting the app to verify persistence
* Checking duplicate handling and history ordering
* Verifying history limit of 10 entries

### DIFF
--- a/src/lib/common/Settings.h
+++ b/src/lib/common/Settings.h
@@ -42,6 +42,7 @@ public:
     inline static const auto XScrollScale = QStringLiteral("client/xScrollScale");
     inline static const auto LanguageSync = QStringLiteral("client/languageSync");
     inline static const auto RemoteHost = QStringLiteral("client/remoteHost");
+    inline static const auto ServerAddressHistory = QStringLiteral("client/serverAddressHistory");
     inline static const auto XdpRestoreToken = QStringLiteral("client/xdpRestoreToken");
   };
   struct Core
@@ -210,6 +211,7 @@ private:
     , Settings::Client::InvertXScroll
     , Settings::Client::LanguageSync
     , Settings::Client::RemoteHost
+    , Settings::Client::ServerAddressHistory
     , Settings::Client::YScrollScale
     , Settings::Client::XScrollScale
     , Settings::Client::XdpRestoreToken
--- a/src/lib/gui/MainWindow.cpp
+++ b/src/lib/gui/MainWindow.cpp
@@ -31,10 +31,12 @@
 #include "net/FingerprintDatabase.h"
 #include "widgets/StatusBar.h"
 
+#include <QAbstractItemView>
 #include <QCheckBox>
 #include <QCloseEvent>
 #include <QDesktopServices>
 #include <QFileDialog>
+#include <QKeyEvent>
 #include <QLocalServer>
 #include <QLocalSocket>
 #include <QMenu>
@@ -283,8 +285,9 @@ void MainWindow::connectSlots()
 
   connect(ui->btnRestartCore, &QPushButton::clicked, this, &MainWindow::resetCore);
 
-  connect(ui->lineHostname, &QLineEdit::returnPressed, ui->btnRestartCore, &QPushButton::click);
-  connect(ui->lineHostname, &QLineEdit::textChanged, this, &MainWindow::remoteHostChanged);
+  connect(ui->comboHostname->lineEdit(), &QLineEdit::returnPressed, ui->btnRestartCore, &QPushButton::click);
+  connect(ui->comboHostname, &QComboBox::editTextChanged, this, &MainWindow::remoteHostChanged);
+  connect(ui->btnRemoveHostname, &QPushButton::clicked, this, &MainWindow::removeHostname);
 
   connect(ui->btnSaveServerConfig, &QPushButton::clicked, this, &MainWindow::saveServerConfig);
   connect(ui->btnConfigureServer, &QPushButton::clicked, this, [this] { showConfigureServer(""); });
@@ -401,6 +404,8 @@ void MainWindow::startCore()
   if (m_coreProcess.mode() == CoreMode::Server && Settings::value(Settings::Core::Interface).toString().isEmpty()) {
     m_serverStartIPs = NetworkMonitor::validAddresses();
     m_serverStartSuggestedIP = m_serverStartIPs.isEmpty() ? "" : m_serverStartIPs.first();
+  } else if (m_coreProcess.mode() == CoreMode::Client) {
+    saveServerAddressHistory(ui->comboHostname->currentText());
   }
 
   m_actionStartCore->setVisible(false);
@@ -649,7 +654,7 @@ void MainWindow::open()
   }
 
   if (Settings::value(Settings::Gui::AutoStartCore).toBool()) {
-    if (ui->rbModeClient->isChecked() && ui->lineHostname->text().isEmpty())
+    if (ui->rbModeClient->isChecked() && ui->comboHostname->currentText().isEmpty())
       return;
     startCore();
   }
@@ -703,8 +708,9 @@ void MainWindow::applyConfig()
     setWindowTitle(kAppName);
   }
 
+  loadServerAddressHistory();
   if (const auto host = Settings::value(Settings::Client::RemoteHost).toString(); !host.isEmpty())
-    ui->lineHostname->setText(host);
+    ui->comboHostname->setCurrentText(host);
 
   updateLocalFingerprint();
   setTrayIcon();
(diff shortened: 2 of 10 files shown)


nbolton (changes requested):
Good start and thanks for the PR.
Would you mind making two improvements?
1. Please provide a way to remove items from history (e.g. in case of typo).
2. Use state instead of config to store the address history data (add your key to `m_stateKeys`).

https://github.com/deskflow/deskflow/blob/f51fb4affbd9c0bfd75268bc46a37cfdf4007b49/src/lib/common/Settings.h#L296

Zish19:
Thanks for the review. I will move the history key from config to state storage and add a way to remove entries from the saved server address history.

Zish19:
Addressed both review comments:

* Moved `ServerAddressHistory` from config to state storage by adding it to `m_stateKeys`
* Added support for removing saved addresses using the Delete key on the hostname dropdown

Please let me know if you'd prefer a more explicit removal UI (e.g. context menu or button) instead of the keyboard shortcut.

src/lib/common/Settings.h
```diff
@@ -311,7 +312,7 @@ class Settings : public QObject
   };
 
   // Settings saved in our State file
-  inline static const QStringList m_stateKeys = { Settings::Gui::WindowGeometry };
+  inline static const QStringList m_stateKeys = { Settings::Gui::WindowGeometry, Settings::Client::ServerAddressHistory };
```
sithlord48: Why save to the state and not the config ?

src/lib/gui/MainWindow.cpp
```diff
@@ -215,6 +217,7 @@ void MainWindow::setupControls()
   ui->lineEditName->setValidator(new QRegularExpressionValidator(m_nameRegEx, this));
   ui->lineEditName->setVisible(false);
   ui->lineEditName->installEventFilter(this);
+  ui->lineHostname->view()->installEventFilter(this);
```
sithlord48: why do you need to have an event filter on the line edits view ?

sithlord48:
iirc you can just do this w/ the combo box's allow insert. Yes a button will be needed to remove them
The ui item should also have its name updated following our naming convention.

sithlord48:
If you have read the contributing guide you would know 
1. we do not allow merges
2. do not fix review comments with new commits.

src/lib/common/Settings.h
nbolton: Because I told them that would be better. Why do you think config is better than state?

Edit: Reviewing the code, it's not completely clear when and why someone should use the state. Perhaps a comment here would help others decide better in future.

Zish19:
Understood regarding the merge commit and review workflow , I’ll clean up the branch history and avoid merge commits going forward.

I’ll remove the event-filter-based Delete shortcut and replace it with a more explicit removal button for the saved hostname history. I’ll also rename the UI widget to match the project naming convention.

For storage, I’ll keep using state for now since that was requested, unless you both decide config is preferred.

Zish19:
Addressed the latest review feedback:

* Removed the merge commit and cleaned branch history (linear commits only)
* Removed the event-filter/Delete-key implementation
* Added an explicit `Remove` button for hostname history removal
* Renamed the hostname widget to follow naming conventions (`comboHostname`)
* Kept server history in state storage as previously requested

I could not run a full local build because Qt6 is unavailable in my current Windows environment, but I performed a manual sanity check:

* `git grep lineHostname` returns no results
* all signal/slot connections were verified
* no dangling references remain
* UI definitions and object names match the implementation

Please take another look when convenient.

sithlord48:
Please rebase upon master and keep the fixes in the one commit.

Zish19:
Rebased onto latest master and squashed the branch into a single commit as requested.

src/lib/common/Settings.h
nbolton: If @sithlord48 agrees with me above, you may also need to add it to m_validKeys.

src/lib/gui/MainWindow.cpp
```diff
@@ -1024,18 +1030,51 @@ void MainWindow::changeEvent(QEvent *e)
...
-    return false;
-  if (const auto keyEvent = static_cast<QKeyEvent *>(event); keyEvent->key() != Qt::Key_Escape)
-    return false;
-  ui->lineEditName->hide();
-  ui->lblComputerName->show();
-  ui->btnEditName->show();
-  ui->lineEditName->setText(Settings::value(Settings::Core::ComputerName).toString());
-  toggleCanRunCore(canRunCore());
-  return true;
+  if (obj == ui->lineEditName && event->type() == QEvent::KeyPress) {
```
sithlord48: This has nothing todo with the feature your changing and we perfer to return early here. 

Please revert.

sithlord48 (changes requested):

src/lib/common/Settings.h
nbolton: Looking at XDG Base Directory spec, `XDG_STATE_HOME` is for "actions history (logs, history, recently used files…)" and app view/layout state... things that persist across restarts but aren't deliberate config. 

We could see it this way: Recently-used server addresses are usage history, not a setting the user authored, so state is the correct store (same thinking as window geometry). 

Docs: https://specifications.freedesktop.org/basedir/latest/

Zish19:
Fixed the CI formatting issue and reverted the unrelated changes in `eventFilter()`. The branch still contains a single commit and is ready for another review.

nbolton:
Comments unresolved - does that mean they're not done yet? Unsure if this PR should be open or draft.

Zish19:
All requested changes have been addressed from my side.

The unresolved comments are not pending work anymore, they correspond to earlier review feedback that has since been implemented:

* moved server address history to state storage
* replaced Delete-key removal with an explicit Remove button
* renamed the UI widget to follow naming conventions
* rebased onto latest `master` and squashed to a single commit
* reverted unrelated `eventFilter()` changes and fixed lint issues

The PR was marked as draft during the review/fix cycle because changes were still in progress. I believe it is now ready for review, so I’m happy for it to be moved out of draft unless there are additional requested changes.

sithlord48:
I would much rather use settings since this history is something users may want to move to a new config. State is more about stuff we don't care about recent files for an editor are a little different than a list of machine you connect to that you across installs will likly continue to connect to .. Putting it in the state file will result in users having to find and edit that file if they want to move the history with them. It can also just be removed by the os as its not really a config file.

nbolton:
> It can also just be removed by the os as its not really a config file.

Do you have a link to official docs that say state files are removed? I can't see anything in XDG spec that says that.

> if they want to move the history with them

I think this the stronger argument for not using state; if a use copies their `.config` dir to another machine, then would they expect the server history to come with it? Maybe, yes.

I'm easy either way, just wanted to follow the XDG docs and I guess this one is down to interpretation/personal preference.

@Zish19 Would you mind flipping it back to config instead of state? Sorry to muck you about!

Zish19:
No problem, and thanks for the clarification.

I’ve moved `ServerAddressHistory` back to config storage as requested, rebased onto the latest `master`, and kept the PR as a single commit.

nbolton:
CI failing: `Unexpected changes to the repo, Often caused by forgetting to commit the updated translation files`

Edit: _Set to draft again as we have a lot of PRs to keep track of, go ahead and set to ready when fixed._

Zish19:
Thanks for flagging this.

It looks like the CI is generating additional repository changes during the build (likely translation-related due to the new UI string for the Remove button), which I did not catch locally because I don’t currently have Qt installed to run the full build pipeline.

I’ll inspect which generated files changed and commit them.

sithlord48:
>  I don’t currently have Qt installed to run the full build pipeline.

Do not make PR's without fully testing locally. I expect you to have fully tested this locally before making a PR...

When you have fixed this and tested locally you can open it back for review.

Zish19:
I can’t complete local GUI testing right now due to environment constraints.

nbolton:
@sithlord48 Do you want to open a new PR based on this or shall I? Would be a nice feature to have. Or maybe we can reopen and fix this ourselves.

sithlord48:
If i am gonna write this it would be from scratch . its not "hard" just have not had time todo it .

PR #9867, opened 2026-06-18
fix: do add existing aliases to the alias list correctly remove items on save

sithlord48:
## Description

  - Don't add existing aliases to the alias list
  - Clear the alias list before saving to avoid stale entries
## AI tools used for this PR

None 

## Fixed/related issues

 Fix an issue where when you have an alias and enter the screen settings screen the aliases are duplicated into the alias settings causing the server to be unable to run 

## How has this been tested?

Locally

### DIFF
--- a/src/lib/gui/dialogs/ScreenSettingsDialog.cpp
+++ b/src/lib/gui/dialogs/ScreenSettingsDialog.cpp
@@ -85,6 +85,8 @@ void ScreenSettingsDialog::accept()
 
   m_screen->setName(ui->lineNameEdit->text());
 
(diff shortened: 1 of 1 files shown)


nbolton (commented):
Bug: I can't remove all aliases from a screen... maybe pre-existing bug though? Fix separately?

1. Added 2 test values:
Image

2. Removed them:
Image

3. Re-open screen, and they're back:
Image

4. Also the OK button is disabled on the server config dialog:
Image

sithlord48:
Its fixed now.

nbolton (approved):

PR #9870, opened 2026-06-18
fix: handle top-down DIB (negative biHeight) in Windows clipboard bitmap converter

### LINKED ISSUE #9869: Windows server: deskflow-core crashes (std::length_error 'string too long') converting a top-down DIB (negative biHeight) clipboard image on screen switch
### Summary

On a Windows **server**, `deskflow-core.exe` reliably crashes with
`FATAL: an error occurred: string too long` (an uncaught `std::length_error`, Windows exception
`0xc0000409` in `ucrtbase.dll`) whenever the cursor leaves the server screen **while a top-down DIB
image is on the Windows clipboard**. The daemon restarts the core, the client reconnects, and any
image left on the clipboard turns this into a continuous crash/reconnect loop (39 crashes in ~20
minutes in my logs). This is the same `string too long` failure reported inside #7329 (by
`@sean5446` and `@spnda`) and #8597 (closed as duplicate) and #8625 — none of which fixed it; it is
still present on `master`.

### Root cause

`src/lib/platform/MSWindowsClipboardBitmapConverter.cpp` → `toIClipboard()` assumes a **positive**
`biHeight`. A `BITMAPINFOHEADER` with **negative `biHeight`** is a *top-down* DIB (origin upper-left),
which is legal per the spec and is produced by macOS/Retina clipboards and many screenshot tools.

```cpp
LONG w = bitmap->bmiHeader.biWidth;
LONG h = bitmap->bmiHeader.biHeight;     // e.g. -1560 for a top-down DIB
...
std::string image((const char *)&info, info.biSize);
image.append((const char *)raw, 4 * w * h);   // 4 * 3186 * (-1560) = -19,880,640
```

`4 * w * h` is computed in signed `LONG` and is **negative**; it then widens to a ~1.8×10¹⁹ `size_t`
in `std::string::append(const char*, size_type)`, exceeds `max_size()`, and throws
`std::length_error("string too long")`. The throw is unhandled on the clipboard worker thread, so the
CRT fail-fasts the process. (Secondary log line `resource deadlock would occur` is the failing
shutdown/join during the abort.)

The 24/32-bit fast path a few lines above only triggers for `BI_RGB`; the crashing screenshots are
`BI_BITFIELDS` (`convert image from: depth=32 comp=3`), so they take the conversion path.
`SetDIBitsToDevice(...)` is also passed the negative `h` as height/scanline count, which is invalid.

### Steps to reproduce

1. Windows server + any client (macOS client here), TLS on.
2. On the **server**, copy an image whose DIB is top-down — easiest source: a macOS/Retina screenshot,
   or many Windows screenshot tools. The log will show a **negative** height, e.g. `bitmap: 3186x-1560 32`.
3. Move the cursor from the server to a client screen (clipboard is sent on leave).
4. `deskflow-core` crashes; with the image still on the clipboard, every crossing repeats it.
   Copying short **text** clears the condition and everything works again.

### Logs

```
[09:39:26.946] INFO: switch from "Ivan-PC11" to "JJJs-MacBook-Pro.local" at 2767,-1022
[09:39:26.946] INFO: leaving screen
[09:39:26.971] INFO: bitmap: 3186x-1560 32
[09:39:26.971] INFO: convert image from: depth=32 comp=3
[09:39:26.971] FATAL: an error occurred: string too long
[09:39:26.972] ERROR: standard exception on thread 0x000028c0: resource deadlock would occur
[09:39:29.825] WARNING: detected application not running, pid=29764
[09:39:30.035] IPC: started server, waiting for clients
```
```
Faulting application name: deskflow-core.exe, version: 1.26.0.0
Faulting module name: ucrtbase.dll
Exception code: 0xc0000409
```

Other reporters' bitmaps in #7329 were also top-down: `268x-58`, `2256x-308`. The negative height is
the common factor.

### Environment

- Server: Windows 11 Pro (build 26200), DeskFlow **v1.26.0, protocol v1.8**, TLSv1.3.
- Client: macOS 26.5.1, DeskFlow v1.26.0, protocol v1.8.
- Confirmed present in `v1.26.0` and on `master` (`MSWindowsClipboardBitmapConverter.cpp` line ~113).

### Proposed fix

Use the absolute height and `size_t` arithmetic for the buffer length (the actual crash fix), pass a
non-negative height to `SetDIBitsToDevice`, and null-check `CreateDIBSection`:

```diff
   LONG w = bitmap->bmiHeader.biWidth;
   LONG h = bitmap->bmiHeader.biHeight;
+  // A negative biHeight denotes a top-down DIB (origin upper-left), legal per spec and
+  // produced by e.g. macOS clipboards/screenshot tools. The buffer is abs(biHeight) rows;
+  // the signed product below goes negative and wraps to a huge size_t in append(),
+  // throwing std::length_error ("string too long") and aborting deskflow-core.
+  const LONG absHeight = (h < 0) ? -h : h;
   info.biSize = sizeof(BITMAPINFOHEADER);
@@
   HBITMAP dst = CreateDIBSection(dc, (BITMAPINFO *)&info, DIB_RGB_COLORS, &raw, nullptr, 0);
+  if (dst == nullptr || raw == nullptr) {
+    LOG_WARN("failed to allocate destination bitmap for clipboard image");
+    ReleaseDC(nullptr, dc);
+    GlobalUnlock(data);
+    return std::string();
+  }
@@
-  SetDIBitsToDevice(dstDC, 0, 0, w, h, 0, 0, 0, h, srcBits, bitmap, DIB_RGB_COLORS);
+  SetDIBitsToDevice(dstDC, 0, 0, w, absHeight, 0, 0, 0, absHeight, srcBits, bitmap, DIB_RGB_COLORS);
@@
   std::string image((const char *)&info, info.biSize);
-  image.append((const char *)raw, 4 * w * h);
+  image.append((const char *)raw, static_cast<size_t>(4) * static_cast<size_t>(w) * static_cast<size_t>(absHeight));
```

I can open a PR with this. Caveat: I don't have a Windows + Qt6 + MSVC build environment, so the patch
is **reasoned but not compiled/tested** — in particular the top-down orientation of the resulting image
should be verified (the size/crash fix is the critical part). Happy to iterate.

### Related

- #7329 (contains this exact `string too long` repro; closed COMPLETED but not fixed)
- #8597 (closed as duplicate), #8625 (closed COMPLETED; users still reproduced)
- #9842 / #9843 fix a *different* clipboard-image crash (use-after-free SIGSEGV on reconnect), not this one

webcare-live:
## Description

`MSWindowsClipboardBitmapConverter::toIClipboard()` sized the destination pixel buffer with `4 * w * h`. When the source clipboard image is a **top-down DIB** — i.e. `biHeight` is **negative**, which is legal per the `BITMAPINFOHEADER` spec and is what macOS/Retina clipboards and several Windows screenshot tools produce — the signed product is negative and wraps to a huge `size_t` in `std::string::append(ptr, count)`. That exceeds `max_size()` and throws `std::length_error("string too long")`. The throw is unhandled on the clipboard worker thread, so `deskflow-core` fail-fasts (Windows exception `0xc0000409`).

**Symptom:** on a Windows server, moving the cursor to a client screen while an **image** is on the Windows clipboard crashes/restarts `deskflow-core`, and keeps doing so for as long as that image stays on the clipboard (copying plain text clears it). Observed bitmaps all had negative heights — e.g. `3186x-1560`, and `268x-58` / `2256x-308` in the reports collected under #7329. The 24/32-bit "canonical form" fast-path above is skipped for these because they are `BI_BITFIELDS` (`convert image from: depth=32 comp=3`), not `BI_RGB`, so they take the conversion path.

This change:
- computes `absHeight = abs(biHeight)` and uses it (with `size_t` arithmetic) for the `append` length — the actual crash fix;
- passes a non-negative height / scan-line count to `SetDIBitsToDevice` (a negative count is invalid);
- null-checks `CreateDIBSection` before dereferencing `raw`.

## Disclosure of AI use

This patch — both the root-cause analysis and the code — was produced with **Claude (Anthropic's Claude Code)**. A human filed #9869 and is coordinating, and (per the offer below) will help test.

## Related Issue

fixes: #9869

## How Has This Been Tested?

**Not yet runtime-tested.** I don't have a Windows + Qt6 + MSVC build environment, so `deskflow-core` could not be compiled/run on my side — which is why this is a **draft** (@nbolton kindly offered to test on #9869). The change is confined to one function, uses an existing log macro (`LOG_WARN`, `base/Log.h`), and is reasoned from the crash and the GDI/DIB semantics rather than observed in a build.

The **buffer-size calculation is the critical fix** (it's the throw site). The thing most worth confirming once built is that **top-down images still come through correctly oriented**, not merely that the crash is gone. Happy to adjust based on your testing.

### DIFF
--- a/src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
+++ b/src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
@@ -72,6 +72,13 @@ std::string MSWindowsClipboardBitmapConverter::toIClipboard(HANDLE data) const
   BITMAPINFOHEADER info;
   LONG w = bitmap->bmiHeader.biWidth;
   LONG h = bitmap->bmiHeader.biHeight;
+  // A negative biHeight denotes a top-down DIB (origin at the upper-left). This is legal
+  // per the BITMAPINFOHEADER spec and is produced by e.g. macOS and some screenshot tools.
+  // The pixel buffer always contains abs(biHeight) rows. Using the signed height in the byte
+  // count below makes the product negative, which then wraps to an enormous size_t in
+  // std::string::append() and throws std::length_error ("string too long"), aborting
+  // deskflow-core (0xc0000409).
+  const LONG absHeight = (h < 0) ? -h : h;
   info.biSize = sizeof(BITMAPINFOHEADER);
   info.biWidth = w;
   info.biHeight = h;
@@ -85,6 +92,12 @@ std::string MSWindowsClipboardBitmapConverter::toIClipboard(HANDLE data) const
   info.biClrImportant = 0;
   HDC dc = GetDC(nullptr);
   HBITMAP dst = CreateDIBSection(dc, (BITMAPINFO *)&info, DIB_RGB_COLORS, &raw, nullptr, 0);
+  if (dst == nullptr || raw == nullptr) {
+    LOG_WARN("failed to allocate destination bitmap for clipboard image");
+    ReleaseDC(nullptr, dc);
+    GlobalUnlock(data);
+    return std::string();
+  }
 
   // find the start of the pixel data
   const char *srcBits = (const char *)bitmap + bitmap->bmiHeader.biSize;
@@ -103,14 +116,14 @@ std::string MSWindowsClipboardBitmapConverter::toIClipboard(HANDLE data) const
   // copy source image to destination image
   HDC dstDC = CreateCompatibleDC(dc);
   HGDIOBJ oldBitmap = SelectObject(dstDC, dst);
-  SetDIBitsToDevice(dstDC, 0, 0, w, h, 0, 0, 0, h, srcBits, bitmap, DIB_RGB_COLORS);
+  SetDIBitsToDevice(dstDC, 0, 0, w, absHeight, 0, 0, 0, absHeight, srcBits, bitmap, DIB_RGB_COLORS);
   SelectObject(dstDC, oldBitmap);
   DeleteDC(dstDC);
   GdiFlush();
 
   // extract data
   std::string image((const char *)&info, info.biSize);
-  image.append((const char *)raw, 4 * w * h);
+  image.append((const char *)raw, static_cast<size_t>(4) * static_cast<size_t>(w) * static_cast<size_t>(absHeight));
 
   // clean up GDI
   DeleteObject(dst);


sithlord48:
replaced by #10023

PR #9878, opened 2026-06-19
Remove server dead corners

### LINKED ISSUE #8355: Port Screens to use new Settings
### Feature Request

Screens uses the QSettingsproxy to write the group at once

Port this to the new Settings

sithlord48:
## Description

 - Removes dead corners from the server config (set them via the screen options)
## AI tools used for this PR

None
## Fixed/related issues

 - The values moved to Deskflow.conf where not being added to the server config
## How has this been tested?

Locally


src/lib/common/Settings.h
```diff
@@ -333,7 +333,9 @@ class Settings : public QObject
     , QStringLiteral("internalConfig/protocol")
     , QStringLiteral("internalConfig/numColumns")
     , QStringLiteral("internalConfig/numRows")
-    , QStringLiteral("interlanConfig/relativeMouseMoves")
+    , QStringLiteral("internalConfig/relativeMouseMoves")
+    , QStringLiteral("internalConfig/switchCorner")
```
nbolton, line 337: Was it `switchCornerArray/*`? (i.e. a bunch of leaf keys)

src/lib/server/Config.h
```diff
@@ -478,6 +478,8 @@ class Config
   inline static const QStringList m_oldNames = {
       QStringLiteral("clipboardSharing"),
       QStringLiteral("clipboardSharingSize"),
+      QStringLiteral("switchCornerSize"),
+      QStringLiteral("switchCorner"),
```
nbolton: I think the old screen one was `switchCorners`? Not sure though.

src/lib/common/Settings.h
sithlord48: Yes but due to how the check is done any key that contains "internalConfig/switchCorner" is removed

src/lib/server/Config.h
sithlord48: it was i've updated it.

nbolton (approved):

PR #9879, opened 2026-06-20
Move aliases to general config

### LINKED ISSUE #9880: Port Aliases to the new settings
Port the alias settings to the general configuration

sithlord48:
## Description

The main goal of this pr is to setup a per screen group in `Deskflow.conf` and provide the needed support methods for moving other screen related settings into the general configuration. Changes are as follows: 

 1. Remove a few unused methods i found in `Config` related to removing aliases from configured screens (may come back as we hot load settings in the future) 
    - Removes `Config::removeAlias`
    - Removes `Config::removeAliases`
    - Removes `Config::removeAllAliases`
 3. Lays the ground work for having a proper screen section
    - Clean up unknown groups when saving the server dialog. 
    - Allow for saving of lists to get past our `Settings::cleanSettings` method
    - New `Settings::removeUnknownGroups(const QString knownScreens)` can bu used to remove unknown groups from the settings file used when saving the screens from the general configuration. 
4. Move Alias to general config
    - Move the keys to Deskflow.conf
    - Remove Config reading of the `alias` section 
       - Use `Config::addScreen` to setup the aliases for each screen
       - Update `Config::renameScreen` to move aliases 
       - Update `Config::removeScreen` to remove screen group
       - Did not update `Config::addAlias` this is unused outside of a test and may be removed 
 5. Modify the Server config to save screen aliases to the settings without having to first start the core 

## AI tools used for this PR

None
## Fixed/related issues

fixes: #9880 
## How has this been tested?

Tested locally with an alias connection
 Should be tested by others using aliases as a connection name.

### DIFF
--- a/docs/user/configuration.md
+++ b/docs/user/configuration.md
@@ -162,6 +162,15 @@ This section contains options used when in server mode it will begin with `[serv
 
  - You can use both the ''switchDelay'' and ''switchDoubleTap'' options at the same time. Deskflow will switch when either requirement is satisfied.
 
+### Screen Settings
+
+Each screen will have a section where its configuration will be stored, if the screen was named "foo" the section will be named `[screen_foo]`
+
+|Option              |    Valid Values    |Description|
+|:-------------------|:------------------:|:-----------|
+| aliases            | Comma separated list of hostnames | Names here will be used as alternatives for the computer. Names must be valid hostnames. |
+
+
 ### InternalConfig
 
 This section contains options used when in server mode it will begin with `[internalConfig]`
@@ -184,7 +193,6 @@ hotkeys\1\keys\1\key=83
 hotkeys\1\keys\size=1
 hotkeys\size=1
 screens\1\name=
-screens\10\aliasArray\size=0
 screens\10\fixArray\1\fix=false
 screens\10\fixArray\2\fix=false
 screens\10\fixArray\3\fix=false
@@ -214,7 +222,6 @@ screens\3\name=
 screens\4\name=
 screens\5\name=
 screens\6\name=
-screens\7\aliasArray\size=0
 screens\7\fixArray\1\fix=false
 screens\7\fixArray\2\fix=false
 screens\7\fixArray\3\fix=false
@@ -234,7 +241,6 @@ screens\7\switchCornerArray\3\switchCorner=false
 screens\7\switchCornerArray\4\switchCorner=false
 screens\7\switchCornerArray\size=4
 screens\7\switchCornerSize=0
-screens\8\aliasArray\size=0
 screens\8\fixArray\1\fix=false
 screens\8\fixArray\2\fix=false
 screens\8\fixArray\3\fix=false
@@ -254,7 +260,6 @@ screens\8\switchCornerArray\3\switchCorner=false
 screens\8\switchCornerArray\4\switchCorner=false
 screens\8\switchCornerArray\size=4
 screens\8\switchCornerSize=0
-screens\9\aliasArray\size=0
 screens\9\fixArray\1\fix=false
 screens\9\fixArray\2\fix=false
 screens\9\fixArray\3\fix=false
@@ -292,11 +297,10 @@ end
 Comments are introduced by ''#'' and continue to the end of the line. ''name'' must be one of the following:
 
 * ''screens''
-* ''aliases''
 * ''links''
 * ''options''
 
-The file is parsed top to bottom and names cannot be used before they've been defined in the <code>screens</code> or <code>aliases</code> sections. So the <code>links</code> and <code>aliases</code> must appear after the <code>screens</code> and <code>links</code> cannot refer to aliases unless the <code>aliases</code> appear before the <code>links</code>.
+The file is parsed top to bottom and names cannot be used before they've been defined in the `screens` or as an alias in the general config. So the `links` must appear after the `screens`.
 
 ### The screens section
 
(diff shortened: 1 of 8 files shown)


src/lib/server/Config.cpp
```diff
@@ -605,8 +566,6 @@ void Config::readSection(ConfigReadContext &s)
     readSectionScreens(s);
   } else if (name == s_links) {
     readSectionLinks(s);
-  } else if (name == s_aliases) {
-    readSectionAliases(s);
```
nbolton, line 609: Thinking out loud: when we delete settings, this will cause older config file parsing to fail for people running the core direct. Is this is the desired behavior? I assume yes because the legacy server config file is going away.

src/lib/common/Settings.h
```diff
@@ -123,6 +123,11 @@ class Settings : public QObject
     inline static const auto XdpRestoreToken = QStringLiteral("server/xdpRestoreToken");
   };
 
+  struct Screen
+  {
+    inline static const auto Aliases = QStringLiteral("%1/aliases");
```
nbolton: I guess something a user could do here is name their computer `server` which if I understand correctly could cause a collision -- do we care?

src/lib/server/Config.cpp
```diff
@@ -46,6 +46,11 @@ bool Config::addScreen(const std::string &name)
   // add name
   m_nameToCanonicalName.insert(std::make_pair(name, name));
 
+  // insertAliases
```
nbolton: I'm not sure what this comment means.

docs/user/configuration.md
```diff
@@ -162,6 +162,15 @@ This section contains options used when in server mode it will begin with `[serv
 
  - You can use both the ''switchDelay'' and ''switchDoubleTap'' options at the same time. Deskflow will switch when either requirement is satisfied.
 
+### Screen Settings
+
+Each screen will have a section where its configuration will be stored, if the screen was named "foo" the section will be named `[foo]`
+
+|Option              |    Valid Values    |Description|
+|:-------------------|:------------------:|:-----------|
+| aliases            | comma seperated list of host names | names here will be used as alternate names for the screen, Names must compliant hostnames.|
```
nbolton:
```suggestion
| aliases            | Comma separated list of hostnames | Names here will be used as alternatives for the computer. Names must be valid hostnames. |
```

docs/user/configuration.md
```diff
@@ -298,11 +303,10 @@ end
 Comments are introduced by ''#'' and continue to the end of the line. ''name'' must be one of the following:
 
 * ''screens''
-* ''aliases''
 * ''links''
 * ''options''
 
-The file is parsed top to bottom and names cannot be used before they've been defined in the <code>screens</code> or <code>aliases</code> sections. So the <code>links</code> and <code>aliases</code> must appear after the <code>screens</code> and <code>links</code> cannot refer to aliases unless the <code>aliases</code> appear before the <code>links</code>.
+The file is parsed top to bottom and names cannot be used before they've been defined in the <code>screens</code> or As an alias in the general config. So the <code>links</code> must appear after the <code>screens</code>.
```
nbolton:
```suggestion
The file is parsed top to bottom and names cannot be used before they've been defined in the <code>screens</code> or as an alias in the general config. So the <code>links</code> must appear after the <code>screens</code>.
```

nbolton (changes requested):

src/lib/server/Config.cpp
sithlord48: correct.

src/lib/common/Settings.h
sithlord48: no the new keys would just be stored under [server/alias]

maybe an issue if they remove that screen but the screen names are tied to host names. and they will also reset the server config... 
we could just ban names that match any of our internal group names used in settings if we like

src/lib/server/Config.cpp
sithlord48: it adds the alias names to the screen updated to 
`// add aliases`

sithlord48:
Gonna draft this for now to think about how to deal with the external config alias section.

src/lib/server/Config.cpp
sithlord48: I have adjusted this to skip the aliases section if found , we just read aliases until we reach the `end`
I have also added a log message saying we move this to general settings.

src/lib/common/Settings.h
sithlord48: Adjusted the screen names to use `screen_%1/...`  so all screens are prefixed with screen_

nbolton:
Will review today.

docs/user/configuration.md
```diff
@@ -597,6 +586,8 @@ Additionally, a name of the form `\uXXXX` where ''XXXX'' is a hexadecimal number
 
 ### Example textual configuration file
 
+the alias section is no longer in the server config
```
nbolton:
```suggestion
The alias section is no longer in the server config.
```

src/lib/server/Config.cpp
```diff
@@ -876,39 +839,13 @@ void Config::readSectionLinks(ConfigReadContext &s)
 
 void Config::readSectionAliases(ConfigReadContext &s)
 {
+  qInfo() << "Your server config has an alias section. Alias have moved to the general config this section will no be "
```
nbolton: Shouldn't this be a warning?

nbolton (changes requested):

nbolton (approved):

PR #9883, opened 2026-06-20
Correctly apply the values stored in the new settings to the server configuration

sithlord48:
## Description

 - Move settings of the server options to before we parse the options section as that will return when end it hit
## AI tools used for this PR

None
## Fixed/related issues

Items saved in in settings were not being added to the server options. 

## How has this been tested?

Tested locally to check all options are being sent as expected.

### DIFF
--- a/src/lib/server/Config.cpp
+++ b/src/lib/server/Config.cpp
@@ -614,6 +614,25 @@ void Config::readSection(ConfigReadContext &s)
 
 void Config::readSectionOptions(ConfigReadContext &s)
 {
+  if (Settings::value(Settings::Server::EnableHeatbeat).toBool()) {
+    addOption("", kOptionHeartbeat, Settings::value(Settings::Server::Heartbeat).toInt());
+  }
+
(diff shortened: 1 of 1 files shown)


cathyjf:
This fixes a pretty important regression. Without this patch, clipboard sharing is silently enabled when upgrading to the latest changeset. Clipboard sharing has significant security implications, so enabling it without notice was problematic.

sithlord48:
its been only a few commits since clipboard was moved. That is one of the risks you take for running continuations. Should have this landed soon as it gets reviewed. If your needing this sooner can install the Ci build for this PR.

cathyjf:
> its been only a few commits since clipboard was moved. That is one of the risks you take for running continuations. Should have this landed soon as it gets reviewed. If your needing this sooner can install the Ci build for this PR.

Thanks for the advice. I only wanted to highlight the security implications of changing the settings relating to the clipboard; they are not ordinary preferences. For example, the clipboard might hold secrets that a person doesn't want shared with other machines.

nbolton (approved):

PR #9884, opened 2026-06-21
refactor: replace use of insert for std::maps with try_emplace

sithlord48:
## Description

Replace uses of `std::map<>::insert` with `std::map<>::try_emplace` as recommended by modern C++ standards.
## AI tools used for this PR

None
## Fixed/related issues

None , maybe some sonar issues ? 
## How has this been tested?

Locally tested with a clean build and connection to a client.

### DIFF
--- a/src/lib/server/Config.cpp
+++ b/src/lib/server/Config.cpp
@@ -41,10 +41,10 @@ bool Config::addScreen(const std::string &name)
   }
 
   // add cell
-  m_map.insert(std::make_pair(name, Cell()));
+  m_map.try_emplace(name, Cell());
 
   // add name
-  m_nameToCanonicalName.insert(std::make_pair(name, name));
+  m_nameToCanonicalName.try_emplace(name, name);
 
   return true;
 }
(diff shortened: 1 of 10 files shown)


src/lib/server/Config.cpp
```diff
@@ -41,10 +41,10 @@ bool Config::addScreen(const std::string &name)
   }
 
   // add cell
-  m_map.insert(std::make_pair(name, Cell()));
+  m_map.try_emplace(name, Cell());
```
nbolton, line 44: Sonar on this line, want to fix or ignore for now?

https://sonarcloud.io/project/issues?id=deskflow_deskflow&issues=AZ7qsbD9XuvsBMnFY9e3&open=AZ7qsbD9XuvsBMnFY9e3&pullRequest=9884

nbolton (approved):

src/lib/server/Config.cpp
sithlord48: That is a different issue so i didn't plan to address that here.

PR #9892, opened 2026-06-24
use recursive mutex for clipboards on x and wayland

### LINKED ISSUE #9890: Commit 0042b43 crashes on OpenBSD

### Deskflow version info

Deskflow 1.26.0.9999 (0042b43b)
Qt: 6.10.2
System: OpenBSD-7.9/amd64
Session X.org

### Build types

### Deskflow configuration

- OpenBSD server, Windows 11 client
- Each computer has a single monitor
- Windows is on the left, and OpenBSD is on the right

1. Start Deskflow server and client
2. Move Mouse Pointer to the left OpenBSD machine
3. When pointer goes to Windows, deskflow-core crashes with SIGABRT.

### Log output

```
uaa@framboise:~/z/deskflow/build$ bin/deskflow-core server
[2026-06-24T23:02:59.714] INFO: initial settings file: /home/uaa/.config/Deskflow/Deskflow.conf
[2026-06-24T23:02:59.715] INFO: starting core
[2026-06-24T23:02:59.717] INFO: using legacy x windows screen
[2026-06-24T23:03:00.063] INFO: accepted secure socket
[2026-06-24T23:03:00.064] INFO: accepted client connection
[2026-06-24T23:03:00.064] INFO: saying hello as Barrier, protocol v1.8
[2026-06-24T23:03:00.085] INFO: local layouts: en
[2026-06-24T23:03:01.638] INFO: switch from "framboise" to "DESKTOP-251U0UF" at 1919,878
[2026-06-24T23:03:01.638] INFO: leaving screen
Abort trap                 (コアダンプ) bin/deskflow-core server
uaa@framboise:~/z/deskflow/build$
```

</details>

### Additional information

It is obvious that src/lib/platform/XWindowsClipboard.cpp changes causes the problem.
I tried with `std::mutex` to `std::recursive_mutex` in src/lib/platform/XWindowsClipboard.h and deskflow-core did not stop with abort trap.

When someone tries to `std::scoped_lock` twice with same mutex, OpenBSD will abort (SIGABRT) to avoid deadlock. I think this might be happen.

sithlord48:
## Description

use recursive mutex in ei and x clipbords
## AI tools used for this PR

None
## Fixed/related issues

fixes: #9890 
## How has this been tested?

locally for wayland server

### DIFF
--- a/src/lib/deskflow/Clipboard.h
+++ b/src/lib/deskflow/Clipboard.h
@@ -55,7 +55,7 @@ public:
 
 private:
   mutable bool m_open = false;
-  mutable std::mutex m_mutex;
+  mutable std::recursive_mutex m_mutex;
   mutable Time m_time;
   bool m_owner = false;
   Time m_timeOwned;
--- a/src/lib/platform/EiClipboard.h
+++ b/src/lib/platform/EiClipboard.h
@@ -44,7 +44,7 @@ private:
   ClipboardID m_id;
   mutable bool m_open = false;
   mutable Time m_time = 0;
-  mutable std::mutex m_mutex;
+  mutable std::recursive_mutex m_mutex;
   bool m_owner = false;
   Time m_timeOwned = 0;
   bool m_added[static_cast<int>(Format::TotalFormats)] = {false, false, false};
--- a/src/lib/platform/XWindowsClipboard.h
+++ b/src/lib/platform/XWindowsClipboard.h
@@ -287,7 +287,7 @@ private:
   Window m_window;
   ClipboardID m_id;
   Atom m_selection;
-  mutable std::mutex m_mutex;
+  mutable std::recursive_mutex m_mutex;
   mutable bool m_open = false;
   mutable Time m_time = 0;
   bool m_owner = false;


sithlord48:
@nbolton  Sonar is not happy with these changes  lets hold off for now.

@jg1uaa You think it would be sane to only use these on openBSD ?

jg1uaa:
@sithlord48 I tried resembled approach, only changing XWindowsClipboard.h (did not modify others) solved my problem. I think XWindowsClipboard issue.

jg1uaa:
Is this test will help to solve the problem?

- use `try { std::scoped_lock lock{m_mutex}; } catch (const std::exception &e) {} catch(...) {}` to avoid SIGABRT -> where lock causes problem

Both XWindowsClipboard::checkCache() and XWindowsClipboard::doFillCache() require this modification. Others are `std::scoped_lock lock{m_mutex};`.

sithlord48:
The problem is more that its only on bsd that this is an issue and recursive mutexes can be problematic.

For this reason i think the fix should be limited in where its used. 

Are you able to try the wayland screen implementation  to see if it has a similar issue ?

jg1uaa:
I agree that using recursive mutex makes unhappy future.
I simply used it to check the issue is caused by re-entered mutex call, and I think it seems to be existed but I could not find it.
I don't have wayland environment so I cannot test on it, sorry.

jg1uaa:
I am trying with `m_mutex.try_lock()` to check the place of attempting double-lock.
(log1.txt is simply tested, log2.txt is remove lock from icccmGetSelection().)

I found icccmGetSelection() is called under locked by checkCache().
It looks not only this, but others not yet found.
And looking the source code, empty() calls clearCache() under lock. I think it also will be deadlock.

How about to modify like this?
clearCache(): interface for external, lock and call doClearCache()
doClearCache(): main body of clear cache, it must be called under lock. (private: recommended)
And, revise icccmGetSelection() related lock.

[log1.txt](https://github.com/user-attachments/files/29402380/log1.txt)
[log2.txt](https://github.com/user-attachments/files/29402391/log2.txt)

sithlord48:
@jg1uaa  i am away from my dev setup for a bit if you can draft a patch or snippit of code i will try that  when i return

sithlord48:
Replaced by #9900

Issue #9121, opened 2025-11-05
Keep clients awake while working on the server


andyearnshaw:
### Feature Request

I am aware of the option to "Prevent this computer from going to sleep". That is overkill for me, as I would still like my computers to go to sleep after a period of inactivity. However, with this option unchecked, I find it annoying that my client machine will turn its screen off if I am doing something that takes more than a few minutes on the server. Usually, I want to be able to glance at the other machine's screen even if I'm not actively working on it.

It would be ideal if actions that keep the server awake also kept clients awake. Actions such as:

* Mouse input
* Keyboard input
* Gamepad input
* Watching media/conference call

I don't know if it could be more simple, perhaps through just synchronising the screen off behaviour, e.g. "keep awake" is on until the server turns its screen off, then turn the screen off on the client too. Likewise, synchronising a wake might also be good - when the server screen comes back on, turn on the client screen. If waking from sleep, send the magic packet to the last connected client (or something).

### Voting Instructions

React with :+1: or :-1: to vote on this request

nbolton:
Fun fact: This used to be called "Screensaver sync" (which did as you describe) but that stopped working properly so we renamed the setting to "Keep client awake" (but that's not exactly ideal).

sithlord48:
What OS support is there for this type of feature ? The only one thing I know about is the prevent sleep style offering

The only way to do this without os support is to send input to the client(s) when the server has focus (or another client, right?) . What do you send some mouse movement? This will result in a bug report when someone is trying to watch a video in full screen and don't' want to see the related controls keep popping up.

andyearnshaw:
My first thought was sending input but you raise a good point about the mouse, though I did wonder if sending a dead key could do it too. I did a little research and there does appear to be some OS support.

For resetting idle timers on client machines:
* Windows has the Win32 API function [`IConsolePower::ResetIdleTimer()`](https://learn.microsoft.com/en-us/windows/win32/api/mmc/nf-mmc-iconsolepower-resetidletimer)
* On Linux, you can [ping the DBus interface](https://stackoverflow.com/a/31504731) `org.freedesktop.ScreenSaver.SimulateUserActivity`
* On macOS, apparently you can [reset `HIDIdleTime`](https://stackoverflow.com/a/58572776)

sithlord48:
The other half of this is knowing of the server has active input from a client.. when the server has input the clients are not getting any info about it. So currently on our side i think the best we can do is react when connected or not.

[...]

PR #9894, opened 2026-06-25
fix: release EIS emulation grab when idle so client screens can sleep

ElCruncharino:
## Description

On the Wayland/libei backend, a deskflow **client** (secondary screen) keeps its
input devices in the EIS *emulating* state for the entire time the cursor is on
that screen: `EiScreen::enter()` calls `ei_device_start_emulating()` and the grab
is only released in `leave()`.

While that grab is held, the compositor will not power the output off via DPMS.
The practical effect on a client is that the screen can **never go to sleep**
(nor be blanked by an idle-lock/screensaver) as long as the deskflow cursor sits
on it, even when no input is actually flowing and the machine is otherwise idle.
Notably the grab does **not** reset the compositor's idle timer, so an idle
DPMS-off is still *requested* on schedule and then immediately aborted/reverted.

This change makes the emulation grab follow *actual input* rather than *cursor
presence*:

- `enter()` no longer eagerly `start_emulating()`s; it positions the cursor via
  `fakeMouseMove()`, which starts emulation on demand.
- A new `ensureEmulating()` starts emulation on the first injected event and
  (re)arms a short one-shot idle timer (`IEventQueue::newOneShotTimer`).
- A new `stopEmulating()` releases the grab when the idle timer fires, or on
  `leave()`.

After `s_idleEmulationTimeout` (4 s) with no relayed input, the grab is released,
so the compositor is free to power the screen off; the next injected event
re-grabs and wakes it. Screens that are actively being used are unaffected (the
grab simply stays alive while events keep arriving). Only `EiScreen.cpp/.h` are
touched; the primary/server path is unchanged.

Open question for maintainers: `s_idleEmulationTimeout` is a hard-coded 4 s. Happy
to make it a setting if preferred.
Moreover, if this behavior of...not allowing DPMS while the mouse/input is active on the client is a feature, not a bug, then so be it, but it ends up not allowing DPMS at all in my instance, so happy to refactor this as a configurable user-facing option if other users would PREFER this behavior.

As far as I can tell, since I have "prevent this computer from going to sleep" disabled, this wouldn't necessarily be expected behavior.

## AI tools used for this PR

Claude (model: Claude Opus 4.8) was used throughout: to diagnose the root cause
(comparing `ei_device_start/stop_emulating` lifetime against the compositor's
DPMS behaviour), to design and write the patch, and to drive the runtime
validation below. All changes were reviewed and tested by me on real hardware.

## Fixed/related issues

Loosely related to #9121 (but I think this is its own behavior bug).

## How has this been tested?

Built from this branch (`cmake -B build -G Ninja && cmake --build build`) and run
as the live client, validated on real hardware:

- Hardware/stack: AMD RX 9070 XT (RDNA4, amdgpu), MSI MPG 274U over DisplayPort,
  KWin Wayland / Plasma 6.7, openSUSE Tumbleweed; deskflow client connected to a
  remote server.
- Method: with the deskflow cursor confirmed parked on the client screen (via the
  `entering screen` log, no `leaving screen` during the test), requested a
  DPMS-off and watched the amdgpu driver's `link_set_dpms_off` / `link_set_dpms_on`
  trace (`drm.debug=0x16`).
- Stock binary: DPMS-off is aborted (no completed `link_set_dpms_off`; the panel
  flashes dark for ~0.5 s and comes back).
- Patched binary: after the 4 s idle-release fires, DPMS-off **completes and holds**
  (`link_set_dpms_off` with no following `link_set_dpms_on`); the panel stays off
  until input arrives, then re-grabs and wakes. Confirmed both in the kernel trace
  and visually (held dark ~10 s until the screen was turned back on).
- Active use sanity check: moving the mouse / typing on the client immediately
  re-grabs and behaves as before.

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -88,6 +88,8 @@ EiScreen::~EiScreen()
   m_events->adoptBuffer(nullptr);
   m_events->removeHandler(EventTypes::System, m_events->getSystemTarget());
 
+  cancelIdleEmulationTimer();
+
   cleanupEi();
 
   delete m_keyState;
@@ -294,6 +296,7 @@ void EiScreen::fakeMouseButton(ButtonID button, bool press)
     break;
   }
 
+  ensureEmulating();
   ei_device_button_button(m_eiPointer, code, press);
   ei_device_frame(m_eiPointer, ei_now(m_ei));
 }
@@ -310,6 +313,7 @@ void EiScreen::fakeMouseMove(int32_t x, int32_t y)
   if (!m_eiAbs)
     return;
 
+  ensureEmulating();
   ei_device_pointer_motion_absolute(m_eiAbs, x, y);
   ei_device_frame(m_eiAbs, ei_now(m_ei));
 }
@@ -319,6 +323,7 @@ void EiScreen::fakeMouseRelativeMove(int32_t dx, int32_t dy) const
   if (!m_eiPointer)
     return;
 
+  ensureEmulating();
   ei_device_pointer_motion(m_eiPointer, dx, dy);
   ei_device_frame(m_eiPointer, ei_now(m_ei));
 }
@@ -332,6 +337,7 @@ void EiScreen::fakeMouseWheel(ScrollDelta delta) const
   // libei and deskflow seem to use opposite directions, so we have
   // to send EI the opposite of the value received if we want to remain
   // compatible with other platforms (including X11).
+  ensureEmulating();
   ei_device_scroll_discrete(m_eiPointer, -delta.x, -delta.y);
   ei_device_frame(m_eiPointer, ei_now(m_ei));
 }
@@ -343,6 +349,7 @@ void EiScreen::fakeKey(uint32_t keycode, bool isDown) const
 
   auto xkbKeycode = keycode + 8;
   m_keyState->updateXkbState(xkbKeycode, isDown);
+  ensureEmulating();
   ei_device_keyboard_key(m_eiKeyboard, keycode, isDown);
   ei_device_frame(m_eiKeyboard, ei_now(m_ei));
 }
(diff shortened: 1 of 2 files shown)


sithlord48:
What about other screen types do they have the same issue ?

ElCruncharino:
> What about other screen types do they have the same issue ?

Good question. I can answer anecdotally and say I haven't seen this for my macOS or Windows clients. Haven't run a X11 environment for awhile and can't speak to that.

Furthermore, I checked the other backends and this _appears to be_ specific to the Wayland/libei (EiScreen) path.

EiScreen is the only backend that holds a persistent grab while the cursor is on the screen: enter() calls ei_device_start_emulating() and it isn't released until leave(), and the compositor won't DPMS-off an output that has an active emulating device.

 The other backends only do a one-time wake when the cursor enters, then let the normal OS idle/screensaver proceed, so an idle client still sleeps:
-  X11 (XWindowsScreen::enter): DPMSForceLevel(DPMSModeOn) once.
- Windows (MSWindowsScreen::enter → ArchMiscWindows::wakeupDisplay): SetThreadExecutionState(ES_DISPLAY_REQUIRED) without ES_CONTINUOUS (the code comments that it intentionally avoids a persistent hold).
 - macOS (OSXScreen::enter): sets IORequestIdle=false on the IODisplayWrangler once (not reset on leave, I guess this would result in being a nudge rather than a held assertion).

So none of them hold the display awake for the duration the cursor is parked. There's no equivalent persistent grab to release, and the fix is scoped to EiScreen.

sithlord48:
We don't allow merge commits only rebase please only use the rebase in the future.

src/lib/platform/EiScreen.cpp
```diff
@@ -361,19 +372,60 @@ void EiScreen::disable()
...
+    ei_device_stop_emulating(m_eiKeyboard);
+  if (m_eiAbs)
+    ei_device_stop_emulating(m_eiAbs);
+  m_isEmulating = false;
+}
+
+void EiScreen::enter()
+{
+  m_isOnScreen = true;
+  if (!m_isPrimary) {
```
sithlord48: i would consider changing the logic on this to 

```
if (m_isPrimary) {
    LOG_DEBUG("releasing input capture at x=%i y=%i", m_cursorX, m_cursorY);
    m_portalInputCapture->release(m_cursorX, m_cursorY);
    return;
}

if (m_eiAbs)
  fakeMouseMove(m_cursorX, m_cursorY);
   
```

sithlord48 (changes requested):

src/lib/platform/EiScreen.cpp
ElCruncharino: Good call, combined them. Also deduped the timer teardown.

src/lib/platform/EiScreen.h
```diff
@@ -151,7 +156,14 @@ class EiScreen : public PlatformScreen
...
 
-  std::uint32_t m_sequenceNumber = 0;
+  mutable std::uint32_t m_sequenceNumber = 0;
+
+  // Lazily-started EIS emulation: only grab while relayed input is actually
+  // flowing, and release after a short idle so the compositor can DPMS-sleep
+  // this screen even while the deskflow cursor logically sits on it.
+  mutable bool m_isEmulating = false;
+  mutable EventQueueTimer *m_idleEmulationTimer = nullptr;
+  static constexpr double s_idleEmulationTimeout = 4.0;
```
nbolton, line 167: Why 4 seconds? Just curious.

ElCruncharino: Picked empirically while testing on my hardware.

It was long enough that it doesn't drop the grab between normal micro-pauses in input, short enough that DPMS-off still happens promptly once actually idle.

Open to making it configurable (or tuning the value) if you'd prefer.

sithlord48: I think just adding a comment there for now should be ok what you u think @nbolton  ? 

Edit: another thought this should not happen if we have the preventSleepOption enabled.

sithlord48:
Since the second commit is "fixing" a problem in the first one it would be nice to have it squashed into commit 1 of this PR 

Edit: updated this to have only the one commit.

ElCruncharino:
Pushed an update to this branch. While testing a fix for another bug, I found that after the idle timer releases emulation, `m_isEmulating` could get out of sync with the actual device state:

- On my KWin setup, `ei_device_stop_emulating()` doesn't pause the device the way the libei API implies. It can trigger a full device teardown and recreation. `removeDevice()` never reset `m_isEmulating` in that case, so a device recreated mid-churn was silently never told to start emulating again (`ei_device_pointer_motion_absolute: device is not emulating`).
- The existing `EI_EVENT_DEVICE_RESUMED` handler also called `ei_device_start_emulating()` directly, bypassing that same bookkeeping, which could cause a redundant start.

Both are fixed by keeping `m_isEmulating` consistent across `removeDevice()` and the `RESUMED` handler (routed through `ensureEmulating()` now). Squashed into the original commit since it's a correctness fix to the same mechanism.

src/lib/platform/EiScreen.h
nbolton: Yeah, a future dev investigating a timing bug will wonder why is 4 seconds significant. If it was an eyeball based on someone's hardware _n_ years ago, that would give a clue.

ElCruncharino: Comment added, and added gates for the `preventSleepOption` being enabled, to minimize churn in that case.

sithlord48 (dismissed):
Was able to test works as expected

nbolton (commented):
When the compositor pauses the devices, nothing appears to clear `m_isEmulating` right? Having a bit of trouble grokking this as its late here and I'm quite tired.

ElCruncharino:
> When the compositor pauses the devices, nothing appears to clear `m_isEmulating` right? Having a bit of trouble grokking this as its late here and I'm quite tired.

Good catch! Will take a look in a little bit.

sithlord48 (approved):
Tested it worked well for me Approving!
@nbolton  i will leave landing up to you after you get a chance to test again.

nbolton (approved):
LGTM

Issue #8575, opened 2025-05-07
Windows client keyboard does not work unless macOS server is set to "ABC" input source


clcocd:
### Project

Deskflow

### Deskflow version info

Deskflow version: 1.21.2.0 (daa09546)

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [x] Windows
- [x] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [ ] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [x] I am not using Wayland on Linux

### Signing on macOS

- [x] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### Continuous build

- [ ] I have tried the latest continuous build and the issue persists
- [ ] I am unable to try the latest continuous build

### OS versions/distros

macOS Sequoia (15.4) , Windows 11 Version 23H2

### Deskflow configuration

macOS as the server and Windows as the client.

### What steps will reproduce the problem?

1. Set up Deskflow with macOS as the server and Windows as the client.
2. On macOS, switch the input source to something other than ABC (e.g., Gureum or Korean).
3. Try typing from the Windows client → no input is sent.

### Log output

```shell
[2025-05-07T13:47:00] INFO: starting core server process (desktop mode)
[2025-05-07T13:47:00] INFO: core config file: /Users/hansiyeong/Library/Deskflow/deskflow-server.conf
[2025-05-07T13:47:00] INFO: running command: /Applications/Deskflow.app/Contents/MacOS/deskflow-server -f --debug INFO --name hansiyeong-ui-MacBookAir.local --enable-crypto --address :24800 -c /Users/hansiyeong/Library/Deskflow/deskflow-server.conf --tls-cert /Users/hansiyeong/Library/Deskflow/tls/deskflow.pem
[2025-05-07T13:47:00] WARNING: can't get the active group, use the first group instead
[2025-05-07T13:47:00] WARNING: can't get the active group, use the first group instead
[2025-05-07T13:47:00] NOTE: started server, waiting for clients
2025-05-07 13:47:00.825 deskflow-server[31317:741870] starting cocoa loop
settings file: /Users/hansiyeong/Library/Deskflow/Deskflow.conf
[...]

Issue #7611, opened 2024-10-03
Language switch is not working on Ubuntu 22.04 and macOS 15.0


iamhslee:
### Project

Deskflow

### Deskflow version number

1.17.0

### Build type

Local developer build (built it yourself)

### Operating systems (OS)

- [ ] Windows
- [X] macOS
- [X] Linux
- [ ] BSD-derived
- [ ] Other

### OS versions/distros

- Ubuntu Desktop 22.04 LTS
- macOS Sequoia 15.0 (24A335)

### Deskflow configuration

- Ubuntu as server, macOS as client
- Both computer using English, and Korean input source
- Ubuntu use `ibus` to use English and Korean
- macOS use native input software to use English and Korean
- Both computer has one display each (Mac using built-in display)
- Ubuntu's display is on center, the Mac display is located right next on Ubuntu's display

### What steps will reproduce the problem?

1. Start Deskflow both computer.
2. Set Deskflow as Server on Ubuntu and Start.
3. Set Deskflow as Client on Mac and Start.
4. Move cursor to right edge of Ubuntu's display to switch to Mac.
5. On Mac's display, focus any text field then hit the language switch key (Currently I'm using Right ALT key to switch both English and Korean).
6. The language is not switching.

### Log output

```shell
[2024-10-03T17:04:45] INFO: starting core client process (desktop mode)
[2024-10-03T17:04:45] INFO: running command: /Applications/Deskflow.app/Contents/MacOS/deskflow-client -f --no-tray --debug INFO --name Hyunseos-MacBook --enable-crypto --sync-language 203.252.118.151:24800
[2024-10-03T17:04:45] NOTE: started client
[2024-10-03T17:04:45] NOTE: connecting to '(Masked)': (Masked):24800
[2024-10-03T17:04:45] NOTE: server fingerprint: DC:DF:B6:7C:3F:DE:08:54:21:E6:4B:8F:4B:AE:6D:16:2C:07:A5:9B:58:BB:D4:3C:5A:F8:8E:D7:90:25:A8:74
2024-10-03 17:04:45.518 deskflow-client[46557:7221655] starting cocoa loop
[2024-10-03T17:04:45] WARNING: can't get the active group, use the first group instead
[2024-10-03T17:04:45] WARNING: cursor may not be visible
[2024-10-03T17:04:45] INFO: connected to secure socket
[2024-10-03T17:04:45] INFO: server tls certificate info: /CN=Deskflow
[...]

Issue #7200, opened 2022-07-12
Keyboard layout not changing when server uses non-english layout


Doumor:
What happened?
So, when you select non-english layout on server and try use english layout on client you cant use it, you typing in non-english layout.

Sometimes it works the other way around: when you select english layout on server and try use non-enghish layout on client you cant use it, you typing in english layout.

I tryed only windows as server and manjaro as client and the other way around.

SymlessBot:
[Bot] Added to Synergy's internal issue tracker with ID: `SYNERGY1-1584`

lofcz:
Doesn't even work with one computer set to QWERTY. Blocking issue for me. Tried Czech/QWERTZ + Czech/QWERTZ and setting client to QWERTY.

chris-hunter206:
I've noticed this issue as well. I run Dvorak on the server and the client must be set to US/Qwerty or the output on the client is totally garbled. This definitely was not an issue in earlier versions. I've had the issue on 1.14.3 client for a while. Server 1.14.4 and 1.14.5 have the problem.

Impact is minimal because the client seems to auto-switch to US when synergy connects, but I have to change it back manually whenever using the physical keyboard.

avijoni:
I would kindly ask people at seamless to build 1.14.1 binaries and put them available for download. This issue is persisting in 1.14.6, just tested it today, and I don't have an old binary to install it back. I want to be able to use my computers as I used to. Everything worked fine on 1.14.1. I updated it because i was installing fresh windows and come to this situation where I am unable to use all 3 computers, since it doesn't switch to right keyboard layout on client.

avijoni:
Btw, why would you release 5 sub versions in meantime with all having same bugs/issues. What is the logic behind that releasing spree of non-functional app? What is the added value in that. I really wonder.

avijoni:
[...]

PR #9898, opened 2026-06-26
feat(server): switch away from the input method (IME) when leaving a screen

HongJunhyoung:
## Description

When the server's keyboard is set to an input method (IME) such as Korean, Japanese or Chinese, the active IME composition swallows keystrokes that are meant for a client. The result is that after switching to a client screen, typing does nothing until you manually switch the server's keyboard back to a plain/ASCII layout. This is a long-standing pain point for CJK users.

This PR adds an **opt-in** server option (off by default): when control leaves the primary screen, the server remembers its current keyboard input source and switches to an ASCII-capable layout; when control returns, the previous input source is restored.

Implementation notes:
- The hook lives in the cross-platform `Screen::enter()` / `Screen::leave()` (only when this screen is the primary), gated by the new setting.
- Two new methods are added to `IPlatformScreen` with **no-op default implementations**, so other platforms are unaffected.
- **macOS** implements them using the Text Input Source (TIS) API (`TISCopyCurrentASCIICapableKeyboardInputSource` so the user's real Latin layout — ABC, US, German, etc. — is chosen, not hardcoded English).
- A checkbox is added to the **Advanced** settings page, next to the existing "Run command on enter/exit" options.
- Windows/Linux are intentionally no-op for now; the hook is in place so they can be implemented later (e.g. via the IMM API on Windows).

## AI tools used for this PR

Claude (Anthropic, **Opus 4.8**) was used to explore the codebase, write the implementation, and assist with debugging. A human author reviewed the full diff, built it locally, installed and ran the resulting build, and verified the behavior described below.

## Fixed/related issues

Related: #8575, #7200, #7611

(This is an opt-in, macOS-first mitigation; it does not fully resolve these cross-platform issues, so it is listed as related rather than `fixes:`.)

## How has this been tested?

- Built locally on **macOS (Apple Silicon, arm64), Qt 6.11.1**, Release.
- Ran the server build and verified with the Korean 2-Set input method active:
  - On leaving the primary screen the input source switches to the ASCII-capable layout, so keystrokes reach the client.
  - On returning to the primary screen the previous (Korean) input source is restored.
- With the option **off** (default), behavior is unchanged.
- Existing unit tests build/run; the only failure observed was the pre-existing, environment-dependent flaky `OSXKeyStateTests::fakePollShift` (unrelated to this change — it passes on re-run and touches code this PR does not modify).
- Windows/Linux were not runtime-tested (no-op there); relying on CI for cross-platform build verification.

### DIFF
--- a/src/lib/deskflow/IPlatformScreen.h
+++ b/src/lib/deskflow/IPlatformScreen.h
@@ -135,6 +135,30 @@ public:
   */
   virtual std::string getSecureInputApp() const = 0;
 
+  //! Remember the current input source and switch to an ASCII-capable one
+  /*!
+  Called when leaving a primary screen so that an active input method
+  (e.g. Korean, Japanese or Chinese) on the server does not swallow
+  keystrokes that are meant for a client. The previous input source is
+  remembered so it can be restored by restoreInputSource(). The default
+  implementation does nothing; platforms that support input source
+  switching may override it.
+  */
+  virtual void switchToAsciiInputSource()
+  {
+    // do nothing
+  }
+
+  //! Restore the input source remembered by switchToAsciiInputSource()
+  /*!
+  Called when returning to a primary screen. The default implementation
+  does nothing.
+  */
+  virtual void restoreInputSource()
+  {
+    // do nothing
+  }
+
   //@}
   //! @name accessors
   //@{
--- a/src/lib/platform/OSXScreen.mm
+++ b/src/lib/platform/OSXScreen.mm
@@ -35,6 +35,7 @@
 
 #include <AppKit/NSEvent.h>
 #include <AvailabilityMacros.h>
+#include <Carbon/Carbon.h>
 #include <IOKit/hidsystem/event_status_driver.h>
 #include <dispatch/dispatch.h>
 #include <libproc.h>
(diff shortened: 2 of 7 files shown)


src/lib/platform/OSXScreen.mm
```diff
@@ -806,6 +807,61 @@
   m_isOnScreen = false;
 }
 
+void OSXScreen::switchToAsciiInputSource()
+{
+  // Remember the current keyboard input source so it can be restored when we
+  // return to this screen.
+  m_savedInputSourceId.clear();
+  if (TISInputSourceRef current = TISCopyCurrentKeyboardInputSource()) {
```
amagus, line 815: calling TIS without locking the needed mutex

src/lib/platform/OSXScreen.mm
```diff
@@ -806,6 +807,61 @@
...
+
+  CFStringRef sourceId =
+      CFStringCreateWithCString(kCFAllocatorDefault, m_savedInputSourceId.c_str(), kCFStringEncodingUTF8);
+  if (sourceId != nullptr) {
+    const void *keys[] = {kTISPropertyInputSourceID};
+    const void *values[] = {sourceId};
+    if (CFDictionaryRef filter = CFDictionaryCreate(
+            kCFAllocatorDefault, keys, values, 1, &kCFTypeDictionaryKeyCallBacks, &kCFTypeDictionaryValueCallBacks
+        )) {
+      if (CFArrayRef list = TISCreateInputSourceList(filter, false)) {
```
amagus, line 850: calling TIS without locking TIS mutex

src/lib/deskflow/IPlatformScreen.h
```diff
@@ -135,6 +135,30 @@ class IPlatformScreen : public IScreen, public IPrimaryScreen, public ISecondary
   */
   virtual std::string getSecureInputApp() const = 0;
 
+  //! Remember the current input source and switch to an ASCII-capable one
+  /*!
+  Called when leaving a primary screen so that an active input method
```
amagus, line 140: Can you please simplify these comments?

amagus (changes requested):
This looks to be fixing only macOS. 
If this is the case, please make this clear in the PR description as I was with the impression that this would be fixed for all OS. If you intend to fix it for other OS, please include the needed changes.

sithlord48:
Drafting until updated for all platforms

PR #9902, opened 2026-06-29
Remember previously used server addresses

### LINKED ISSUE #8114: Remember previously used server address (hostname/IP)

Borosnyai:
## Description

This pull request changes the server hostname input from a `QLineEdit` to an editable `QComboBox`.

The combo box stores previously used server addresses in the application settings and reloads them when the application starts. Existing addresses are kept unique, the most recently used address is moved to the top of the list, and the history is limited to 10 entries.

## AI tools used for this PR

* ChatGPT (GPT-5.5)

Used to discuss the implementation approach and review the C++/Qt code changes.

## Fixed/related issues

Fixes #8114

## How has this been tested?

* Verified the code changes manually.
* Ran `git diff --check` to ensure there are no whitespace errors.
* Attempted a local build. The project could not be built completely because the required dependencies (including the required Qt version and additional system packages) are not available in my Debian 12 environment.

### DIFF
--- a/src/lib/gui/MainWindow.cpp
+++ b/src/lib/gui/MainWindow.cpp
@@ -33,10 +33,12 @@
 
 #include <QCheckBox>
 #include <QCloseEvent>
+#include <QComboBox>
 #include <QDesktopServices>
 #include <QFileDialog>
 #include <QLocalServer>
 #include <QLocalSocket>
+#include <QLineEdit>
 #include <QMenu>
 #include <QMenuBar>
 #include <QMessageBox>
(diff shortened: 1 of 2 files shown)


sithlord48 (commented):
Cleaner than the other impl.. 

Ill test this when i get some time

sithlord48 (changes requested):
@Borosnyai  you will need to correct your commit message it should begin with`feat: `

It appears you also have some clang-format issues. please correct them (without adding a new commit) 

I will mark this as a draft when you have updated it and your ready to have us look at it again please mark it ready for review.

sithlord48:
> Attempted a local build. The project could not be built completely because the required dependencies (including the required Qt version and additional system packages) are not available in my Debian 12 environment.

We do not accept prs that are not locally tested.

PR #9916, opened 2026-06-30
feat(macos): preserve the left/right side of modifier keys on the client

rNoz:
## Description

On macOS the client collapsed every right-hand modifier onto its left-hand virtual key and only emitted the left device-dependent event flags, so a right-side modifier sent by the server (Right Control/Shift/Option/Command) arrived as the left-hand key.

Apps that distinguish the two never saw the right key (e.g. a hotkey bound to Right Control, or input methods that use Right Shift).

Maps the right-hand kKey*_R KeyIDs to their macOS right virtual keys, teaches isModifier()/setKeyboardModifiers() about them, tracks which side is held, and emits the matching NX_DEVICER*KEYMASK flag. No behavioural change for left-hand modifiers.

## Disclosure of AI use

Developed with the help of an AI coding assistant (Claude Opus 4.8) using tab expansion as I was coding, and helping me exploring the codebase as I was testing this software for my own purposes, specifically checking rules and mimicking code/contributing styles.

All code reviewed, built, and tested on real hardware by me (author). I run deskflow (server) in a Linux with wayland (KDE, Manjaro), and deskflow (client) in macOS (m1).

## Related Issue
fixes/relates to: #8486 (may be worded as cross-platform; but my fixes are macOS-only)

## How Has This Been Tested?

Built on macOS (Apple Silicon, Qt 6.9, clang) as a client with a Linux server.

Right Control now triggers a macOS app listening for kVK_RightControl (0x3E). Before it was delivered as Left Control. Left-hand modifiers and normal typing unaffected.

### DIFF
--- a/src/lib/platform/OSXKeyState.cpp
+++ b/src/lib/platform/OSXKeyState.cpp
@@ -26,6 +26,10 @@ static const uint32_t s_controlVK = kVK_Control;
 static const uint32_t s_altVK = kVK_Option;
 static const uint32_t s_superVK = kVK_Command;
 static const uint32_t s_capsLockVK = kVK_CapsLock;
+static const uint32_t s_shiftRightVK = kVK_RightShift;
+static const uint32_t s_controlRightVK = kVK_RightControl;
+static const uint32_t s_altRightVK = kVK_RightOption;
+static const uint32_t s_superRightVK = kVK_RightCommand;
 static const uint32_t s_numLockVK = kVK_ANSI_KeypadClear; // 71
 
 static const uint32_t s_brightnessUp = 144;
@@ -95,18 +99,18 @@ static const KeyEntry s_controlKeys[] = {
     // to map to.  also the enter key with numlock on is a modifier but i
     // don't know which.
 
-    // modifier keys.  OS X doesn't seem to support right handed versions
-    // of modifier keys so we map them to the left handed versions.
+    // modifier keys.  map the left and right variants to their respective
+    // macOS virtual keys so the modifier side is preserved on the client.
     {kKeyShift_L, s_shiftVK},
-    {kKeyShift_R, s_shiftVK}, // 60
+    {kKeyShift_R, s_shiftRightVK},
     {kKeyControl_L, s_controlVK},
-    {kKeyControl_R, s_controlVK}, // 62
+    {kKeyControl_R, s_controlRightVK},
     {kKeyAlt_L, s_altVK},
-    {kKeyAlt_R, s_altVK},
+    {kKeyAlt_R, s_altRightVK},
     {kKeySuper_L, s_superVK},
-    {kKeySuper_R, s_superVK}, // 61
+    {kKeySuper_R, s_superRightVK},
     {kKeyMeta_L, s_superVK},
-    {kKeyMeta_R, s_superVK}, // 61
+    {kKeyMeta_R, s_superRightVK},
 
     // toggle modifiers
     {kKeyNumLock, s_numLockVK},
@@ -168,7 +172,8 @@ io_connect_t getEventDriver()
 
 bool isModifier(uint8_t virtualKey)
 {
-  static std::set<uint8_t> modifiers{s_shiftVK, s_superVK, s_altVK, s_controlVK, s_capsLockVK};
+  static std::set<uint8_t> modifiers{s_shiftVK,      s_controlVK,      s_altVK,      s_superVK,     s_capsLockVK,
+                                     s_shiftRightVK, s_controlRightVK, s_altRightVK, s_superRightVK};
 
   return (modifiers.find(virtualKey) != modifiers.end());
 }
@@ -228,6 +233,10 @@ void OSXKeyState::init()
   m_altPressed = false;
   m_superPressed = false;
   m_capsPressed = false;
+  m_shiftRightPressed = false;
+  m_controlRightPressed = false;
+  m_altRightPressed = false;
+  m_superRightPressed = false;
 
   // build virtual key map
   for (size_t i = 0; i < sizeof(s_controlKeys) / sizeof(s_controlKeys[0]); ++i) {
@@ -549,19 +558,19 @@ CGEventFlags OSXKeyState::getDeviceDependedFlags() const
   CGEventFlags modifiers = 0;
 
   if (m_shiftPressed) {
-    modifiers |= NX_DEVICELSHIFTKEYMASK;
+    modifiers |= m_shiftRightPressed ? NX_DEVICERSHIFTKEYMASK : NX_DEVICELSHIFTKEYMASK;
   }
 
   if (m_controlPressed) {
-    modifiers |= NX_DEVICELCTLKEYMASK;
+    modifiers |= m_controlRightPressed ? NX_DEVICERCTLKEYMASK : NX_DEVICELCTLKEYMASK;
   }
 
   if (m_altPressed) {
-    modifiers |= NX_DEVICELALTKEYMASK;
+    modifiers |= m_altRightPressed ? NX_DEVICERALTKEYMASK : NX_DEVICELALTKEYMASK;
   }
 
   if (m_superPressed) {
-    modifiers |= NX_DEVICELCMDKEYMASK;
+    modifiers |= m_superRightPressed ? NX_DEVICERCMDKEYMASK : NX_DEVICELCMDKEYMASK;
   }
 
   return modifiers;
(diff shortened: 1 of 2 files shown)


sithlord48:
If this is ready to be reviewed make sure to mark it ready for review.  @amagus  can you test this when you get sometime @rNoz This should if possible be expanded to work on all platforms also please fix the commit message it should be `feat(macos): ....`

rNoz:
> If this is ready to be reviewed make sure to mark it ready for review. @amagus can you test this when you get sometime @rNoz This should if possible be expanded to work on all platforms also please fix the commit message it should be `feat(macos): ....`

Sure, done. Totally fine to test/update as you wish. I don't have more devices to test, but let me know if you need some other test from my end.
I am leaving the other PR in draft until this is merged, but already applying the same change to the git message.

src/lib/platform/OSXKeyState.cpp
```diff
@@ -168,7 +168,8 @@ io_connect_t getEventDriver()
 
 bool isModifier(uint8_t virtualKey)
 {
-  static std::set<uint8_t> modifiers{s_shiftVK, s_superVK, s_altVK, s_controlVK, s_capsLockVK};
+  static std::set<uint8_t> modifiers{s_shiftVK,      s_superVK,        s_altVK,         s_controlVK,     s_capsLockVK,
```
amagus: mixing different types of constants is kind of weird...

amagus:
Applications should not differentiate between left or right modifiers in macOS. They are supposed to be interpreted exactly the same.
The official keystroke APIs not even provide this information natively (detecting which side is pressed requires using private APIs).
Can you please exemplify applications that differentiate between those keys so that I can test and understand how these applications detect which key is being pressed?

Anyway, this change looks to me like to be related on using macOS as a client alone, right? I don't see any private API usage to detect these modifiers as server.

src/lib/platform/OSXKeyState.cpp
```diff
@@ -95,18 +95,18 @@ static const KeyEntry s_controlKeys[] = {
     // to map to.  also the enter key with numlock on is a modifier but i
     // don't know which.
 
-    // modifier keys.  OS X doesn't seem to support right handed versions
```
amagus, line 98: this comment remains correct. I am still puzzled on what you are trying to fix here. Changing these will perhaps break usage of macOS as server. Have you tried using macOS as server with your changes?

src/lib/platform/OSXKeyState.h
```diff
@@ -162,4 +162,11 @@ class OSXKeyState : public KeyState
   bool m_altPressed;
   bool m_superPressed;
   bool m_capsPressed;
+  // track whether the right-hand variant of a modifier is held so the
+  // device-dependent event flags report the correct side (e.g. Right Shift
+  // for CJK input methods).
+  bool m_shiftRightPressed = false;
```
amagus: why initializing these here if all the others are not initialized here?

rNoz:
Before answering the other questions, I am focusing first on this because it is a decision to consider the PR or not.

> Can you please exemplify applications that differentiate between those keys so that I can test and understand how these applications detect which key is being pressed?

There are apps (like TypeWhisper) where I can define RCtrl to dictate (toggle, record/pause), and others like Karabiner that allows defining L/R modifiers. In Linux I have the same type of modifiers (LCtrl/RCtrl) for some apps.

> Anyway, this change looks to me like to be related on using macOS as a client alone, right? I don't see any private API usage to detect these modifiers as server.

Right, I tested in that way: server (linux) sends the appropriate information, but client (macos) is not using it. 

It is my current use case, where I am normalizing how I type in every machine, no matter if in linux/macos, and having that distinction between LCtrl/RCtrl is helpful, but if you don't find it useful even having specific apps that support that, then feel free to close this PR (or mutate this branch as you wish).

src/lib/platform/OSXKeyState.cpp
rNoz: No, but I can include as part of the tests if this functionality makes sense for all (distinguishing L/R)

src/lib/platform/OSXKeyState.h
rNoz: Right, just being explicit. Fine as they are 0 by default. I'll update based on the main questions.

amagus:
> There are apps (like TypeWhisper) where I can define RCtrl to dictate (toggle, record/pause), and others like Karabiner that allows defining L/R modifiers. In Linux I have the same type of modifiers (LCtrl/RCtrl) for some apps.

None of the cited apps are available on the mac app store so I am tending to say that they are most likely using private APIs.

> It is my current use case, where I am normalizing how I type in every machine, no matter if in linux/macos, and having that distinction between LCtrl/RCtrl is helpful, but if you don't find it useful even having specific apps that support that, then feel free to close this PR (or mutate this branch as you wish).

I understand that this is very helpful for power users like us.

I am just not sure that supporting left/right modifiers as client only is a good option... 
If we can support it as server AND as client then the support for this feature should be delivered together...

We may want to deliver deskflow on the Mac App Store and using private APIs there is not an option. 
Perhaps we should raise this discussion and decide about such cases...

rNoz:
> I am just not sure that supporting left/right modifiers as client only is a good option...
> If we can support it as server AND as client then the support for this feature should be delivered together...

I can test/work on that if in the end it will be useful for the project, although I only have linux and macos. And Linux (server) has this functionality already supported (I need no changes on that side). So afaik, I will need to work on the macOS (Server) side to verify it (although it will be with Linux as client). Let me know if you are interested, otherwise I will wait for the discussion (see below).

> We may want to deliver deskflow on the Mac App Store and using private APIs there is not an option.
> Perhaps we should raise this discussion and decide about such cases...

Ok, "til" of Mac App Store public APIs vs private APIs, this is important and thanks for bringing this up. Totally open to you what to do here, since this is your project. I will be curious about that conversation and trade-offs. Because I understand that maintaining this project with the functionality for the Mac App Store or for "development mode"/private APIs isn't viable as it is a corner case.

If I can help with anything, please, let me know. 

Thanks for your time!

sithlord48:
I don't want to use any private APIs as they tend to  change without warning and things are suddenly broken

rNoz:
@sithlord48  @amagus  Quick clarification on the private API point, since I think there was a mix-up: this diff does not use any private macOS API. The only new symbols are kVK_Right* (from <Carbon/Carbon.h>) and the NX_DEVICER*KEYMASK flags (from <IOKit/hidsystem/IOHIDLib.h>), both public SDK headers. The private-API concern was about how other apps (Karabine, TypeWhisper, etc) detect left/right, not about this code or this project. Deskflow already posts events via CGEventPost/IOHIDPostEvent on master, and this PR does not add to that surface.

That said, I fully accept the design question is separate from the API one. **If the project's position is that a macOS client should not distinguish left/right modifiers, that is your call and I am happy to close this** (the server, deskflow in Linux, already sends those separately). In that case I will keep my fork for such behaviors.

On testing: I can only test macOS client with a Linux server, or viceversa, since those are the only machines I have. Supporting and verifying left/right as a macOS server is doable but more effort for me (as it was implementing this, I like to test thoroughly), so before I invest in that I would like to know whether the feature is wanted at all. If it is, I will add tests and work the server side; if not, no problem, I will keep this just for me.

Note this does not block the AltGr fix: #9917 is now a standalone single commit that no longer depends on this PR.

rNoz:
Meanwhile, addressed Luiz feedback: consistent s_*RightVK constants instead of mixing raw kVK_Right*, and the right-side state flags are initialised in init() like the others. On macOS-as-server: this change only touches the client output path (s_controlKeys, setKeyboardModifiers, getDeviceDependedFlags); the server key map (getKeyMap) is untouched, so server behaviour is unchanged. Just waiting on above decision to decide if testing that or closing.

amagus:
@rNoz if there’s no side effect on server mode, I think we can land these changes. My only additional comment would be seeing if we could test this to avoid breaking this feature in the future (in all platforms).

rNoz:
> @rNoz if there’s no side effect on server mode, I think we can land these changes. My only additional comment would be seeing if we could test this to avoid breaking this feature in the future (in all platforms).

Yes. That seems like willing to accept. So in that case, I will allocate time these days to test the macOS server side -> linux client for this PR ("the reverse direction"). If by the time I am testing this we have already the AltGr fix merged, I will rebase and test all again in both directions.

Per the additional comment: I suppose when you mean "test this" you mean programmatically, to detect regressions, however, when dealing with key events like that, usually it is not straightforward as it involves many OS (-dependent) events, but I will give it a try.
Also, I lack Windows, so I cannot help there. Any guidance so I can try to do so? Maybe some previous feature/test that I can use for inspiration?

tuxtard:
I tried it the other way around, using macOS 26.5 as the host and Linux KDE as the client, but the client only detected the left modifier keys.

amagus:
@tuxtard This is expected. As I mentioned previously, detecting which side of the keyboard is in use on macOS requires private APIs.

tuxtard:
> @tuxtard This is expected. As I mentioned previously, detecting which side of the keyboard is in use on macOS requires private APIs.

Hammerspoon can correctly detect and distinguish between the left and right modifier keys, which likely means other third-party apps should be able to do the same as well.

```
2026-07-09 12:44:54: keyCode=59 name=ctrl
2026-07-09 12:44:54: keyCode=59 name=ctrl
2026-07-09 12:44:58: keyCode=58 name=alt
2026-07-09 12:44:58: keyCode=58 name=alt
2026-07-09 12:45:00: keyCode=61 name=rightalt
2026-07-09 12:45:00: keyCode=61 name=rightalt
2026-07-09 12:45:01: keyCode=62 name=rightctrl
2026-07-09 12:45:01: keyCode=62 name=rightctrl
```

amagus:
@tuxtard We can detect it if we want, that's not the problem... It just means that we would be using private APIs (unable to submit to Mac App Store if we want + not stable, i.e. may cause breakages between versions)

amagus:
Tested on my side, works fine, but I wanted to be sure to mention on the PR/commit title that this fixes it as client only. People may look on the changelogs and assume it changes the behavior for server mode as well.
@rNoz can you please update the title of the PR and the first line of your commit message?

rNoz:
> Tested on my side, works fine, but I wanted to be sure to mention on the PR/commit title that this fixes it as client only. People may look on the changelogs and assume it changes the behavior for server mode as well. @rNoz can you please update the title of the PR and the first line of your commit message?

Done. Title and commit updated to clarify this is client-only (let me know if you want it more obvious), PR desc was already with it. Server-side behavior is unchanged: this feature only touches the client output path (`s_controlKeys`, `setKeyboardModifiers`, `getDeviceDependedFlags`).

amagus (approved):
Tested locally and it works. Code looks good to me.

rNoz:
Rebased as master had changes.

sithlord48 (approved):
@nbolton  i will leave this to you to land

nbolton (approved):
Tested with https://keytest.io/ since the macOS native keyboard viewer doesn't seem to differentiate between L/R shift.

PR #9917, opened 2026-06-30
feat(macos): type AltGr (Level3) characters via the right Option key

### LINKED ISSUE #4411: AltGr not sent to client from server
I'm pretty sure that the AltGr support has been lost on every OS (actually I can't remember if we ever supported it). This issue is to consolidate all issues into one.
#3327, #3929, #4339, #3476, #582, #1832, #1849, #2163, #1850

rNoz:
## Description

macOS has no AltGr modifier: the characters on the Option layer are produced with the Option key. Because of that, an AltGr modifier coming from a remote server never matched the macOS key map (those entries require Alt), so AltGr characters such as `~ @ # [ ] { }` on European layouts were dropped on the client with a "no mapping" message.

This PR maps the AltGr KeyID and the AltGr modifier to the right Option key, and registers an AltGr-required variant of every Option (Alt) key-map entry, so an incoming AltGr both matches those characters and is synthesized through the right Option key. Left Alt and AltGr stay distinct, so left-Option shortcuts and word navigation keep working.

### Relationship to #9916

This PR is now standalone. It previously sat on top of #9916, but it has been reduced to a single commit that no longer depends on it. It only carries the minimum Alt_R to right-Option wiring that AltGr synthesis needs (`s_controlKeys`, `isModifier()`, `setKeyboardModifiers()`, `getDeviceDependedFlags()`, and one state flag). The full left/right side preservation for Shift/Control/Super/Meta stays in #9916, which is still open pending the private API discussion, and is intentionally not part of this change.

The change is macOS client only. It does not touch server behaviour and does not add any private API: the symbols used (`kVK_RightOption`, `NX_DEVICERALTKEYMASK`) are public and come from `<Carbon/Carbon.h>` and `<IOKit/hidsystem/IOHIDLib.h>`.

## Related issue

fixes: #4411

## How has this been tested?

Built on macOS (Apple Silicon, Qt 6.9, clang) as client, with a Linux (Wayland, KDE) server. Tested with Spanish layouts: AltGr+key now types `~ @ # [ ] { }` in GUI apps and in Alacritty (`option_as_alt = "OnlyLeft"`). Unit tests `OSXKeyStateTests` (6/6) and `KeyMapTests` (13/13) pass, and the code is formatted with clang-format 20.1.0 to match CI.

### DIFF
--- a/src/lib/platform/OSXKeyState.cpp
+++ b/src/lib/platform/OSXKeyState.cpp
@@ -97,12 +97,16 @@ static const KeyEntry s_controlKeys[] = {
 
     // modifier keys.  OS X doesn't seem to support right handed versions
     // of modifier keys so we map them to the left handed versions.
+    // For the client output path, AltGr is mapped to the right Option key
+    // so it can be synthesized without colliding with left-Option shortcuts
+    // (e.g. Meta word-nav).
     {kKeyShift_L, s_shiftVK},
     {kKeyShift_R, s_shiftVK}, // 60
     {kKeyControl_L, s_controlVK},
     {kKeyControl_R, s_controlVK}, // 62
     {kKeyAlt_L, s_altVK},
-    {kKeyAlt_R, s_altVK},
+    {kKeyAlt_R, kVK_RightOption},
+    {kKeyAltGr, kVK_RightOption}, // macOS has no AltGr; treat it as right Option
     {kKeySuper_L, s_superVK},
     {kKeySuper_R, s_superVK}, // 61
     {kKeyMeta_L, s_superVK},
@@ -168,7 +172,9 @@ io_connect_t getEventDriver()
 
 bool isModifier(uint8_t virtualKey)
 {
-  static std::set<uint8_t> modifiers{s_shiftVK, s_superVK, s_altVK, s_controlVK, s_capsLockVK};
+  // kVK_RightOption is included because a remote AltGr is synthesized via
+  // the right Option key (see s_controlKeys).
+  static std::set<uint8_t> modifiers{s_shiftVK, s_superVK, s_altVK, s_controlVK, s_capsLockVK, kVK_RightOption};
 
   return (modifiers.find(virtualKey) != modifiers.end());
 }
@@ -199,6 +205,7 @@ void OSXKeyState::init()
   m_shiftPressed = false;
   m_controlPressed = false;
   m_altPressed = false;
+  m_altRightPressed = false;
   m_superPressed = false;
   m_capsPressed = false;
 
@@ -531,7 +538,7 @@ CGEventFlags OSXKeyState::getDeviceDependedFlags() const
   }
 
   if (m_altPressed) {
-    modifiers |= NX_DEVICELALTKEYMASK;
+    modifiers |= m_altRightPressed ? NX_DEVICERALTKEYMASK : NX_DEVICELALTKEYMASK;
   }
 
   if (m_superPressed) {
@@ -565,6 +572,13 @@ void OSXKeyState::setKeyboardModifiers(CGKeyCode virtualKey, bool keyDown)
     break;
   case s_altVK:
     m_altPressed = keyDown;
+    if (keyDown)
+      m_altRightPressed = false; // left Option takes precedence for the Alt flag
+    break;
+  case kVK_RightOption:
+    // right Option carries the AltGr state
+    m_altPressed = keyDown;
+    m_altRightPressed = keyDown;
     break;
   case s_superVK:
     m_superPressed = keyDown;
@@ -790,6 +804,16 @@ bool OSXKeyState::getKeyMap(deskflow::KeyMap &keyMap, int32_t group, const IOSXK
       for (std::set<uint32_t>::iterator k = required.begin(); k != required.end(); ++k) {
         item.m_required = mapModifiersFromOSX(*k << 16);
         keyMap.addKeyEntry(item);
+
+        // register an AltGr variant of each Alt entry so a remote AltGr
+        // matches the Option-layer characters and is synthesized via the
+        // right Option key.
+        if ((item.m_required & KeyModifierAlt) != 0) {
+          auto altGrItem = item;
+          altGrItem.m_required = (item.m_required & ~KeyModifierAlt) | KeyModifierAltGr;
+          altGrItem.m_sensitive = item.m_sensitive | KeyModifierAltGr;
+          keyMap.addKeyEntry(altGrItem);
+        }
       }
     }
   }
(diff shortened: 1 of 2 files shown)


sithlord48 (changes requested):
Can you fix the commit messages to have an expected prefix.

please note this pr requires the other one landing first 
`Requires: #...` should be added to the description . you can leave this as draft until the other lands.

rNoz:
> Can you fix the commit messages to have an expected prefix.
> 
> please note this pr requires the other one landing first `Requires: #...` should be added to the description . you can leave this as draft until the other lands.

Done. I will set ready for review when the other is merged (and if by that time there are some conflicts, I will solve them). As in the other, feel free to update/test this PR as much as you want.

rNoz:
If the other PR (#9916) is finally not integrated in this repo (discussion around private APIs), but you are interested in this AltGr functionality, I can try to isolate this fix to ship it independently. Please, let me know if waiting for the other decision, or start working on this.

sithlord48:
go for it. we can have this reviewed separately

src/lib/platform/OSXKeyState.cpp
```diff
@@ -565,6 +572,15 @@ void OSXKeyState::setKeyboardModifiers(CGKeyCode virtualKey, bool keyDown)
     break;
   case s_altVK:
     m_altPressed = keyDown;
+    if (keyDown)
```
amagus, line 575: I am not sure I understand the logic behind this if. Can you please provide some more details?

src/lib/platform/OSXKeyState.cpp
```diff
@@ -790,6 +806,18 @@ bool OSXKeyState::getKeyMap(deskflow::KeyMap &keyMap, int32_t group, const IOSXK
       for (std::set<uint32_t>::iterator k = required.begin(); k != required.end(); ++k) {
         item.m_required = mapModifiersFromOSX(*k << 16);
         keyMap.addKeyEntry(item);
+
+        // macOS has no AltGr modifier; characters on the Option (Alt) layer are
```
amagus: Too much detail. Please simplify.

src/lib/platform/OSXKeyState.h
```diff
@@ -162,4 +162,8 @@ class OSXKeyState : public KeyState
   bool m_altPressed;
   bool m_superPressed;
   bool m_capsPressed;
+  // set while the right Option key is held so the device-dependent flag
```
amagus: remove comments from header file, the variable name should suffice to indicate what it represents.

src/lib/platform/OSXKeyState.cpp
```diff
@@ -96,13 +96,17 @@ static const KeyEntry s_controlKeys[] = {
     // don't know which.
 
     // modifier keys.  OS X doesn't seem to support right handed versions
-    // of modifier keys so we map them to the left handed versions.
+    // of modifier keys so we map them to the left handed versions, with
```
amagus: same issue as in other PR. There's no exception. What happens here is that you can map TO any key you want... You can't map FROM the right handed versions.

amagus (changes requested):
Please remove unnecessary comments.

src/lib/platform/OSXKeyState.cpp
rNoz: If the user presses right Option, both flags become true. Later, if the user presses left Option while still holding right Option (or after a quick release/press strokes), the altVK branch runs. Afaik, we want the synthesized Alt modifier to be reported as left, not right. So we clear m_altRightPressed.

Without this guard:
- left Option press would set m_altPressed = true
- but m_altRightPressed could still be true from an earlier right Option press
-  getDeviceDependedFlags would emit NX_DEVICERALTKEYMASK even though the active key is left Option

But if you prefer, I drop the guard

src/lib/platform/OSXKeyState.cpp
rNoz: Reworded the comment. The point is that this table maps incoming KeyIDs to macOS virtual keys (the client output path), so we can map to any key we want. Right, mapping is not an exception, it is AltGr to kVK_RightOption so a remote AltGr can be synthesized on macOS through the right Option key. 

And as I understand, the limitation you mentioned about detecting the right-hand side applies when macOS is the server reading a physical keyboard, not when macOS is the client synthesizing events.

src/lib/platform/OSXKeyState.cpp
rNoz: Sure!

src/lib/platform/OSXKeyState.h
rNoz: Done as well

rNoz (commented):
Rebased and suggestions/comments applied here as well.

rNoz:
Rebased as master had changes.

sithlord48:
@amagus  can you check this when you get some time and let me know

amagus:
I am particularly not sure if only the right alt should be mapped to option key variants in mac os... When using a keyboard directly on a mac both option keys trigger the variant... 
If it would depend on me, I would say that we need to take a look on why any of the alt keys are not directly mapping to the option keys on mac.
I will test the change to understand it better, but I think it should apply any alt key from the server as the option key when using mac as a client.

rNoz:
> I am particularly not sure if only the right alt should be mapped to option key variants in mac os... When using a keyboard directly on a mac both option keys trigger the variant... If it would depend on me, I would say that we need to take a look on why any of the alt keys are not directly mapping to the option keys on mac. I will test the change to understand it better, but I think it should apply any alt key from the server as the option key when using mac as a client.

Both Option keys do produce the variants, and they still do here: `kKeyAlt_L` → left Option, `kKeyAlt_R` → right Option. Any Alt arriving from the server maps to Option exactly as it did before this PR; nothing is restricted to the right side, afaik, right?

I would say tghe gap is a different key. On X11/Wayland the right-hand ISO key does not send Alt at all; it sends `ISO_Level3_Shift`, which reaches the client as `kKeyAltGr` / `KeyModifierAltGr`, with no Alt bit set. On the other side the macOS client had nothing for it: no row in `s_controlKeys`, no `KeyModifierAltGr` branch in `handleModifierKeys()`, and `getKeyMap()` registered only entries requiring `KeyModifierAlt`. So the lookup failed with "no mapping" and `~ @ # [ ] { }` were dropped (#4411). Mapping more Alt keys to Option cannot fix that, because the incoming event carries AltGr, not Alt.

Which Option we synthesize for AltGr is then a "free choice", exactly as you say. I picked right Option so a held remote AltGr does not disturb left-Option behaviour (word navigation, Option+click), and so the device flag matches a real right Option. If you would rather it be left Option, it is a two-line change (`{kKeyAltGr, s_altVK}` plus the `handleModifierKeys()` line) with identical glyph output... happy to switch.

amagus:
> Both Option keys do produce the variants, and they still do here: `kKeyAlt_L` → left Option, `kKeyAlt_R` → right Option. Any Alt arriving from the server maps to Option exactly as it did before this PR; nothing is restricted to the right side, afaik, right?

Have you tried using option key variants (with both alt keys) with your PR with mac as client?

I unfortunately didn't have time to do that and I think that again we are going against the public apple API.

To the best of my knowledge, only the altgr flag trigger the option variants on macOS. It's not only about mapping keys.
If you don't have the correct flags, it won't trigger accents (or other option variants) on mac.

PR #9919, opened 2026-06-30
fix: don't run dead-key ToUnicode translation on the server's own screen

guillepinho:
## Description

This PR fixes **local dead-key accent composition being broken on the server's own screen**, Issue [#9254](https://github.com/deskflow/deskflow/issues/9254).

On Windows, the low-level keyboard hook (`MSWindowsHook.cpp`) runs `ToUnicode()` (together with `setDeadKey()`) to translate key events into characters. `ToUnicode()` has the side effect of consuming the pending dead-key state in the keyboard layout. When the cursor is on the server's *own* screen, deskflow is not relaying keystrokes to any client — it is only watching for the cursor to reach a jump zone. Running the dead-key translation there seems unnecessary, adding overhead, and it destroyed the dead-key state, so typing an accent (e.g. `´` + `a` = `á`) into local apps no longer composed correctly.

- When deskflow is on the server's own screen (`g_mode != kHOOK_RELAY_EVENTS`), stop running the dead-key / `ToUnicode()` composition path.
- Post the raw key event with no composed character and let the local OS handle dead-key composition normally.
- Keep the existing relay behavior (`kHOOK_RELAY_EVENTS`) untouched, so character composition for remote clients is unchanged.
- Early, character-less return on the server's own screen (`src/lib/platform/MSWindowsHook.cpp`)

Since I had the same problem as the original user on the related issue, I tried this approach that would be simple enough to fix without the need of any complex decision at this moment, since the existing relay behavior remains untouched.

## AI tools used for this PR

Claude Opus 4.8, used to identify where the workaround should be placed and used to generate this pull request description.

## Fixed/related issues

- related: #9254 , fixes for the server not clients
## How has this been tested?

This was tested with the generated .msi on the Continuous Integration for the Windows Server (where the issue seems to happen).

The client (Linux) didn't have any issues on the test, everything worked as it should:

1. Ran a deskflow **server** on Windows with a keyboard layout that uses dead keys for accents (ABNT).
2. With the cursor on the **server's own** screen, opened a local text editor and typed a dead-key accent sequence, e.g. `´` then `a`.
   - **Before:** the accent was dropped / not composed (`á` did not appear) if the text was typed too fast.
   - **After:** `á` is composed correctly.
3. Switched the cursor to a **client** screen and confirmed accented characters are still relayed and composed as before (no regression in the relay path).

### DIFF
--- a/src/lib/platform/MSWindowsHook.cpp
+++ b/src/lib/platform/MSWindowsHook.cpp
@@ -292,6 +292,12 @@ static bool keyboardHookHandler(WPARAM wParam, LPARAM lParam)
     if (menu && (sc >= 0x47u && sc <= 0x52u && sc != 0x4au && sc != 0x4eu)) {
       return false;
     }
+
+    // We're on the server's own screen: we're only watching for the
+    // cursor to reach a jump zone, we are NOT relaying keystrokes to a
+    // client, so we have no need for the composed character here.
+    PostThreadMessage(g_threadID, DESKFLOW_MSG_KEY, makeKeyMsg((UINT)wParam, 0, false), lParam);
+    return false;
   }
 
   // map the key event to a character.  we have to put the dead
--- a/src/lib/platform/MSWindowsHook.cpp
+++ b/src/lib/platform/MSWindowsHook.cpp
@@ -309,6 +309,12 @@ static bool keyboardHookHandler(WPARAM wParam, LPARAM lParam)
     if (menu && (sc >= 0x47u && sc <= 0x52u && sc != 0x4au && sc != 0x4eu)) {
       return false;
     }
+
+    // We're on the server's own screen: we're only watching for the
+    // cursor to reach a jump zone, we are NOT relaying keystrokes to a
+    // client, so we have no need for the composed character here.
+    PostThreadMessage(g_threadID, DESKFLOW_MSG_KEY, makeKeyMsg((UINT)wParam, 0, false), lParam);
+    return false;
   }
 
   // map the key event to a character.  we have to put the dead


sithlord48:
Can you please fill out  the description, I also don't think you need a separate commit to remove comments you can just do that in the first commit.

guillepinho:
> Can you please fill out the description, I also don't think you need a separate commit to remove comments you can just do that in the first commit.

I was typing the description, sorry for the delay.

sithlord48:
Gonna let the actions run this will need to be squashed together before landing.

sithlord48:
@guillepinho  can you combine your two commits into one and rebase upon current master so we can look into landing this

guillepinho:
> @guillepinho can you combine your two commits into one and rebase upon current master so we can look into landing this

sure, will do.

guillepinho:
@sithlord48 Done! squashed into a single commit and rebased onto current master

guillepinho:
Btw, wanted to provide a quick note on scope: this fixes the dead-key/accent composition when typing locally on the server machine, where I think the low-level hook was an overhead.

The problem will still occur in client machines, so maybe we can classify it as a partial fix for #9254

sithlord48:
This has to be rebased upon master we do not allow for merge commits in our history please update this

guillepinho:
> This has to be rebased upon master we do not allow for merge commits in our history please update this

Done, I'll pay attention to this next time. Sorry!

src/lib/platform/MSWindowsHook.cpp
```diff
@@ -292,6 +292,12 @@ static bool keyboardHookHandler(WPARAM wParam, LPARAM lParam)
     if (menu && (sc >= 0x47u && sc <= 0x52u && sc != 0x4au && sc != 0x4eu)) {
       return false;
     }
+
+    // We're on the server's own screen: we're only watching for the
+    // cursor to reach a jump zone, we are NOT relaying keystrokes to a
+    // client, so we have no need for the composed character here.
+    PostThreadMessage(g_threadID, DESKFLOW_MSG_KEY, makeKeyMsg((UINT)wParam, 0, false), lParam);
```
nbolton, line 299: With `keyboardBroadcast` turned on, do letters typed on the server's screen still show up on the clients? From the code it looks like this may be impacted?

guillepinho: yes, nice catch, it is. keyboardBroadcast doesn't change the hook mode, so keys typed on the server screen still reach Server::onKeyDown() and get relayed when broadcasting is on.

The fix would be to keep the translation on the server screen but pass wFlags and drop the setDeadKey() put-back there. The character survives for broadcast and the layout's pending dead key is never consumed.

Want me to pursue that path?

gitfmc:
Confirming this from another angle, with one detail I haven't seen reported anywhere and which I think corroborates the diagnosis in this PR.

**Setup:** Windows 10 Pro server, deskflow 1.26.0 (portable), physical **PT-PT** keyboard layout with dead keys (´ ` ~ ^). Linux Mint / macOS clients.

**Symptom:** typing accents on the **server's own screen** produces wrong or dropped characters — *but only in Chromium-based applications* (Chrome, Edge, Brave). **Word and VS Code on the same machine, same layout, same moment are completely unaffected.** The clients are fine too; accented typing composes correctly there.

It happens whenever the **server is running**, regardless of whether a client is connected or where the cursor is — which matches `keyboardHookHandler()` doing the `setDeadKey()` + `ToUnicode()` work unconditionally, as this PR describes.

**Why I think the Chromium-only split is evidence rather than noise:** Chromium's own Windows key handling deliberately avoids per-keystroke `ToUnicode` for exactly this reason. From `ui/events/keycodes/platform_key_map_win.cc`, it pre-builds a layout map with `ToUnicodeEx` instead, because calling it per keystroke *"will consume the dead keys in keyboard buffer"*.

So on a machine running the deskflow server there are **two independent consumers poking the same shared kernel-mode dead-key buffer**, neither aware of the other. Microsoft's own [`ToUnicodeEx` remarks](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-tounicodeex) warn about precisely this:

> As ToUnicodeEx translates the virtual-key code, it also changes the state of the kernel-mode keyboard buffer... It might also cause undesired side-effects if used in conjunction with TranslateMessage (which also changes the state of the kernel-mode keyboard buffer).

Applications using the TSF/Cicero path (Word) never touch that buffer the same way, which would explain why they're immune.

**FWIW this is long-standing** — I've been working around it across Synergy, Barrier, Input-Leap and now deskflow for years, by never letting the server run unless I actively need it. This PR is the first explanation I've seen that accounts for all of the observed behaviour, including the local-only and Chromium-only aspects.

I've backported this diff onto the `v1.26.0` tag (so the fix is the only delta from the release) and built it via CI. I'll report back here with the result — happy to test any variation if that's useful.

Related: #9254.

gitfmc:
**Confirmed working.** I backported this diff onto the `v1.26.0` tag (the fix is the only delta from the release), built it via CI, and have been running it as my server.

**Before:** typing `´` then `a` in Chrome on the server machine gave a dropped or doubled accent — reliably, every time.
**After:** `á`, correctly, in Chrome and in a web-based notepad. No dead key inserted, no doubling.

Word and VS Code were unaffected before and remain unaffected, as expected.

**Relay path still works** — the client (macOS) connects and accented typing over the link is unchanged, which is the regression I was most concerned about given the patch skips translation on the server's own screen. One gotcha for anyone else testing a fresh build rather than an upgrade: the new build generates its own TLS certificate, so existing clients refuse it on fingerprint mismatch until you either re-accept the new fingerprint or carry `settings/tls/` across from the old install. That is a packaging artefact of side-by-side testing, not anything to do with this change.

For context on how long-lived this is: I have been working around this across Synergy, Barrier, Input-Leap and deskflow for more than a decade, by never leaving the server running unless I actively needed it. This is the first thing that has actually fixed it.

Thank you @guillepinho. Would be great to see this land.

PR #9926, opened 2026-07-02
i18n(ja): update Japanese(ja) translation

ykasap:
## Description

Check and update the unfinished ja translation, including some mistranslations.

## AI tools used for this PR

None used.

## How has this been tested?

Published the translation by Qt linguist and tested locally with deskflow 1.26.0.296.
I noticed that there is a "LogLevel" context, but these are not translated in the settings UI.


sithlord48 (approved):

sithlord48:
@ykasap  nice catch fixed that in https://github.com/deskflow/deskflow/pull/9928

PR #9931, opened 2026-07-02
Const log levels and network protocol method returns

sithlord48:
## Description

Make the returns from the logLevel and Network protcol objects const.
## AI tools used for this PR

None
## Fixed/related issues

Sonar is upset we can change the log level string
## How has this been tested?

Local smoke test

### DIFF
--- a/src/lib/common/LogLevel.h
+++ b/src/lib/common/LogLevel.h
@@ -30,19 +30,19 @@ public:
   };
   Q_ENUM(Level)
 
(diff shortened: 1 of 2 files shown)


amagus (approved):
If CI passes, these changes look safe to me.

nbolton:
No idea why, but SC is not happy with any of that.

image

sithlord48:
you can close this if you want . this was done in response to sonar saying log level can have injected '\n`

PR #9939, opened 2026-07-04
fix: preserve client key ids and lock modifiers

MEIZU16:
## Description

Fix client-side key translation so normal key IDs and lock modifier bits are not clamped to the modifier ID range.

The current translation path clamps non-modifier key IDs to `0..6`, which turns normal keys such as `a` or `p` into invalid key IDs. It also clamps the translated modifier mask to `0..6`, which can turn high lock bits such as `KeyModifierNumLock` (`0x2000`) into `Control|Alt` (`0x0006`). On a Wayland/libei client controlled by a Windows server, this caused plain keys to either fail to inject or trigger shortcuts such as `Ctrl+Alt+p` instead of typing `p`.

This keeps clamping scoped to the modifier remapping table index only, while preserving normal key IDs and non-remapped modifier mask bits. It also adds `ServerProxyTests` coverage for regular key IDs, modifier remapping, and preserving `NumLock`.

## AI tools used for this PR

OpenAI Codex (GPT-5) was used to inspect local Deskflow logs, identify the regression in `ServerProxy`, implement the patch, and run the local build/tests. The PR description was manually reviewed and edited.

## Fixed/related issues

No existing issue found.

## How has this been tested?

Tested on KDE Plasma Wayland with Deskflow client controlled from a Windows server. Before the fix, pressing `p` arrived as `id=0x70, mask=0x2000` and was translated to mask `0x0006`, triggering a `Ctrl+Alt+p` input method shortcut instead of typing. After the fix, normal keys type correctly.

Also ran:

```sh
cmake -S /home/saki/deskflow-keyfix -B /home/saki/deskflow-keyfix/build
cmake --build /home/saki/deskflow-keyfix/build --target run_tests
```

Result: 25/25 tests passed, including the new `ServerProxyTests`.

### DIFF
--- a/src/lib/client/ServerProxy.cpp
+++ b/src/lib/client/ServerProxy.cpp
@@ -387,6 +387,11 @@ void ServerProxy::sendInfo(const ClientInfo &info)
 
 KeyID ServerProxy::translateKey(KeyID id) const
 {
+  return translateKey(id, m_modifierTranslationTable);
+}
+
+KeyID ServerProxy::translateKey(KeyID id, const KeyModifierID *modifierTranslationTable)
+{
   static const KeyID s_translationTable[kKeyModifierIDLast][2] = {
       {kKeyNone, kKeyNone},     {kKeyShift_L, kKeyShift_R}, {kKeyControl_L, kKeyControl_R}, {kKeyAlt_L, kKeyAlt_R},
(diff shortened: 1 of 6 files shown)


sithlord48:
What locale are you using for this ?

sithlord48:
The new tests are broken on windows.

sithlord48:
Closing for #9945

Issue #9075, opened 2025-10-24
Cursor jumps to center of the client screen


tobiashugo:
### Sanity checks

- [ ] I have done the sanity checks, and my issue persists
- [x] These sanity checks are not relevant to the bug

### Deskflow version info

Deskflow Deskflow: 1.24.0.139 (b5c188ab) (also 1.24.0) (also Synergy 3/Barrier)
Qt: 6.10.0
System server: Windows 11 Version 25H2
System client: MacOS Sequioa 15.6.1

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [x] Windows
- [x] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Deskflow configuration

Windows 11 Server, macos 15 client
Windows has two monitors (2560x1440) middle (secondary) and on the right (primary), Macos 1920x1200 on the left.

### What steps will reproduce the problem?

The cursor on the mac client jumps away from the right border of the screen when transitioning with speed. Only when approached slowly, the transition remains natural and continuous. 
When configuring the middle monitor as primary everything is fine.

### Log output

```shell

```

andyearnshaw:
@tobiashugo is this the same as the bug I describe in #9072?

tobiashugo:
They might be related but are not identical. I can add, the cursor jumps to the middle of the client screen or even further.

verglor:
Hi, I have similar problem as https://github.com/deskflow/deskflow/issues/9072#issuecomment-3442660829 which seems to better fit to this issue since it's not about wrong screen but rather wrong position on correct screen.

PR #9942, opened 2026-07-04
fix(win): drop stale hook motion delivered after leave warps the cursor to the center

MEIZU16:
## Description

When the server switches to a client, `MSWindowsScreen::leave()` warps the (hidden) cursor to the center of the primary display so that relative motion can be accumulated. Hardware events generated just before that warp can still be delivered by the low-level mouse hook after it. Their positions are relative to the pre-warp cursor near the screen edge, so the delta computed against the new center baseline amounts to roughly half a screen, and it is relayed to the client: the cursor teleports away from the entry point, typically landing near the middle of the client screen (with multi-monitor servers it can land even further away).

Two defects combine here:

1. The existing bogus-motion filter only drops deltas within 10 px of the exact center-to-edge distance. Stale events produced during fast movement report positions further inside the screen and pass through.
2. `onMouseMove()` saves every event's position as the baseline for the next delta *before* classifying the event. A dropped event therefore poisons the baseline: the next event (often the warp echo at the center) yields a huge delta in the opposite direction, passes the filters, and pushes the accumulated server position back across the shared edge - the server instantly switches back to the screen it just left. With fast crossings this repeats as a ping-pong storm. Log from the affected setup (Windows server, 3840x2160 primary; client entered at its left edge):

```
[11:57:48.372] INFO: switch from "DESKTOP-7BV50ER" to "mujica" at 0,685
[11:57:48.372] INFO: leaving screen
[11:57:48.392] DEBUG: dropped bogus delta motion: +2037,+89
[11:57:48.395] INFO: switch from "mujica" to "DESKTOP-7BV50ER" at 3729,1136   <- bounced back after 23 ms
(repeats every ~35 ms while the user keeps moving)
```

The fix:

- While off screen the cursor is parked at the center, so genuine relayed events always report a position near the center. For 500 ms after leaving, a motion event reporting a position further from the warp center than two thirds of the primary screen's half width/height can only be a stale pre-warp event; drop it (`isStaleMotionAfterLeave()`).
- Events dropped by either filter re-baseline the saved mouse position to the warp center instead of keeping the dropped event's position, so warp echoes can no longer produce reverse jumps.

## Disclosure of AI use

This PR was developed with the help of an AI coding assistant (Claude Code). The root-cause analysis, the change and the testing were verified on real hardware before submission.

## Related Issue

fixes: #9072

The "cursor jumps towards the middle of the client screen when crossing fast" reports in #9075 (closed as a duplicate) describe the same mechanism.

## How Has This Been Tested?

- Windows 11 Pro for Workstations (build 26200) server with a 3840x2160 primary display, KDE Plasma (Wayland) client, built with MSVC / Qt 6.10.3 on top of v1.26.0 and current master.
- Before the change: fast edge crossings intermittently teleported the client cursor to mid-screen or bounced control back to the server within ~25 ms (VERBOSE logs show the stale deltas of about half the screen width).
- After the change: extended fast back-and-forth crossing sessions behave correctly; the new `dropped stale motion after leave` DEBUG line confirms stale events are intercepted and no reverse switches occur.


sithlord48 (changes requested):
Way to long of commit message and way to many comments in obvious code. Keep the descriptions in the headeers.

src/lib/platform/MSWindowsScreen.cpp
```diff
@@ -1418,6 +1426,19 @@ bool MSWindowsScreen::ignore() const
...
+bool MSWindowsScreen::isStaleMotionAfterLeave(int32_t mx, int32_t my) const
+{
+  using namespace std::chrono_literals;
+  if (std::chrono::steady_clock::now() - m_leaveTime > 500ms) {
+    return false;
+  }
+
+  // genuine events report a position near the parked cursor at the center;
+  // events generated before the leave() warp report one near a screen edge
+  const int32_t staleZone = (std::min)(m_xCenter, m_yCenter) * 2 / 3;
```
nbolton: Create consts for the magic numbers, `2`, and `3`

src/lib/platform/MSWindowsScreen.cpp
```diff
@@ -1418,6 +1426,19 @@ bool MSWindowsScreen::ignore() const
   return (m_mark != m_markReceived);
 }
 
+bool MSWindowsScreen::isStaleMotionAfterLeave(int32_t mx, int32_t my) const
+{
+  using namespace std::chrono_literals;
+  if (std::chrono::steady_clock::now() - m_leaveTime > 500ms) {
```
nbolton: Create a const for the magic number `500ms` and add a small comment explaining why you chose that number.

src/lib/platform/MSWindowsScreen.cpp
```diff
@@ -1418,6 +1426,19 @@ bool MSWindowsScreen::ignore() const
...
+    return false;
+  }
+
+  // genuine events report a position near the parked cursor at the center;
+  // events generated before the leave() warp report one near a screen edge
```
nbolton: > `leave()`

Putting a function name in a comment is fragile as it tends not to survive across refactors.

src/lib/platform/MSWindowsScreen.cpp
nbolton: Pretty sure this doesn't seem to match the PR description: `min()` gives one threshold for both axes, the description reads as per-axis.

nbolton (changes requested):

PR #9944, opened 2026-07-04
fix: clamp modifier translation table index, not the translated value

MEIZU16:
## Description

Commit 205a3c80 ("fix: clamp mapped modifiers") applied `std::clamp` to the *results* of the client's key/modifier translation instead of the table index that needed bounds checking. On any client built from a revision containing it:

- `translateKey()` clamps the returned `KeyID` to `[0, kKeyModifierIDLast - 1]` = `[0, 6]`: translated modifier keys turn into bogus key ids, and ordinary key ids collapse to 6, mangling key events.
- `translateModifierMask()` clamps the rebuilt modifier mask the same way, stripping `KeyModifierMeta`/`KeyModifierSuper` and all lock bits (num/caps/scroll lock) from every key event.

**First commit** clamps the translation table lookup index instead and returns translated values unchanged. It adds unit tests for both functions; the translation helpers get static overloads taking the table as a parameter so they can be tested without a server connection. The new test binary links `net` explicitly because the client library references `NetworkAddress`, which MSVC resolves strictly.

**Second commit**, independent of the regression: if a key event arrives whose modifier mask omits modifiers that are still physically held on the client (key down delivered, no key up yet), the client synthesizes modifier releases around the key and shortcuts like Ctrl+X stop working. Track active non-lock modifier presses per key button and merge them into subsequent key event masks when the incoming mask omits them. The tracked state is cleared on enter, leave and resetOptions to avoid stale modifier state.

## Disclosure of AI use

This PR was developed with the help of AI coding assistants (Claude Code and OpenAI Codex). The analysis and changes were reviewed and tested on real hardware before submission.

## Related Issue

Regression introduced by 205a3c80; no open issue found for it yet.

## How Has This Been Tested?

- New unit tests: 6/6 pass on Windows (MSVC, Qt 6.10.3); the change also builds and its tests pass on Linux.
- Real setup: Windows 11 server + KDE Plasma (Wayland) client. With a client built from a post-205a3c80 revision, key ids and lock/modifier masks arrive mangled; with this change they arrive intact and modifier shortcuts work.
- `deskflow-core` builds cleanly on Windows with the change applied.

### DIFF
--- a/src/lib/client/ServerProxy.cpp
+++ b/src/lib/client/ServerProxy.cpp
@@ -387,6 +387,11 @@ void ServerProxy::sendInfo(const ClientInfo &info)
 
 KeyID ServerProxy::translateKey(KeyID id) const
 {
+  return translateKey(id, m_modifierTranslationTable);
+}
+
+KeyID ServerProxy::translateKey(KeyID id, const KeyModifierID *modifierTranslationTable)
+{
   static const KeyID s_translationTable[kKeyModifierIDLast][2] = {
       {kKeyNone, kKeyNone},     {kKeyShift_L, kKeyShift_R}, {kKeyControl_L, kKeyControl_R}, {kKeyAlt_L, kKeyAlt_R},
       {kKeyMeta_L, kKeyMeta_R}, {kKeySuper_L, kKeySuper_R}, {kKeyAltGr, kKeyAltGr}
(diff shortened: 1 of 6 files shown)


nbolton:
I made a simpler version of this PR: 
- https://github.com/deskflow/deskflow/pull/9945

What does your PR solve that mine doesn't? The change seems very large.

Edit: Maybe we should just drop one of the commits? Keeping the other (the held-modifier / Ctrl+X tracking fix) -- need to make sure that the Ctrl+X bug still reproduces once the bad output clamps are gone.

MEIZU16:
Agreed - the minimal version is the right call. The write-site clamp in `setOptions()` already covers the OOB concern, so let's go with #9945. Closing this PR in favor of it.

If useful, the unit tests from this PR could still land separately on top of #9945. As for the held-modifier commit: it was diagnosed on a build that also had the output clamps, so I'll only re-propose it if the Ctrl+X symptom still reproduces on a clean build.

PR #9945, opened 2026-07-06
fix: simplify key translation logic by removing clamping

nbolton:
## Description
Fixes a client keyboard regression from 205a3c80 without weakening its security fix.

205a3c80 clamped the modifier index at its write site in `setOptions` (the real OOB fix, kept here), but also clamped the *outputs* of `translateKey()` and `translateModifierMask()` to `[0, 6]`. Those output clamps broke the keyboard: key codes collapsed to `6` and Meta/Super plus lock bits were dropped. They guarded nothing, so this PR just removes them and keeps the index clamp.

Minimal alternative to #9944.

## AI tools used for this PR
Claude Code (Claude Opus 4.8), to find the regression and reduce the fix.

## Fixed/related issues
Regression from 205a3c80. Must not regress GHSA-8rcq-7w87-h64j (PoC: deskflow/scripts#12).

## How has this been tested?
PoC and client keyboard test.

### DIFF
--- a/src/lib/client/ServerProxy.cpp
+++ b/src/lib/client/ServerProxy.cpp
@@ -455,11 +455,9 @@ KeyID ServerProxy::translateKey(KeyID id) const
   }
 
   if (id2 != kKeyModifierIDNull) {
-    return std::clamp<KeyModifierMask>(
-        s_translationTable[m_modifierTranslationTable[id2]][side], 0, kKeyModifierIDLast - 1
-    );
+    return s_translationTable[m_modifierTranslationTable[id2]][side];
   } else {
-    return std::clamp<KeyModifierMask>(id, 0, kKeyModifierIDLast - 1);
+    return id;
   }
 }
 
@@ -489,7 +487,7 @@ KeyModifierMask ServerProxy::translateModifierMask(KeyModifierMask mask) const
   if ((mask & KeyModifierSuper) != 0) {
     newMask |= s_masks[m_modifierTranslationTable[kKeyModifierIDSuper]];
   }
-  return std::clamp<KeyModifierMask>(newMask, 0, kKeyModifierIDLast - 1);
+  return newMask;
 }
 
 void ServerProxy::enter()


sithlord48 (approved):
Good with me as long as the issue we tried to prevent was not re introduced

nbolton:
> Good with me as long as the issue we tried to prevent was not re introduced

Don't worry, I tested it against the PoC
```
./cve_XXXX_XXXXX_mod_map_oob.py

CVE-XXXX-XXXXX -- dsop modifier-map out-of-bounds read
malicious server on 127.0.0.1:24800, poison index 0xfbffff04
start a deskflow client (tls disabled) pointed at this address
[*] waiting for client
client connected from 127.0.0.1:38738, handshake complete
sent poisoned dsop, mapped shift modifier to out-of-range index
sent left-shift key down to trigger the translation
[PASS] client survived -- index clamped, fix in place
```

nbolton:
@manop55555 Mind just checking I didn't regress any vulnerability here? I ran it against my [`cve_XXXX_XXXXX_mod_map_oob.py`](https://github.com/deskflow/scripts/blob/28b627e696098494b2f5a5ccc08e4a93c9833673/security/poc/cve_XXXX_XXXXX_mod_map_oob.py) PoC but thought I'd double check with you to be safe.

manop55555:
@nbolton Hey, thanks for pinging me on this. I ran the PoC against master to double check and can confirm it's clean, no crash, client survives fine. For comparison I also tested the same PoC against v1.26.0 (pre-fix) just to make sure the exploit itself still behaves as expected there, and it crashed with SIGSEGV like before, so the test setup checks out. Also poked around a bit with normal keyboard remapping on master to make sure nothing broke, and that's working fine too. So from my side this looks good, no regression, the fix holds up. Appreciate you following up on this.

nbolton:
> @nbolton Hey, thanks for pinging me on this.

No worries! Really appreciate your efforts.

PR #9950, opened 2026-07-08
fix(mac): allow screen switching at interior display edges

stefan-caraiman:
On macOS, the screen shape reported to the server is the bounding box of all displays. With a non-rectangular arrangement, the OS pins the cursor at display edges that lie inside that bounding box, so the cursor can never reach the bounding box edge and no screen switch happens from those edges.

  ## Description

 Setup that triggers the bug: a server with two external displays (A, B) above the built-in display (C), and a client (D) physically to the right of C:

```
Physical arrangement (server = one machine with displays A, B, C):

+---------------+---------------+
|               |               |
|       A       |       B       |
|  (external)   |  (external)   |
|               |               |
+-------+-------+-------+-------+--------+
		|               |                |
		|       C       |        D       |
		|  (built-in)   |    (client)    |
		|               |                |
		+---------------+----------------+
						^
		   cursor should cross here (C <-> D)

Screen shape reported to the server (bounding box of A+B+C):

+---------------+---------------+
|       A       |       B       |
|               |               |
+-------+-------+-------+-------+ <-- D attaches to this edge
|.......|       C       |.......|
|.......|               |.......| <-- ". . ." = dead zone: no display
+-------+---------------+-------+
						^
	  macOS pins the cursor at C's right edge, which is inside
	  the bounding box -- the switch to D never triggers
```

Because B extends further right than C, the bounding box's right edge is B's right edge, which the cursor can never reach from C. The reverse trip has the same problem: entry positions computed on the bounding box can land in the dead zone where no display exists.

Changes (all in `src/lib/platform/OSXScreen.{h,mm}`):

  - The CGEvent tap now forwards the raw mouse deltas to `onMouseMove`. When the cursor is pushed against a display edge with no display beyond it (checked against cached per-display rects), the reported position is promoted to the bounding box edge on that axis, so the server's existing jump logic can trigger. For rectangular arrangements the promoted position equals the actual position, so behavior is unchanged.
  - `warpCursor` clamps targets that lie on no display onto the nearest display, preserving the coordinate mapped along the entered edge (new `clampToDisplays` helper).
  - `updateScreenShape` caches the individual display bounds in `m_displayRects`.

Known limitations: the crossing height is still mapped over the full bounding-box height (flat screen model), so exiting from C enters D proportionally lower rather than perfectly straight; and since the bounding box edge is shared, leaving from B's right edge also switches to D. Equivalent fixes for the Windows/X11/libei backends are out of scope here.

## AI tools used for this PR

  This PR was created with Claude Code using the Claude Fable 5 model (claude-fable-5). It was used for the root-cause analysis & writing the patch itself. I'm not familiar with C++ myself, I validated the behavior on my own setup (see below), but the code deserves an extra-careful review.

## Fixed/related issues

No existing issue found for this; happy to link one if a maintainer knows of a duplicate.

## How has this been tested?

Manually, on the arrangement shown above:

  - Server: MacBook Pro (Apple Silicon), macOS 26, two external displays above the built-in display; built from source (Release, Qt 6.11).
  - Client: MacBook Pro, macOS.
  - Before the patch, the cursor could not leave C's right edge; the client was only reachable via B's right edge. After the patch, the cursor crosses C ↔ D directly in both directions, and returning from D lands on C (not B).
  - Verified existing behavior is unchanged where the display edge coincides with the bounding box edge (crossing from B to D still works).

No automated tests added, the behavior depends on live CoreGraphics display geometry and cursor pinning, which the existing test setup doesn't simulate.

### DIFF
--- a/src/lib/platform/OSXScreen.h
+++ b/src/lib/platform/OSXScreen.h
@@ -111,8 +111,16 @@ private:
   void sendEvent(EventTypes type, void * = nullptr) const;
   void sendClipboardEvent(EventTypes type, ClipboardID id) const;
 
+  // returns true if the given point lies on one of the displays
+  bool isPointOnDisplay(double x, double y) const;
+
+  // moves a point that lies inside the bounding box of all displays but
+  // on no display (a "dead zone" of a non-rectangular arrangement) onto
+  // the nearest display
+  void clampToDisplays(int32_t &x, int32_t &y) const;
+
   // message handlers
-  bool onMouseMove();
+  bool onMouseMove(double deltaX, double deltaY);
   // mouse button handler.  pressed is true if this is a mousedown
   // event, false if it is a mouseup event.  macButton is the index
   // of the button pressed using the mac button mapping.
@@ -231,6 +239,7 @@ private:
   int32_t m_x, m_y;
   int32_t m_w, m_h;
   int32_t m_xCenter, m_yCenter;
+  std::vector<CGRect> m_displayRects;
 
   // mouse state
   mutable int32_t m_xCursor, m_yCursor;
--- a/src/lib/platform/OSXScreen.mm
+++ b/src/lib/platform/OSXScreen.mm
@@ -35,6 +35,7 @@
 
 #include <AppKit/NSEvent.h>
 #include <AvailabilityMacros.h>
+#include <algorithm>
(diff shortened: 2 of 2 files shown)


sithlord48:
you can do this already using the server config "links"  https://github.com/deskflow/deskflow/wiki/Server-Text-Config#links

PR #9957, opened 2026-07-10
fix(mac): prevent tray icon crash on macOS 26+

hsbt:
On macOS 26+ status items are scene-based, so opening the tray menu posts a notification where `NSApp.currentEvent` is not a mouse event. Qt's `QCocoaSystemTrayIcon::emitActivated()` still calls `-[NSEvent clickCount]` on it, which raises an `NSInternalInconsistencyException` and aborts the app. This is unfixed in Qt as of 6.11.

`installMacOSTrayCrashWorkaround()` swizzles the mouse-only `NSEvent` accessors `clickCount` and `buttonNumber` to return 0 for non-mouse events instead of throwing, guarded to run only on macOS 26 and later. A 0 click count makes Qt treat the activation as a plain trigger, so the existing menu behavior is preserved.

Built with Xcode 27 beta 3 and verified on macOS 27 beta 3 that clicking the menu bar tray icon opens the menu without crashing.

### DIFF
--- a/src/apps/deskflow-gui/deskflow-gui.cpp
+++ b/src/apps/deskflow-gui/deskflow-gui.cpp
@@ -24,6 +24,7 @@
 #include <QSharedMemory>
 
 #if defined(Q_OS_MACOS)
+#include "gui/OSXHelpers.h"
 #include <Carbon/Carbon.h>
 #include <cstdlib>
 #endif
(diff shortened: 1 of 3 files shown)


sithlord48:
No issues with the tray on my MBP using os 26..  If its a Qt bug report it to Qt's bug tracker. When mac os 27 gets closer to release we will look into it

hsbt:
Okay. Please don't use my fix when macOS 27 released. Good luck.

PR #9958, opened 2026-07-10
fix: defer client disconnect cleanup

happys2333:
## Description
Keep event handlers alive during dispatch and defer `ServerProxy` disconnect/refusal cleanup until the current callback returns. Remove `ClientApp` handlers using the same event target used during registration.

## AI tools used for this PR
Codex gpt5.6

## Fixed/related issues
Related: #9639

## How has this been tested?
- macOS 26.5.2, AppleClang 21.0.0, Qt 6.11.1
- Full local Debug build
- 26/26 QTest tests
- 27/27 legacy tests
- `EventQueueTests` and `ServerProxyTests` under AddressSanitizer
- Modified targets compiled with warnings as errors
- `clang-format` 20.1.0

### DIFF
--- a/src/lib/base/EventQueue.cpp
+++ b/src/lib/base/EventQueue.cpp
@@ -167,12 +167,12 @@ bool EventQueue::getEvent(Event &event, double timeout)
 bool EventQueue::dispatchEvent(const Event &event)
 {
   void *target = event.getTarget();
-  if (const auto *type_handler = getHandler(event.getType(), target); type_handler) {
-    (*type_handler)(event);
+  if (auto typeHandler = getHandler(event.getType(), target); typeHandler.has_value()) {
+    (*typeHandler)(event);
     return true;
   }
-  if (const auto *any_handler = getHandler(EventTypes::Unknown, target); any_handler) {
-    (*any_handler)(event);
+  if (auto anyHandler = getHandler(EventTypes::Unknown, target); anyHandler.has_value()) {
+    (*anyHandler)(event);
     return true;
   }
   return false;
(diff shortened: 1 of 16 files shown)


src/lib/base/EventQueue.cpp
```diff
@@ -167,12 +167,12 @@ bool EventQueue::getEvent(Event &event, double timeout)
 bool EventQueue::dispatchEvent(const Event &event)
 {
   void *target = event.getTarget();
-  if (const auto *type_handler = getHandler(event.getType(), target); type_handler) {
-    (*type_handler)(event);
+  if (auto typeHandler = getHandler(event.getType(), target)) {
```
sithlord48: This makes the IF pointless.. If you can assign a variable will be 100 % of the time..

sithlord48 (changes requested):

src/lib/base/EventQueue.cpp
```diff
@@ -295,17 +295,17 @@ void EventQueue::removeHandlers(void *target)
   }
 }
 
-const EventQueue::EventHandler *EventQueue::getHandler(EventTypes type, void *target) const
+EventQueue::EventHandler EventQueue::getHandler(EventTypes type, void *target) const
```
amagus: perhaps this should return std::optional<EventHandler> instead?

src/lib/base/EventQueue.cpp
amagus: One other possible approach if defining an operator bool on the EventHandler object... Or making it a std::optional as I suggested on the other file

happys2333: Implemented the std::optional approach, with an explicit has_value() check at each dispatch site.

src/lib/base/EventQueue.cpp
happys2333: Implemented. getHandler() now returns std::optional<EventHandler> and uses std::nullopt when no handler is registered.

src/lib/base/EventQueue.cpp
happys2333: Good point. I changed getHandler() to return std::optional<EventHandler> and now check has_value() explicitly before dispatch. I also added coverage for the no-handler and Unknown-fallback paths.

sithlord48:
Failing windows build tests. Drafting for now , please undraft when its fixed up and ready for review.

sithlord48:
Remember to not fix review issues with new commits instead fix them in the commit they were made
 
I've squashed your fixes into the second commit.

Gonna run to check windows tests are working now

nbolton (approved):

sithlord48 (approved):

PR #9962, opened 2026-07-11
Performance & reliability: logging, TLS socket state, wake reconnect, and Windows tweaks

nxrobins:
## Summary

Six independent performance/reliability improvements, each in its own commit so they can be reviewed or cherry-picked separately. Found via an audit focused on input latency and connection stability.

## Changes

1. **`perf(log)` — skip argument evaluation for filtered-out log messages.** In release builds `LOG_*` expanded to an unconditional `Log::print()`, so every call evaluated its arguments (e.g. per-input-event `getName()` string construction) and took the global log mutex via `getFilter()` before the level check discarded the message. The filter level is now atomic (lock-free) and the macros gate on a cheap `willPrint()` before evaluating arguments.

2. **`fix(net)` — per-connection SecureSocket I/O state.** `doWrite()`/`secureRead()`/`secureWrite()`/`secureAccept()`/`secureConnect()` held their scratch write buffer and SSL retry counters in function-local `static`s, shared across all instances. With multiple TLS clients (or across a reconnect) one connection's partial-write/handshake state could bleed into another; the write buffer was also leaked. Now per-connection members.

3. **`fix(server)` — consume deprecated file-transfer/drag message bodies.** `ClientProxy1_5` recognised `kMsgDFileTransfer`/`kMsgDDragInfo` but read none of the body, so a foreign/older client sending them desynced the message stream. The bodies are now read and discarded, keeping the stream framed.

4. **`fix(client)` — reconnect on wake when the suspend signal is missed.** On Windows 10/11 the suspend notification is often dropped and `WM_TIMECHANGE` stands in for resume, so `m_suspended` was never set and `handleResume()` (gated on it) did nothing — leaving the dead connection until keep-alive death detection (~9s). The resume decision is extracted into a pure, unit-tested policy that drops a stale connection so the existing auto-reconnect re-establishes in ~1s. The normal suspend→resume path and all other platforms are unchanged.

5. **`perf(daemon)` — adaptive backoff for the watchdog output poll.** The non-blocking output pipe was polled with a fixed 100ms sleep, costing a steady ~10 Hz wakeup in the always-running daemon. It now backs off (20→200 ms) while the child is silent and polls faster right after output; the pipe stays non-blocking so shutdown behaviour is unchanged.

6. **`perf(win)` — center the cursor as soon as it's hidden on screen leave.** Replaces a fixed 30 ms sleep with a bounded poll of cursor visibility, so the transition proceeds as soon as Windows hides the cursor (capped at the same ~30 ms).

## Verification

Built against Qt 6.10.1 + OpenSSL 3.0.16 (MSVC 2022). The full unit-test suite passes (**24/24**), including two new tests:
- `LogTests::verboseArgsNotEvaluatedWhenFiltered` — the lazy-evaluation property (#1)
- `ResumePolicyTests` — the resume-decision policy over its full input space (#4)

## Notes for reviewers

- **#4** deliberately trades a rare, cheap redundant reconnect on a spurious `WM_TIMECHANGE` (e.g. an NTP clock change) for reliable sub-second recovery after every real wake. The **server-side** resume path is intentionally left alone (its listener survives sleep; restarting on spurious signals would drop clients) and would benefit from a separate, integration-tested fix.
- **#5 and #6** are Win32 I/O / screen-transition code not covered by unit tests, so they are compile-verified and reasoned about, but their runtime behaviour (daemon wakeup cadence; cursor flicker) has not been measured end-to-end.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/lib/base/Log.cpp
+++ b/src/lib/base/Log.cpp
@@ -219,14 +219,17 @@ bool Log::setFilter(const QString &maxPriority)
 
 void Log::setFilter(LogLevel::Level maxPriority)
 {
-  std::scoped_lock lock{m_mutex};
-  m_maxPriority = maxPriority;
+  m_maxPriority.store(maxPriority, std::memory_order_relaxed);
 }
 
 LogLevel::Level Log::getFilter() const
 {
-  std::scoped_lock lock{m_mutex};
-  return m_maxPriority;
+  return m_maxPriority.load(std::memory_order_relaxed);
+}
+
+bool Log::willPrint(LogLevel::Level level) const
+{
+  return level <= getFilter();
 }
 
 void Log::output(LogLevel::Level priority, const char *msg)
--- a/src/lib/base/Log.h
+++ b/src/lib/base/Log.h
@@ -10,6 +10,7 @@
 
 #include "common/LogLevel.h"
 
+#include <atomic>
 #include <list>
 #include <mutex>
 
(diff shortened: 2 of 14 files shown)


sithlord48:

nxrobins:
withdrawing, thanks

Issue #9779, opened 2026-05-24
macOS server: cursor jumps on first mouse move after switching to client


MomoyamaSawa:
### Before reporting a bug, please complete the sanity checks above.

- [x] I have done the sanity checks, and my issue persists.

### Deskflow version info

**Server**
Deskflow: 1.26.0.207 (62a8ee74)
Qt: 6.10.2
System: macOS Sequoia (15.7.5)

**Client**
Deskflow: 1.26.0.207 (62a8ee74)
Qt: 6.10.2
System: Windows 10 Version 21H2

### Build types

Official release (installer/package from GitHub releases)

### Deskflow configuration

- macOS server (primary), Windows client (secondary)
- `relativeMouseMoves = false`
- `defaultLockToScreenState = true`
- Two keystrokes to switch in direction between two devices

### What steps will reproduce the problem?

1. Set up Deskflow with macOS as server and Windows as client.
2. Move the macOS cursor to a position clearly offset from the screen center (e.g. near a corner).
3. Use a hotkey (`switchInDirection` or `switchToScreen`) to switch to the Windows client.
4. Confirm the cursor appears at the expected Windows position — no jump yet.
5. Move the mouse slightly.
6. Observe: the cursor jumps by a vector roughly equal to `(macOS cursor position) − (macOS screen center)` at the moment of switching.

The issue does **not** reproduce in the reverse direction (Windows → macOS).  
Natural edge-crossing with `defaultLockToScreenState = false` does not reproduce the jump; hotkey switching always does.

### Log output

<details><summary><h3>Debug1 logs — server and client</h3></summary>

macOS server (relevant excerpt):
```
[2026-05-24T21:36:16.603] DEBUG1: hotkey: switchInDirection(left)
[2026-05-24T21:36:16.604] INFO: switch from "MAC.local" to "DESKTOP-UADIEQA" at 0,0
[2026-05-24T21:36:16.607] DEBUG1: send enter to "DESKTOP-UADIEQA", 0,0 2 0017
[2026-05-24T21:36:18.246] DEBUG1: try to leave "DESKTOP-UADIEQA" on left
[2026-05-24T21:36:18.247] DEBUG1: no neighbor left
[2026-05-24T21:36:37.955] DEBUG1: hotkey: switchInDirection(right)
[2026-05-24T21:36:37.955] INFO: switch from "DESKTOP-UADIEQA" to "MAC.local" at 777,689
```

Windows client (relevant excerpt):
```
[...]

PR #9963, opened 2026-07-12
fix: capture the mouse on macOS while the cursor is on a client

### LINKED ISSUE #5871: Fullscreen Game / Shooter issue (Server PC)
When you have a fullscreen game running on Server PC.
And you switch to Client PC with a hotkey it works when you enable:
Server configuration - Advanced: "Do not enable Windows Server window in foreground"

So basically, when I play a fullscreen game with a "curstor" like StarCraft 2, or League of Legends on the Server PC. And I switch to the Client PC. The cursor stand still.
Like it have to be.

When you play a fullscreen game without a "cursor" like a shooter, for example Overwatch.
And you switch with a hotkey from Server to Client PC. And you move the Mouse Cursor on the Client PC you still move the "crosshair" (aim) of the shooter...
And even when you left click something on the Client PC, you SHOOOOT on the Server PC.
I know that is because the game have "no cursor".
Only "workaround" at the moment is, to open "menu" to get a mouse cursor, and then switch to Client PC with a hotkey. So the cursor stand still.
**Maybe Synergy can be optimized for shooter games aswell?!**

## And there is another "Bug"?!
When I play (for example) Witcher 3 in fullscreen on the Server PC.
I open the "Menu" to get a mouse cursor. Then I switch to the Client PC via hotkey.
When I move the mouse around. The mouse cursor on the Server PC (Witcher 3) stand still.
Like it have to be. Perfect so far :)
BUT when I left click something on the Client PC !!!
It left click on the Server PC (Witcher 3) aswell. So it removes the Menu from the game.
Because the cursor is hoverd: Continue playing

Why is a left click on the "Client PC" still left clicking on the "Server PC"?!
I thought when I'm on the Client PC. All the inputs (keyboard / mouse) will be on the Client PC only. And not on the Server PC. So basically this should be a bug.

I know Synergy is not made for "Games".
But it should be. Synergy should be working with playing a game on one of the PCs, while controlling other PCs.

This will give Synergy a MUCH wider spectrum of people who wants to use Synergy.
Like "Dual PC Streamer". Means:
- Playing on PC1
- Streaming on PC2 (Capture Card build in to get image / sound from PC 1)

tehreet:
## Description

On the macOS server, while the cursor is on a client, every physical mouse move warped the local cursor back to screen center and the event-tap returned the mouse-move event instead of consuming it. Local apps sitting under the centered cursor therefore kept reacting to the movement — a fullscreen game or a centered video responds as if the mouse were active at center.

This captures the mouse the way native macOS games do:

- `leave()` calls `CGAssociateMouseAndMouseCursorPosition(false)`, so the cursor freezes in place and hides; the hardware still emits deltas.
- `onMouseMove()` reads the raw delta straight from the triggering event instead of diffing cursor position and warping.
- the tap callback consumes off-screen moves (returns `nullptr`) rather than force-returning them — safe now that the cursor is frozen.
- `enter()` re-associates the mouse (`disable()` already does via `showCursor()`).

The old center-warp and the "bogus motion" edge heuristic are removed, since they only existed to paper over the warp.

## Disclosure of AI use

AI (Claude) was used to help analyze the root cause and draft the patch. I reviewed the change, built it, and tested it on real hardware myself.

## Related Issue

fixes: #5871

May also relate to the cursor center-jump reports #9075 and #9779.

## How Has This Been Tested?

Built on macOS (Apple Silicon) with `-DCMAKE_COMPILE_WARNING_AS_ERROR=ON`; `clang-format` clean. Ran the patched server with a second Mac as client: moved the cursor onto the client and confirmed a centered/fullscreen local app no longer reacts to mouse movement, the client still receives motion normally, and returning to the server re-couples the cursor with no dead-mouse.

### DIFF
--- a/src/lib/platform/OSXScreen.h
+++ b/src/lib/platform/OSXScreen.h
@@ -112,7 +112,7 @@ private:
   void sendClipboardEvent(EventTypes type, ClipboardID id) const;
 
   // message handlers
-  bool onMouseMove();
+  bool onMouseMove(CGEventRef event);
   // mouse button handler.  pressed is true if this is a mousedown
   // event, false if it is a mouseup event.  macButton is the index
   // of the button pressed using the mac button mapping.
--- a/src/lib/platform/OSXScreen.mm
+++ b/src/lib/platform/OSXScreen.mm
@@ -769,6 +769,8 @@ void OSXScreen::enter()
   showCursor();
 
   if (m_isPrimary) {
+    // re-couple the mouse to the cursor, undoing the capture from leave()
+    CGAssociateMouseAndMouseCursorPosition(true);
     setZeroSuppressionInterval();
   } else {
     // reset buttons
@@ -798,8 +800,11 @@ void OSXScreen::leave()
 
   if (m_isPrimary) {
     avoidHesitatingCursor();
-    LOG_VERBOSE("centering cursor on leave: %+d, %+d", m_xCenter, m_yCenter);
-    warpCursor(m_xCenter, m_yCenter);
+
+    // capture the mouse: freeze the cursor so no local app sees motion while on
+    // a client (onMouseMove reads raw deltas instead). must follow hideCursor(),
+    // which re-associates. re-coupled in enter()/disable().
+    CGAssociateMouseAndMouseCursorPosition(false);
   }
 
   // now off screen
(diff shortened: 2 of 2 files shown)


sithlord48:
Seems to have alot of comments, Try to keep them to a minimum and only when needed the code should be written clearly enough to follow the logic (most of the time)

tehreet:
Thanks — trimmed the comments down to just the non-obvious bits (the mouse-capture rationale in leave/onMouseMove). Force-pushed.

amagus (approved):
Works flawlessly. Fixes movement detection on the main screen. Tested and it is working on macOS 26 and 27.

sithlord48 (approved):

sithlord48:
This is looking good i am gonna land it if CI passes @amagus thanks for the review and testing

tehreet:
LFG

PR #9965, opened 2026-07-13
fix(package): improve msvc runtime version check on Windows

nbolton:
## Description

The runtime guard read its minimum from the host registry (installed redist version), but a debug build loads vcruntime140d.dll from the VS toolset, which often lags the redist. So a fully up-to-date machine fails with "runtime outdated" and the daemon exits. 

This PR derives the floor from the actual compiler version (cl 19.44 -> CRT 14.44), which tracks the runtime the toolset actually ships. This is more precise.

## AI tools used for this PR

Claude Code (Claude Opus 4.8) to find the CMake config that needed changing and fix it.

## Fixed/related issues

Not sure? Not many of us open source Windows devs out there!

## How has this been tested?

CMake reconfigure, bulild and smoke test on Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,18 +119,20 @@ else()
 endif()
 
 if (MSVC)
-  # On Windows, require that the same MSVC runtime is used as on the host.
-  # Mitigates things like access violations caused by accidental ABI-compatibility breakage.
+  # Require the same MSVC runtime that built the binary. Mitigates access violations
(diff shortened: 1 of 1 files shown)


sithlord48 (approved):
A lot of work for something that is a non issue on all other platforms...

nbolton:
> A lot of work for something that is a non issue on all other platforms...

Windows! :D

PR #9972, opened 2026-07-15
fix: don't recenter cursor or notify server on transient EiScreen shape churn

ElCruncharino:
## Description

On the Wayland/libei backend, `EiScreen::updateShape()` recenters the cursor to
the middle of the screen and notifies the server of a "shape changed" event
every single time it's called, including during transient device-list churn
when no device is currently reporting a region.

The compositor can renegotiate the EIS device list (tear down and recreate a
device) for reasons that have nothing to do with deskflow. On KDE/KWin, for
example, simply opening the on-screen brightness/volume slider popup is enough
to trigger this. Each time it happens, `updateShape()` briefly sees a degenerate
"no region" state, recenters the tracked cursor to the middle of the screen, and
tells the server the shape changed. The server reacts to that, producing a
visible cursor jump with no user input involved.

This reproduces on stock `master`, with no other patches applied.

The fix: only touch tracked shape state (`m_x`/`m_y`/`m_w`/`m_h`, and the cursor
position) once a real region is actually found, and only notify the server when
the shape genuinely changed, not on every churn round-trip back to the same
values.

## AI tools used for this PR

Claude Code was used to help verify this fix (building, running the client on
real hardware, and drafting this description). The bug and the fix itself were
identified and written by me.

## Fixed/related issues

Found while testing #9894. Not tied to a filed issue, and not caused by that
PR (confirmed by reproducing it against stock `master` with none of that PR's
changes applied).

## How has this been tested?

Built from this branch (`cmake -B build -G Ninja && cmake --build build`, 26/26
unit tests passing) and run as the live client on real hardware (KWin Wayland /
Plasma, openSUSE Tumbleweed):

- Reproduced first against unpatched `master`: with the deskflow cursor parked
  on the client screen, opening the KDE brightness OSD reliably snapped the
  cursor to the middle of the screen.
- Applied this fix, rebuilt, redeployed the live client, repeated the same
  trigger: cursor stays exactly where it was.

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -481,26 +481,49 @@ bool EiScreen::isPrimary() const
 
 void EiScreen::updateShape()
 {
-  m_w = 1;
-  m_h = 1;
-  m_x = std::numeric_limits<uint32_t>::max();
-  m_y = std::numeric_limits<uint32_t>::max();
+  std::uint32_t newW = 1;
+  std::uint32_t newH = 1;
+  std::uint32_t newX = std::numeric_limits<uint32_t>::max();
+  std::uint32_t newY = std::numeric_limits<uint32_t>::max();
+  bool foundRegion = false;
   for (auto it = m_eiDevices.begin(); it != m_eiDevices.end(); it++) {
     auto idx = 0;
     struct ei_region *r;
     while ((r = ei_device_get_region(*it, idx++)) != nullptr) {
-      m_x = std::min(ei_region_get_x(r), m_x);
-      m_y = std::min(ei_region_get_y(r), m_y);
-      m_w = std::max(ei_region_get_x(r) + ei_region_get_width(r), m_w);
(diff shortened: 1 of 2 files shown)


sithlord48 (approved):
Looks alright but not sure how exactly to trigger a test of this with my x11 client . @nbolton  ill leave landing to you 

Edit:

I Tested this with a wayland client (plasma 6.7.3) and i can reproduce this . You need to try to drag the brightness slider.

With this PR it no longer does that and you can use the slider correctly !! 

@ElCruncharino  Nice find!

sithlord48:
> Reproduced first against unpatched master: with the deskflow cursor parked
on the client screen, opening the KDE brightness OSD reliably snapped the
cursor to the middle of the screen.

I Tested this with a wayland client (plasma 6.7.3) and i can reproduce this . You need to try to drag the brightness slider.

sithlord48 (approved):
Was able to trigger the issue w/ a wayland client and checked this also fixes it !

nbolton (approved):

nbolton:
Small tweak to the commit message: added `fix(wayland):`

PR #9977, opened 2026-07-17
fix(win): stop TLS sockets busy-spinning the socket poll loop

### LINKED ISSUE #9364: Windows high cpu use when TLS is enabled
Split from #9116

Hello =)

This issue still exists in version 1.25 and also in latest continuous version (from this night 2026/01/18) :
 - Deskflow: 1.25.0.0 (d4ff55da)       /  Qt: 6.10.0  /  System: Windows 11 Version 24H2
 - Deskflow: 1.25.0.165 (aca274e9)  /  Qt: 6.10.1  /  System: Windows 11 Version 24H2

CPU goes to 12/13% on my Win_11 x64

Using VisualStudio attaching the process gave me a lot of calls to:
 - ws2_32.dll

Using SysInternals ProcessExplorer gave me those calls mainly on the call stack:
 - ucrtbase.dll ! wcsrchr
 - ws2_32.dll
 - deskflow-core.exe ! OPENSSL_Applink
 
 
 So **I disabled TLS** on both server (Win11) and client (ubuntu 25.10) and **the windows server dropped to 0% instantly**.
 
 I didn't want to open a new thread, so I am just adding an answer here.
 Maybe admins could open it again?
 
 Hope this helps. For now I will use without TLS !
 Have a great day

_Originally posted by @castaway- in https://github.com/deskflow/deskflow/issues/9116#issuecomment-3765929465_

ElCruncharino:
## Description
I noticed this issue tracked on the [1.27.0 milestone](https://github.com/deskflow/deskflow/milestone/92), so I thought I'd take a stab at it.

While looking into #9364, I think I found a plausible explanation for the CPU behavior, though I'd welcome a second pair of eyes on this.

On Windows, `ArchNetworkWinsock::pollSocket()` caches a socket's writability (`m_pollWrite`): once `FD_WRITE` has fired, it stops waiting for writability and assumes the socket stays writable, forcing the poll timeout to `0` instead of blocking. As far as I can tell, the only place that resets this cache is `ArchNetworkWinsock::writeSocket()`, on `WSAEWOULDBLOCK`.

`SecureSocket` appears to never go through `writeSocket()` for actual I/O, since `SSL_write()`/`SSL_read()` operate directly on the raw socket. So my working theory is that once TLS hits any real backpressure, the cache gets stuck and the socket multiplexer's event loop stops blocking and busy-spins calling `SSL_write()` as fast as possible instead of waiting for the next real writable notification. This lines up with the reported symptoms as best I can tell: elevated CPU only on Windows, only with TLS enabled, with call stacks through `OPENSSL_Applink`/`ws2_32.dll`.

My proposed fix: add `IArchNetwork::resetPollWriteOnSocket()` (no-op by default) and call it from `SecureSocket::checkResult()`'s `SSL_ERROR_WANT_WRITE` case, mirroring what `writeSocket()` already does.

While in this code, I also converted `SecureSocket::doWrite()`/`secureRead()`/`secureWrite()`'s retry state from function-local `static` variables to instance members/plain locals. Those statics looked like shared mutable state across every `SecureSocket` instance in the process rather than per connection, which seemed worth fixing at the same time.

## Fixed/related issues
fixes: #9364

## How has this been tested?
I had a tough time proving this out on a Windows machine (that's not my primary platform, but I did try to replicate the issue on the one I had), so I couldn't reproduce the reported CPU spike directly on the affected platform myself. Here's what I did instead:
- Traced the control flow in `ArchNetworkWinsock.cpp`/`SecureSocket.cpp` to confirm `m_pollWrite` is only ever reset in `writeSocket()`, and that `SecureSocket` never calls it.
- Built a standalone harness reimplementing the same `pollSocket()`/`writeSocket()`/`checkResult()` control flow against a simulated backpressuring socket. The results looked promising: the unpatched logic pins ~100% CPU busy-spinning after a single would-block, while the fix matches the plain-TCP baseline (near-0% CPU, blocking properly).

This is encouraging but still just a hypothesis validated in isolation, not a confirmed fix on real hardware. If anyone hitting the original issue (or anyone with a Windows box handy) could give this a try and report back, that'd help a lot before this gets merged. Happy to adjust based on feedback.

## AI tools used for this PR
Claude (Sonnet 5) via Claude Code: used to investigate the root cause and write the fix, and to build the verification harness described above.

### DIFF
--- a/src/lib/arch/win32/ArchNetworkWinsock.cpp
+++ b/src/lib/arch/win32/ArchNetworkWinsock.cpp
@@ -1,5 +1,6 @@
 /*
  * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
  * SPDX-FileCopyrightText: (C) 2012 - 2016 Synergy App Ltd
  * SPDX-FileCopyrightText: (C) 2002 Chris Schoeneman
  * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
@@ -555,6 +556,12 @@ size_t ArchNetworkWinsock::writeSocket(ArchSocket s, const void *buf, size_t len
     throwError(err);
   }
   return static_cast<size_t>(n);
+}
+
+void ArchNetworkWinsock::resetPollWriteOnSocket(ArchSocket s)
+{
+  assert(s != nullptr);
+  s->m_pollWrite = true;
 }
 
 void ArchNetworkWinsock::throwErrorOnSocket(ArchSocket s)
--- a/src/lib/arch/IArchNetwork.h
+++ b/src/lib/arch/IArchNetwork.h
@@ -211,6 +211,16 @@ public:
   */
   virtual size_t writeSocket(ArchSocket s, const void *buf, size_t len) = 0;
 
+  //! Reset the writable poll hint for a socket
+  /*!
+  Tells pollSocket() to wait for a fresh writable notification instead of
+  assuming the socket is still writable. Needed by callers that write to a
+  socket without going through writeSocket(), e.g. OpenSSL's SSL_write().
+  */
+  virtual void resetPollWriteOnSocket(ArchSocket)
+  {
+  }
+
   //! Check error on socket
   /*!
   If the socket \c s is in an error state then throws an appropriate
--- a/src/lib/arch/win32/ArchNetworkWinsock.h
+++ b/src/lib/arch/win32/ArchNetworkWinsock.h
@@ -1,5 +1,6 @@
 /*
  * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
  * SPDX-FileCopyrightText: (C) 2012 - 2016 Synergy App Ltd
  * SPDX-FileCopyrightText: (C) 2002 Chris Schoeneman
  * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
@@ -71,6 +72,7 @@ public:
   void unblockPollSocket(ArchThread thread) override;
   size_t readSocket(ArchSocket s, void *buf, size_t len) override;
   size_t writeSocket(ArchSocket s, const void *buf, size_t len) override;
+  void resetPollWriteOnSocket(ArchSocket s) override;
   void throwErrorOnSocket(ArchSocket) override;
   bool setNoDelayOnSocket(ArchSocket, bool noDelay) override;
   void setKeepAliveOnSocket(ArchSocket, bool keepAlive) override;
--- a/src/lib/net/SecureSocket.cpp
+++ b/src/lib/net/SecureSocket.cpp
@@ -556,8 +556,8 @@ void SecureSocket::checkResult(int status, int &retry)
     break;
 
   case SSL_ERROR_WANT_WRITE:
-    // Need to make sure the socket is known to be writable so the impending
-    // poll action actually triggers on a write.
+    // SSL_write() bypasses writeSocket(), so reset the poll hint ourselves.
+    ARCH->resetPollWriteOnSocket(getSocket());
     setWritable(true);
     retry++;
     LOG_VERBOSE("want to write, error=%d, attempt=%d", errorCode, retry);
--- a/src/lib/net/SecureSocket.cpp
+++ b/src/lib/net/SecureSocket.cpp
@@ -66,6 +66,7 @@ SecureSocket::SecureSocket(
 SecureSocket::~SecureSocket()
 {
   freeSSL();
+  free(m_writeStaticBuffer);
 }
 
 void SecureSocket::close()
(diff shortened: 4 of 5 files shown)


sithlord48:
@ElCruncharino  We will be looking to freeze for the 1.27.0 release soon please update this and get it ready to test asap.

ElCruncharino:
> @ElCruncharino We will be looking to freeze for the 1.27.0 release soon please update this and get it ready to test asap.

Thanks for the heads up. About as ready as it's going to be, per testing on my end. Marking as ready for review.

src/lib/arch/win32/ArchNetworkWinsock.cpp
```diff
@@ -1,5 +1,7 @@
 /*
  * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
+ * SPDX-FileCopyrightText: (C) 2026 Nick Haghiri <nick@haghiri.net>
```
amagus: I would suggest to only include your name directly on the header file if you massively contributed to the functionalities provided by this file. 
Your name will still be recorded as part of Deskflow developers and mentioned on the release note if your PR is approved.
It would be very bad if every single person that modify a given file add their name to its header.

ElCruncharino: > I would suggest to only include your name directly on the header file if you massively contributed to the functionalities provided by this file. Your name will still be recorded as part of Deskflow developers and mentioned on the release note if your PR is approved. It would be very bad if every single person that modify a given file add their name to its header.

Fair suggestion. Happy to bump that out of there before merge. No difference to me!

sithlord48 (approved):
Looks ok the best i can say is it didn't break my linux / mac machines, 

Someone will need to see if this helps the windows bug maybe @nbolton can try it out..

nbolton (approved):
Code looks good. Smoke tested on a Windows server with 2 clients, TLS seems fine. I couldn't reproduce the CPU spike here though, so I can't say either way whether it helps.

I would feel more comfortable landing this if someone else could say it fixes it, but I know PRs are piling up, so... @sithlord48 your call.

mmcxujie-cpu:
**Where to post:** https://github.com/deskflow/deskflow/pull/9977
**Why:** the PR author asked for real-hardware confirmation of the hypothesis and never got it. This is that data.

---

Independent confirmation on real Windows hardware of the busy-spin mechanism described in this PR. This is measured on **1.26.0 (pre-fix)**, so it documents the *before* state — I have not yet tested v1.26.0.443.

**Setup:** Windows 11 Pro (build 26200) client, Intel Core Ultra 9 285K (24 logical CPUs), Deskflow 1.26.0 client mode, TLS enabled (`INFO: network encryption protocol: TLSv1.3`), macOS 1.26.0 server.

Your description — "once TLS hits any real backpressure, the cache gets stuck and the socket multiplexer's event loop stops blocking and busy-spins calling `SSL_write()` as fast as possible instead of waiting for the next real writable notification" — lines up exactly with what the process is doing here.

Thread-level evidence while the desktop is completely idle, one session established:

```
deskflow-core.exe  pid=11104  PriorityClass=RealTime  BasePriority=24  threads=5

tid=26436   cpu%=99   state=2 (Running)   waitReason=6    prio=24   <-- the spinner
tid=31056   cpu%=0    state=5 (Waiting)   waitReason=15   prio=25
tid=23944   cpu%=0    state=5 (Waiting)   waitReason=13   prio=24
tid=26600   cpu%=0    state=5 (Waiting)   waitReason=6    prio=24
tid=10804   cpu%=0    state=5 (Waiting)   waitReason=6    prio=24

process CPU, 12 s delta on an idle desktop: 12.02 s => 100.1% of one core
```

So: **exactly one thread**, permanently in `Running` state, pegged at 100 % of a single core — which is what a non-blocking poll timeout of `0` on a stuck-writable socket would produce.

Two control experiments that narrow it down the same way your analysis predicts:

| Change | Result |
|---|---|
| `remoteHost` → `127.0.0.1` (no session, so no `SSL_write()` traffic) | CPU delta **0.00 s / 12 s = 0.0 %** |
| Reconnect (encrypted session re-established) | **100.1 % of one core** — deterministic, 100 % reproducible |
| `clipboardSharing=false` (rules out clipboard traffic as the driver) | still **100 %** |

The spin starts the moment the secure session comes up and stops the moment the session is gone, with nothing in between. On this machine it is also 100 % reproducible rather than intermittent, which I hope makes it a useful data point: every one of my session reconnects reproduced it, and I have never once seen it not reproduce.

One thing your PR does **not** cover, which I have reported separately: on Windows `deskflow-core` runs the whole process at `REALTIME_PRIORITY_CLASS` (all five threads at base priority 24–25), so this in-process spin preempts the input stack and DWM — on a 24-thread CPU the spin is only ~4 % of total CPU, yet the entire machine stutters (mouse stutter, browser sluggishness) and Task Manager looks almost idle. Fixing the spin removes the trigger; the priority class will still turn the next in-process spin or block into a system-wide stall.

ElCruncharino:
@mmcxujie-cpu 
Can you test post fix?

I get your point about the thread priority, which itself isn't a hard fix I'd imagine. But may have other implications, so probably not something to rush into merging.

Issue #9852, opened 2026-06-12
macOS 27 beta: clicking inactive window body via Deskflow does not focus or raise window


ZeroSegFault:
### Before reporting a bug, please complete the sanity checks above.

- [x] I have done the sanity checks, and my issue persists.

### Deskflow version info

Deskflow: 1.26.0.236 (76673b8c)
Qt: 6.11.1
System: macOS 27.0 beta (26A5353q)
Session: macOS desktop

### Build types

Local developer build (built it myself), Official release (installer/package from GitHub releases)

### Deskflow configuration

- macOS 27.0 beta machine is the Deskflow client/controlled machine.
- Server controls the macOS client normally.
- Mouse movement and clicks are delivered to the macOS client.
- Issue occurs when clicking an inactive/background window on the macOS client.

### What steps will reproduce the problem?

1. Run Deskflow with a macOS 27 beta machine as the client/controlled machine.
2. On the macOS client, open two normal app windows, for example Finder and TextEdit.
3. Put one window in front.
4. Leave part of the other inactive/background window visible.
5. Move the Deskflow-controlled pointer over the inactive window's body/content area.
6. Click inside the inactive window's body/content area.
7. Observe that the click is delivered to the window content, but the window does not become active or come to the foreground.
8. Repeat the same test by clicking the inactive window's title bar.
9. Observe that clicking the title bar does activate and raise the window.
10. Repeat using a physical mouse or trackpad directly connected to the Mac.
11. Observe that a physical click on the window body/content area does activate and raise the inactive window.

### Log output

No crash or obvious Deskflow log error is produced.

The click event is delivered to the macOS client. The problem appears to be that macOS 27 beta does not activate/raise inactive windows when the synthetic click targets the window body/content area. Synthetic title-bar clicks do activate the window.

### Additional information

This appears to be a macOS 27 beta behavior change or regression affecting synthetic mouse events.

Important observations:
- Deskflow mouse movement works.
- Deskflow synthetic clicks are delivered.
- Clicking an inactive window's title bar through Deskflow activates and raises the window.
- Clicking an inactive window's body/content area through Deskflow can select/interact with content, but the window remains inactive/behind the foreground window.
- A physical mouse/trackpad click at the same content-area coordinate activates and raises the window normally.

This suggests hit-testing and event delivery are working, but normal click-to-focus behavior is not applied by macOS for synthetic content-area clicks.

A local Deskflow workaround was tested successfully:
[...]

PR #9978, opened 2026-07-19
fix(mac): ensure background clicks work correctly on macOS 27

amagus:
## Description
Fixes clicks not bringing up background application to foreground on macOS 27.
Investigation showed that click events need to be numbered to be properly processed on apps using Carbon APIs on this new version of macOS.

Thanks @bho3538 and @ZeroSegFault for helping providing their version of this fix. It was of great help.
I tried their fix first but realized that the first click upon connecting to a server was still not working (any subsequent clicks worked fine). After a lot of trial and error I realized that there are even stricter rules about the event number tag on the first click sent by a given process.

## AI tools used for this PR
Claude Fable 5 was used to discuss possible solutions and create a small reproducer.
Code was fully human written.

## Fixed/related issues
fixes:#9852

## How has this been tested?
Tested on macOS 27 as client, and certified that there was no side effects on macOS 26 (as client and server)

### DIFF
--- a/src/lib/platform/OSXScreen.h
+++ b/src/lib/platform/OSXScreen.h
@@ -236,6 +236,9 @@ private:
   mutable int32_t m_xCursor, m_yCursor;
   mutable bool m_cursorPosValid;
 
+  // mouse event number (needed for macOS 27)
+  uint32_t m_mouseEventNumber = 0;
+
   /* FIXME: this data structure is explicitly marked mutable due
      to a need to track the state of buttons since the remote
      side only lets us know of change events, and because the
--- a/src/lib/platform/OSXScreen.mm
+++ b/src/lib/platform/OSXScreen.mm
@@ -547,6 +547,22 @@ void OSXScreen::fakeMouseButton(ButtonID id, bool press)
 
   CGEventRef event = CGEventCreateMouseEvent(nullptr, type, pos, static_cast<CGMouseButton>(index));
 
+  if (NSProcessInfo.processInfo.operatingSystemVersion.majorVersion >= 27) {
+    // Apps that use Carbon APIs on macOS 27 requires an incrementing event number for each synthetic click
+    if (press) {
+      if (m_mouseEventNumber == 0) {
+        // For some unknown reason, the first click event number must be unique and greater (but in the same ballpark)
+        // than the last event number used by the system.
+        m_mouseEventNumber =
+            CGEventSourceCounterForEventType(kCGEventSourceStateHIDSystemState, kCGEventLeftMouseDown) +
+            CGEventSourceCounterForEventType(kCGEventSourceStateHIDSystemState, kCGEventRightMouseDown) +
+            CGEventSourceCounterForEventType(kCGEventSourceStateHIDSystemState, kCGEventOtherMouseDown);
+      }
+      ++m_mouseEventNumber;
+    }
+    CGEventSetIntegerValueField(event, kCGMouseEventNumber, m_mouseEventNumber);
+  }
+
   CGEventSetIntegerValueField(event, kCGMouseEventClickState, m_clickState);
 
   // Fix for sticky keys


sithlord48 (approved):
@nbolton  this looks ok to me but i am not able to test on mac os 27 yet.

geneccx:
This is very similar to what I ended up with, except my version tracks each button with their own unique counters. No idea which is correct, but it's been working for me. The Apple docs are unfortunately quite unclear as to what is expected...

Looking at OpenJDK, they just start it at an [arbitrarily high number](https://github.com/openjdk/jdk/blob/e487b0dd02c33078b94768ca0cb29ae3baabdd62/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m#L46). It also looks like this has been [an issue for at least 15 years](https://github.com/openjdk/jdk/blame/e487b0dd02c33078b94768ca0cb29ae3baabdd62/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m#L60) 😅
```objc
// Apparently, for mouse up/down events we have to set an event number
// that is incremented on each button press. Otherwise, strange things
// happen with z-order.
static int gsEventNumber;
static int* gsButtonEventNumber;
```

```objc
// Starting number for event numbers generated by Robot.
// Apple docs don't mention at all what are the requirements
// for these numbers. It seems that they must be higher
// than event numbers from real events, which start at some
// value close to zero. There is no API for obtaining current
// event number, so we have to start from some random number.
// 32000 as starting value works for me, let's hope that it will
// work for others as well.
#define ROBOT_EVENT_NUMBER_START 32000

...

gsEventNumber = ROBOT_EVENT_NUMBER_START;
```

```objc
  int eventNumber = gsEventNumber;

    if (isMouseMove) {
        // any mouse movement resets click count
        gsLastClickTime = 0;
    } else {
        clickCount = GetClickCount(isButtonsDownState);

        if (isButtonsDownState) {
            gsButtonEventNumber[button] = gsEventNumber++;
        }
        eventNumber = gsButtonEventNumber[button];
    }
```

amagus:
> This is very similar to what I ended up with, except my version tracks each button with their own unique counters. No idea which is correct, but it's been working for me. The Apple docs are unfortunately quite unclear as to what is expected...
> 
> Looking at OpenJDK, they just start it at an [arbitrarily high number](https://github.com/openjdk/jdk/blob/e487b0dd02c33078b94768ca0cb29ae3baabdd62/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m#L46). It also looks like this has been [an issue for at least 15 years](https://github.com/openjdk/jdk/blame/e487b0dd02c33078b94768ca0cb29ae3baabdd62/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m#L60) 😅
> 
> ```objc
> // Apparently, for mouse up/down events we have to set an event number
> // that is incremented on each button press. Otherwise, strange things
> // happen with z-order.
> static int gsEventNumber;
> static int* gsButtonEventNumber;
> ```
> 
> ```objc
> // Starting number for event numbers generated by Robot.
> // Apple docs don't mention at all what are the requirements
> // for these numbers. It seems that they must be higher
> // than event numbers from real events, which start at some
> // value close to zero. There is no API for obtaining current
> // event number, so we have to start from some random number.
> // 32000 as starting value works for me, let's hope that it will
> // work for others as well.
> #define ROBOT_EVENT_NUMBER_START 32000
> 
> ...
> 
> gsEventNumber = ROBOT_EVENT_NUMBER_START;
> ```
> 
> ```objc
>   int eventNumber = gsEventNumber;
> 
>     if (isMouseMove) {
>         // any mouse movement resets click count
>         gsLastClickTime = 0;
>     } else {
>         clickCount = GetClickCount(isButtonsDownState);
> 
>         if (isButtonsDownState) {
>             gsButtonEventNumber[button] = gsEventNumber++;
>         }
>         eventNumber = gsButtonEventNumber[button];
>     }
> ```

My goal was to make the minimal change that works. Hence I tried to simplify the counters.

Thanks for the JDK pointer, it's reassuring that we are not the first ones doing something similar (but we are 15 years apart and there are now some new APIs to see some of the event counters...)

Still... maybe it's an issue that was solved before that got reintroduced on macOS 27? Very weird.

nbolton:
> @nbolton this looks ok to me but i am not able to test on mac os 27 yet.

Upgrading now.

nbolton (approved):
Repro'd bug on 27 Beta and confirmed this fixes it.

ZeroSegFault:
I just started testing the master branch with this. One thing i noticed differently over my build is that I can no longer drag windows around from their title bars.

The window background clicking is more reliable so that is an improvement.

I have drafted a fix PR here https://github.com/deskflow/deskflow/pull/9991

@nbolton @sithlord48  let me know if there are any issues with it.

PR #9991, opened 2026-07-21
fix(mac): preserve mouse event number during dragging - only affects macOS 27

ZeroSegFault:
## Description

On macOS 27, synthetic mouse-down and mouse-up events use an incrementing mouse event number, but the corresponding drag events retained the default event number of zero.

This prevented windows from being moved by their title bars. Set `kCGMouseEventNumber` on drag events while a mouse button is held so the complete mouse-down, drag, and mouse-up sequence uses the same event number.

## AI tools used for this PR

OpenAI Codex (GPT-5.6) was used to investigate the regression and validate a candidate fix. I manually typed the code, built and signed the application, and tested the submitted change.

## Fixed/related issues

None currently.

## How has this been tested?

Tested on:

- Apple Silicon Mac
- macOS 27 beta 4
- Xcode 27 beta
- Release build using CMake and Ninja
- Application and embedded `deskflow-core` signed with an Apple Development certificate
- Code signature verified using `codesign --verify --deep --strict`

Manual testing confirmed that:

- Windows can be moved by dragging their title bars.
- Clicking controls in background windows continues to work.
- Dragging across desktop icons continues to select multiple icons.

### DIFF
--- a/src/lib/platform/OSXScreen.mm
+++ b/src/lib/platform/OSXScreen.mm
@@ -61,6 +61,12 @@ enum
 
 static const double kCarbonLoopWaitTimeout = 10.0;
 
+// Synthetic mouse button and drag events require event numbers on macOS 27 and later.
+static inline bool needsEventNumber()
+{
+  return NSProcessInfo.processInfo.operatingSystemVersion.majorVersion >= 27;
+}
+
 int getSecureInputEventPID();
 std::string getProcessName(int pid);
 
@@ -463,6 +469,10 @@ void OSXScreen::postMouseEvent(CGPoint &pos) const
 
   CGEventRef event = CGEventCreateMouseEvent(nullptr, type, pos, static_cast<CGMouseButton>(button));
 
+  if (button != -1 && needsEventNumber()) {
+    CGEventSetIntegerValueField(event, kCGMouseEventNumber, m_mouseEventNumber);
+  }
+
   // Dragging events also need the click state
   CGEventSetIntegerValueField(event, kCGMouseEventClickState, m_clickState);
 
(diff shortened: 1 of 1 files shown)


amagus:
@ZeroSegFault Thanks for the quick fix and sorry for missing testing this scenario on my testing! Can you please mention that this affects macOS 27 alone on the PR and commit description title?

ZeroSegFault:
> @ZeroSegFault Thanks for the quick fix and sorry for kissing testing this scenario on my testing! Can you please mention that this affects macOS 27 alone on the PR and commit description title?

Done :)

src/lib/platform/OSXScreen.mm
```diff
@@ -463,6 +463,10 @@
 
   CGEventRef event = CGEventCreateMouseEvent(nullptr, type, pos, static_cast<CGMouseButton>(button));
 
+  if (button != -1 && NSProcessInfo.processInfo.operatingSystemVersion.majorVersion >= 27) {
```
amagus: I wonder if we should create an inoinefunction “needsEventNumber” that returns this OS 27… now that we use in two places I think it would be better

amagus (changes requested):

src/lib/platform/OSXScreen.mm
ZeroSegFault: Maybe something like this?

`inline bool needsMouseEventNumber()
{
return NSProcessInfo.processInfo.operatingSystemVersion.majorVersion >= 27;
} `

Excuse the formatting I'm on my phone. If that looks good I can make the change when I get back later today

amagus: Exactly. And use it on this new block and the old one I introduced on my recent commit to use that.

amagus (approved):

src/lib/platform/OSXScreen.mm
amagus: Resolved.

sithlord48 (approved):
thanks for checking this out @amagus, leaving landing to @nbolton after his review.

nbolton (approved):
How did I miss this bug? Great fix, thanks!

PR #9993, opened 2026-07-21
fix(wayland): fix mouse scroll jitter and missing scroll events on a wayland host to a mac client

### LINKED ISSUE #9120: The mouse scroll wheel doesn't work correctly with a Wayland server and macOS client
### Sanity checks

- [x] I have done the sanity checks, and my issue persists
- [ ] These sanity checks are not relevant to the bug

### Deskflow version info

Deskflow: 1.24.0
DE: KDE Plasma 6.5.1 on Wayland
OS: CachyOS (but I have the same problem on Arch and Debian)

### Build type

Community package (apt, dnf, brew, etc.)

### Operating systems (OS)

- [ ] Windows
- [x] macOS
- [ ] Linux (X11)
- [x] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Deskflow configuration

Linux Server (CachyOS), macOS Client (26.0 Tahoe)
Each system has a single display attached.
macOS to the left, linux as the server

With KDE's default mouse scrolling speed which is set at ~4 tick marks, scrolling on macOS more or less doesn't work.  To get it to scroll with the default KDE config it requires violently spinning the mouse wheel to get it to scroll down a tiny bit.  Changing the scrolling speed to the 11th tick mark in KDE makes scrolling on the macOS client work as expected.  Changing it to anything less than that requires violent scrolling.

### Log output

```shell

```

### Additional information

_No response_

sawicky:
## Description
Hi-res mouse wheels (like my logitech G502) send scrolls as fractional clicks. On a mac client, fractional events are truncated to zero lines. This fix accumulates those fractional events then emits the entire click 120 units at a time

## AI tools used for this PR
Qwen 3.6 27b to find relevant file, provide relevant function to edit (Discrete event vs normal event). no code was written by AI

## Fixed/related issues
Fixes #9120

## How has this been tested?
Server: Arch Linux kernel 7.1.3
DE: KDE Plasma 6.7.3 Wayland
Mouse: Logitech G502 X Lightspeed

Client:
Macbook Pro M3
macOS Tahoe 26.3.1
Deskflow 1.26.0 release build

Moving mouse over to Macbook client, then scrolling lines in VS Code without increasing the scroll sensitivity to 10 in the client settings like suggested in #9120 . These settings remained at 1

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -28,7 +28,8 @@
 #include <unistd.h>
 #include <vector>
 
-// Values are in fractional wheel-click units (1.0 == one full 120-unit click)
+// Values are in fractional wheel-click units (1.0 == one full 120-unit click) for trackpad
+// and pixels in scroll wheel events.
 struct ScrollRemainder
 {
   double x;
@@ -813,8 +814,7 @@ void EiScreen::onPointerScrollEvent(ei_event *event)
 void EiScreen::onPointerScrollDiscreteEvent(ei_event *event)
 {
   // both libei and deskflow use multiples of 120 to represent
-  // one scroll wheel click event so we can just forward things
-  // as-is.
+  // one scroll wheel click event
 
   assert(m_isPrimary);
 
@@ -823,10 +823,26 @@ void EiScreen::onPointerScrollDiscreteEvent(ei_event *event)
 
   LOG_VERBOSE("event: scroll discrete (%d, %d)", dx, dy);
 
+  // accumulate fractional ticks, then emit 120 units at a time
+  struct ei_device *device = ei_event_get_device(event);
+  auto *remainder = static_cast<struct ScrollRemainder *>(ei_device_get_user_data(device));
+  if (!remainder) {
+    remainder = new ScrollRemainder();
+    ei_device_set_user_data(device, remainder);
+  }
+
+  double ax = remainder->x + dx;
+  double ay = remainder->y + dy;
+  auto cx = static_cast<int32_t>(ax / s_scrollDelta) * s_scrollDelta;
+  auto cy = static_cast<int32_t>(ay / s_scrollDelta) * s_scrollDelta;
+  remainder->x = ax - cx;
+  remainder->y = ay - cy;
+
   // libei and deskflow seem to use opposite directions, so we have
   // to send the opposite of the value reported by EI if we want to
   // remain compatible with other platforms (including X11).
-  sendEvent(EventTypes::PrimaryScreenWheel, WheelInfo::alloc(-dx, -dy));
+  if (cx != 0 || cy != 0)
+    sendEvent(EventTypes::PrimaryScreenWheel, WheelInfo::alloc(-cx, -cy));
 }
 
 void EiScreen::onMotionEvent(ei_event *event)


src/lib/platform/EiScreen.cpp
```diff
@@ -822,10 +821,26 @@ void EiScreen::onPointerScrollDiscreteEvent(ei_event *event)
...
+  auto cx = static_cast<int32_t>(ax / s_scrollDelta) * s_scrollDelta;
+  auto cy = static_cast<int32_t>(ay / s_scrollDelta) * s_scrollDelta;
+  remainder->x = ax - cx;
+  remainder->y = ay - cy;
+
   // libei and deskflow seem to use opposite directions, so we have
   // to send the opposite of the value reported by EI if we want to
   // remain compatible with other platforms (including X11).
-  sendEvent(EventTypes::PrimaryScreenWheel, WheelInfo::alloc(-dx, -dy));
+  if (cx != 0 || cy != 0)
```
sithlord48, line 844: won't this always be true ? you don't scroll diagonally and if its always true do we even need it ?

sithlord48:
This PR is also fixing what https://github.com/deskflow/deskflow/pull/9975 does . 
We need to pick one approach.

sithlord48 (commented):
Upon a quick test this seems to work no differently with my mouse. (i guess that is good ? ) Since i have a "normal" mouse ?

src/lib/platform/EiScreen.cpp
sawicky: heya
this line does not deal with diagonal deltas
this is where we check if our scroll fraction is 0 or not
when i scroll with my mouse wheel 1 tiny amount, ax becomes 30, then 60, 90 etc until 120
`auto cx = static_cast<int32_t>(ax / s_scrollDelta) * s_scrollDelta;`
if ax is 30, then this line is trying to cast 30/120 (which is 0.25) to integer first, before multiply happens, so ax becomes 0 * 120 = 0; this is what is causing scrolling to feel bad and have events eaten

src/lib/platform/EiScreen.cpp
```diff
@@ -822,10 +821,26 @@ void EiScreen::onPointerScrollDiscreteEvent(ei_event *event)
...
+    remainder = new ScrollRemainder();
+    ei_device_set_user_data(device, remainder);
+  }
+
+  double ax = remainder->x + dx;
+  double ay = remainder->y + dy;
+  auto cx = static_cast<int32_t>(ax / s_scrollDelta) * s_scrollDelta;
+  auto cy = static_cast<int32_t>(ay / s_scrollDelta) * s_scrollDelta;
+  remainder->x = ax - cx;
+  remainder->y = ay - cy;
```
nbolton, line 839: I'd prefer less cryptic names:

```suggestion
  double accX = remainder->x + dx;
  double accY = remainder->y + dy;
  auto sendX = static_cast<int32_t>(accX / s_scrollDelta) * s_scrollDelta;
  auto sendY = static_cast<int32_t>(accY / s_scrollDelta) * s_scrollDelta;
  remainder->x = accX - sendX;
  remainder->y = accY - sendY;
```

I'd accept `acc` or `accum` but `a` is a bit too cryptic for my liking.

nbolton (changes requested):

src/lib/platform/EiScreen.cpp
sawicky: done, thanks

sawicky:
> This PR is also fixing what #9975 does . We need to pick on approach.

I tried to pull this users branch and test it myself, but my mouse still acted in the same way (skipping inputs). I don't know codebase that well but from quick debug I think his code is for KWin touchpad scrolling whereas Discrete scrolling is for hardware wheels

but i did like his truncate to 0 instead of floor, so i have cherry picked it

sithlord48 (approved):
No changes locally for my mouse. I have a non fancy scroll wheel on this mouse so i think that is to be expected.

neon12345:
> This PR is also fixing what #9975 does .
> We need to pick one approach.

The two PRs handle different code paths:

void EiScreen::onPointerScrollEvent(ei_event *event)
void EiScreen::onPointerScrollDiscreteEvent(ei_event *event)

you can combine them here if you want

sithlord48:
@neon12345  can you try this pr to see if it also fixes the issue you were attempting to fix

neon12345:
> @neon12345 can you try this pr to see if it also fixes the issue you were attempting to fix

As i understand it, these methods handle different events a device can send in any combination it likes. So there could be devices that only use one event type. I guess both at the same time would not make sense, but i can test it.

neon12345:
> but i can test it.

does not work - I guess touchpads commonly use one path and mouse scroll the other.

sithlord48 (dismissed):
This seems ok to land for me we can look into the other event in the other PR. 
@nbolton  ill leave landing to you so you can test

src/lib/platform/EiScreen.cpp
```diff
@@ -823,10 +822,28 @@ void EiScreen::onPointerScrollDiscreteEvent(ei_event *event)
...
+  }
+
+  double accX = remainder->x + dx;
+  double accY = remainder->y + dy;
+
+  // trunc toward zero, not floor: floor(-0.3) == -1 would emit a spurious click
+  auto sendX = static_cast<int32_t>(std::trunc(accX / s_scrollDelta)) * s_scrollDelta;
+  auto sendY = static_cast<int32_t>(std::trunc(accY / s_scrollDelta)) * s_scrollDelta;
+  remainder->x = accX - sendX;
+  remainder->y = accY - sendY;
```
nbolton: Now that #9975 has landed, both handlers store their leftover scroll in the same remainder, one in clicks and one in 120ths. Doesn't a wheel leftover of 15 get replayed as 15 clicks on the next touchpad scroll?

sithlord48: unsure i don't have both devices at once..

neon12345: It can clash if you use mouse scroll and touch scroll at the same time. It would make sense to align the remainder units but at the same time, how common is this use case?

sithlord48: i would think small enough where we can fix it after this lands.

neon12345: > // Values are in fractional wheel-click units (1.0 == one full 120-unit click)

The comment from my PR should be changed....

nbolton: OK lets just get that nit then land.

nbolton (dismissed):

src/lib/platform/EiScreen.cpp
sithlord48: remove d

src/lib/platform/EiScreen.cpp
```diff
@@ -28,7 +28,6 @@
 #include <unistd.h>
 #include <vector>
 
-// Values are in fractional wheel-click units (1.0 == one full 120-unit click)
```
neon12345, line 31: I changed the original comment about pixel units. It would be better to state this is used both as pixel and fractional wheel-click units from different methods.

Edit: It is not pixels but different scale now i guess.

sithlord48: updated thanks

sithlord48 (approved):

adrianinsaval:
thanks for this fix! this was driving me mad, I wasted so much time trying to fix this on my desktop and my mac playing around with scroll sensitivity

PR #9998, opened 2026-07-23
build: place mac os docs in resources/docs

sithlord48:
## Description

Move docs to mac os's app / resources folder 
## AI tools used for this PR

None
## Fixed/related issues

share in dmg


amagus:
I am still not seeing the docs on the generated dmg :(
Perhaps we would need to update the dmg script as well

sithlord48:
> I am still not seeing the docs on the generated dmg :( Perhaps we would need to update the dmg script as well

Its in the app bundle Resouces/docs

amagus (approved):
It's not necessarily easily accessible, but at least it's there. As discussed over chat, maybe the best idea for later would be to add a menu item that opens finder on that directory.

sithlord48:
> It's not necessarily easily accessible, but at least it's there. As discussed over chat, maybe the best idea for later would be to add a menu item that opens finder on that directory.

I will address this in a future PR.

sithlord48:
Closing, To make a better pr later.

Issue #9999, opened 2026-07-24
macOS: quitting the GUI can leave deskflow-core orphaned in desktop process mode


rNoz:
### Before reporting a bug, please complete the sanity checks above.

- [x] I have done the sanity checks, and my issue persists.

### Deskflow version info

Deskflow: current upstream master / local developer build (2026-07-24)
Qt: Qt 6 on macOS
System: macOS (Tahoe 26.5)
Session: Deskflow GUI running in desktop process mode

### Build types

Local developer build (built it myself)

### Deskflow configuration

- macOS machine running the Deskflow GUI as the server.
- Core process mode is `Desktop` (GUI starts `deskflow-core` as a child process).
- The problem is about application shutdown, not a particular client layout.

### What steps will reproduce the problem?

1. On macOS, configure Deskflow to run the core in desktop process mode.
2. Start the server from the GUI so a `deskflow-core` child process is running.
3. Quit the GUI through an OS/application quit path, for example Dock **Quit**, Cmd-Q, logout, or another path that reaches `[NSApp terminate:]`.
4. Check for `deskflow-core` after the GUI exits, or immediately launch Deskflow again.

## Expected result

The GUI shutdown path stops the desktop-mode `deskflow-core` synchronously before the GUI process exits. The core releases the IPC socket and the server listening port, so the next launch starts cleanly.

## Actual result

The GUI can exit while `deskflow-core` remains alive, reparented to init (`ppid=1`). The orphan keeps the IPC socket and the server port open, corrupting or blocking the next Deskflow launch.

### Log output

<details><summary><h3> Show log </h3></summary>

```ceylon
# typical external observation after quitting the GUI
# pgrep/ps still shows deskflow-core with ppid=1
# a relaunch may find the old IPC endpoint or fail to bind the server port
```

</details>

### Additional information

Root cause: on macOS, an OS-initiated quit (`[NSApp terminate:]`) can call `exit()` without unwinding the C++ stack, so `MainWindow::~MainWindow()` and its `CoreProcess::cleanup()` path do not reliably run. The existing cleanup also only stopped the desktop core when the tracked state was exactly `Started`, missing live children in other states such as `Stopping`, `RetryPending`, or an inconsistent `Stopped` state.

The desktop-mode code path and the `CoreProcess::cleanup()` condition (`isDesktop && isRunning`) both exist in clean upstream master, so this is not fork-only behavior.

sithlord48:
[...]

PR #10000, opened 2026-07-24
fix(macOS): stop desktop core during GUI quit

rNoz:
## Description

In desktop process mode on macOS, quitting the GUI can leave the `deskflow-core` child process orphaned. The orphan keeps the core IPC socket and the server listening port alive, which can corrupt or block the next launch.

This change stops the desktop core from the quit path itself and hardens `CoreProcess::cleanup()` so a live child is stopped even when the tracked state is not exactly `Started`.

### What changed / why

- Stop the core synchronously in `MainWindow::closeEvent()` before accepting a real quit.
- Also connect `QCoreApplication::aboutToQuit` to `CoreProcess::cleanup()` for non-destructor quit paths.
- Change `CoreProcess::cleanup()` to stop any live desktop child (`isRunning` or `hasLiveChild`), not only the exact `Started` state.
- Make foreground desktop shutdown bounded: ask the core to stop over IPC, then terminate, then kill if needed.
- Detach the process as a last resort so `~QProcess` can never block GUI shutdown indefinitely.
- Add a unit test suite for the desktop child shutdown behavior.

The macOS-specific part matters because an OS-initiated quit (`[NSApp terminate:]`) can call `exit()` without unwinding the C++ stack. Relying on `MainWindow::~MainWindow()` to run `CoreProcess::cleanup()` is therefore not sufficient.

## AI tools used for this PR

GPT-5.6 (Sol) / Opus 4.8 was used to help debugging and iterating on some fixes (various commits) as I was progressing testing edge-cases. I was developing/fixing/testing and required real e2e verifications, being tedious but required, as needed to act normally with the app for many times (Quit, Start, Force Quit, Start, etc). All e2e tests done by me, but capturing Verbose logs to be helpful for AIs.

## Fixed/related issues

No relevant existing upstream issue was found from read-only searches for `macos quit`, `core process orphan`, `macos quit deskflow-core`, or similar terms so I needed to create https://github.com/deskflow/deskflow/issues/9999.

## How has this been tested?

The branch adds `src/unittests/gui/core/CoreProcessTests.*`, covering:

- `stop_terminatesDesktopChild`
- `stop_sendsGracefulIpcStop`
- `stop_killsDesktopChildIgnoringTerminate`

Before submitting upstream, run the targeted test suite from the clean rebased branch, for example with the project's normal CMake/CTest workflow and a selector for `CoreProcessTests`.

I am using these changes locally (built) to be able to use  Deskflow w/o issues in my mac.

Rebased to have latest changes from upstream.

### DIFF
--- a/src/lib/deskflow/ipc/CoreIpcServer.cpp
+++ b/src/lib/deskflow/ipc/CoreIpcServer.cpp
@@ -15,7 +15,8 @@ namespace deskflow::core::ipc {
 
 static CoreIpcServer *s_instance = nullptr;
 
-CoreIpcServer::CoreIpcServer(QObject *parent) : IpcServer(parent, kCoreIpcName, QStringLiteral("core"))
+CoreIpcServer::CoreIpcServer(QObject *parent, QString socketName)
+    : IpcServer(parent, socketName.isEmpty() ? kCoreIpcName : socketName, QStringLiteral("core"))
 {
   assert(s_instance == nullptr);
   s_instance = this;
--- a/src/lib/deskflow/ipc/CoreIpcServer.h
+++ b/src/lib/deskflow/ipc/CoreIpcServer.h
@@ -20,7 +20,7 @@ class CoreIpcServer : public IpcServer
   Q_OBJECT
 
 public:
-  explicit CoreIpcServer(QObject *parent);
+  explicit CoreIpcServer(QObject *parent, QString socketName = {});
 
   static CoreIpcServer &instance();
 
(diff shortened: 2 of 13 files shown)


amagus (changes requested):
This is way too much to fix a simple problem.

I think you are trying to piggy back other unrelated changes on top of the simple issue you found.

Indeed the dtors are not called, but the fix should be as simple as possible for that problem. Adding an aboutToQuit handler that kills the core process, if alive, is more than enough.

Please simplify your PR.

sithlord48:
Replaced by #10005

PR #10002, opened 2026-07-26
Better help

sithlord48:
## Description

- Install mac os docs into resources
- Add Help-> View Help menu entry to show help 
- Make a simple help viewer to render our help via the markdown files we write.
- Remove user docs target , only docs and dev docs now
- doxygen is only used for dev docs.

Screenshot_20260726_214606

## AI tools used for this PR

None
## Fixed/related issues

User docs are included but hard to find

## How has this been tested?

Local build. 
Checked ; 
  - Flatpak
  - Arch
  - Mac os

External links will open in your browser internal help ones in the help viewer.


CMakeLists.txt
```diff
@@ -229,6 +229,7 @@ if (WIN32)
 elseif(BUILD_OSX_BUNDLE)
   set(CMAKE_INSTALL_BINDIR ${CMAKE_PROJECT_PROPER_NAME}.app/Contents/MacOS)
   set(CMAKE_INSTALL_LICENSE_DIR ${CMAKE_PROJECT_PROPER_NAME}.app/Contents/Resources)
+  set(CMAKE_INSTALL_DOCDIR ${CMAKE_INSTALL_LICENSE_DIR}/docs)
```
amagus, line 232: Is the ordering of the commits correct? I see that on the previous one you make the help menu point to a file on this folder, wouldn't that commit need to come first?

sithlord48: #9998 is needs to land first

src/lib/gui/MainWindow.cpp
```diff
@@ -1282,6 +1292,40 @@ void MainWindow::updateTimeoutDelay(int newDelay)
...
 }
 
+void MainWindow::setHelpFilePath()
+{
+  const QString appPath = QCoreApplication::applicationDirPath();
+  auto buildPath = QString("%1/../docs/HelpMain.md").arg(appPath);
+  auto installPath = QString("%1/../share/doc/%2/HelpMain.md").arg(appPath, kAppId);
+  if (deskflow::platform::isMac()) {
+    installPath = QString("%1/../Resources/docs/HelpMain.md").arg(appPath);
+    buildPath = QString("%1/../../../../docs/HelpMain.md");
```
nbolton:
```suggestion
    buildPath = QString("%1/../../../../docs/HelpMain.md").arg(appPath);
```

docs/dev/CMakeLists.txt
```diff
@@ -1,15 +1,28 @@
...
+  doxygen_add_docs(dev-docs
+    ${CMAKE_CURRENT_SOURCE_DIR}
+    ${CMAKE_SOURCE_DIR}/src
+    COMMENT "Generating developer documentation" ALL)
+else()
+  message(STATUS "Doxygen not found, skipping docs build")
+endif()
 
 # HACK Only these will show in your IDE
 target_sources(dev-docs PRIVATE
```
nbolton: Does this need to move inside the if?

nbolton (changes requested):
Since no CI runner installs Doxygen anymore, nothing ever builds the dev docs, so docs/dev/ config rot won't be caught. Was that intentional?

src/lib/gui/MainWindow.cpp
sithlord48: oh yeah that is gonna be a problem for sure!

docs/dev/CMakeLists.txt
sithlord48: No if someone has turned this on but does not have doxygen we can still show the source files, We could also block including it unless doxygen is found. I just figured this was more covenant

sithlord48:
> Since no CI runner installs Doxygen anymore, nothing ever builds the dev docs, so docs/dev/ config rot won't be caught. Was that intentional?

the arch runner still installs doxygen and this is the one that builds the dev docs.

docs/dev/CMakeLists.txt
sithlord48: Moved them in the if block

nbolton (approved):

PR #10005, opened 2026-07-26
fix(mac): properly handle system initiated quit events

### LINKED ISSUE #9999: macOS: quitting the GUI can leave deskflow-core orphaned in desktop process mode

amagus:
## Description
Intercepts and correct handle quit requests coming from the system on mac OS (Upon dock clicking, for example).

If deskflow is configured to go to the tray, it will do so even if the system requests the deskflow app to quits.
Special treatment is given for shutdown requests (which includes logoff/restarts as well).

This also ensures that the proper Qt preparation to quit runs. 

## AI tools used for this PR
Claude Code (Opus 5) on guidance and suggestions.
Code was manually written.

## Fixed/related issues
replaces: #10000
fixes: #9999 

## How has this been tested?
Tested on macOS Tahoe 26.5.2 (25F84) as server

### DIFF
--- a/src/lib/gui/OSXHelpers.mm
+++ b/src/lib/gui/OSXHelpers.mm
@@ -13,11 +13,18 @@
 #import <UserNotifications/UNNotificationContent.h>
 #import <UserNotifications/UNNotificationTrigger.h>
 #import <UserNotifications/UNUserNotificationCenter.h>
+#import <objc/runtime.h>
 
 #import <QtGlobal>
 
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
+
+namespace {
+std::function<bool()> s_shouldQuit;
+IMP s_originalShouldTerminate = nullptr;
+BOOL s_isSystemShuttingDown = NO;
+} // namespace
 
 void requestOSXNotificationPermission()
 {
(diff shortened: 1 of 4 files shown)


sithlord48 (approved):

src/lib/gui/OSXHelpers.mm
```diff
@@ -107,3 +114,44 @@ void macOSNativeHide()
...
+  Method method = class_getInstanceMethod(cls, selector);
+  if (method) {
+    s_originalShouldTerminate = method_getImplementation(method);
+    method_setImplementation(method, (IMP)deskflow_applicationShouldTerminate);
+  } else {
+    class_addMethod(cls, selector, (IMP)deskflow_applicationShouldTerminate, "l@:@");
+  }
+
+  // shutdown is also triggered for logout/restart
+  [[[NSWorkspace sharedWorkspace] notificationCenter] addObserverForName:NSWorkspaceWillPowerOffNotification
```
nbolton, line 150: Is s_isSystemShuttingDown guaranteed to be set before applicationShouldTerminate: runs? Objective-C not my strong suit.

amagus: Since system actions all run in the cocoa loop, it is indirectly "guaranteed": if we don't get the logoff request before, the GUI will refuse to be terminated and it will try again shortly (in which the notification will have arrived and now the termination will be allowed). 
I stress tested this on macOS 26 and 27 (tried logging off on all situations and it always killed deskflow as expected)

nbolton: Ah cool I was going to ask about logging off. Sounds like you did a thorough test! :)

nbolton (approved):
Haven't tested log out myself but @amagus did and that's good enough for me.

PR #9929, opened 2026-07-02
fix: Claim Wayland portal clipboard after updates

zfxkj:
## Description
Claim the portal clipboard after the Wayland input-capture clipboard cache is updated.

## AI tools used for this PR
Codex / ChatGPT was used to inspect the code path, draft the patch, and run build checks.

## Fixed/related issues
related: #9288
related: #9600
related: #8031
related: #7600

## How has this been tested?
Tested clipboard sharing with Ubuntu 26.04 LTS + KDE Plasma + Wayland and Windows 11 Pro 25H2.

Built the Linux platform target with libportal clipboard support enabled.

macOS and Windows builds were not run locally.

PR #10006, opened 2026-07-27
feat(wayland): server-side clipboard via a RemoteDesktop portal sidecar

ryanlawler:
## Description

On GNOME 50 (and any compositor that offers only the InputCapture portal at **version 1**) the Wayland **server** has no clipboard, even after #9431 landed and when built against libportal 0.10.

Root cause is structural rather than a bug in the portal clipboard code:

- `EiScreen` constructs `PortalInputCapture` for the server and `PortalRemoteDesktop` only for the client (`src/lib/platform/EiScreen.cpp`).
- At InputCapture portal **version 1**, `PortalInputCapture::initSession()` takes the legacy session path and never calls `xdp_session_request_clipboard()`.
- `xdg-desktop-portal-gnome` additionally refuses clipboard on non-RemoteDesktop sessions, so even InputCapture v2 would not be sufficient there.

Net effect on a GNOME server: on activation it logs `no current clipboard selection` → `clipboard cache empty, nothing to claim`, then broadcasts an empty clipboard to its clients on every screen crossing (overwriting theirs).

**This PR** adds `PortalServerClipboard`. When the input-capture session reports that it carries no clipboard, `EiScreen` lazily opens a second, **clipboard-only RemoteDesktop portal session** alongside input capture — the exact portal the client side already uses successfully. The session is opened `PERSISTENT`, and its restore token is stored in a new `server/xdpClipboardRestoreToken` setting (kept distinct from the input-capture token and the client token, since one machine can be both server and client). Input capture is untouched, and the sidecar is created lazily only when the input-capture session lacks clipboard — so a future portal that *does* grant an input-capture clipboard gets no second session and no second consent dialog. If the RemoteDesktop session is unavailable or denied, the KVM behaves exactly as it does today.

Scope note: this restores the server's ability to **read and share its own clipboard**. Cross-platform image/format negotiation (e.g. BMP vs PNG) is a separate, pre-existing concern tracked elsewhere and is not addressed here; text is what I verified end to end.

## AI tools used for this PR

Disclosed in full. The patch was AI-assisted, then built and verified by a human (me) on real hardware:

- **Claude (Anthropic, "Fable" / Claude 5 family)** — mapped the relevant code paths, produced and synthesized the patch across several agents, and ran an adversarial code-review pass (three independent reviewers checking: does it actually fix the read, does it break input capture, would a maintainer merge it).
- **Codex (OpenAI, GPT-5.x)** — independent second-opinion agents that analyzed the same source and produced an alternate build for cross-checking.
- **Human** — I built it against libportal 0.10 in a clean Ubuntu 26.04 container, deployed it to my own machines, and verified the full loop (see testing below). I've reviewed the diff.

Happy to answer for any part of it, and to add unit tests / adjust to your Code Style before this is considered for merge — I didn't want to guess at your test harness conventions without a maintainer signalling this is a direction you want.

## Fixed/related issues

related: #8031
related: #9929
related: #7600

## How has this been tested?

Real full-loop test on hardware (not a build-only check):

- **Environment:** Ubuntu 26.04 / GNOME Shell 50.1 / Wayland, as the Deskflow **server**; clients = a second Ubuntu 26.04 / GNOME 50 / Wayland box and a macOS (Apple Silicon) box. Built from master + this patch against a bundled libportal 0.10 (`xdp_session_request_clipboard` present).
- **Before:** across 96 activations the server logged `no current clipboard selection` every time — 0 successful local reads — and sent empty (4-byte) clipboards to clients.
- **After:** on first activation the server logs `input capture session carries no clipboard, opening a clipboard-only remote desktop session`, then `clipboard read local selection, formats: 1`. Text copied on the GNOME server now pastes on both the Linux and macOS clients; text copied on a client pastes on the GNOME server. Consent is a one-time RemoteDesktop dialog; the restore token persists it across restarts (verified `server/xdpClipboardRestoreToken` is written and re-used with no re-prompt).
- Input capture / screen switching continued to work throughout (18+ crossings logged after the change).

**Not yet tested / out of scope:** KDE Plasma as server, Windows/macOS as server, image-format round-trips, and automated unit tests (offered above).

### DIFF
--- a/src/lib/base/EventTypes.h
+++ b/src/lib/base/EventTypes.h
@@ -238,5 +238,10 @@ enum class EventTypes : uint32_t
 
   /// Stop libei
   EISessionClosed,
+
+  /// The portal session that carries input cannot also carry the clipboard, so
+  /// the screen has to open a session that can. Posted by PortalInputCapture on
+  /// its glib thread, handled by EiScreen on the event queue thread.
+  PortalClipboardUnavailable,
 };
 } // namespace deskflow
--- a/src/lib/client/ServerProxy.cpp
+++ b/src/lib/client/ServerProxy.cpp
@@ -545,14 +545,17 @@ void ServerProxy::setClipboard()
   // parse
   ClipboardID id;
   uint32_t seq;
+  const auto maximumSize = m_client->getMaximumClipboardReceiveSizeBytes();
 
-  auto r = ClipboardChunk::assemble(
-      m_stream, m_clipboardDataCached, id, seq, m_clipboardChunkState, m_client->getMaximumClipboardReceiveSizeBytes()
-  );
+  auto r = ClipboardChunk::assemble(m_stream, m_clipboardDataCached, id, seq, m_clipboardChunkState, maximumSize, true);
 
   if (r == TransferState::Started) {
     size_t size = ClipboardChunk::getExpectedSize(m_clipboardChunkState);
-    LOG_DEBUG("receiving clipboard %d size=%zu", id, size);
+    if (ClipboardChunk::isDiscarding(m_clipboardChunkState)) {
+      LOG_WARN("dropping oversize clipboard from server (size %zu > limit %zu), staying connected", size, maximumSize);
+    } else {
+      LOG_DEBUG("receiving clipboard %d size=%zu", id, size);
+    }
   } else if (r == TransferState::Finished) {
     LOG_DEBUG("received clipboard %d size=%zu", id, m_clipboardDataCached.size());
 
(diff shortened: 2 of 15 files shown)


sithlord48:
IIrc this is fixed in Gnome 51. I will tag @KoljaFrahm  and @nbolton to confirm.

KoljaFrahm:
Yes, this is not needed with Gnome 51

nbolton:
> Yes, this is not needed with Gnome 51

Should this PR be closed then?

sithlord48:
I think so we should look at this again after gnome 51 if its needed we can always bring it back

JupiterFeng:
Fedora44 user passed out in the bathroom from crying

PR #10007, opened 2026-07-28
fix(wayland): Corrects the cursor coordinates mappings at primary desktop

### LINKED ISSUE #9990: Wrong mouse coordinates when barrier edge size is different than the screen geometry on multi monitor setup.

### Deskflow version info

Deskflow: 1.26.0.325
QT: 6.11.1
System: Slackware Current
Session: KDE Plasma (Wayland)

### Build types

### Geometry

|  Client 1  |  Client 2  |  Server   |  Client 3  |  Client 4  |
|-----------|-----------|-----------|-----------|-----------|
| Plasma X11 | CDE X11 | Plasma Wayland | Windows | Plasma Wayland |
| Monitor 2 | Monitor 1 | Monitor 2 | Monitor 2 | Monitor 2 |
| Monitor 1 |           | Monitor 1 | Montior 1 | Monitor 1 |

1. I named the main monitor at each desktop as *Monitor 1*.
2. There is also a sixth desktop (*Client 5*) on top of *Client 2*, but that will make the table hard to make (and understand).
3. With the exception of *Client 3*, all desktops are Slackware Linux hosts running Slackware Current
4. All the desktops are connected by the first row and all *Monitor 2* are bigger than *Monitor 1*
5. _Client 1_ and _Client 3_ are using the Deskflow 1.26.0, the others are all with today's continuous build.

### Configuration

```
section: links
        Client_1:
                right(0,20) = Client_5(60,100)
                right(24,53) = Client_2(0,47)
        Client_2:
                right(0,94) = Server(12,63)
                left(0,47) = Client_1(24,53)
                up(0,100) = Client_5(0,100)
        Client_5:
                right(90,100) = Server(0,4)
                left(60,100) = Client_1(0,20)
                down(0,100) = Client_2(0,100)
        Server:
                left(0,4) = Client_5(90,100)
                left(12,63) = Client_2(0,94)
                right(0,50) = Client_3(41,78)
        Client_3:
                left(41,78) = Server(0,50)
                right(42,78) = Client_4(0,58)
        Client_4:
                left(0,58) = Client_3(42,78)

```

1. Configure systems to link only over part of its height
2. Bring the mouse from Primary to any Secondary
3. When the mouse goes to Secondary, all goes Ok.
4. Bring the mouse back to Primary from Secondary
5. Instead using the percentage of full height (adding Monitor 1 and Monitor 2) the mouse coordinates at Primary will be calculated counting only the Monitor 2 height at Primary, so the Cursor will be displaced.
6. Move from Secondary to Secondary don't have this problem
7. Deskflow 1.26.0 don't have this problem

### Log output

Sadly, no relevant logs, only the "leaving screen" and "entering screen"

### Additional information

- No idea of why this is happening, only know that is something between 1.26.0 and now.
- No workarounds
- I found the issue today and it's very annoying, but I will not stop using Deskflow because it
- Restart computer or deskflow doesn't helps
- Locale is "the same" on all Linux machines, the Windows host is in Portuguese

piterpunk:
Remapping the cursor coordinates to the portal edge size makes the mouse position being relative to the edge size, in a multi-monitor environment this edge can be in only one of the monitors and the cursor coordinates got wrong.

## Description

This change uses the full height and the full width of the combined monitors to set the mouse cursor position, cropped by the portal edge size.

## Fixed/related issues
fixes: #9990 

## How has this been tested?

The test was done by compiling the Deskflow tree of this PR, packaging, upgrading and restarting the Deskflow on three machines of my environment, namely: Client 1, Server and Client 4. Server is the important part as the wrong mouse position only happens at server. Client 2, Client 3 and Client 5 are kept with Deskflow 1.26.0.

After that, the test was made moving the mouse between the different desktop and screens, with the mouse showing in the correct coordinates.

### Geometry

|  Client 1  |  Client 2  |  Server   |  Client 3  |  Client 4  |
|-----------|-----------|-----------|-----------|-----------|
| Plasma X11 | CDE X11 | Plasma Wayland | Windows | Plasma Wayland |
| Monitor 2 | Monitor 1 | Monitor 2 | Monitor 2 | Monitor 2 |
| Monitor 1 |           | Monitor 1 | Montior 1 | Monitor 1 |

1. I named the main monitor at each desktop as *Monitor 1*.
2. There is also a sixth desktop (*Client 5*) on top of *Client 2*, but that will make the table hard to make (and understand).
3. With the exception of *Client 3*, all desktops are Slackware Linux hosts running Slackware Current
4. All the desktops are connected by the first row and all *Monitor 2* are bigger than *Monitor 1*
5. _Client 1_ and _Client 3_ are using the Deskflow 1.26.0, the others are all with today's continuous build.

### Configuration
```
section: links
        Client_1:
                right(0,20) = Client_5(60,100)
                right(24,53) = Client_2(0,47)
        Client_2:
                right(0,94) = Server(12,63)
                left(0,47) = Client_1(24,53)
                up(0,100) = Client_5(0,100)
        Client_5:
                right(90,100) = Server(0,4)
                left(60,100) = Client_1(0,20)
                down(0,100) = Client_2(0,100)
        Server:
                left(0,4) = Client_5(90,100)
                left(12,63) = Client_2(0,94)
                right(0,50) = Client_3(41,78)
        Client_3:
                left(41,78) = Server(0,50)
                right(42,78) = Client_4(0,58)
        Client_4:
                left(0,58) = Client_3(42,78)
```

### DIFF
--- a/src/lib/platform/PortalInputCapture.cpp
+++ b/src/lib/platform/PortalInputCapture.cpp
@@ -481,8 +481,8 @@ std::pair<double, double> PortalInputCapture::mapPortalReleasePosition(double x,
     return {x, y};
   }
 
-  auto mappedX = scaleCoordinateBetweenRanges(x, screenLeft, screenRight, portalBounds.left, portalBounds.right);
-  auto mappedY = scaleCoordinateBetweenRanges(y, screenTop, screenBottom, portalBounds.top, portalBounds.bottom);
+  auto mappedX = static_cast<std::int32_t>(std::lround(x));
+  auto mappedY = static_cast<std::int32_t>(std::lround(y));
 
   if (BarrierInfo releaseBarrier;
       getClosestReleaseBarrier(x, y, screenLeft, screenTop, screenRight, screenBottom, portalBounds, releaseBarrier)) {


sithlord48:
quick smoke test seems to work the same for "normal" displays i.e no percent based screens . I have to try to mock this setup to do a proper test

piterpunk:
Minimal_Setup_Deskflow

This is the minimal setup to test. The blue machine is the Primary, running Wayland and the green one is the client. Because the two monitors of the blue machine have different sizes, only one of those touch the edge of the rectangular logical display area, and only this one have a barrier set up (the red line).

sithlord48:
Thank you that visual helps ALOT! ill try to test that when i get time.

sithlord48 (commented):
I have set up a test like you have said above

Screenshot_20260729_205826

and with or without this patch i am still only able to have an effective barrier on the RED line

i have tried w/ the follow links 

```
server: 
   right = client
client: 
  left = server
```

and 

```
server: 
   right(0,60) = client
client: 
  left = server(0,60)
```

and both result in the barrier being at the red line with or without this PR applied. (server was updated client was not)

piterpunk:
Yeah, without the patch (and with the patch) the barrier itself doesn´t change. This is what the e8e9d1c193c5050720cf62a2389876c1786762f9 does and it's a correct behavior, the previously the barrier have the full size of its screen side, not only the one at the edge.

What changes is the mapping of cursor position. I made two videos of the behavior with and without (1.26.0.325) the patch, in both videos the Primary is at right and the `links` section is:
```
section: links
        cameron.mylab:
                left(12,63) = melchior.mylab(0,94)
        melchior.mylab:
                right(0,94) = cameron.mylab(12,63)
end
```
Yes, these are the names of my machines, they appears in the video, so there isn't any reason to mask them now.

Cameron, at right, is the Primary Desktop with two monitors, and Melchior, at left, is the Secondary with only one.

https://github.com/user-attachments/assets/4b7f1931-68f4-4716-80ed-11761dd95ce6

This is what happens using 1.26.0.325, you can see that the cursor goes to almost the same height between Primary to Secondary, but it jumps half size up from Secondary to Primary. That is because it's applying the `(12,63)` relative to the barrier itself, not the full height.

https://github.com/user-attachments/assets/faf8bef9-d6b6-474d-ab35-1591bc00d832

This one is with the patch applied, the same behaviour of 1.26.0, the cursor goes forth and back at almost the same height position.

sithlord48 (dismissed):
Ok i have managed to get a set up to test this correctly and its workings correctly now.

I tested by moving my top monitor right and can see clearly on the current master this is moving to a much lower return location than expected while it works correctly if the right border is lined up . This PR fixes that so that the return is always the correctly vertical return space (and not space - the height of that non border screen)

src/lib/platform/PortalInputCapture.cpp
```diff
@@ -469,6 +469,8 @@ std::pair<double, double> PortalInputCapture::mapPortalReleasePosition(double x,
   std::int32_t screenY;
   std::int32_t screenW;
   std::int32_t screenH;
+  std::int32_t mappedX;
+  std::int32_t mappedY;
```
nbolton: Lets init these to avoid undefined behavior. I have a feeling not 0 though? Something like rounded x/y. Also we should probably just call `static_cast<int32_t>(` once.

nbolton: Not saying this the right fix but would something like this be better?

```patch
diff --git a/src/lib/platform/PortalInputCapture.cpp b/src/lib/platform/PortalInputCapture.cpp
index a62f57d481..45115e8ecb 100644
--- a/src/lib/platform/PortalInputCapture.cpp
+++ b/src/lib/platform/PortalInputCapture.cpp
@@ -469,8 +469,6 @@ std::pair<double, double> PortalInputCapture::mapPortalReleasePosition(double x,
   std::int32_t screenY;
   std::int32_t screenW;
   std::int32_t screenH;
-  std::int32_t mappedX;
-  std::int32_t mappedY;
   m_screen->getShape(screenX, screenY, screenW, screenH);
 
   const auto screenLeft = screenX;
@@ -483,6 +481,9 @@ std::pair<double, double> PortalInputCapture::mapPortalReleasePosition(double x,
     return {x, y};
   }
 
+  auto mappedX = static_cast<std::int32_t>(std::lround(x));
+  auto mappedY = static_cast<std::int32_t>(std::lround(y));
+
   if (BarrierInfo releaseBarrier;
       getClosestReleaseBarrier(x, y, screenLeft, screenTop, screenRight, screenBottom, portalBounds, releaseBarrier)) {
     const Bounds releaseBounds = {
@@ -494,21 +495,23 @@ std::pair<double, double> PortalInputCapture::mapPortalReleasePosition(double x,
     switch (releaseBarrier.side) {
     case Left:
       mappedX = std::min(releaseBounds.left + jumpZoneSize, releaseBounds.right);
-      mappedY = std::clamp(static_cast<int32_t>(y), releaseBounds.top, releaseBounds.bottom);
+      mappedY = std::clamp(mappedY, releaseBounds.top, releaseBounds.bottom);
       break;
     case Right:
       mappedX = std::max(releaseBounds.right - jumpZoneSize, releaseBounds.left);
-      mappedY = std::clamp(static_cast<int32_t>(y), releaseBounds.top, releaseBounds.bottom);
+      mappedY = std::clamp(mappedY, releaseBounds.top, releaseBounds.bottom);
       break;
     case Top:
-      mappedX = std::clamp(static_cast<int32_t>(x), releaseBounds.left, releaseBounds.right);
+      mappedX = std::clamp(mappedX, releaseBounds.left, releaseBounds.right);
       mappedY = std::min(releaseBounds.top + jumpZoneSize, releaseBounds.bottom);
       break;
     case Bottom:
-      mappedX = std::clamp(static_cast<int32_t>(x), releaseBounds.left, releaseBounds.right);
+      mappedX = std::clamp(mappedX, releaseBounds.left, releaseBounds.right);
       mappedY = std::max(releaseBounds.bottom - jumpZoneSize, releaseBounds.top);
       break;
     }
+  } else {
+    LOG_DEBUG("no release barrier matched, releasing at the unmapped position");
   }
 
   return {mappedX, mappedY};
```

Added in that `LOG_DEBUG` at the end to turn up anything we might not expect.

nbolton: Or we could set the value in the else + default case... but that feels less safe.

nbolton (changes requested):

src/lib/platform/PortalInputCapture.cpp
piterpunk: The changes looks Ok for me, I don't have knowledge of C++ best practices, so I am happy to be corrected. I will test it here and then update the PR with the proposed changes.

sithlord48: I made these changes .

sithlord48 (approved):
:shipit:

nbolton:
> The changes looks Ok for me, I don't have knowledge of C++ best practices, so I am happy to be corrected. I will test it here and then update the PR with the proposed changes.

@piterpunk Could you confirm if the bug is still fixed after @sithlord48's improvements?

nbolton (approved):

PR #10009, opened 2026-07-28
fix(macos): null-check the input source languages before counting them

### LINKED ISSUE #10008: macOS: getKeyboardLayoutList() dereferences a NULL CFArrayRef before its own guard

### Deskflow version info

Deskflow: current upstream master / local developer build
Qt: Qt 6 on macOS
System: macOS server (tahoe)
Session: Deskflow GUI running in desktop process mode

### Build types

### Deskflow configuration

- macOS machine running as the server, single Linux (Arch, Wayland) client.
- Several keyboard input sources installed, including layouts that expose no languages property.
- Not layout-specific: the defect is in how the array pointer is tested.

1. On macOS, have an input source installed whose `kTISPropertyInputSourceLanguages` property is absent.
2. Start Deskflow as the server and let a client connect, so `ClientProxy1_8` calls `synchronizeLanguages()` → `AppUtil::instance().getKeyboardLayoutList()`.
3. The process dereferences `NULL` inside `CFArrayGetCount()`.

### Log output

It is by inspection.

</details>

### Additional information

- Cause: operand order in the loop condition. Testing the pointer first (`layoutLanguages && index < CFArrayGetCount(layoutLanguages)`) restores the intent with no behaviour change when the array is present.
- Distinct from #9924, which is the `dispatch_assert_queue` SIGTRAP in the same function. This one is the NULL pointer, and it is still there.

rNoz:
## Description

`layoutLanguages` comes from `TISGetInputSourceProperty()`, which returns `NULL` for an input source that carries no `kTISPropertyInputSourceLanguages` value. In `getCurrentLanguageCode()` it also stays `NULL` when `TISCopyCurrentKeyboardInputSource()` returns nothing, because the assignment sits behind `if (source)`.

Both loops evaluate `CFArrayGetCount(layoutLanguages)` *before* the `&& layoutLanguages` guard in the same condition, so the guard is dead code and `CFArrayGetCount()` dereferences the null pointer first. This swaps the operands at both sites. No behaviour change when the array is present.

## AI tools used for this PR

While working on other feature for my system, found this issue upstream, and Opus helped debugging and wrapping this quick side PR.

## Fixed/related issues

Fixes https://github.com/deskflow/deskflow/issues/10008

## How has this been tested?

- Built `deskflow-core` from this branch on macOS (Apple Silicon, Qt 6).
- Ran the Mac as the server for a Linux client and confirmed layout synchronisation on client connect still reports the expected languages, i.e. the loops still iterate normally when the array is non-null.
- The null path is the one that used to crash; it now exits the loop immediately.

### DIFF
--- a/src/lib/deskflow/unix/AppUtilUnix.cpp
+++ b/src/lib/deskflow/unix/AppUtilUnix.cpp
@@ -77,7 +77,7 @@ std::vector<std::string> AppUtilUnix::getKeyboardLayoutList()
       layoutLanguages = (CFArrayRef)TISGetInputSourceProperty(keyboardLayout, kTISPropertyInputSourceLanguages);
     }
     char temporaryCString[128] = {0};
-    for (CFIndex index = 0; index < CFArrayGetCount(layoutLanguages) && layoutLanguages; index++) {
+    for (CFIndex index = 0; layoutLanguages && index < CFArrayGetCount(layoutLanguages); index++) {
       auto languageCode = (CFStringRef)CFArrayGetValueAtIndex(layoutLanguages, index);
       if (!languageCode || !CFStringGetCString(languageCode, temporaryCString, 128, kCFStringEncodingUTF8)) {
         continue;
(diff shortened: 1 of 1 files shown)


sithlord48 (approved):
Looks ok, Has my approval. Will leaving for @amagus to test and @nbolton  to land

nbolton (approved):

Issue #9984, opened 2026-07-19
Chinese IME intermittently stops composing on macOS client after Deskflow screen switches


yi-of-heng:
### Before reporting a bug, please complete the sanity checks above.

- [x] I have done the sanity checks, and my issue persists.

### Deskflow version info

Client:
Deskflow: 1.26.0
Qt: 6.10.1
System: macOS 15.7.7 (24G720), arm64
Session: Aqua

Server:
Deskflow: 1.26.0
System: Windows 10

### Build types

Official release (installer/package from GitHub releases), Community package (apt, dnf, pacman/aur, brew, etc.)

### Deskflow configuration

- Windows 10 is the Deskflow server.
- macOS 15.7.7 is the Deskflow client.
- The keyboard and mouse are physically connected to Windows and control the Mac through Deskflow.
- The problem occurs across multiple applications and both web and native text controls.

### What steps will reproduce the problem?

The issue is intermittent, but it has occurred repeatedly with this setup.

1. Start Deskflow 1.26.0 on the Windows server and macOS client.
2. Move the pointer from Windows onto the macOS client.
3. Select a Chinese input method on macOS.
4. Type in text fields and switch between applications or text controls.
5. After some screen/focus switches, one text field may stop accepting Chinese composition even though macOS still shows the Chinese input method as selected. Another text field may continue working.
6. Click the macOS input source icon in the menu bar and select the input method again.
7. The previously broken text field immediately starts accepting Chinese composition again.

Actual behavior:
The selected Chinese input method intermittently stops composing in the focused text control. Depending on the occurrence, only Latin characters are produced or Chinese composition does not begin. This reproduces with Apple Simplified Chinese Pinyin, Sogou Input Method, and WeChat Input Method.

Expected behavior:
Chinese IME composition should remain active after the Deskflow pointer enters the macOS client and while switching applications or text controls.

### Log output

<details><summary><h3>Show sanitized log</h3></summary>

```text
deskflow-core: Loading Preferences From User CFPrefsD
WindowServer: HW Cursor Disabled: deactivating and swap completed
```

</details>

### Additional information

Diagnostic observations:

- macOS still reported com.apple.inputmethod.SCIM.ITABC as the selected input source during the failure.
- SCIM, InputMethodKit, and TextInput services were alive and had not crashed.
- In one normal two-minute sample, unified logging recorded 97 'Setting marked text' operations. During the incident window it recorded only 4, while Cursor UI/input-mode-switcher activity increased.
[...]

PR #10010, opened 2026-07-28
fix(mac): invalidate event tap CFMachPort on disable

MrMage:
## Description

Fixes a macOS resource leak in `OSXScreen::disable()`: the CGEventTap teardown does `CGEventTapEnable(false)` + `CFRelease` but never calls `CFMachPortInvalidate`. CoreGraphics holds internal references to the `CFMachPort` returned by `CGEventTapCreate` (observed retain count 4 immediately after creation), so `CFRelease` alone never deallocates it — **the window server keeps the (disabled) event-tap registration until the process exits**. Every screen `enable()`/`disable()` cycle — i.e. every connect, disconnect, and reconnect (network blips, sleep/wake, config reloads) — therefore leaks one stale registration.

Accumulated stale registrations measurably degrade **system-wide** input latency (the window server's per-event tap-table processing scales with table size): on a test machine, ~1,000 stale disabled registrations made all input visibly laggy within seconds, and releasing them restored responsiveness immediately. Since Deskflow machines typically run 24/7 with periodic reconnects, the table grows without bound until the process restarts. This may contribute to otherwise-unexplained "macOS gets sluggish over time" reports.

The same bug class was recently root-caused and fixed in Karabiner-Elements ([pqrs-org/Karabiner-Elements#4505](https://github.com/pqrs-org/Karabiner-Elements/pull/4505) — includes a demonstration video and standalone repro scripts: `taplist.swift` enumerates the window server tap table via `CGGetEventTapList`, `bloatgen.swift` demonstrates the latency impact), Mos ([Caldis/Mos#999](https://github.com/Caldis/Mos/pull/999) — with a red/green regression test of this exact fix pattern), pynput (PR pending), and Easydict ([tisfeng/Easydict#1170](https://github.com/tisfeng/Easydict/issues/1170), maintainer-confirmed). Long-lived tap-heavy projects (Hammerspoon, LinearMouse, libuiohook) all invalidate in their teardown paths.

The fix is one call inserted before the existing `CFRelease` in `OSXScreen::disable()`, plus an explanatory comment. No behavior change on the event path — the tap is already disabled at that point.

Note: input-leap has the identical inherited code; a matching PR is being submitted there.

## AI tools used for this PR

Claude Code (model: Claude Fable 5, `claude-fable-5`) was used for the leak analysis, the mechanism verification scripts, and drafting this patch and description. All measurements are from a real machine (macOS 26.5, Apple Silicon) and the final diff was human-reviewed.

## Fixed/related issues

No tracked issue matches this leak directly. Possibly related context: #9984 concerns the event tap being disabled after wake — a different tap issue, but the same subsystem; and the leak fixed here may contribute to gradual "macOS gets sluggish over long uptimes" reports that never got far enough to file.

## How has this been tested?

- Mechanism verified standalone on macOS 26.5 (Apple Silicon): a tap torn down with `CGEventTapEnable(false)` + `CFRelease` (no invalidate) leaves its mach receive right alive and its `CGGetEventTapList` entry present indefinitely; adding `CFMachPortInvalidate` reaps the entry immediately. Repeating the no-invalidate teardown grows the process's tap-table entry count by exactly 1 per cycle.
- The patched code path is structurally identical to the verified clean variant (and to the fixes merged/filed in the sibling projects above).
- Not built locally (no Qt toolchain on the test machine) — the change is a single CoreFoundation call adjacent to the existing `CFRelease`, using an already-included API family.
- Suggested verification on a built binary: connect/disconnect a client a few times, then enumerate `CGGetEventTapList` — the Deskflow process's disabled-entry count should stay at 0 instead of growing by 1 per cycle.

### DIFF
--- a/src/lib/platform/OSXScreen.mm
+++ b/src/lib/platform/OSXScreen.mm
@@ -768,6 +768,7 @@ void OSXScreen::disable()
 
   if (m_eventTapPort) {
     CGEventTapEnable(m_eventTapPort, false);
+    CFMachPortInvalidate(m_eventTapPort);
     CFRelease(m_eventTapPort);
     m_eventTapPort = nullptr;
   }


sithlord48:
wow so much comments for this can we simplify those ?

MrMage:
Done — trimmed to two lines; the full mechanism write-up stays in the PR description.

src/lib/platform/OSXScreen.mm
```diff
@@ -768,6 +768,9 @@ static inline bool needsEventNumber()
 
   if (m_eventTapPort) {
     CGEventTapEnable(m_eventTapPort, false);
+    // CoreGraphics retains the CFMachPort internally, so CFRelease alone never
```
amagus: no need for these comments, please remove

amagus (changes requested):

amagus:
> Done — trimmed to two lines; the full mechanism write-up stays in the PR description.

I think no comment is needed at all. The function being called is clear. 
A git blame can be used to understand reasoning if really needed.

MrMage:
Fair enough. Removed.

amagus:
please modify the commit message and PR title.
It should start with "fix:" or "fix(mac):"

MrMage:
Done.

sithlord48 (approved):
OK with me

MrMage:
Anything left to do before this can be merged?

amagus:
@MrMage It's pending stress testing this change as client on my side.

sithlord48:
@MrMage  Never press the update button always use the rebase option

amagus (approved):
Tested it on macOS Tahoe and Golden Gate, as server and client, worked as intented.

Issue #8857, opened 2025-08-14
Numpad not working on Wayland


Astorsoft:
### Project

Deskflow

### Deskflow version info

Deskflow: 1.23.0.0 (56f7a0d7)
Qt: 6.9.0
System: KDE Flatpak runtime (Bazzite)
Session: KDE (wayland)

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [ ] Windows
- [x] macOS
- [ ] Linux (X11)
- [x] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [x] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [ ] I am not using Wayland on Linux

### Signing on macOS

- [x] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### Continuous build

- [x] I have tried the latest continuous build and the issue persists
- [ ] I am unable to try the latest continuous build

### OS versions/distros

Hi, while there are other issues raised regarding special characters or external usb pads I couldn't find anything related explicitely to supporting the builtin numpad of a keyboard.

Using the numpad on the host (bazzite) works just fine but if I try to press numpad keys on my mac, they behave like arrow keys (as if numlock was off) and I get the following console error on the server (bazzite): WARNING: failed to open x11 default display

### Deskflow configuration

- Server: Bazzite Nvidia edition, KDE
- Bazzite has one monitor on the right
- CLient: MacOS Sequoia 15.6
- MacOS has 3 monitors on the left

### What steps will reproduce the problem?

1. Start Deskflow on server (Bazzite) and client (MacOS)
2. Try to use numpad keys on MacOS

### Log output

```shell
[2025-08-14T12:16:07] INFO: starting core server process (desktop mode)
[2025-08-14T12:16:07] INFO: core config file: /home/bazzite/.var/app/org.deskflow.deskflow/config/Deskflow/deskflow-server.conf
[2025-08-14T12:16:07] INFO: running command: /app/bin/deskflow-server -f --debug INFO --name bazzite --address :24800 -c /home/bazzite/.var/app/org.deskflow.deskflow/config/Deskflow/deskflow-server.conf
[2025-08-14T12:16:07] INFO: using ei screen for wayland
[2025-08-14T12:16:07] WARNING: can't open xkb display during reading languages
[2025-08-14T12:16:07] NOTE: clipboard sharing is disabled
[...]

Issue #8437, opened 2025-04-02
Modifier states are not synced to client when changed on Wayland server


gabeklavans:
### Project

Deskflow

### Deskflow version info

Deskflow: 1.20.0.0 (3ef5793b)
Qt: 6.8.2
System: KDE Flatpak runtime
Session: KDE (wayland)

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [x] Windows
- [x] macOS
- [ ] Linux (X11)
- [x] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [x] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [ ] I am not using Wayland on Linux

### Signing on macOS

- [x] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### Continuous build

- [ ] I have tried the latest continuous build and the issue persists
- [ ] I am unable to try the latest continuous build

### OS versions/distros

- Windows 11
- macOS 15
- Fedora 41

### Deskflow configuration

- Fedora 41 server, 2 monitors horizontal
- macOS 15 client to the left
- Windows 11 to the left of macOS 15

### What steps will reproduce the problem?

1. Start deskflow
2. Toggle numlock "on" while on server
3. Move to client
4. numlock is still off on client
5. toggle numlock while on client
6. numlock is now "on" on client, but "off" again on server

### Log output

```shell

```

### Additional information

As of 3/1/26, additionally, when moving from the primary screen back to a secondary screen, no matter the previous toggled modifier state, all states will be back to toggled "off". This is technically due to a different part of the code, but will be fixed by the same solution.

HalJordan:
I use X, since I'm annoyed as fuck at it asking for permission to run EVERY TIME IT'S STARTED on wayland, and have been having this same issue.

nbolton:
> I use X, since I'm annoyed as fuck at it asking for permission to run EVERY TIME IT'S STARTED on wayland, and have been having this same issue.

Yes, it's quite annoying, please watch this other issue:
- #8032

sithlord48:
Since this can happen with any modifier i have updated the title to reflect this.

xobs:
[...]

PR #10011, opened 2026-07-28
fix: preserve locked modifiers in Wayland key state

cjones26:
## Description

When Deskflow runs as the server on KDE Plasma Wayland, the libei backend rebuilds its XKB state when moving between screens. That clears stale pressed modifiers, which is useful for keys like Shift.

The same rebuild also clears locked modifiers. NumLock is locked state, so dropping it here can leave the backend reporting the wrong modifier state after a screen transition.

This patch keeps the existing stale-modifier cleanup, but preserves the locked modifier and locked layout masks across the XKB state rebuild. `pollActiveModifiers()` also reads locked modifiers so NumLock, CapsLock, and ScrollLock are reported from the state they actually live in.

For locked modifiers, Mod2 is mapped to NumLock. That matches the default evdev keymap used by this path and keeps the change inside the Linux Wayland/libei backend.

This does not implement full server-to-client lock-state synchronization. It fixes the smaller state-preservation bug in `EiKeyState`.

## AI tools used for this PR

I used Codex GPT-5 while tracing the Wayland/libei key state path, preparing the patch, and checking the test coverage.

## Related issues

Related to #8437
Related to #8857

## How has this been tested?

I tested this on KDE Plasma Wayland on x86_64 with Deskflow running as the Linux server and a macOS client. Before the patch, switching screens caused NumLock on the Mac side to stop matching the Linux server. With the patched Flatpak, NumLock stayed enabled across screen switches and the numpad kept working.

I also added EiKeyStateTests coverage for the state bug. The test presses Shift, toggles NumLock, calls `clearStaleModifiers()`, then checks that Shift is cleared and NumLock is still reported.

Validation passed on branch `agent/fix-wayland-locked-modifiers-validation` in run 30393192958. That run built the core target, built and ran EiKeyStateTests, built the Flatpak, and validated the bundle.

### DIFF
--- a/src/lib/platform/EiKeyState.cpp
+++ b/src/lib/platform/EiKeyState.cpp
@@ -97,8 +97,9 @@ bool EiKeyState::fakeCtrlAltDel()
 
 KeyModifierMask EiKeyState::pollActiveModifiers() const
 {
-  const auto xkbMask = xkb_state_serialize_mods(m_xkbState, XKB_STATE_MODS_EFFECTIVE);
-  return convertModMask(xkbMask);
+  const auto effectiveMask = xkb_state_serialize_mods(m_xkbState, XKB_STATE_MODS_EFFECTIVE);
+  const auto lockedMask = xkb_state_serialize_mods(m_xkbState, XKB_STATE_MODS_LOCKED);
+  return convertModMask(effectiveMask) | convertModMask(lockedMask, true);
 }
 
 std::int32_t EiKeyState::pollActiveGroup() const
@@ -112,7 +113,7 @@ void EiKeyState::pollPressedKeys(KeyButtonSet &) const
   return;
 }
 
-std::uint32_t EiKeyState::convertModMask(xkb_mod_mask_t xkbModMaskIn) const
+std::uint32_t EiKeyState::convertModMask(xkb_mod_mask_t xkbModMaskIn, bool mapMod2ToNumLock) const
 {
   // This is our own modifier mask, not xkb's.
   std::uint32_t modMaskOut = 0;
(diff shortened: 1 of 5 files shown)


sithlord48:
When ready to be reviewed please mark it ready for review we don't look at drafts

cjones26:
@sithlord48 seems to be working now, not a C++ engineer by trade so please let me know if anything looks off!  Thank you!

sithlord48:
Upon a quick test (on my server) this changes no behavior at all. Is this a client mode only fix ?

cjones26:
> Upon a quick test (on my server) this changes no behavior at all. Is this a client mode only fix ?

Previously, every time I switched from my host (KDE Neon) -> Mac, numlock state got lost.  With this patch I was able to resolve that and maintain numlock across the client as well.

sithlord48:
I am not seeing this w/ Plasma (6.7.3, wayland)  -> x11

cjones26:
OK.  

> I am not seeing this w/ Plasma (6.7.3, wayland) -> x11

The client is a Mac.

sithlord48:
I see no change with this pr or with out it
wayland server -> (mac, wayland , x11) client, numlock does not follow the servers state but must be toggled on the client

cjones26:
My bad, @sithlord48 , this PR doesn't (fully) fix the NumLock sync behavior from #8437 / #8857, it's a smaller fix...

`clearStaleModifiers()` is supposed to clear stale pressed modifiers after moving between screens (which makes sense for Shift), but the current code rebuilds the whole XKB state, which also drops locked modifiers, and I don't think those should be cleared by the cleanup path.

The test is trying to assert that:

  - press Shift
  - toggle NumLock
  - call `clearStaleModifiers()`
  - Shift should be gone
  - NumLock should still be reported

Do you think that's the correct behavior to you?  I think it at least my help with part of the linked bugs.

sithlord48:
I would guess it should be the same for caps lock here also ?

cjones26:
Yeah, it should apply to locked modifiers in general, as the patch saves the locked modifier mask before rebuilding the XKB state then restores it afterward:

https://github.com/deskflow/deskflow/pull/10011/changes#diff-3f5cc69b29d36a312a61ecb68d3c7092cf2b1d5baff7a90bf1c3b94c136eca26R350

sithlord48:
@cjones26  We do not allow merges DO not use the update button only rebase

nbolton (approved):

tinohendricks:
Loving this, but it kills clipboard sync KDE <-> macOS for me

cjones26:
> Loving this, but it kills clipboard sync KDE <-> macOS for me

What build were you running before this?

sithlord48:
@tinohendricks Nothing in this pr changes the clipboard. But if your coming from 1.26.0 and were using wl-clipboard backend for the clipboard that has been removed. You will need to use Plasma 6.7.2+ and libportal 0.10.0+ to for clipboard support via the portals

tinohendricks:
> @tinohendricks Nothing in this pr changes the clipboard. But if your coming from 1.26.0 and were using wl-clipboard backend for the clipboard that has been removed. You will need to use Plasma 6.7.2+ and libportal 0.10.0+ to for clipboard support via the portals

YES!!! After complete upgrade to debian forky works like charm! Thank you so much!

Issue #8597, opened 2025-05-15
FATAL: an error occurred: string too long


Avaviel:
### Project

Deskflow

### Deskflow version info

Server:
Deskflow: 1.21.2.0 (daa09546)
Qt: 6.9.0
System: Windows 10 Version 22H2

Mac:
Deskflow: 1.21.2.0 (daa09546)
Qt: 6.9.0
System: macOS Sequoia (15.4)

Error:
[2025-05-15T21:38:42] INFO: accepted secure socket
[2025-05-15T21:38:42] INFO: network encryption protocol: TLSv1.3
[2025-05-15T21:38:42] NOTE: accepted client connection
[2025-05-15T21:38:42] INFO: local languages: en
[2025-05-15T21:38:42] NOTE: client "Pauls-Mac-mini.local" has connected
[2025-05-15T21:38:50] INFO: switch from "DESKTOP-NCDPHG2" to "Pauls-Mac-mini.local" at 0,339
[2025-05-15T21:38:50] INFO: leaving screen
[2025-05-15T21:38:50] INFO: bitmap: 1024x-1024 32
[2025-05-15T21:38:50] INFO: convert image from: depth=32 comp=3
[2025-05-15T21:38:50] FATAL: an error occurred: string too long
[2025-05-15T21:38:50] WARNING: detected application not running, pid=27064
[2025-05-15T21:38:51] NOTE: started server, waiting for clients
[2025-05-15T21:38:52] NOTE: peer fingerprint: (SHA1) 
[2025-05-15T21:38:52] INFO: accepted secure socket
[2025-05-15T21:38:52] INFO: network encryption protocol: TLSv1.3
[2025-05-15T21:38:52] NOTE: accepted client connection
[2025-05-15T21:38:52] INFO: local languages: en
[2025-05-15T21:38:52] NOTE: client "Pauls-Mac-mini.local" has connected
[2025-05-15T21:38:53] INFO: switch from "DESKTOP-NCDPHG2" to "Pauls-Mac-mini.local" at 0,262
[2025-05-15T21:38:53] INFO: leaving screen
[2025-05-15T21:38:53] INFO: bitmap: 1024x-1024 32
[2025-05-15T21:38:53] INFO: convert image from: depth=32 comp=3
[2025-05-15T21:38:53] FATAL: an error occurred: string too long

Made screenshot with Greenshot, tried to move to mac, and mouse reset to middle of screen. 

FIX:
Cleared the clipboard by copying text.

[...]

PR #10023, opened 2026-07-31
fix: handle top-down DIB images in clipboard bitmap converter

### LINKED ISSUE #9869: Windows server: deskflow-core crashes (std::length_error 'string too long') converting a top-down DIB (negative biHeight) clipboard image on screen switch

mmustafasenoglu:
## Description
A `BITMAPINFOHEADER` with negative `biHeight` (top-down DIB, legal per spec — produced by macOS/Retina clipboards and many screenshot tools) caused `std::length_error` crash in `toIClipboard()`. The signed product `4 * w * h` went negative and wrapped to a huge `size_t` in `string::append`, exceeding `max_size()`.

Changes in `MSWindowsClipboardBitmapConverter.cpp`:
- Compute `absHeight = abs(h)` and use it for the `SetDIBitsToDevice` scanline count and the pixel buffer size
- Use `static_cast<size_t>` arithmetic to prevent signed-to-unsigned overflow in the append call
- Add null check for `CreateDIBSection` to prevent a secondary crash if bitmap allocation fails

## AI tools used for this PR
opencode/mimo-v2.5 was used to help draft the code change. I reviewed and verified the logic myself.

## Fixed/related issues
fixes: #9869
Also related: #7329, #8597, #8625 (all reported the same `string too long` crash with negative biHeight)

## How has this been tested?
This is a Windows-only code path (`MSWindowsClipboardBitmapConverter`). I do not have a Windows + MSVC build environment available to compile/test locally. The fix is straightforward arithmetic (`abs()`) and the root cause is well-understood from the crash logs and DIB spec. Happy to iterate if a maintainer can test on Windows.

### DIFF
--- a/src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
+++ b/src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
@@ -1,6 +1,7 @@
 /*
  * Deskflow -- mouse and keyboard sharing utility
  * SPDX-FileCopyrightText: (C) 2012 - 2016 Synergy App Ltd
+ * SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
  * SPDX-FileCopyrightText: (C) 2004 Chris Schoeneman
  * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
  */
@@ -72,6 +73,7 @@ std::string MSWindowsClipboardBitmapConverter::toIClipboard(HANDLE data) const
   BITMAPINFOHEADER info;
   LONG w = bitmap->bmiHeader.biWidth;
   LONG h = bitmap->bmiHeader.biHeight;
+  const LONG absHeight = (h < 0) ? -h : h;
   info.biSize = sizeof(BITMAPINFOHEADER);
   info.biWidth = w;
   info.biHeight = h;
@@ -85,6 +87,12 @@ std::string MSWindowsClipboardBitmapConverter::toIClipboard(HANDLE data) const
   info.biClrImportant = 0;
   HDC dc = GetDC(nullptr);
   HBITMAP dst = CreateDIBSection(dc, (BITMAPINFO *)&info, DIB_RGB_COLORS, &raw, nullptr, 0);
+  if (dst == nullptr || raw == nullptr) {
+    LOG_WARN("failed to allocate destination bitmap for clipboard image");
+    ReleaseDC(nullptr, dc);
+    GlobalUnlock(data);
+    return std::string();
+  }
 
   // find the start of the pixel data
   const char *srcBits = (const char *)bitmap + bitmap->bmiHeader.biSize;
@@ -103,14 +111,14 @@ std::string MSWindowsClipboardBitmapConverter::toIClipboard(HANDLE data) const
   // copy source image to destination image
   HDC dstDC = CreateCompatibleDC(dc);
   HGDIOBJ oldBitmap = SelectObject(dstDC, dst);
-  SetDIBitsToDevice(dstDC, 0, 0, w, h, 0, 0, 0, h, srcBits, bitmap, DIB_RGB_COLORS);
+  SetDIBitsToDevice(dstDC, 0, 0, w, absHeight, 0, 0, 0, absHeight, srcBits, bitmap, DIB_RGB_COLORS);
   SelectObject(dstDC, oldBitmap);
   DeleteDC(dstDC);
   GdiFlush();
 
   // extract data
   std::string image((const char *)&info, info.biSize);
-  image.append((const char *)raw, 4 * w * h);
+  image.append((const char *)raw, static_cast<size_t>(4) * static_cast<size_t>(w) * static_cast<size_t>(absHeight));
 
   // clean up GDI
   DeleteObject(dst);


sithlord48:
Replaces #9870

nbolton (approved):

PR #10027, opened 2026-08-02
Simplify our inner linkage

sithlord48:
## Description

 We have a bunch of unneeded links in our Cmake this will aim to remove those
 We should look to move some files around as this would allow us to further simplify the links.
## AI tools used for this PR

  None
## Fixed/related issues

  Clean up the code base even further, and help to place requirements into their correct places in to make our lib more modular
## How has this been tested?

CI Builds

### DIFF
--- a/src/lib/gui/CMakeLists.txt
+++ b/src/lib/gui/CMakeLists.txt
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: (C) 2024 Deskflow Developers
+# SPDX-FileCopyrightText: (C) 2024 - 2026 Deskflow Developers
 # SPDX-FileCopyrightText: (C) 2024 Symless Ltd
 # SPDX-License-Identifier: MIT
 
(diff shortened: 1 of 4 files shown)


src/lib/gui/CMakeLists.txt
```diff
@@ -122,19 +122,14 @@ add_library(${target} STATIC
...
-if(WIN32)
-  target_link_libraries(
-    ${target}
-    base
-    arch
-    net
-  )
+if(WIN32 OR APPLE)
+  target_link_libraries(${target} PRIVATE net)
 endif()
```
nbolton: Seems GUI uses `net` on every platform, e.g. 
https://github.com/deskflow/deskflow/blob/fe29194b64277c425af13bcf67bc1d24e6d30c95/src/lib/gui/TlsUtility.cpp#L11

nbolton (approved):

PR #10032, opened 2026-08-06
Move server external config to advanced tab

sithlord48:
## Description

  - Moves the options for external config file to the server dialog's advanced tab 
     - Retain the note as a tooltip for the section. 
  - Removes the Config file tab
  - Remove the forced setting of servers tab widget to 0 allowing to be set to "Advanced" if the Computers tab is disabled. 

Screenshot_20260805_221538
Screenshot_20260805_221719

## AI tools used for this PR

None
## Fixed/related issues

Currently we open to the "Computers" tab if an external file is in use 
## How has this been tested?

Locally

### DIFF
--- a/src/lib/gui/dialogs/ServerConfigDialog.cpp
+++ b/src/lib/gui/dialogs/ServerConfigDialog.cpp
@@ -42,11 +42,6 @@ ServerConfigDialog::ServerConfigDialog(QWidget *parent, ServerConfig &config)
   ui->lblNewScreen->setPixmap(QIcon::fromTheme("video-display").pixmap(QSize(64, 64)));
   ui->btnBrowseConfigFile->setIcon(QIcon::fromTheme(QIcon::ThemeIcon::DocumentOpen));
 
-  if (ui->groupExternalConfig->isChecked())
-    ui->tabWidget->setCurrentIndex(3);
-  else
-    ui->tabWidget->setCurrentIndex(0);
-
   if (!deskflow::platform::isWindows())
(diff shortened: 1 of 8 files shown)


src/lib/gui/dialogs/ServerConfigDialog.cpp
```diff
@@ -42,10 +42,6 @@ ServerConfigDialog::ServerConfigDialog(QWidget *parent, ServerConfig &config)
   ui->lblNewScreen->setPixmap(QIcon::fromTheme("video-display").pixmap(QSize(64, 64)));
   ui->btnBrowseConfigFile->setIcon(QIcon::fromTheme(QIcon::ThemeIcon::DocumentOpen));
 
-  // force the first tab, since qt creator sets the active tab as the last one
-  // the developer was looking at, and it's easy to accidentally save that.
-  ui->tabWidget->setCurrentIndex(0);
```
nbolton, line 47: Strongly recommend moving `setCurrentIndex(0)` above `loadFromConfig()` rather than deleting it. You still land on Advanced when the tabs get disabled, and you keep the guard against Qt Creator baking a currentIndex into the .ui... which has caught us out so many times I lost count.

nbolton (approved):
Small nit which you can ignore, but I hope you don't.

src/lib/gui/dialogs/ServerConfigDialog.cpp
sithlord48: That is a review failure, but added to both our tabwidget in https://github.com/deskflow/deskflow/pull/10037

PR #10033, opened 2026-08-06
Mv clear settings action to settings dialog

sithlord48:
## Description

 - Move the Clear Settings Action to the Settings dialog's Advanced tab
 - Relabel it to "Remove all settings"
 - Settings dialog can emit a signal to request removing all settings
 - Move the question dialog to settings from messages
Screenshot_20260805_225333

## AI tools used for this PR

None
## Fixed/related issues

N/A
## How has this been tested?

Locally

### DIFF
--- a/src/lib/gui/MainWindow.cpp
+++ b/src/lib/gui/MainWindow.cpp
@@ -70,7 +70,6 @@ MainWindow::MainWindow()
       m_menuView{new QMenu(this)},
       m_menuHelp{new QMenu(this)},
       m_actionAbout{new QAction(this)},
-      m_actionClearSettings{new QAction(this)},
       m_actionMinimize{new QAction(this)},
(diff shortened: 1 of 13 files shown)


src/lib/gui/MainWindow.cpp
nbolton: Bug: because it's parented on main window, if you click off the message box, then you become trapped and can't interact with the GUI any more. You can't see it, but the message box is behind the dialog:

Image

Image

nbolton (changes requested):

src/lib/gui/MainWindow.cpp
sithlord48: Ill update after #10035  lands, by moving the dialog to the settings and removing it from `messages`

nbolton (approved):

PR #10034, opened 2026-08-06
Message rm settings message

sithlord48:
## Description

 -  Move the final of the `messages` that the Settings Dialog uses to settings.
## AI tools used for this PR

None
## Fixed/related issues

N/A
## How has this been tested?

Locally

### DIFF
--- a/src/lib/gui/dialogs/SettingsDialog.cpp
+++ b/src/lib/gui/dialogs/SettingsDialog.cpp
@@ -13,7 +13,6 @@
 
 #include "common/I18N.h"
 #include "common/Settings.h"
-#include "gui/Messages.h"
 #include "gui/TlsUtility.h"
 #include "gui/core/NetworkMonitor.h"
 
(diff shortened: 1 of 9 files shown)


src/lib/gui/dialogs/SettingsDialog.cpp
```diff
@@ -81,6 +80,7 @@ SettingsDialog::SettingsDialog(QWidget *parent, const ServerConfig &serverConfig
 
   setButtonBoxEnabledButtons();
   initConnections();
+  showReadOnlyMessage();
```
nbolton: On master, the message is triggered by showing, not constructing.
Upshot: the user won't see what is read only so the message has a bit less context.
Was that your intention?

Master:

Image

This PR:

Image

nbolton (dismissed):
Minor nit, can be ignored.

sithlord48:
I've dropped the first commit the dialog should act the same now

nbolton (approved):

Issue #8899, opened 2025-08-27
Incorrect keyboard layout conversion


d3npa:
### Project

Deskflow

I am having issues using **Deskflow v1.2.3** on my **Linux Desktop (X11; Server)** and **Macbook (Client)**. Both have Japanese keyboards, though Mac keyboards are always a little different.

The observed behaviour depends on the keyboard layout settings (the keyboard language) of each system. 

1. When Linux is `ja` and the mac only has the Japanese layout configured (it still shows up as `ja, en` in logs), none of the keys typed on the server work on the mac:

```
[2025-08-27T17:10:45] INFO: local languages: ja, en
[2025-08-27T17:10:45] INFO: remote languages: ja
[2025-08-27T17:10:45] DEBUG1: no keyboard resource for group 0
[2025-08-27T17:10:45] DEBUG1: no keyboard resource for group 1
[2025-08-27T17:10:45] DEBUG1: no keyboard resource for group 2
[2025-08-27T17:10:45] DEBUG1: mask=0000 outMask=0000
[2025-08-27T17:10:45] DEBUG1: modifiers on update: 0x0000
[2025-08-27T17:10:45] WARNING: can't get the active group, use the first group instead
```
```
[2025-08-27T17:10:46] WARNING: can't get the active group, use the first group instead
[2025-08-27T17:10:46] DEBUG1: mapKey 0061 (97) with mask 2000, start state: 0000, group: 0
[2025-08-27T17:10:46] DEBUG1: key 0061 is not on keyboard
```

2. Many guides online recommend adding an English keyboard language in Mac settings, but if I do that, it still reports `ja, en` and now the behaviour is so: letters work, but some keys from the server are needlessly translated on the client, for example, `:` becomes `+`. Some other keys, like 無変換 (英数 on mac keyboard) do not work:

```
2025-08-26T10:20:06] DEBUG1: recv key down id=0x0000003a, mask=0x2000, button=0x0030, lang="ja"
[2025-08-26T10:20:06] DEBUG1: mapKey 003a (58) with mask 2000, start state: 0000, group: 3
[2025-08-26T10:20:06] DEBUG1: language ja has group id 0
[2025-08-26T10:20:06] DEBUG1: best key index 1 of 1 (1 modifiers)
[2025-08-26T10:20:06] DEBUG1: found key in group 3
[2025-08-26T10:20:06] DEBUG1: state: 0000,0001,0005
[...]

PR #10039, opened 2026-08-07
fix: use ASCII-capable layout when an input method is active on macOS

### LINKED ISSUE #6599: Keyboard Language support is broken
### Operating Systems ###

Server: RHEL Workstation 7.5
Client: Win10

### Synergy Version ###
1.1.13

In RHEL, I have English and Russian, per application (not system-wide).
When I go to Windows (it also has English/Russian per application), it picks up the language from the last application in RHEL instead of using the language set in Windows.

For example, in RHEL, I have App1 with English and App2 with Russian. In Windows, I have App 3 in English.
I go to App2 in RHEL and start typing - it's in Russian.
I click App3 in Windows and start typing - now it's also Russian (what?)
I click App1 in RHEL and start typing - it's in English.
I click App3 in Windows and start typing - now it's also English (why?)

Note that none of these steps included switching languages. I only click between applications.
So the language in Windows completely depends on the language of the current application in RHEL - this is just plain wrong.
Please fix the bug.

BTW, the same problem (only worse) happens if Win is server and RHEL is client: when last app in Win was in English then in RHEL the languages are correct but when it was in Russian - then in RHEL it types question marks, commas etc - i.e. doesn't work at all.

vicyu0612:
## Description

While an input method is active on macOS, `TISCopyCurrentKeyboardLayoutInputSource()` returns the input method's *companion* keyboard layout. For CJK input methods that layout maps the physical keys to marks such as Bopomofo, so `UCKeyTranslate()` succeeds but produces key IDs the client cannot type. The keystroke is therefore silently lost — it is mistranslated rather than dropped.

Measured on macOS 26.6 with Traditional Chinese Zhuyin selected:

| API | returns | type | has uchr |
|---|---|---|---|
| `TISCopyCurrentKeyboardInputSource()` | `com.apple.inputmethod.TCIM.Zhuyin` | `TISTypeKeyboardInputMode` | no |
| `TISCopyCurrentKeyboardLayoutInputSource()` | `com.apple.keylayout.ZhuyinBopomofo` | `TISTypeKeyboardLayout` | **yes** |
| `TISCopyCurrentASCIICapableKeyboardLayoutInputSource()` | `com.apple.keylayout.ABC` | `TISTypeKeyboardLayout` | yes |

Because the Bopomofo layout does carry Unicode key layout data, existing null checks never trigger.

The input method belongs to the server only — the client runs its own — so `mapKeyFromEvent()` now translates with the ASCII-capable layout whenever the current input source is an input *mode* rather than a keyboard *layout*. That keeps the change narrow: plain non-Latin layouts (Cyrillic, Greek, ...) report `kTISTypeKeyboardLayout`, so their behaviour and their reliance on language sync are unchanged.

Also guards `CFDataGetBytePtr()` against a null `CFDataRef`, which is reachable whenever the selected layout carries no Unicode key layout data.

## AI tools used for this PR

Claude Opus 5, via Claude Code. It read the key-translation path, proposed an initial hypothesis (that keys were dropped because the layout had no `uchr` data), disproved that hypothesis by writing a small probe against the TIS APIs — which produced the table above — then wrote the patch and drafted this description.

I directed the investigation, decided on the approach, and did all testing on my own hardware. I have reviewed the diff and understand it.

## Fixed/related issues

Related: #6599

This addresses the macOS-server-with-input-method symptom reported in #8575, #9332 and #8899, which were closed as duplicates of #6599. It does not claim to close #6599, which covers keyboard language handling more broadly.

## How has this been tested?

Built from this branch and installed as the server on my own machine, then verified interactively.

- **Server:** macOS 26.6, Apple silicon, input sources ABC + Traditional Chinese Zhuyin
- **Client:** Windows 11, stock Deskflow 1.26.0, protocol v1.8 (unmodified)

Before the change, selecting Zhuyin on the server meant no keystroke at all reached the client; switching back to ABC restored typing. After the change, typing on the client works with either input source selected, and switching between them mid-session needs no action.

`ctest` passes 26/26, including `OSXKeyStateTests`. `clang-format` reports no changes.

### DIFF
--- a/src/lib/platform/OSXKeyState.cpp
+++ b/src/lib/platform/OSXKeyState.cpp
@@ -297,7 +297,26 @@ KeyButton OSXKeyState::mapKeyFromEvent(KeyIDs &ids, KeyModifierMask *maskOut, CG
   CFDataRef ref = nullptr;
   {
     std::lock_guard<std::mutex> lock(g_tisMutex);
-    currentKeyboardLayout = AutoTISInputSourceRef(TISCopyCurrentKeyboardLayoutInputSource(), CFRelease);
+    // When an input method is active (Chinese, Japanese, Korean, ...) the
+    // current keyboard layout is the input method's companion layout, which
+    // maps the physical keys to marks such as Bopomofo rather than to Latin
+    // characters. Translating with it produces key IDs the client cannot type,
+    // so the keystroke is silently lost. The input method belongs to this
+    // computer only -- the client runs its own -- so translate with the
+    // ASCII-capable layout, which is what the physical keys are engraved with.
+    // Plain non-Latin layouts (Cyrillic, Greek, ...) are reported as keyboard
+    // layouts, not input modes, and keep their existing behaviour.
+    bool inputMethodActive = false;
+    if (AutoTISInputSourceRef currentInputSource(TISCopyCurrentKeyboardInputSource(), CFRelease); currentInputSource) {
+      auto type = (CFStringRef)TISGetInputSourceProperty(currentInputSource.get(), kTISPropertyInputSourceType);
+      inputMethodActive = (type != nullptr) && !CFEqual(type, kTISTypeKeyboardLayout);
+    }
+
+    currentKeyboardLayout = AutoTISInputSourceRef(
+        inputMethodActive ? TISCopyCurrentASCIICapableKeyboardLayoutInputSource()
+                          : TISCopyCurrentKeyboardLayoutInputSource(),
+        CFRelease
+    );
     if (currentKeyboardLayout)
       ref = (CFDataRef)TISGetInputSourceProperty(currentKeyboardLayout.get(), kTISPropertyUnicodeKeyLayoutData);
   }
@@ -332,7 +351,7 @@ KeyButton OSXKeyState::mapKeyFromEvent(KeyIDs &ids, KeyModifierMask *maskOut, CG
   }
 
   // translate via uchr resource
-  const UCKeyboardLayout *layout = (const UCKeyboardLayout *)CFDataGetBytePtr(ref);
+  const UCKeyboardLayout *layout = ref ? (const UCKeyboardLayout *)CFDataGetBytePtr(ref) : nullptr;
   const bool layoutValid = (layout != nullptr);
 
   if (layoutValid) {


sithlord48:
Way to many comments and the commit message is way way to long a single line for the commit message is fine

src/lib/platform/OSXKeyState.cpp
```diff
@@ -297,7 +297,26 @@ KeyButton OSXKeyState::mapKeyFromEvent(KeyIDs &ids, KeyModifierMask *maskOut, CG
   CFDataRef ref = nullptr;
   {
     std::lock_guard<std::mutex> lock(g_tisMutex);
-    currentKeyboardLayout = AutoTISInputSourceRef(TISCopyCurrentKeyboardLayoutInputSource(), CFRelease);
+    // When an input method is active (Chinese, Japanese, Korean, ...) the
```
amagus, line 300: This is too verbose... Please simplify the comments.

src/lib/platform/OSXKeyState.cpp
```diff
@@ -297,7 +297,26 @@ KeyButton OSXKeyState::mapKeyFromEvent(KeyIDs &ids, KeyModifierMask *maskOut, CG
...
+    // current keyboard layout is the input method's companion layout, which
+    // maps the physical keys to marks such as Bopomofo rather than to Latin
+    // characters. Translating with it produces key IDs the client cannot type,
+    // so the keystroke is silently lost. The input method belongs to this
+    // computer only -- the client runs its own -- so translate with the
+    // ASCII-capable layout, which is what the physical keys are engraved with.
+    // Plain non-Latin layouts (Cyrillic, Greek, ...) are reported as keyboard
```
amagus, line 307: Are you sure about this?
Unless this is documented somewhere, relying on this information may cause breakages on users of very specific keyboard layouts...

amagus (changes requested):

sithlord48:
@vicyu0612  This is very similar to another open pr, Can you try #9792  to see if this fixes the issues your trying to fix here and let us know.

vicyu0612:
Tested #9792 — it fixes this. Closing in favour of it.

Setup: macOS 26.6 on Apple silicon as server with Traditional Chinese Zhuyin selected, Windows 11 client on stock 1.26.0. Before, no keystroke reached the client at all; with #9792 typing works with Zhuyin active. `ctest` passed 26/26 on that branch here.

Measured properties, in case they're useful:

| API | returns | uchr | selectCapable |
|---|---|---|---|
| `TISCopyCurrentKeyboardInputSource` | `com.apple.inputmethod.TCIM.Zhuyin` | no | yes |
| `TISCopyCurrentKeyboardLayoutInputSource` | `com.apple.keylayout.ZhuyinBopomofo` | yes | no |

The Bopomofo layout does carry key layout data, so null checks never fire, but it isn't select-capable — which is what #9792 keys off. That condition is narrower than mine and correctly leaves user-selected non-Latin layouts alone, so @amagus's concern about other layouts was justified.

Thanks both.

MomoyamaSawa:
@vicyu0612 Thanks for testing #9792 and confirming the fix — really appreciate it! Glad to see it also works with Zhuyin 🙌

PR #10041, opened 2026-08-08
fix: prevent modifier keys from being stuck (Windows)

barrelltitor:
## Description

How to reproduce
- Hold LCTRL down on the client and move your mouse offscreen
- Release lctrl when you're on your host and click once
- Return to the client screen with your mouse
- CTRL is still down, up was never received, or it was received but for the wrong key, or it was received but then another down came in 

The modifier keys got stuck for several reasons, basically a bunch of stuck keys in different places:

1. a modifier up on Windows was discarded in a defensive fix 
2. some malformed windows events with scan code 0 figured out the button but then sent button id 0 anyway
3. GetKeyboardState() lagged behind the actual physical state somehow and reported ctrl as being down after the windows hook changed it to up

The PR guards for down repeat events specifically so up can still get through in 1), preserves the figured out button in 2) and gets the physical state from the windows hook with a lock for safety to handle 3)

I haven't seen or written C++ since high school, so I don't know if there are better solutions to these, but after many hours of testing I needed all 3 solutions to fully make it stop happening.

All fixes were made after logs confirmed them happening, though some were rare to happen

## AI tools used for this PR

Codex 5.6 Terra in vscode extension

I used it to:
- investigate the debug and verbose logs 
- add more debug logs in relevant places since verbose made it unusable, 
- search & explain the codebase
- write code
- build the flatpak and windows portable versions so I can actually test it out 

## How has this been tested?

With blood, sweat, tears, and an entire day lost. I have tested this for many hours with different combinations of fixes.

Just to sanity check, I switched to the unfixed version again and it immediately happened. I could consistently reproduce it with the steps above.

### DIFF
--- a/src/lib/platform/MSWindowsHook.cpp
+++ b/src/lib/platform/MSWindowsHook.cpp
@@ -10,6 +10,9 @@
 #include "base/Log.h"
 #include "deskflow/ScreenException.h"
 
+#include <cstring>
+#include <mutex>
+
 #ifndef WM_MOUSEHWHEEL
 #define WM_MOUSEHWHEEL 0x020E
 #endif
@@ -34,6 +37,8 @@ static WPARAM g_deadRelease = 0;
 static LPARAM g_deadLParam = 0;
 static BYTE g_deadKeyState[256] = {0};
 static BYTE g_keyState[256] = {0};
+static bool g_keyStateValid = false;
+static std::mutex g_keyStateMutex;
 static DWORD g_hookThread = 0;
 static bool g_fakeServerInput = false;
 static BOOL g_isPrimary = TRUE;
@@ -142,6 +147,15 @@ void MSWindowsHook::setMode(EHookMode mode)
   g_mode = mode;
 }
 
+bool MSWindowsHook::getPhysicalKeyState(BYTE keys[256])
+{
+  std::lock_guard<std::mutex> lock(g_keyStateMutex);
+  if (g_keyStateValid) {
+    std::memcpy(keys, g_keyState, sizeof(g_keyState));
+  }
+  return g_keyStateValid;
+}
+
 static void keyboardGetState(BYTE keys[256], DWORD vkCode, bool kf_up)
 {
   // we have to use GetAsyncKeyState() rather than GetKeyState() because
@@ -151,6 +165,8 @@ static void keyboardGetState(BYTE keys[256], DWORD vkCode, bool kf_up)
   if (vkCode < 0 || vkCode >= 256) {
     return;
   }
+
+  std::lock_guard<std::mutex> lock(g_keyStateMutex);
 
   // Keep track of key state on our own in case GetAsyncKeyState() fails
   g_keyState[vkCode] = kf_up ? 0 : 0x80;
@@ -179,6 +195,7 @@ static void keyboardGetState(BYTE keys[256], DWORD vkCode, bool kf_up)
 
   key = GetKeyState(VK_CAPITAL);
   keys[VK_CAPITAL] = (BYTE)(((key < 0) ? 0x80 : 0) | (key & 1));
+  g_keyStateValid = true;
 }
 
 static WPARAM makeKeyMsg(UINT virtKey, WCHAR wc, bool noAltGr)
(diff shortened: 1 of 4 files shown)


sithlord48:
We don't allow merge requests. please correct it with a rebase. This also maybe better broken into two parts one for the x11 changes and one for the windows ones.

barrelltitor:
My bad, it's muscle memory from work, i see an update branch button I can't help myself.  Will rebase in the morning 

The x11 fix is for the client while windows fixes are for server, but they are to fix the same issue. There's only 3 lines for the x11 fix

Should I go ahead and split them up in the morning as well like you said or leave them like this?

sithlord48:
Yes please split them since i can test the X11 changes but not the windows ones and others maybe only able to test the windows changes.

src/lib/platform/XWindowsScreen.cpp
```diff
@@ -1902,6 +1902,11 @@ void XWindowsScreen::refreshKeyboard(XEvent *event)
       return;
     }
   }
+  // release all keys before refresh to prevent key stuck in down state
```
sithlord48: This comment will not be needed we can see what happens clearly in the code.

sithlord48:
> How to reproduce
> - Hold LCTRL down on the X11 client and move your mouse offscreen
> - Release lctrl when you're on your host and click once
> - Return to the X11 client screen with your mouse
> - CTRL is still down, up was never received, or it was received but for the wrong key, or it was received but then another down came in 

This does not trigger the holding of modifers for me. 

| Server  | Client | 
|:----------:| :-------: |
|Deskflow: 1.26.0.366 (f765e829) | Deskflow: 1.26.0.315 (857cc219)|
|Qt: 6.11.1|Qt: 6.11.1 |
|System: Arch Linux|System: Arch Linux|
| Session: KDE (wayland)|Session: KDE (x11)|

barrelltitor:
@sithlord48 rebased and left only the windows changes in this one. 

The X11 client issue specifically needs the keyboard map refresh to happen between down and up so that's not going to be as easy to test. There's also probably other issues with the code for the host people had the bug on, like it was for the Windows host

The repro steps work all the time on a windows host though. Still haven't had the issue once since I switched to the custom build

I'll make a draft pr for x11

sithlord48:
Can you try with just your windows fixes and see if it also fixes if for you . (it may not and that is also ok)

src/lib/platform/XWindowsScreen.cpp
barrelltitor: i'll leave the explanation in about the modifiers since that's not very clear, but will need to test this more and see if there's some sure reproduction methods like for the windows host

barrelltitor:
Yeah, building it now for Windows and for linux I'll just switch to the continuous version I had before 

I'll use this combination today while working and doing other stuff to see if I get this again 

One thing is I suspected there's more places in the code where this could happen for x11 from the getgo and had the AI look through everything, but since I don't know the codebase very well and it decided to modify 12 different files I didn't explore it more. I didn't want to just push random AI code and go "see if any of this helps" for obvious reasons

I'll be switching my main computer to Linux as soon as Playnite is available for it, so I'll be able to test there more on the wayland/x11 host side

sithlord48:
> I didn't want to just push random AI code and go "see if any of this helps"  

Thank  you for being considerate !!!!

barrelltitor:
With the Windows host changes alone and using the continuous build I was using before(unmodified)I haven't experienced the issue again so far

This does fix it on Windows, but it will not actually close the issue if merged since it happens also on Wayland and X11 host. Should I remove the fixed issue from the pr description?

sithlord48:
Yes you should remove the fixed issue since this does not fix that one.

barrelltitor:
Removed

nbolton:
One minor thing that we may want to fix, now that the X11 fix was split out into a different PR, the description for this PR is slightly confusing (we should probably drop all X11 references [just in this PR]).

Also the commits could do with being `fix(win32):` for clarity.

sithlord48:
@barrelltitor  can you please up this this description for us

nbolton:
Managed to reproduce shift key stuck on `master` with Windows server and Wayland client with these steps!

1. On the server keyboard, press and hold left shift. Client receives shift down.
2. While holding it, also press and hold right shift. On master this down is swallowed server-side, which is already the bug's signature: nothing appears in the client log.
3. Release left shift first, still holding right shift. This is the stuck-key moment: the up is swallowed, and the client's left shift stays down.
4. Release right shift, then type letters while still on the client screen. Capitals prove shift is stuck. On the PR build, all four transitions reach the client and typing is normal.

Only managed to do it once though, 2nd time I tried I was unable to repro. 

Sneaky/evasive bug!

nbolton (approved):

PR #10046, opened 2026-08-10
fix(macos): build key map with UCKeyTranslate to honor keyboard type

fruitriin:
## Description
On a macOS client with a Japanese (JIS) keyboard, the key map was built with the US layout, so forwarded punctuation was mistyped (e.g. " → *). 

Replaced the hand-rolled `uchr` parser, which ignores the physical keyboard type, 
with `UCKeyTranslate` (already used on the server side), which honors it.

Verified on a JIS-Mac ↔ JIS-Mac setup. 
By the same mechanism, other macOS-client bugs caused by the hand-rolled parser may also be resolved.

## AI tools used for this PR

Claude Code (Opus 4.8) — root-cause investigation and drafting the patch, translated description.
Reviewed, built and tested by me.

## Fixed/related issues

related: #6599 — the macOS JIS-keyboard case only.
Non-macOS clients, dead keys / AltGr and non-Latin scripts are out of scope, so this does not close it.

## How has this been tested?

- `ctest` 26/26 pass (incl. `OSXKeyStateTests`).
- macOS JIS client ↔ macOS server: `"` `@` `&` `'` `( ) = ~ [ ] _` etc. correct in both kana and alphanumeric modes.


sithlord48:
Can you update your commit message to remove the co-author claude line. It can not hold copyright so its pointless spam in our commit messages. We will only be looking for you to fix and support issues (if any ) with this code.

fruitriin:
Thank you! I fixed it

amagus:
Have you tried with the latest continuous if this issue was not already fixed?
We had similar issues with pinyin keyboards and maybe keyboard translation issues got already fixed.

src/lib/platform/OSXUchrKeyResource.cpp
```diff
@@ -1,257 +0,0 @@
-/*
```
amagus, line 1: Why deleting this file instead of fixing what is wrong on it? Please explain why we need the new class

fruitriin:
@amagus I tried brew/tap/deskflow-dev yesterday but still need this fix.

amagus:
Since you are not exactly fixing an issue directly, I think we need more information on how to reproduce the issue first before continuing with the review. 

Maybe it would be easier for you to fill an issue and explain exactly the operating system used (to the exact version) and the exact keyboard settings used (output for `defaults read ~/Library/Preferences/com.apple.HIToolbox.plist AppleSelectedInputSources` should suffice).

I really liked the cleaner code but I am not sure if we should rename the class like you did. 
I want to reproduce the issue so that I can confirm it works.

src/lib/platform/OSXUchrKeyResource.cpp
fruitriin: `OSXUchrKeyResource` is a hand-rolled `uchr` parser. 
Its table choice doesn't reflect the physical keyboard, so it picks the US layout, not JIS.

To fix this properly we would have to rebuild the keyboard-type handling ourselves. 
This parser also leaves some `uchr` cases "not supported yet".

The server code (`mapKeyFromEvent`) already uses `UCKeyTranslate`,
so I think it's better to use `UCKeyTranslate` here too than to patch the hand-rolled parser.

Once it calls `UCKeyTranslate` it no longer parses the `uchr` resource, 
so the name `OSXUchrKeyResource` would be misleading. If you'd rather keep the original file/name, 
I'm happy to do that instead.

fruitriin:
Repro:
- Client input sources: ABC + Apple Japanese IME (no custom layouts). "Use server's keyboard language" is off.
- `defaults read ~/Library/Preferences/com.apple.HIToolbox.plist AppleSelectedInputSources` → ABC selected (Japanese IME also enabled).
- Type `"` (Shift+2 on a JIS keyboard) on the server → it arrives as `*` on the client. `@` arrives as `` ` ``.

I also checked the latest origin/master (dd3944cab): the client key map still uses the US arrangement — `"` is assigned to the US quote key (vk 0x27), while on JIS it should be the "2" key (vk 0x13). So it is not fixed upstream yet.

Environment:
- Client: macOS Sequoia 15.7.4 (24G517), JIS keyboard, ABC + Apple Japanese IME (no custom layouts). Deskflow origin/master (dd3944cab).
- Server: macOS Tahoe 26.3.1 (25D771280a), JIS keyboard, ABC + Google Japanese Input (no custom layouts). Deskflow 1.26.0.366 (f765e829), Homebrew deskflow/tap/deskflow-dev.
- "Use server's keyboard language on this computer" is off.

More detailed logs: https://github.com/deskflow/deskflow/issues/6599#issuecomment-5204242793

nbolton (approved):
Tested by setting both layouts to French and pressing the "2" key to get `é` on client.

PR #10048, opened 2026-08-10
refactor: no need to call setAppId for gui as the App does it on its own

sithlord48:
## Description

 - Only call `setAppId` in deskflow-gui when we are using a Qt version < 6.10
## AI tools used for this PR

None
## Fixed/related issues

Removes the warning on start up `WARNING: Failed to register with host portal QDBusError("org.freedesktop.portal.Error.Failed", "Could not register app ID: Connection already associated with an 
## How has this been tested?

Local test

### DIFF
--- a/src/apps/deskflow-gui/deskflow-gui.cpp
+++ b/src/apps/deskflow-gui/deskflow-gui.cpp
@@ -52,7 +52,8 @@ int main(int argc, char *argv[])
 #endif
 
 #if !defined(Q_OS_MAC) && !defined(Q_OS_WIN)
-  deskflow::platform::setAppId();
+  if (QT_VERSION < QT_VERSION_CHECK(6, 10, 0))
+    deskflow::platform::setAppId();
 #endif
 
   QCoreApplication::setApplicationName(kAppName);


src/apps/deskflow-gui/deskflow-gui.cpp
```diff
@@ -51,10 +51,6 @@ int main(int argc, char *argv[])
   QLoggingCategory::setFilterRules(QStringLiteral("*.debug=true\nqt.*=false"));
 #endif
 
-#if !defined(Q_OS_MAC) && !defined(Q_OS_WIN)
-  deskflow::platform::setAppId();
```
nbolton, line 55: Doesn't Qt only do this itself from 6.10? Our minimum is 6.7.

sithlord48: good point

src/apps/deskflow-gui/deskflow-gui.cpp
```diff
@@ -52,7 +52,8 @@ int main(int argc, char *argv[])
 #endif
 
 #if !defined(Q_OS_MAC) && !defined(Q_OS_WIN)
-  deskflow::platform::setAppId();
+  if (QT_VERSION < QT_VERSION_CHECK(6, 10, 0))
```
nbolton, line 55: Could also do:
```suggestion
  if (QLibraryInfo::version() < QVersionNumber(6, 10))
```

But that would need...
```
#include <QLibraryInfo>
#include <QVersionNumber>
```

nbolton (approved):

src/apps/deskflow-gui/deskflow-gui.cpp
sithlord48: Changed this to do the check if qt < 6.10

src/apps/deskflow-gui/deskflow-gui.cpp
sithlord48: yeah but mine just uses the macros those are calling

PR #10049, opened 2026-08-11
Add read only messages

### LINKED ISSUE #10047: Warn when settings cannot be saved because the settings file is read-only
### Feature Request

If `Deskflow.conf` is read-only, GUI changes are silently lost: `QSettings::sync()` fails without any error, so things like screen aliases just never take effect and there's nothing in the GUI or logs to explain why. The `SettingsDialog` already shows a read-only warning on open; it would be great to show the same warning when the `ServerConfigDialog` opens, and to log a warning in `Settings::save()` when the file is not writable.

### Voting Instructions

sithlord48:
## Description

 - Revert #10034 
 - Use `deskflow::gui::message::showReadOnlyMessage` in Client Config Dialog 
 - Use `deskflow::gui::message::showReadOnlyMessage` in Server Settings Dialog 
 - Settings Warn on save if not writable
## AI tools used for this PR

None
## Fixed/related issues

fixes: #10047
## How has this been tested?

Locally

### DIFF
--- a/src/lib/gui/dialogs/ServerConfigDialog.cpp
+++ b/src/lib/gui/dialogs/ServerConfigDialog.cpp
@@ -7,6 +7,7 @@
  */
 
 #include "ServerConfigDialog.h"
+#include "Messages.h"
 #include "ui_ServerConfigDialog.h"
 
 #include "common/Constants.h"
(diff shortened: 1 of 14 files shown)


src/lib/gui/dialogs/ServerConfigDialog.cpp
```diff
@@ -498,6 +499,21 @@ void ServerConfigDialog::initConnections() const
...
+  connect(this, &ServerConfigDialog::shown, this, &ServerConfigDialog::showReadOnlyMessage, Qt::QueuedConnection);
+}
+
+void ServerConfigDialog::showEvent(QShowEvent *event)
+{
+  QDialog::showEvent(event);
+  Q_EMIT shown();
+}
+
+void ServerConfigDialog::showReadOnlyMessage()
```
nbolton, line 511: SC suggests using auto and moving the declare inside the if statement.

nbolton (changes requested):
Bug: the message doesn't show when I make a change to the screen layout

1. `chmod 444 ~/.config/Deskflow/Deskflow.conf`
2. Start GUI
3. Edit screen layout
4. Click 'OK'

Expect: Read only message
Actual: No message, and when I restart the GUI the old settings are restored

src/lib/gui/dialogs/ServerConfigDialog.cpp
sithlord48: it would we are using C++ 20..

sithlord48:
> Bug: the message doesn't show when I make a change to the screen layout
> 
>     1. `chmod 444 ~/.config/Deskflow/Deskflow.conf`
> 
>     2. Start GUI
> 
>     3. Edit screen layout
> 
>     4. Click 'OK'
> 
> 
> Expect: Read only message Actual: No message, and when I restart the GUI the old settings are restored

The server settings looks to see if the external settings is writable. it guess it should check both

PR #10052, opened 2026-08-11
refactor: move config to qt config file and remove legacy config file

nbolton:
## Description

The GUI wrote a legacy `deskflow-server.conf` on every core start and the server parsed it
straight back, all duplicating what's already in the settings INI. The server now builds its
config directly from settings (screens/hotkeys from `[internalConfig]`, links from grid
adjacency); the .conf parser/writer, external config option, and export button are removed.
Also fixes mouse-button hotkeys, which the core silently skipped when parsing the old options section.

## AI tools used for this PR

Claude Code (Claude Fable 5) - wrote the refactor, docs, and the new unit test.

## Fixed/related issues

None.

## How has this been tested?

Smoke tested on Linux (Fedora 44, Qt 6.11). New `ServerConfigTests::loadFromSettings` covers
screens, aliases, links, options, and hotkeys from an isolated INI; full suite passes. Windows
and macOS not yet tested.

- [ ] Test on Windows
- [ ] Test on macOS
- [ ] Remove core mode arg (just read from settings)


sithlord48:
My biggest issue with this is the loss of some of the more advanced screen layout being able to be set manually.. I am looking at getting this updated.

PR #10055, opened 2026-08-12
feat(macos): forward navigation gestures as mouse buttons

andy5090:
## Description

On macOS, back and forward navigation actions can arrive as gesture events instead of standard extra mouse buttons. This change adds an opt-in macOS server setting that forwards those horizontal navigation gestures as standard back and forward mouse buttons while controlling a client. The setting is disabled by default.

The issue was discovered with physical back and forward buttons configured through Logi Options+, but the implementation is based on the public macOS event representation and contains no vendor detection or vendor-specific event fields.

### Implementation note

The Core Graphics event is converted with the public `+[NSEvent eventWithCGEvent:]` API. Events are forwarded only when AppKit exposes them as `NSEventTypeSwipe` with a non-zero public `deltaX` value, the option is enabled, and the cursor is on a remote screen. Positive and negative horizontal deltas are mapped to standard back and forward mouse buttons respectively.

Local input and the default disabled behavior remain unchanged. No undocumented `CGEventField` values or private APIs are used.

## AI tools used for this PR

OpenAI Codex (GPT-5.6-sol) was used for code drafting, test generation, code review, and local build and test automation. The behavior was manually verified with Logi Options+ on macOS and Deskflow clients on Linux and Windows.

## How has this been tested?

- Built and signed a Release configuration on macOS 26.5.2 (arm64).
- Verified through temporary diagnostics that AppKit exposes the observed back and forward actions as public swipe events with `deltaX` values `+1` and `-1`.
- Verified back and forward navigation on a Linux client running Deskflow 1.26.0 and on a Windows client using the public API implementation.
- Ran `OSXScreenTests`: 12 tests passed.
- Ran `ServerConfigTests`: 10 tests passed.
- Built the GUI, core, and translations, and verified the signed application bundle.
- Linux and Windows builds were not run locally and remain to be verified by CI.


sithlord48:
I really don't care for these hyper specific application / device workarounds

src/lib/gui/dialogs/SettingsDialog.cpp
```diff
@@ -298,6 +305,11 @@ void SettingsDialog::loadFromConfig()
 
   if (!deskflow::platform::isWindows())
     ui->groupService->setVisible(false);
+#if defined(Q_OS_MACOS)
```
sithlord48: you can use 
`deskflow::platform::isMac()` you don't need to use the if def for this

src/lib/gui/dialogs/SettingsDialog.ui
```diff
@@ -92,6 +92,16 @@
          </property>
         </widget>
        </item>
+       <item>
```
sithlord48: This would belong in the server settings not the general settings.

amagus:
Does this work while using other operating systems?
Can't we work on something that would work for all of them?
What about official macOS gestures? Can't we support them as well?

src/lib/platform/OSXScreen.mm
```diff
@@ -60,6 +60,12 @@
 };
 
 static const double kCarbonLoopWaitTimeout = 10.0;
+static constexpr auto kNavigationGestureEventType = static_cast<CGEventType>(NSEventTypeGesture);
+// CoreGraphics does not publish this field. Its value is reverse engineered
+// and matches the gesture events observed from Logi Options+.
+static constexpr auto kNavigationGestureSwipeDirectionField = static_cast<CGEventField>(117);
```
amagus: This is private API usage which may prevent us to ever being Notarized and/or in the App Store. Can't you make it work without that?

src/lib/gui/dialogs/SettingsDialog.cpp
andy5090: Updated in 0619bafc3. The control now lives in `ServerConfigDialog` and uses `deskflow::platform::isMac()` for visibility. The compile-time guard and the Options+ installation-detection helper were removed.

src/lib/gui/dialogs/SettingsDialog.ui
andy5090: Updated in 0619bafc3. The opt-in control and its persistence logic have been moved from the general settings dialog to the server configuration dialog.

src/lib/platform/OSXScreen.mm
andy5090: Yes, this now works without the undocumented field. In 0619bafc3 the `CGEvent` is converted with the public `+[NSEvent eventWithCGEvent:]` API, accepted only when AppKit exposes it as `NSEventTypeSwipe`, and classified using the documented horizontal `deltaX`. Temporary diagnostics observed `+1` and `-1` for the two directions, and the public-API implementation was then manually verified with Linux and Windows clients. Field 117 and all private/undocumented API usage have been removed.

andy5090:
Thanks @sithlord48 and @amagus for the feedback. I have reworked the PR in 0619bafc3:

- The title and setting are now vendor-neutral macOS navigation gesture functionality.
- The Logi Options+ installation check and other vendor-detection code were removed.
- The setting now belongs to the server configuration dialog and uses `deskflow::platform::isMac()`.
- The undocumented Core Graphics field was removed. Direction now comes from the public `NSEventTypeSwipe` representation and its documented `deltaX` property.

Logi Options+ remains the motivating and manually tested case. I considered the interoperability issue worth investigating because Logitech is a major mouse vendor. I agree, however, that Deskflow should not carry a private or vendor-specific event implementation for it.

The source-side detection is macOS-specific, but the result is sent as standard Deskflow back/forward mouse buttons, so it is client-OS independent. I manually verified the revised public-API implementation on Linux and Windows clients.

Native trackpad gesture UX is not being explicitly expanded or claimed by this PR. Forwarding system trackpad gestures can introduce separate local-gesture conflicts, so I think that behavior should remain a separate discussion. This PR only forwards an event when macOS publicly exposes the received navigation action as a horizontal `NSEventTypeSwipe`, the opt-in setting is enabled, and the cursor is on a client.

src/lib/gui/dialogs/SettingsDialog.cpp
```diff
@@ -283,11 +276,6 @@ void SettingsDialog::loadFromConfig()
 
   if (!deskflow::platform::isWindows())
     ui->groupService->setVisible(false);
-#if defined(Q_OS_MACOS)
```
amagus, line 286: don't fix a commit with another commit... Use amend

src/lib/platform/OSXScreen.mm
```diff
@@ -1794,20 +1790,19 @@ static inline bool needsEventNumber()
...
-ButtonID OSXScreen::classifyNavigationGestureButton(CGEventType type, bool isOnScreen, bool enabled, int64_t direction)
+ButtonID OSXScreen::classifyNavigationGestureButton(CGEventType type, bool isOnScreen, bool enabled, double deltaX)
 {
   if (type != kNavigationGestureEventType || isOnScreen || !enabled) {
     return kButtonNone;
   }
 
-  switch (direction) {
-  case kNavigationGestureSwipeLeft:
+  if (deltaX > 0.0) {
```
amagus, line 1799: what if that button is mapped to something else in the client machine?

amagus:
IMHO, this PR is too specific.

My question was not about you claiming it works on native trackpad but maybe that supporting native trackpads you would get support for this specific mouse for free. We should focus on general solutions instead.

Does the support for the gestures work when using windows or linux as a server and connecting your mouse to these OS? 
If not, can't you provide support for all platforms instead of focusing on mac alone?

Also, I am not sure that triggering a mouse button in lieu of the gesture is a good ideia. What if the user mapped that button to something else? Can't you reproduce the exact same gesture that happens when you use your mouse directly to the other machine there as well?

src/lib/gui/OSXHelpers.h
```diff
@@ -1,5 +1,6 @@
 /*
  * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
```
sithlord48: Why are you adding copyright changes to a file you do not otherwise edit ?

src/lib/gui/OSXHelpers.mm
```diff
@@ -1,5 +1,6 @@
 /*
  * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
```
sithlord48: Why are you adding copyright changes to a file you do not otherwise edit ?

src/lib/platform/OSXScreen.h
```diff
@@ -43,6 +43,8 @@ class Mutex;
 //! Implementation of IPlatformScreen for OS X
 class OSXScreen : public PlatformScreen
 {
+  friend class OSXScreenTests;
```
sithlord48: Generally not a fan of friend classes.

src/lib/deskflow/OptionTypes.h
```diff
@@ -59,6 +60,7 @@ static const OptionID kOptionDefaultLockToScreenState = OPTION_CODE("LTSS");
 static const OptionID kOptionDisableLockToScreen = OPTION_CODE("DLTS");
 static const OptionID kOptionClipboardSharing = OPTION_CODE("CLPS");
 static const OptionID kOptionClipboardSharingSize = OPTION_CODE("CLSZ");
+static const OptionID kOptionMacNavigationGestures = OPTION_CODE("MNGS");
```
sithlord48: Idk you can just add an option to the server without modifying the protocol as well to handle this message.

andy5090:
I understand the preference for implementing a general gesture path first, and I agree that it is a worthwhile longer-term direction.

However, this PR addresses a concrete problem that users experience today: with the standard Logi Options+ Back and Forward configuration, the actions work on the local Mac but stop working when controlling a Deskflow client. Using Logi Options+ to configure these buttons is a normal setup for MX Master series users, so this is unlikely to be unique to my environment. Without this change, affected users have no practical built-in solution in the standard Deskflow release. I do not think the fact that the use case is specific is, by itself, a sufficient reason to leave that usability problem unresolved.

The behavior is opt-in and disabled by default. Enabling it is the user’s explicit decision to interpret these macOS events as Back and Forward. Users with different mappings simply leave it disabled, so the change does not impose that interpretation on existing users.

Supporting Windows and Linux servers or reproducing the original gesture would be a separate and substantially broader project. There is no common gesture representation and injection path across the operating systems, so that direction would require a new protocol and platform-specific implementations rather than a small extension of this PR.

I support pursuing that general design separately, and this macOS path could be migrated to it later. However, requiring the complete cross-platform solution first could leave affected macOS users without a practical solution indefinitely, even though the current fix is isolated and optional.

amagus:
Could you please give direct answers to the questions asked? (and perhaps reduce the amount of AI generated text?)

Adding options to stuff that not many people use is a major risk for a project like deskflow. 
The related code is not used that much and it's not possible for the project maintainers to ensure the code works as intended (not only right now, but in the future). 
Unit tests don't cover OS updates or future refactorings... General solutions is the way to go.

Also, I don't think your code really works without external tools. 
Have you really tested it? 
Are you using any external special Logitech tool?
Based on my recent testing, macOS does not provide NSSwipe actions to the event tap unless there's an active cocoa application listening for it under the mouse cursor. Deskflow can't handle these as we just stay in the background.

src/lib/platform/OSXScreen.mm
```diff
@@ -1737,25 +1732,26 @@ static inline bool needsEventNumber()
...
   case kNavigationGestureEventType: {
-    // Leave local and opt-out behavior untouched, including avoiding access to
-    // the undocumented field below.
     if (screen->m_isOnScreen || !screen->m_navigationGesturesEnabled) {
       break;
     }
 
-    const auto direction = CGEventGetIntegerValueField(event, kNavigationGestureSwipeDirectionField);
+    NSEvent *nativeEvent = [NSEvent eventWithCGEvent:event];
+    if (nativeEvent == nil || nativeEvent.type != NSEventTypeSwipe) {
```
amagus, line 1740: how does this work? 
The CG event type is Gesture but the NSEvent is a Swipe?
Is Logitech software/driver injecting something completely bogus?
All events that I can generate on my side keep their type if their base type is NS (and this is one, given you define kNavigationGestureEventType at the very top of this file)

amagus (changes requested):

andy5090:
Yes, I tested it, and the mouse is configured through Logi Options+.

With Options+ running, Back and Forward work on the local Mac. On Linux and Windows clients, they do not work with this option off, but they do work with it on.

I also tested with Options+ stopped and this option off. In that case, the buttons work both locally and remotely through Deskflow's existing extra-button handling.

In my setup, Deskflow does receive the type 29 event while running in the background, and `eventWithCGEvent:` reports it as a swipe. I do not know why AppKit converts it that way.

You are also right about custom button mappings. Supporting those properly would probably require adding more configuration, and I am not convinced that expanding this PR that far is the right direction.

I will close this PR for now and may revisit the problem if there is a cleaner, more general approach.

Issue #9790, opened 2026-05-27
macOS screenshots (images) cannot be pasted to Windows 11 via shared clipboard — only text works


metaversetimes:
### Before reporting a bug, please complete the sanity checks above.

- [x] I have done the sanity checks, and my issue persists.

### Deskflow version info

Server (Windows 11):
Deskflow: 1.26.0.210 (bf65080b)
Qt: 6.10.2
System: Windows 11 Version 25H2

Client (macOS):
Deskflow: 1.26.0.210 (bf65080b)
Qt: 6.9.3
System: macOS Sequoia (15.7.3)

### Build types

Continuous v1.26.0.210 Pre-release

### Deskflow configuration

Windows 11 (Server) + macOS Sequoia (Client)
- Each computer has a single monitor
- Windows server in the center, Mac client is on the side
- Mouse and keyboard shared via Deskflow — text clipboard works fine
- Using Continuous Build from GitHub Actions on both machines

### Log output

<details><summary><h3> Show log </h3></summary>

```ceylon
[2026-05-27T11:03:08.480] INFO: entering screen
[2026-05-27T11:03:11.085] INFO: switch from "Meta" to "Metas-MacBook" at 0,287
[2026-05-27T11:03:11.085] INFO: leaving screen
[2026-05-27T11:03:15.853] INFO: screen "Metas-MacBook" grabbed clipboard 0 from "Metas-MacBook"
[2026-05-27T11:03:15.853] INFO: screen "Metas-MacBook" grabbed clipboard 1 from "Metas-MacBook"
[2026-05-27T11:03:16.085] INFO: switch from "Metas-MacBook" to "Meta" at 1913,649
[2026-05-27T11:03:16.086] INFO: entering screen
[2026-05-27T11:03:16.151] INFO: screen "Metas-MacBook" updated clipboard 0
[2026-05-27T11:03:16.183] INFO: screen "Metas-MacBook" updated clipboard 1
[2026-05-27T11:03:26.152] INFO: screen "Meta" grabbed clipboard 0 from "Metas-MacBook"
[2026-05-27T11:03:26.152] INFO: screen "Meta" grabbed clipboard 1 from "Metas-MacBook"



Mac log：
[2026-05-27T11:02:29.128] WARNING: can't get the active group, use the first group instead
[2026-05-27T11:02:37.458] INFO: leaving screen
[2026-05-27T11:02:37.461] WARNING: cursor may not be visible
[...]

PR #9728, opened 2026-05-10
Fix image copy/paste from macOS

nbolton:
## Description

Fixes copy/paste of images from macOS to Windows and Linux. macOS puts images on the clipboard in a format the old code didn't fully handle, so the receiving side saw a distorted (Windows) or blank (Linux/Wayland) image. The full image data is now preserved.

## AI tools used for this PR

Claude Opus 4.7 - Found root-cause and suggested fix.

## Related Issue

Fixes: #9759 
Fixes: #9409
Fixes: #7255

Related: #8412, #9640, #9146, #9288, #9279

## How Has This Been Tested?

Tests:
- macOS client -> Windows client: image pastes correctly (was distorted).
- macOS client -> Linux/Wayland server: image pastes correctly (was blank).

Environment:
- [Paintbush](https://paintbrush.sourceforge.io/) app on macOS client
- [Drawing](https://maoschanz.github.io/drawing/) on Linux server
- MS Paint on Windows client

Issue #8412, opened 2025-03-28
Distorted Images Pasting from Mac Client Clipboard to Windows Server


throwawayburner69:
### Project

Deskflow

### Deskflow version info

1.21 - both Windows server and Mac Client

### Build type

Deskflow package (downloaded from Deskflow)

### Operating systems (OS)

- [x] Windows
- [x] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

- [ ] I have reviewed the Wayland [known issues](https://github.com/deskflow/deskflow/discussions/7499) and my issue is new
- [ ] I am not using Wayland on Linux

### Signing on macOS

- [x] I have authorized the app to run on my Mac
- [ ] I am not using macOS

### Continuous build

- [ ] I have tried the latest continuous build and the issue persists
- [ ] I am unable to try the latest continuous build

### OS versions/distros

- Windows 11
- MacOS 15.3.2

### Deskflow configuration

- Windows 11 server
- MacOS client

### What steps will reproduce the problem?

Copying any image from the Mac client and attempting to paste it in virtually any Windows app results in a corrupt image. Corrupt = either unable to display all together or "incomplete" missing pixels (think getting 8-bit versions on paste) into MS Paint.  Note: I've already set the server clipboard size to 300mb with no change.

Also, I'm not certain if directly related, but when I attempt to paste bigger screenshots, that can almost always trigger the Mac client to stop/error out (see https://github.com/deskflow/deskflow/issues/8411).

Below is the original image with 1) results after paste in MS Paint and 2) the same paste into the web version of Teams (note this is similar in most other apps).

Original (from Mac client):

![Image](https://github.com/user-attachments/assets/705863ef-6d00-41e7-989f-10b9b4cd15aa)

When pasting into MS Paint:

![Image](https://github.com/user-attachments/assets/7fef26dd-d53a-4a82-85da-959e74913972)

When pasting into the web version of Teams:

![Image](https://github.com/user-attachments/assets/ab96b424-1d71-442e-9467-8ae1805c4f4e)

### Log output

```shell

```

throwawayburner69:
> There is a workaround to this issue, which might also provide insight on how to fix it permanently.
> 
> For some reason, the default .png screenshot format in macOS causes this bug (when pasting in Windows after switching).
> 
> The workaround is to change the default screenshot format to jpeg/jpg on macOS:
> 
[...]

PR #10057, opened 2026-08-13
fix(win): normalise malformed macOS clipboard DIB

zlkent:
## Description

Normalise a malformed 32-bit DIB received from macOS before publishing it as `CF_DIB` on Windows.

Some macOS clipboard images declare a `BITMAPV5HEADER` and `BI_BITFIELDS`, but contain only a `BITMAPINFOHEADER` followed directly by BGRA pixels. Windows then interprets the first pixel bytes as V5 colour masks, which can make images appear blank, transparent, or severely distorted in paste targets such as WeChat.

The converter now detects this contradictory layout from the header fields and exact 32-bit pixel payload size, rewrites it as a canonical `BITMAPINFOHEADER` with `BI_RGB`, and preserves the pixel bytes.

## AI tools used for this PR

OpenAI Codex (GPT-5) was used to inspect clipboard DIB metadata, prepare the patch, and add the regression test. The final behavior was validated manually on a Windows 11 server with a macOS client.

## Fixed/related issues

Related: #8412, #9409, #9790, #9728

## How has this been tested?

- Added `MSWindowsClipboardTests::normalisesMalformedMacBitmap` to cover a DIB that declares a V5 header while containing only an INFOHEADER-sized payload.
- Built the Windows x64 package and ran the Windows unit-test suite in GitHub Actions.
- Manual verification: macOS screenshots and copied images pasted correctly on the Windows 11 server, including WeChat, after installing the build.

### DIFF
--- a/src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
+++ b/src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
@@ -10,6 +10,47 @@
 
 #include "base/Log.h"
 
+#include <QtEndian>
+
+#include <cstdlib>
+#include <limits>
+
+namespace {
+bool normaliseMalformedMacDib(const std::string &data, std::string &normalisedData)
+{
+  if (data.size() < sizeof(BITMAPINFOHEADER)) {
+    return false;
+  }
+
+  const auto *header = reinterpret_cast<const BITMAPINFOHEADER *>(data.data());
+  if (header->biWidth <= 0) {
+    return false;
+  }
+
+  const auto width = static_cast<size_t>(header->biWidth);
+  const auto height = static_cast<size_t>(std::abs(static_cast<int64_t>(header->biHeight)));
+  if (height == 0 || width > (std::numeric_limits<size_t>::max() - sizeof(BITMAPINFOHEADER)) / 4 / height) {
+    return false;
+  }
+  const auto expectedSize = sizeof(BITMAPINFOHEADER) + width * height * 4;
+
+  // macOS can describe an INFOHEADER-sized 32-bit pixel payload as a V5 DIB.
+  // Windows then interprets the first pixels as V5 colour masks. The pixel
+  // bytes are ordinary BGRA, so publish a canonical BI_RGB DIB instead.
+  if (header->biSize <= sizeof(BITMAPINFOHEADER) || header->biPlanes != 1 || header->biBitCount != 32 ||
+      header->biCompression != BI_BITFIELDS || expectedSize != data.size()) {
+    return false;
+  }
+
+  normalisedData = data.substr(0, sizeof(BITMAPINFOHEADER));
+  qToLittleEndian<quint32>(sizeof(BITMAPINFOHEADER), reinterpret_cast<quint8 *>(&normalisedData[0]));
+  qToLittleEndian<quint32>(BI_RGB, reinterpret_cast<quint8 *>(&normalisedData[0]) + 16);
+  normalisedData += data.substr(sizeof(BITMAPINFOHEADER));
+  LOG_INFO("normalised malformed macOS clipboard image to BI_RGB");
+  return true;
+}
+} // namespace
+
 //
 // MSWindowsClipboardBitmapConverter
 //
@@ -27,13 +68,19 @@ UINT MSWindowsClipboardBitmapConverter::getWin32Format() const
 HANDLE
 MSWindowsClipboardBitmapConverter::fromIClipboard(const std::string &data) const
 {
+  std::string normalisedData;
+  const auto *clipboardData = &data;
+  if (normaliseMalformedMacDib(data, normalisedData)) {
+    clipboardData = &normalisedData;
+  }
+
   // copy to memory handle
-  HGLOBAL gData = GlobalAlloc(GMEM_MOVEABLE | GMEM_DDESHARE, data.size());
+  HGLOBAL gData = GlobalAlloc(GMEM_MOVEABLE | GMEM_DDESHARE, clipboardData->size());
   if (gData != nullptr) {
     // get a pointer to the allocated memory
     char *dst = (char *)GlobalLock(gData);
     if (dst != nullptr) {
-      memcpy(dst, data.data(), data.size());
+      memcpy(dst, clipboardData->data(), clipboardData->size());
       GlobalUnlock(gData);
     } else {
       GlobalFree(gData);
--- a/src/unittests/platform/MSWindowsClipboardTests.cpp
+++ b/src/unittests/platform/MSWindowsClipboardTests.cpp
@@ -9,6 +9,9 @@
 #include "MSWindowsClipboardTests.h"
 
 #include "platform/MSWindowsClipboard.h"
+#include "platform/MSWindowsClipboardBitmapConverter.h"
+
+#include <QtEndian>
 
 void MSWindowsClipboardTests::initTestCase()
 {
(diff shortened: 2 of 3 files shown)


sithlord48:
didnt  #10023 fix this ?

zlkent:
No, this is a separate (but related) DIB issue, so I believe this PR should remain open.

#10023 changes `MSWindowsClipboardBitmapConverter::toIClipboard()`. It fixes a crash while Windows is reading and serialising a top-down DIB: a negative `biHeight` previously caused the output-buffer size calculation to wrap.

This PR changes `fromIClipboard()`, the macOS-to-Windows receive path. In the failing captures, the payload declares a V5/`BI_BITFIELDS` header (`biSize = 124`) but its total byte count is only `sizeof(BITMAPINFOHEADER) + width * abs(height) * 4`. In other words, the BGRA pixels begin at byte 40. Passing that contradictory payload to `CF_DIB` makes Windows paste targets interpret the first pixels as V5 colour masks, producing blank, transparent, or distorted images.

#10023 leaves `fromIClipboard()` as a raw copy, so it does not normalise this malformed layout. I reproduced the problem with macOS screenshots on a Windows 11 server and verified that this conversion fixes paste into WeChat as well as other Windows targets. I also added a regression test for the malformed V5 declaration.

sithlord48:
@nbolton @amagus  since this is a mac os / windows thing can one of you look at this please.

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
```diff
@@ -10,6 +10,41 @@
...
+std::string normaliseMalformedMacDib(const std::string &data)
+{
+  if (data.size() < sizeof(BITMAPINFOHEADER)) {
+    return data;
+  }
+
+  const auto *header = reinterpret_cast<const BITMAPINFOHEADER *>(data.data());
+  const auto width = static_cast<int64_t>(header->biWidth);
+  const auto height = static_cast<int64_t>(header->biHeight);
+  const auto absoluteHeight = height < 0 ? -height : height;
```
amagus: use std::abs instead

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
```diff
@@ -27,13 +62,15 @@ UINT MSWindowsClipboardBitmapConverter::getWin32Format() const
 HANDLE
 MSWindowsClipboardBitmapConverter::fromIClipboard(const std::string &data) const
 {
+  const auto normalisedData = normaliseMalformedMacDib(data);
```
amagus: this always creates a copy of the data... you will be consuming a double the memory all the time... this bitmap can be big
Please change the approach to only duplicate the data if normalization is indeed needed.

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
```diff
@@ -10,6 +10,41 @@
...
+    return data;
+  }
+
+  const auto *header = reinterpret_cast<const BITMAPINFOHEADER *>(data.data());
+  const auto width = static_cast<int64_t>(header->biWidth);
+  const auto height = static_cast<int64_t>(header->biHeight);
+  const auto absoluteHeight = height < 0 ? -height : height;
+  const auto expectedSize =
+      width > 0 && absoluteHeight > 0
+          ? sizeof(BITMAPINFOHEADER) + static_cast<uint64_t>(width) * static_cast<uint64_t>(absoluteHeight) * 4
```
amagus: this looks weird... width variable is signed... converting it to uint here looks wrong... don't you need to do absolute conversion on it as well?

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
amagus: why not declare as uint64_t?

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
amagus: absoluteHeight may not need the static_cast at all.

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
```diff
@@ -10,6 +10,41 @@
...
+{
+  if (data.size() < sizeof(BITMAPINFOHEADER)) {
+    return data;
+  }
+
+  const auto *header = reinterpret_cast<const BITMAPINFOHEADER *>(data.data());
+  const auto width = static_cast<int64_t>(header->biWidth);
+  const auto height = static_cast<int64_t>(header->biHeight);
+  const auto absoluteHeight = height < 0 ? -height : height;
+  const auto expectedSize =
```
amagus: use size_t type given it will be compared to a string::size()

src/unittests/platform/MSWindowsClipboardTests.cpp
```diff
@@ -136,4 +139,30 @@ void MSWindowsClipboardTests::isOwnedByDeskflow()
...
+  constexpr qsizetype headerSize = sizeof(BITMAPINFOHEADER);
+  std::string dib(headerSize + 4, '\0');
+  auto *raw = reinterpret_cast<quint8 *>(&dib[0]);
+  qToLittleEndian<quint32>(sizeof(BITMAPV5HEADER), raw); // claims V5 but only has an INFOHEADER
+  qToLittleEndian<quint32>(1, raw + 4);
+  qToLittleEndian<quint32>(-1, raw + 8);
+  qToLittleEndian<quint16>(1, raw + 12);
+  qToLittleEndian<quint16>(32, raw + 14);
+  qToLittleEndian<quint32>(BI_BITFIELDS, raw + 16);
+  raw[headerSize + 3] = 0xff;
```
amagus, line 154: this looks like magic... can you please add a short explanation of what this data represents? (one single line stating something like "simple 1x1 mac v5 bitmap" is more than enough)

amagus (changes requested):

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
zlkent: Updated to use `std::abs`.

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
zlkent: Changed this so the original buffer is used by default. A second buffer is created only when the DIB actually needs normalization.

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
zlkent: The code now rejects non-positive widths before converting to `size_t`, so the size calculation cannot reinterpret a negative width as a large unsigned value.

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
zlkent: I used `size_t` for the checked dimensions and expected byte count because the result is compared with `std::string::size()`. The overflow guard is expressed in the same type.

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
zlkent: I keep the widening cast before `std::abs`: `biHeight` is signed 32-bit, and widening first avoids the minimum-negative-value edge case before converting the result to `size_t`.

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
zlkent: Updated the dimensions and expected byte count to `size_t`, matching `std::string::size()` and the allocation API.

src/unittests/platform/MSWindowsClipboardTests.cpp
zlkent: Added a short comment: this fixture represents a 1x1 top-down macOS DIB that incorrectly declares a V5 header.

zlkent:
@amagus All seven review points have been addressed in efa8bf1. The updated change passed a Windows x64 package build and 25/25 unit tests, and the resulting package was also verified with the original macOS-to-Windows clipboard scenario. Could you please re-review when convenient?

amagus:
don't fix a commit with another one, use --amend

zlkent:
@amagus Done — I amended the original fix commit and force-pushed the PR branch, so the review changes are now contained in a single commit.

src/unittests/platform/MSWindowsClipboardTests.cpp
amagus: resolved.

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
amagus: resolved

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
amagus: resolved

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
amagus: resolved

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
amagus: resolved

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
amagus: resolved

src/lib/platform/MSWindowsClipboardBitmapConverter.cpp
amagus: resolved

src/unittests/platform/MSWindowsClipboardTests.cpp
```diff
@@ -136,4 +139,31 @@ void MSWindowsClipboardTests::isOwnedByDeskflow()
   QVERIFY(clipboard.isOwnedByDeskflow());
 }
 
+void MSWindowsClipboardTests::normalisesMalformedMacBitmap()
```
nbolton, line 142: The test covers a broken image getting repaired, but it seems to me repairing a good V5 DIB would corrupt it, and only the size check stops that. 

Could a second case prove a good image passes through unchanged? i.e. Give it a healthy image and check it comes out byte-for-byte identical, proving the repair code never touches images that aren't broken.

zlkent: Added `preservesHealthyMacV5Bitmap`, which builds a complete 1x1 top-down V5 DIB with BGRA masks and a pixel, then verifies that `fromIClipboard()` returns it byte-for-byte unchanged. The Windows x64 validation build compiled the test and passed all 25 tests. Could you please re-review when convenient?

nbolton: Thank you!

nbolton (approved):

PR #10058, opened 2026-08-13
Better readonly settings

### LINKED ISSUE #10047: Warn when settings cannot be saved because the settings file is read-only

sithlord48:
## Description

 Provide a better more standard way to show non writable settings
 - Settings new signal emitting when the read only state of the setting file chages
 - New Settings Dialog button Box with info about read only state
 - Use SettingsDialogButtonBox in SettingsDialog and Client Config Dialog
 - Client Config Dialog enable / disable controls if not writable 
 - Add missing translation for the clear settings button
 - Add log if settings trying to save to a readonly file.
 - Basic report of readonly settings in server config dialog 
 - disable items on ServerConfigDialog advanced tab for readonly settings

## AI tools used for this PR

None
## Fixed/related issues

fixes: #10047 

## How has this been tested?

Locally
open the settings and make the settings file (Deskflow.conf) readonly and back you should see the dialog react to the file being readonly

### DIFF
--- a/src/lib/common/Settings.cpp
+++ b/src/lib/common/Settings.cpp
@@ -12,6 +12,7 @@
 
 #include <QCoreApplication>
 #include <QFile>
+#include <QFileSystemWatcher>
 #include <QRect>
 #include <QRegularExpression>
 #include <QStandardPaths>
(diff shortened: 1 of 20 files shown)


nbolton:
Sorry mate, still seeing this bug from #10049 (closed):

> Bug: the message doesn't show when I make a change to the screen layout
> 
> 1. `chmod 444 ~/.config/Deskflow/Deskflow.conf`
> 2. Start GUI
> 3. Edit screen layout
> 4. Click 'OK'
> 
> Expect: Read only message Actual: No message, and when I restart the GUI the old settings are restored

What happens when you do these steps?

sithlord48:
This does not include changes to the server settings dialog but i can toggle them to be readonly or not via file permission and the dialog for settings / client config is updated.

nbolton (approved):

PR #10060, opened 2026-08-13
Avoid macOS 27 system tray crash

rafiki270:
## Summary

- avoid creating a system-tray status item on macOS 27, where opening its menu aborts the Qt Cocoa plugin
- keep Deskflow's main window visible when the tray is unavailable
- do not hide the main window into an unavailable tray

## Root cause

On macOS 27, the Qt Cocoa platform plugin can receive a non-mouse current event while tracking an `NSStatusItem` menu. It unconditionally reads `clickCount`, which raises an Objective-C exception and terminates the Deskflow GUI. The sharing core remains alive, but the user loses the Deskflow controls.

The Qt 6.12.0 package used for this reproduction still exhibited the crash, so this change avoids the unsafe status-item path only on macOS 27. A normal Deskflow window remains available for starting, stopping, and configuring the sharing core.

## Testing

- built `Deskflow` for Apple Silicon macOS 27 with Qt 6.12.0
- reproduced the pre-fix crash in `-[NSEvent clickCount]` from `libqcocoa.dylib`
- launched the signed, self-contained patched application and verified that both `Deskflow` and `deskflow-core` remained running
- ran the existing unit-test binaries with Qt 6.11.1 and Qt 6.12.0: 25 passed; `OSXKeyStateTests::fakePollShift` failed identically on macOS 27 before this change
- not tested on Intel macOS or GitHub Actions

### DIFF
--- a/src/lib/gui/MainWindow.cpp
+++ b/src/lib/gui/MainWindow.cpp
@@ -43,6 +43,7 @@
 #include <QMessageBox>
 #include <QNetworkAccessManager>
 #include <QNetworkInterface>
+#include <QOperatingSystemVersion>
 #include <QPushButton>
 #include <QRegularExpression>
 #include <QRegularExpressionValidator>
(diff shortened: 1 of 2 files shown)


sithlord48:
No problem bumping to Qt 6.12 when its released. But it maybe a bit too soon to start shipping it while its in RC

sithlord48:
Replaced by #10088 .

Issue #8583, opened 2025-05-09
Can the pointer in deskflow be hidden when it is not on a specific device?


half-lang:
### Feature Request

when the point leave one screen ,the pointer always stay in the position ,and when I move back the pointer will teleportation to another position

sithlord48:
the cursor should hide when off screen., Without providing any information no one will be able to help you.

nbolton:
@half-lang What OS is your server/client?

sithlord48:
I have seen this recently on mac os
It seams to randomly happen if the log is minimized (collapsed)

Joe-Schmuccatelli:
I have the same issue with Deskflow 1.22.0 on Fedora 42 with Wayland. I did NOT see this on Fedora 42 with X11 or Windows Server 2025.

sithlord48:
@Joe-Schmuccatelli  when it happens do you have the log collapsed or not ? for me it seams to happen only if the log is in a collapsed state.

Joe-Schmuccatelli:
It happens all of the time. It happens when the log is open, collapsed or
when the app is hidden.

Oddly enough, when I noticed this issue earlier today, it was happening on
both the server and the clients. I only see it on the client now. I do NOT
see it on the server. Fedora 42 pushed a half gig of updates on my machines
tonight. They're all different hardware. Something may have "healed" on the
server. Who knows?

nbolton:
> I only see it on the client now.

What OS is the client and are you saying you only get cursor visibility issues on the client now?

Joe-Schmuccatelli:
The client and server are Fedora 42 & KDE Plasma with the latest
generally available updates as of tonight EST. I only see it on the client
when running Wayland.

FYI: I installed the Fedora 42 OS that I'm using several months ago with the Fedora 42 Everything GUI based installer. All of my machines are bare bones with little more than KDE, X11, LibreOffice, Brave browser and some diagnostic tools. Deskflow was installed and is updated via the Fedora Discover software management app. If the Discover app doesn't push it, then it doesn't get updated. The Deskflow version that I'm currently running is 1.22. 

As soon as I see that updates are available, I apply the updates to one or more of the machines that I happen to be using at that time.

nbolton:
OK, I see. So on a Wayland client, mouse cursor hiding is not yet implemented. This is something we would like to do, but I haven't dug into it yet so don't know if it's possible with the current Wayland landscape. Hopefully it's just a dbus call.

[...]

PR #10066, opened 2026-08-15
feat(wayland): moves mouse pointer to the center of screen when leaving

piterpunk:
## Blocked

- #8583

## Description
The mouse pointer is centralized on screen in X11 and Windows when it leaves the screen. In some situations (locked screen on X11 or an window at center of screen in Windows) the pointer is still visible, but when it disappears at edge of screen we got a good visual feedback that the mouse switched screens.

The same doesn´t  happens for Wayland clients, with the mouse pointer visible at its last position at edge of screen. This PR just makes the Wayland behaviour the same of X11 and Windows, with the mouse pointer going to the center of screen of the inactive screen when leaving one screen to another.

## AI tools used for this PR
None

## Fixed/related issues
This is related to #8583 , the mouse will not be hidden but, at least, it will disappear at screen edge.

## How has this been tested?
Yes, it is tested with both server and client using Plasma on Wayland and a deskflow package generated by this PR's tree.

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -438,6 +438,7 @@ bool EiScreen::canLeave()
 void EiScreen::leave()
 {
   if (!m_isPrimary) {
+    fakeMouseMove(m_w / 2, m_h / 2);
     stopEmulating();
   }
 


sithlord48:
I know this is what happens on other platforms but idk that i like that (even on those) without it hiding its makes it even worse, but again we are also doing this on all other platforms..

nbolton (commented):
Hmm, so I think until we have cursor hiding widely available on Wayland, this could get quite annoying for users... @sithlord48 thoughts?

sithlord48:
> Hmm, so I think until we have cursor hiding widely available on Wayland, this could get quite annoying for users... @sithlord48 thoughts?

This is what i was thinking its distracting now when i leave my client due to the motion

nbolton:
Hmm, I think we might need to draft this until the cursor hiding bug is fixed:
- #8583

normanr:
btw, on X11 the mouse cursor is only centered when _connecting_ (not disconnecting). I've always considered it a bug, and would want a way to turn off the behavior (if it were implemented for Wayland).

sithlord48:
@normanr if its made into an option it would be for all backends

PR #10070, opened 2026-08-16
ci: macOS artifacts are now signed and notarized

amagus:
## Description
This allows CI to deliver a signed and notarized Deskflow.app for macOS

**Depends on #10087 landing first.**

Before continuing, a codesigning identity is needed.

A paid Apple Developer is required for signing and notarizing.

Follow the following instructions to generate a developer id certificate: https://developer.apple.com/help/account/certificates/create-developer-id-certificates/

Then, to use it, there's the need to define the following secrets in the repository:

**APPLE_CODESIGN_DEV**

Should contain the full name of the signing identity to be used to sign the application.

After installing your identity locally you can run `security find-identity -v` to find the full name of your identity.
It should look something like:
`Developer ID Application: <dev name> (<team id>)`

**APPLE_CODESIGN_P12**

Should contain the P12 container file in base64. 

Please ensure both the certificate and private key are exported.

In order to generate this, in your local machine where you have the identity installed, head to "Keychain Access", select your identity, right click on it and chose export. Please ensure to use a strong password to export it. Save the p12 file somewhere. Then run `base64 <p12file>` on a terminal to get its content in base64. You can run `base64 <p12file> | pbcopy` to directly copy its content to the clipboard.

**APPLE_CODESIGN_P12_PWD**

Should contain the P12 password in plain text. 

This is what was used in "Keychain Access" in order to generate APPLE_CODESIGN_P12.

**APPLE_NOTARIZE_ID**

Should contain the Apple ID of an user from the team of the Developer ID certificate.

We recommend not to use the main Apple ID of the developer account, instead create a dedicated user account on the App Store connect with the "App manager" role. Even individual developers may create those secondary accounts and they will have notarization permissions.

**APPLE_NOTARIZE_PWD**

Should contain a application specific password for that given Apple ID
It won't work with the real account password.
Head to https://account.apple.com to generate one for that account.

## Fixed/related issues
N/A

## How has this been tested?
Tested in my local fork ([amagus/deskflow](https://www.github.com/amagus/deskflow))

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -113,15 +113,15 @@ jobs:
 
           - name: "macos-arm64"
             runs-on: macos-15
-            timeout: 10
+            timeout: 20
             qt-version: 6.11.2
             config-args: '-DCMAKE_OSX_ARCHITECTURES="arm64" -DCMAKE_OSX_DEPLOYMENT_TARGET=14'
 
           - name: "macos-x64"
             runs-on: macos-15-intel
-            timeout: 20
+            timeout: 30
             qt-version: 6.9.3
-            config-args: '-DCMAKE_OSX_ARCHITECTURES="x86_64" -DCMAKE_OSX_DEPLOYMENT_TARGET=12 -DCMAKE_OSX_SYSROOT=/Applications/Xcode_16.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk'
+            config-args: '-DCMAKE_OSX_ARCHITECTURES="x86_64" -DCMAKE_OSX_DEPLOYMENT_TARGET=12 -DCMAKE_OSX_SYSROOT=/Applications/Xcode_16.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk ${APPLE_CODESIGN_DEV:+-DAPPLE_CODESIGN_DEV="$APPLE_CODESIGN_DEV"}'
 
           - name: "debian-x86_64"
             runs-on: ubuntu-latest
@@ -271,7 +271,17 @@ jobs:
         uses: ./.github/actions/get-version
 
       - name: Configure
+        env:
+          APPLE_CODESIGN_DEV: ${{ secrets.APPLE_CODESIGN_DEV }}
         run: ${{env.CMAKE_CONFIGURE}} ${{ matrix.target.config-args }} ${{ steps.get-deps.outputs.vcpkg-cmake-config }} -DPACKAGE_VERSION_LABEL="${{env.DESKFLOW_PACKAGE_VERSION}}"
+
+      - name: Import code-signing certificate
+        if: runner.os == 'macOS' && ((github.ref == 'refs/heads/master') || (contains(github.ref, '/tags/v')))
+        uses: ./.github/actions/codesign-keychain
+        with:
+          action: import
+          apple-codesign-p12: ${{ secrets.APPLE_CODESIGN_P12 }}
+          apple-codesign-p12-pwd: ${{ secrets.APPLE_CODESIGN_P12_PWD }}
 
       - name: Build
         shell: bash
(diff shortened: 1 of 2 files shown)


nbolton:
Still waiting on Apple support to fix the typo in the cert name. Will set the repo secrets when I have them.

amagus:
I still need to work on removing the references for all the things users had to do in order to make Deskflow work without a signature. But I wanted you guys to review it and request changes before, if needed.

.github/workflows/build.yml
```diff
@@ -0,0 +1,222 @@
...
+      - name: Import codesign certificate
+        if: ${{ inputs.use-apple-codesign }}
+        shell: bash
+        env:
+          APPLE_CODESIGN_P12: ${{ secrets.APPLE_CODESIGN_P12 }}
+          APPLE_CODESIGN_BUILD_PWD: ${{ secrets.APPLE_CODESIGN_BUILD_PWD }}
+          APPLE_CODESIGN_P12_PWD: ${{ secrets.APPLE_CODESIGN_P12_PWD }}
+        run: |
+          CERTIFICATE_PATH=$RUNNER_TEMP/build_certificate.p12
+          KEYCHAIN_PATH=$RUNNER_TEMP/build.keychain
```
nbolton: A temporary password would mean less secrets to maintain and would work the same.

```suggestion
          KEYCHAIN_PATH=$RUNNER_TEMP/build.keychain
          KEYCHAIN_PWD=$(openssl rand -base64 24)
```

nbolton (changes requested):

.github/workflows/build.yml
amagus: Indeed, in one of my previous versions I was reusing the keychain later (in order to sign the DMG and notarize it). 
But I found out that a app manager account is safer...
I will update it with your suggestion and with the suggestions from @sithlord48 as soon as I have some free time

nbolton:
These secrets are set now:
- `APPLE_CODESIGN_P12`
- `APPLE_CODESIGN_P12_PWD`

Edit: We'll need to run the test on a PR using a branch from the repo, as your personal branch wont have access to the secrets. We can do this after you make the `KEYCHAIN_PWD` change.

Edit: Corrected env name.

amagus:
Sure. I am working on making the changes to use actions instead of workflows as suggested by @sithlord48. 
Recent changes work fine on my fork (using my credentials).
I expect to have some free time on the weekend to finish it up before we test it with official credentials in the official repo.
Anyway, please note that notarization requires 3 extra secrets now (and I am trying to make it 2, given we can get the team id from the developer id name).

One extra question, @nbolton: when generating the developer id certificate, have you generated it with the extended validity? 
The Apple website defaults usually to the old CA that generates certificates valid only until Feb/2027.

nbolton:
> extended validity

I used the defaults. Apparently it expires 22 Aug 2031, and I assumed this would just carry through with the p12 export. Let me know if I need to do something different.

.github/workflows/build.yml
nbolton: No rush :)

amagus:
That's exactly what is needed. Not sure why on my side it defaults to the old one. (Perhaps because I have old apps still compiling on old xcode...)

.github/workflows/build.yml
amagus: Should be resolved on the most recent version.

amagus:
The new version is up. I reduced the number of secrets needed and everything. The CI on this PR is unable to fetch the secrets, not sure if on purpose. Perhaps we need to setup a new branch with CI, as @nbolton  suggested.

.github/workflows/continuous-integration.yml
```diff
@@ -271,44 +271,75 @@ jobs:
         uses: ./.github/actions/get-version
 
       - name: Configure
+        if: runner.os != 'macOS'
         run: ${{env.CMAKE_CONFIGURE}} ${{ matrix.target.config-args }} ${{ steps.get-deps.outputs.vcpkg-cmake-config }} -DPACKAGE_VERSION_LABEL="${{env.DESKFLOW_PACKAGE_VERSION}}"
 
+      - name: Configure (macOS)
+        if: runner.os == 'macOS'
+        env:
+          APPLE_CODESIGN_DEV: ${{ secrets.APPLE_CODESIGN_DEV }}
```
sithlord48, line 275: can't we just add `${APPLE_CODESIGN_DEV:+-DAPPLE_CODESIGN_DEV="$APPLE_CODESIGN_DEV"}` to the os "config-args"  for mac os ?

amagus: will do the needed changes, just be aware this requires migrating configure to use bash (windows defaults to powershell)
I wished my changes to touch as little as possible the current working flow, that’s why it has the different steps with one exclusive to mac

sithlord48: That is fine we are using bash in a few other places for windows shell as well.

amagus: In the end I wouldn't need that. I could edit the config args directly on the target matrix.

sithlord48 (dismissed):
I like this much much better !!!! 
Can land (IMHO) when the secrets are ready

nbolton:
> These secrets are set now:
> 
> * `APPLE_CODESIGN_P12`
> * `APPLE_CODESIGN_P12_PWD`

I realised I didn't set `APPLE_CODESIGN_DEV` -- fixed now.

Edit: Added: `APPLE_NOTARIZE_ID` and `APPLE_NOTARIZE_PWD`

image

nbolton (approved):

sithlord48 (approved):

PR #10079, opened 2026-08-21
fix(wayland): sync desktop shape from portal zones under scale≠1

songlairui:
## Summary

- On scaled Wayland compositors (notably **Hyprland** with fractional/integer scale such as 1.25 / 1.6 / 2), Input Capture **portal zones/barriers** use **logical** pixels while libei device regions often report **physical** pixels.
- Deskflow previously used the EIS size for `EiScreen` shape, so barrier activation (e.g. x=2400 at scale 1.6) never matched edge leave (shape width 3840) → activate → immediate release loop; cursor/keyboard got stuck.
- Prefer the portal zone union via `setPortalDesktopShape`, keep that shape when EIS later reports physical size, and only `release()` capture if the server is still on the primary after handling the edge motion (`DeliverImmediately`).

Verified locally: Omarchy/Hyprland server @ scale 1.6 + macOS Deskflow client; edge switch works with logical shape `2400×1350`.

## Test plan

- [ ] Wayland server (Hyprland) with monitor scale **1** — edge switch still works
- [ ] Same setup with scale **1.6** or **2** — cursor leaves to client without activate/release loop
- [ ] Logs show `portal desktop shape (logical): …` matching barrier coordinates
- [ ] Multi-monitor server (if available) — outer barriers still apply; no regression vs existing union-boundary logic
- [ ] Return from client to primary still releases capture via `enter()`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -8,6 +8,7 @@
 
 #include "platform/EiScreen.h"
 
+#include "base/Event.h"
 #include "base/IEventQueue.h"
 #include "base/Log.h"
 #include "common/Constants.h"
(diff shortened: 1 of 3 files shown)


sithlord48 (changes requested):

sithlord48:
Have you tried this on other compositors, This sounds like an issue for hyperland. 

This is working fine on Plasma I have two scaled and one non scaled screen on my server.. I also have clients that are both wayland and scaled. (also Plasma). I know others have similar setups with GNOME as well.

PR #10087, opened 2026-08-22
build(mac): enable options needed for the upcoming support of signed/notarized releases

amagus:
## Description

This PR introduces several minor changes to the build process that are required in order to obtain properly signed and notarized versions of Deskflow for macOS.

This effectively:
- Ensures no files are added to the application folder after calling qt's deploy
- Enables deep signing: may cause some extra delay on building (if configured with APPLE_CODESIGN_DEV)
- Enables hardened runtime: may cause macOS to block certain functionalities like external command execution or JIT

## AI tools used for this PR
N/A

## Fixed/related issues
N/A

## How has this been tested?
Tested locally on macOS Tahoe 26.6.2 (as server)

Locally I was not able to hit any real problem.
Reviewers, please do wider testing on different configurations if possible, particularly with external command execution (as triggered by entering/exiting screens, for example).


sithlord48 (approved):
Tested on mac os client enter / exit commands are working still

PR #10092, opened 2026-08-24
fix(mac): reveal cursor on local client input

### LINKED ISSUE #8709: Make mouse visible on clients when the client's own mouse is moved.
### Feature Request

Currently, with the MacOS client at least, when the deskflow pointer is on a different computer, if the MacOS client mouse is moved, the client's pointer (which was made invisible when the mouse left the client screens) will stay invisible, which is inconvenient, because it requires a few clicks (in unknown places, since the pointer is invisible) to get the pointer to be visible once again. 

My current workaround: move the client pointer (using the client's mouse) to a corner of the screen where it is safe to click, then click until the pointer appears. 

Suggested short-term workaround: have an option for the client's pointer to not become invisible when the deskflow pointer moves off screen. This should typically result in the pointer staying visible near the boundary with the other computer.

Suggested solution: allow the client software to listen for mouse movement from any mice attached to the client, the same way the server software listens for movement from the server's mouse. If movement is detected from the client's own mouse, then make the pointer visible.

Changroro:
## Description

Remove the early return from the existing macOS secondary input handler so local mouse or trackpad movement reveals the hidden cursor while the shared cursor is on another screen.

The only other change corrects the format specifiers in the debug log that becomes reachable.

## AI tools used for this PR

OpenAI Codex (GPT-5 and GPT-5.6) was used for code analysis and review.

## Fixed/related issues

fixes: #8709

## How has this been tested?

- Compared clean upstream `c20870cc7` and patched builds with an Ubuntu 24.04 server and macOS 27.0 client
- Verified that the cursor stays hidden without local input and appears on the first small local trackpad movement
- Repeated Linux → Mac → Linux cursor transitions five times
- Built Release with warnings as errors
- Passed all 26 macOS tests, clang-format 20.1.0, REUSE lint, and `git diff --check`

### DIFF
--- a/src/lib/platform/OSXScreen.mm
+++ b/src/lib/platform/OSXScreen.mm
@@ -1660,17 +1660,13 @@ bool OSXScreen::HotKeyItem::operator<(const HotKeyItem &x) const
 CGEventRef
 OSXScreen::handleCGInputEventSecondary(CGEventTapProxy proxy, CGEventType type, CGEventRef event, void *refcon)
 {
-  // this fix is really screwing with the correct show/hide behavior. it
-  // should be tested better before reintroducing.
-  return event;
-
   OSXScreen *screen = (OSXScreen *)refcon;
   if (screen->m_cursorHidden && type == kCGEventMouseMoved) {
 
     CGPoint pos = CGEventGetLocation(event);
     if (pos.x != screen->m_xCenter || pos.y != screen->m_yCenter) {
 
-      LOG_DEBUG("show cursor on secondary, type=%d pos=%d,%d", type, pos.x, pos.y);
+      LOG_DEBUG("show cursor on secondary, type=%d pos=%.0f,%.0f", static_cast<int>(type), pos.x, pos.y);
       screen->showCursor();
     }
   }


src/lib/platform/OSXScreenCursor.h
```diff
@@ -0,0 +1,28 @@
+/*
```
sithlord48: Why not just add these to the OSXScreen itself?

src/lib/platform/OSXScreenCursor.h
```diff
@@ -0,0 +1,28 @@
...
+
+#pragma once
+
+#include <ApplicationServices/ApplicationServices.h>
+
+#include <cstdint>
+
+namespace deskflow::platform {
+
+inline constexpr int64_t SyntheticMouseEventTag = 0x4465736b666c6f77;
```
sithlord48: This magic number.., what does it do where did it come from ?

sithlord48 (changes requested):

amagus:
Is this ready for review? I see it's on draft.

sithlord48:
Idk a lot of ppl are putting up drafts and then tagging us for review... If this it not expected to be reviewed yet they should not make a pr yet .

src/lib/platform/OSXScreen.mm
```diff
@@ -1660,19 +1663,12 @@ static inline bool needsEventNumber()
 CGEventRef
 OSXScreen::handleCGInputEventSecondary(CGEventTapProxy proxy, CGEventType type, CGEventRef event, void *refcon)
 {
-  // this fix is really screwing with the correct show/hide behavior. it
-  // should be tested better before reintroducing.
-  return event;
```
amagus, line 1665: wouldn't the fix be just removing this return? why all the rest of the code you wrote is needed?

src/lib/platform/OSXScreenCursor.h
Changroro: Fixed in fedc372b0. The cursor-classification helper now lives on OSXScreen, and the separate header was removed.

src/lib/platform/OSXScreenCursor.h
Changroro: Fixed in fedc372b0. The arbitrary constant and custom event tag were removed; the handler now uses the standard macOS kCGEventSourceUnixProcessID field.

Changroro:
Hello, and I sincerely apologize for opening this PR before it was ready for review. I should have completed the isolated implementation and validation before creating a public PR.

The PR now contains only the macOS cursor fix, incorporates both review comments, and has been rebuilt and tested from a clean upstream revision.

When you have time, could you please review the updated change and approve the pending workflow runs? I understand that review and CI approval take maintainer time, and I appreciate your consideration.

I will ensure that future contributions are fully ready for review before opening a PR.

amagus:
Could you please answer the questions on the review? It's still not clear to me why do you need such changes.

Changroro:
You're right — my earlier replies explained what I changed, but not why the change is needed. Sorry about that.

Even while Deskflow is running, someone may still use the MacBook's own trackpad for a moment without first moving the shared pointer back to the Mac. In my setup, the Deskflow pointer stays on the Linux screen, but moving the MacBook trackpad still moves the Mac's local pointer. Hover and click targets change, but the pointer itself remains invisible, sometimes until the trackpad is moved around a lot.

This change does not move Deskflow focus back to the Mac. It only makes the Mac's local pointer visible when the user actually moves the local trackpad.

There were two separate points in the review that I did not answer clearly:

- I originally put the event check in a separate header because it made the logic easier to test without constructing `OSXScreen`. It did not need a separate header, so I moved it into `OSXScreen`.
- `0x4465736b666c6f77` was simply the word "Deskflow" encoded as hexadecimal and used as a custom marker. It was not a value provided by macOS. I removed both the value and the custom marker. The current code checks the standard `kCGEventSourceUnixProcessID` field that macOS already provides.

The old secondary event handler is currently disabled with an early return. Re-enabling it as-is can also react to mouse events generated by Deskflow, which is why the updated version ignores events coming from the Deskflow process and only shows the cursor after the pointer has moved away from the center position.

I tested this with a Linux server and a Mac client. The Deskflow pointer remained on Linux, and a small movement on the MacBook trackpad made the Mac cursor visible immediately.

If using the Mac's local trackpad this way is not something Deskflow intends to support, please let me know and I will close the PR.

amagus:
This is bug we are aware and we are looking for a proper fix. Anyway, you still haven’t answered the question I asked on the review. Please answer it shortly there. There’s no need for long texts (which, by the way, are likely AI-generated). I want to understand why the fix wouldn’t be just removing the early return on the function I marked. Why there’s all these checks including some pid checks? Keep the changes as minimal as possible.

Changroro:
Removing the early return is sufficient. I added the PID checks because I mistakenly thought Deskflow-generated mouse events needed to be filtered, but they are unnecessary and have been removed. Sorry.

src/lib/platform/OSXScreen.mm
amagus: resolved

amagus (approved):
Approved. Works on my local machine.

@nbolton you added this early exit in the past in order to avoid some kind of bug. I can't figure out exactly what is the bug you tried to avoid with that. Do you by any chance remember and can you check to see if the bug is still reproducible? Otherwise, please land this at your earliest convenience.

src/lib/platform/OSXScreen.mm
```diff
@@ -1660,17 +1660,13 @@ static inline bool needsEventNumber()
...
```
nbolton, line 1665: So, this early return was a fix from 2012 for a random show/hide bug (c34918fca1). We post the entry mouse move before `enter()` clears `m_cursorHidden`, so the tap fires on our own warp. 2x `CGDisplayShowCursor` per `CGDisplayHideCursor` on every entry, and Quartz needs those to equal otherwise it seems random. 

To test this properly: 10x round trips, does the cursor still hide?

sithlord48: Thanks Nick for your input , @amagus  can you try that ill try to do the same also

amagus: Tried with the patch on the client (macOS 27 Golden Gate, beta 7), I was able to move in and out without any problem. Perhaps the new approach we have that doesn't lock the cursor on the center of the screen fixed that bug?

sithlord48 (approved):

Changroro:
After using the patched build for a few days, I noticed that the cursor would occasionally remain visible on the Mac after moving the shared cursor back to Linux, so I tested repeated transitions again.

In my logs, it looked like a mouse-move event generated by Deskflow reached the secondary event tap and triggered an extra cursor show. I prepared a small candidate change that ignores only mouse moves from Deskflow’s own PID and keeps the show/hide calls balanced.

I tested the candidate with an Ubuntu 24.04.4 server and a macOS 27.0 client. All 26 tests passed. Local trackpad movement still revealed the cursor immediately, and the cursor did not remain on the Mac during 58 repeated transitions.

I have not pushed this change yet. Could you please check whether the same issue is reproducible on your setup? If so, would it be okay for me to update this PR with the candidate for another review? Sorry I did not catch this earlier.

amagus:
Why would the deskflow PID matter in this case? 
Which other process/PID would cause mouse movements? I am confused on why this is relevant.

I will try to reproduce the issue over here... 
Perhaps I didn't notice it because the transition out of my mac is on the right side (where cursor is almost always hidden). Maybe I will notice the problem if the cursor is on left or top side.
I will keep you posted.

Changroro:
The PID check is not intended to identify another application. It only distinguishes Deskflow’s own synthetic mouse movement from local hardware input. In my diagnostic, Deskflow-generated events had a source PID equal to `getpid()`, while physical trackpad events had PID 0. A queued Deskflow-generated event was causing the extra cursor show, so the candidate ignores only events from its own PID.

amagus:
@Changroro then you don't need to compare to getpid(), you can identify physical movements using pid == 0.

I still couldn't reproduce the issue you mentioned. But I see that the mouse pointer stays visible only on the first time you move it out of the screen to the top or down. But this is also true before the changes you suggested. Looks like mac os don't like hiding the cursor while it is over the dock or the menu bar (but just once).

sithlord48:
Since this needs a bit more work i am going to make this a draft. When your happy with the fix @Changroro  you can make it ready for review again .

PR #10104, opened 2026-08-27
fix(clipboard): check oob for clipboard before reading number of formats

sithlord48:
## Description

 Check the number of formats after oob check on clipboard header 
## AI tools used for this PR

None
## Fixed/related issues

https://github.com/deskflow/deskflow/security/advisories/GHSA-xpjj-c4qw-mv8q
CVE-2026-41476
## How has this been tested?

Locally built copied a few things to / from the clipboard on my server / client

### DIFF
--- a/src/lib/deskflow/IClipboard.cpp
+++ b/src/lib/deskflow/IClipboard.cpp
@@ -27,13 +27,13 @@ void IClipboard::unmarshall(IClipboard *clipboard, const std::string_view &data,
     // clear existing data
     clipboard->empty();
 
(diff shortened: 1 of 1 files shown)


nbolton (approved):
Worth adding a test that unmarshalls a 1-byte payload?

sithlord48:
> Worth adding a test that unmarshalls a 1-byte payload?

if you want to add or suggest the test for PR I would add it

nbolton:
> if you want to add or suggest the test for PR I would add it

We can do later.

PR #10106, opened 2026-08-28
fix(mac): exclude keyboard HID tap on local screen for EVKey IME

nnhan0719:
## Description

Fix Vietnamese Telex corruption (U+202F narrow no-break space) when Deskflow macOS server runs alongside EVKey.

When the cursor is on the local screen, the server no longer registers keyboard events on its `kCGHIDEventTap`. The tap uses a mouse/system-only mask (`kCGEventMaskForAllEvents` minus KeyDown/KeyUp/FlagsChanged), is recreated on `enter()`/`leave()`, and still uses `kCGTailAppendEventTap` so EVKey stays ahead in the HID chain when keyboard capture is required off-screen.

Additional hardening:
- Early-return keyboard events in `handleCGInputEvent` while on-screen (defense during tap mask transitions)
- Reset `OSXKeyState::m_deadKeyState` on enable/enter/leave
- Skip stale dead-key state when the active input source has no Unicode layout (EVKey), improving key forwarding to remote clients

## AI tools used for this PR

Cursor (Auto agent) — investigation, implementation, local build/deploy verification, and PR description update.

## Fixed/related issues

No GitHub issue filed. Repro: EVKey Telex on macOS 15.7.7 + Deskflow 1.26.0 server inserts U+202F (e.g. `nguyễn thành nhân` → `nguy ễn t ành n ân`). Related upstream discussion: AltTab macOS EVKey interference (lwouis/alt-tab-macos#5766).

## How has this been tested?

Environment:
- macOS 15.7.7 (24G720), arm64
- Deskflow 1.26.0.9999 built from this branch (Qt 6.10.1)
- EVKey Telex Unicode, Vietnamese + English layouts
- Server: Nguyens-MacBook-Pro.local; client: harness-desktop (192.168.1.147 ↔ 192.168.1.224)

Manual:
- [ ] Re-grant Accessibility after ad-hoc codesign (`System Settings → Privacy & Security → Accessibility`)
- [ ] Local typing on server (cursor on Mac): `nguyeenx thanhf nhaan` in Chrome → Cmd+C → `pbpaste` hex check has no `0x202f`
- [ ] Baseline without Deskflow server: same input still correct (rules out EVKey-only bug)
- [ ] Mouse handoff at screen edge remains smooth after `recreateEventTap()` on enter/leave
- [ ] Remote control: move cursor to client, type Telex from server keyboard — client receives full text (no truncated `ng tan`)
[...]

PR #10108, opened 2026-08-28
fix(mac): exclude keyboard HID tap on local screen for EVKey IME

nnhan0719:
## Description

Fix Vietnamese Telex corruption (U+202F narrow no-break space) when Deskflow macOS server runs alongside EVKey.

When the cursor is on the local screen, the server no longer registers keyboard events on its `kCGHIDEventTap`. The tap uses a mouse/system-only mask (`kCGEventMaskForAllEvents` minus KeyDown/KeyUp/FlagsChanged), is recreated on `enter()`/`leave()`, and still uses `kCGTailAppendEventTap` so EVKey stays ahead in the HID chain when keyboard capture is required off-screen.

Additional hardening:
- Early-return keyboard events in `handleCGInputEvent` while on-screen (defense during tap mask transitions)
- Reset `OSXKeyState::m_deadKeyState` on enable/enter/leave
- Skip stale dead-key state when the active input source has no Unicode layout (EVKey), improving key forwarding to remote clients

Supersedes #10106 (could not be reopened after branch history fix).

## AI tools used for this PR

Cursor (Auto agent) — investigation, implementation, local build/deploy verification, and PR description.

## Fixed/related issues

No GitHub issue filed. Repro: EVKey Telex on macOS 15.7.7 + Deskflow 1.26.0 server inserts U+202F (e.g. `nguyễn thành nhân` → `nguy ễn t ành n ân`). Related upstream discussion: AltTab macOS EVKey interference (lwouis/alt-tab-macos#5766).

## How has this been tested?

Environment:
- macOS 15.7.7 (24G720), arm64
- Deskflow 1.26.0.9999 built from this branch (Qt 6.10.1)
- EVKey Telex Unicode, Vietnamese + English layouts
- Server: Nguyens-MacBook-Pro.local; client: harness-desktop (192.168.1.147 ↔ 192.168.1.224)

Manual:
- [ ] Re-grant Accessibility after ad-hoc codesign (`System Settings → Privacy & Security → Accessibility`)
- [ ] Local typing on server (cursor on Mac): `nguyeenx thanhf nhaan` in Chrome → Cmd+C → `pbpaste` hex check has no `0x202f`
- [ ] Baseline without Deskflow server: same input still correct (rules out EVKey-only bug)
- [ ] Mouse handoff at screen edge remains smooth after `recreateEventTap()` on enter/leave
- [ ] Remote control: move cursor to client, type Telex from server keyboard — client receives full text (no truncated `ng tan`)
- [ ] `lsof -i :24800` shows server listening; no `assistive devices does not trust this process` after TCC re-grant

Unit tests:
- `OSXKeyStateTests`: pre-existing `fakePollShift` failure unchanged (not introduced by this PR)

### DIFF
--- a/src/lib/platform/OSXKeyState.cpp
+++ b/src/lib/platform/OSXKeyState.cpp
@@ -330,6 +330,11 @@ KeyButton OSXKeyState::mapKeyFromEvent(KeyIDs &ids, KeyModifierMask *maskOut, CG
 
   // get keyboard info
   AutoTISInputSourceRef currentKeyboardLayout = copyKeyboardLayoutForKeyTranslation();
+  AutoTISInputSourceRef inputSource(TISCopyCurrentKeyboardInputSource(), CFRelease);
+  if (inputSource == nullptr ||
+      TISGetInputSourceProperty(inputSource.get(), kTISPropertyUnicodeKeyLayoutData) == nullptr) {
+    m_deadKeyState = 0;
+  }
   CFDataRef ref = nullptr;
   {
     std::lock_guard<std::mutex> lock(g_tisMutex);
@@ -382,7 +387,8 @@ KeyButton OSXKeyState::mapKeyFromEvent(KeyIDs &ids, KeyModifierMask *maskOut, CG
 
     // get the characters
     if (status == 0) {
-      if (count != 0 || m_deadKeyState == 0) {
+      if (count != 0 || m_deadKeyState == 0 || inputSource == nullptr ||
+          TISGetInputSourceProperty(inputSource.get(), kTISPropertyUnicodeKeyLayoutData) == nullptr) {
         m_deadKeyState = 0;
         for (UniCharCount i = 0; i < count; ++i) {
           ids.push_back(IOSXKeyResource::unicharToKeyID(chars[i]));
--- a/src/lib/platform/OSXKeyState.h
+++ b/src/lib/platform/OSXKeyState.h
@@ -67,6 +67,11 @@ public:
   */
   KeyButton mapKeyFromEvent(KeyIDs &ids, KeyModifierMask *maskOut, CGEventRef event) const;
 
+  void clearDeadKeyState() const
+  {
+    m_deadKeyState = 0;
+  }
+
   //! Map key and mask to native values
   /*!
   Calculates mac virtual key and mask for a key \p key and modifiers
--- a/src/lib/platform/OSXScreen.mm
+++ b/src/lib/platform/OSXScreen.mm
@@ -689,6 +689,49 @@ void OSXScreen::hideCursor()
   m_cursorHidden = true;
 }
 
+void OSXScreen::recreateEventTap(CGEventTapOptions option, CGEventMask mask)
+{
+  if (m_eventTapRunLoop) {
+    CFRunLoopStop(m_eventTapRunLoop);
+  }
+  if (m_eventTapThread.joinable()) {
+    m_eventTapThread.join();
+  }
+  if (m_eventTapRLSR) {
+    CFRelease(m_eventTapRLSR);
+    m_eventTapRLSR = nullptr;
+  }
+  if (m_eventTapPort) {
+    CGEventTapEnable(m_eventTapPort, false);
+    CFMachPortInvalidate(m_eventTapPort);
+    CFRelease(m_eventTapPort);
+    m_eventTapPort = nullptr;
+  }
+
+  CGEventTapCallBack cb = m_isPrimary ? handleCGInputEvent : handleCGInputEventSecondary;
+  m_eventTapPort = CGEventTapCreate(kCGHIDEventTap, kCGHeadInsertEventTap, option, mask, cb, this);
+  if (!m_eventTapPort) {
+    return;
+  }
+
+  m_eventTapRLSR = CFMachPortCreateRunLoopSource(kCFAllocatorDefault, m_eventTapPort, 0);
+  if (!m_eventTapRLSR) {
+    return;
+  }
+
+  auto sem = dispatch_semaphore_create(0);
+  m_eventTapThread = std::thread([this, sem]() {
+    m_eventTapRunLoop = CFRunLoopGetCurrent();
+    CFRunLoopAddSource(m_eventTapRunLoop, m_eventTapRLSR, kCFRunLoopDefaultMode);
+    dispatch_semaphore_signal(sem);
+    CFRunLoopRun();
+    CFRunLoopRemoveSource(CFRunLoopGetCurrent(), m_eventTapRLSR, kCFRunLoopDefaultMode);
+    m_eventTapRunLoop = nullptr;
+  });
+  dispatch_semaphore_wait(sem, DISPATCH_TIME_FOREVER);
+  dispatch_release(sem);
+}
+
 void OSXScreen::enable()
 {
   // watch the clipboard
@@ -700,12 +743,7 @@ void OSXScreen::enable()
 
   if (m_isPrimary) {
     // FIXME -- start watching jump zones
-
-    // kCGEventTapOptionDefault = 0x00000000 (Missing in 10.4, so specified literally)
-    m_eventTapPort = CGEventTapCreate(
-        kCGHIDEventTap, kCGHeadInsertEventTap, kCGEventTapOptionDefault, kCGEventMaskForAllEvents, handleCGInputEvent,
-        this
-    );
+    recreateEventTap(kCGEventTapOptionDefault, kCGEventMaskForAllEvents);
   } else {
     // FIXME -- prevent system from entering power save mode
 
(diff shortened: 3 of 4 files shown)


src/lib/platform/OSXKeyState.cpp
```diff
@@ -178,6 +178,12 @@ bool isModifier(uint8_t virtualKey)
   return (modifiers.find(virtualKey) != modifiers.end());
 }
 
+bool inputSourceLacksUnicodeKeyLayout(TISInputSourceRef source)
```
amagus: no need for this function if it's used just once, please check the values directly on the if.

src/lib/platform/OSXScreen.mm
```diff
@@ -689,62 +706,79 @@ static inline bool needsEventNumber()
   m_cursorHidden = true;
 }
 
+void OSXScreen::recreateEventTap(CGEventTapOptions option, CGEventMask mask)
```
amagus, line 698: why do we need to move the code to this new function? please do refactors in refactor commits, keep the fixes to the minimal in fix commits.

src/lib/platform/OSXScreen.mm
```diff
@@ -1681,6 +1722,12 @@ static inline bool needsEventNumber()
 CGEventRef OSXScreen::handleCGInputEvent(CGEventTapProxy proxy, CGEventType type, CGEventRef event, void *refcon)
 {
   OSXScreen *screen = (OSXScreen *)refcon;
+  // Defense in depth while the tap mask is switched during enter()/leave().
+  if (screen->m_isOnScreen) {
```
amagus: why we need to do this here? Is your problem the call to screen->onKey(event)?

src/lib/platform/OSXScreen.mm
amagus, line 1749: if the problem is the call to onKey, why not just add an if here to avoid that call? (the event will still be emitted back to the tap after this switch)

amagus (changes requested):
Please refrain from mixing refactors and actual fixes in the same commit. If you want to refactor, do it in a separated commit. This way it will be clear what you are really fixing and what you refactoring.

nnhan0719:
Addressed all four review comments in 7413b64 (squashed):

**1. `inputSourceLacksUnicodeKeyLayout()` (OSXKeyState.cpp)** — Removed; the TIS property check is inlined in `mapKeyFromEvent()`.

**2. `recreateEventTap()` extraction (OSXScreen.mm)** — Kept because the tap mask is fixed at creation time; `enter()`/`leave()` must tear down and recreate with keyboard-inclusive vs keyboard-exclusive masks. The helper avoids duplicating teardown/thread setup in three call sites. Squashed to one commit here; no separate refactor commit.

**3–4. Early keyboard return in `handleCGInputEvent` (lines ~1726 / ~1760)** — You're right: EVKey corruption comes from the HID tap *listening* for keyboard events, not from `onKey()`. Removed the top-of-handler early return. Keyboard events on the local screen are skipped at the `onKey()` call site instead; the handler still falls through and returns the event when `m_isOnScreen` (unchanged pass-through behavior).

Also fixed `leave()` ordering: `m_isOnScreen = false` now runs *before* recreating the tap with the full keyboard mask, so we do not briefly register keyboard listening while still considered on-screen.

src/lib/platform/OSXKeyState.cpp
nnhan0719: Done — removed the helper, check is inline in the if now.

src/lib/platform/OSXScreen.mm
nnhan0719: The mask is fixed at CGEventTapCreate time, so enter()/leave() have to tear down and recreate the tap when switching between local (no keyboard in mask) and remote (keyboard in mask). Extracted recreateEventTap() so we don't copy the teardown/thread setup three times — no behavior change on its own.

Squashed to one commit on this branch; happy to split refactor/fix if you prefer that for review.

amagus:
You are still refactoring and fixing at the same commit. If your fix require a refactor to apply the fixes, refactor it first and then the fix with a later commit. What you are fixing needs to be completely clear, without refactors. Your PR can contain multiple commits.

src/lib/platform/OSXScreen.mm
amagus: Why do you need to recreate the event tap? Why can't we just ignore the keyboard events?

And yes, if you want to refactor the code so that the tap creation can be reused, please create a different commit specific to this refactor. The refactor commit should contain no fix and just move certain parts of the code to the reusable function(s).

nnhan0719:
Clean, pls help review

src/lib/platform/OSXScreen.mm
amagus: Question is not answered. Why do you need a new event tap port? Why can't you just ignore the events?
Recreating the event tap that often is not ideal.

amagus:
Please don't mark conversations as resolved without answering the asked questions.

src/lib/platform/OSXKeyState.cpp
```diff
@@ -330,6 +330,11 @@ KeyButton OSXKeyState::mapKeyFromEvent(KeyIDs &ids, KeyModifierMask *maskOut, CG
 
   // get keyboard info
   AutoTISInputSourceRef currentKeyboardLayout = copyKeyboardLayoutForKeyTranslation();
+  AutoTISInputSourceRef inputSource(TISCopyCurrentKeyboardInputSource(), CFRelease);
```
amagus, line 333: why can't you use currentKeyboardLayout instead of having to create yet another copy of the layout?

src/lib/platform/OSXKeyState.h
```diff
@@ -141,6 +143,8 @@ class OSXKeyState : public KeyState
   void postKeyboardKey(CGKeyCode virtualKey, bool keyDown);
 
 private:
+  friend class OSXScreen;
```
amagus: why do you need to friend? Create accessors to the values you need.

src/lib/platform/OSXKeyState.cpp
nnhan0719: Still addressed in `37fb2b7` — the TIS property check is inline in the `if`, no helper function.

src/lib/platform/OSXScreen.mm
nnhan0719: Sorry for the earlier unclear reply.

**Why not just ignore keyboard events?**

Ignoring them in the callback (early return or skipping `onKey()`) does not fix EVKey. The corruption happens because the HID tap is *registered* with keyboard events in its mask. EVKey breaks even if our callback returns immediately and never calls `onKey()`.

**Why recreate the tap?**

`CGEventTapCreate` fixes the event mask at creation time — it cannot be changed later. On the local screen we must not listen for keyboard at all (`kLocalScreenEventMask`). When the cursor moves to a client we need the full mask again to forward keys remotely. That switch only happens in `enter()` / `leave()`, not on every key press.

**Latest `37fb2b7`:** single fix-only commit, no extracted helper — tap teardown/creation is inlined in `enter()` / `leave()`.

src/lib/platform/OSXKeyState.cpp
nnhan0719: `currentKeyboardLayout` comes from `copyKeyboardLayoutForKeyTranslation()` — for EVKey it may already be an ASCII fallback layout used for `UCKeyTranslate`, not the active input source.

We need the *current keyboard input source* to detect an IME like EVKey that has no `kTISPropertyUnicodeKeyLayoutData`. That is what stale dead-key state is tied to. Using `currentKeyboardLayout` for this check would miss EVKey when a fallback layout was selected for translation.

src/lib/platform/OSXScreen.mm
```diff
@@ -61,6 +61,12 @@
 
 static const double kCarbonLoopWaitTimeout = 10.0;
 
+// EVKey/IME corrupts when a kCGHIDEventTap listens for keyboard events, even if the
```
amagus: what exactly gets corrupted? is it the status from deskflow code?

src/lib/platform/OSXScreen.mm
```diff
@@ -791,6 +797,49 @@ static inline bool needsEventNumber()
...
+      m_eventTapRLSR = nullptr;
+    }
+    if (m_eventTapPort) {
+      CGEventTapEnable(m_eventTapPort, false);
+      CFMachPortInvalidate(m_eventTapPort);
+      CFRelease(m_eventTapPort);
+      m_eventTapPort = nullptr;
+    }
+
+    m_eventTapPort = CGEventTapCreate(
```
amagus: I would prefer event taps to be created in a single place, please refactor the code first and then introduce your fix.

src/lib/platform/OSXScreen.mm
```diff
@@ -1710,7 +1802,9 @@ static inline bool needsEventNumber()
   case kCGEventKeyDown:
   case kCGEventKeyUp:
   case kCGEventFlagsChanged:
-    screen->onKey(event);
+    if (!screen->m_isOnScreen) {
```
amagus: this shouldn't be needed if you drop the event tap... Are you sure this isn't the entire fix that you need? And maybe some flag reset on the screen enter?

nnhan0719: Agreed — removed the onKey guard. With the keyboard-exclusive tap mask on the local screen, keyboard events no longer reach our callback there. Kept clearDeadKeyState() on enter()/leave() for remote forwarding with EVKey.

src/lib/platform/OSXKeyState.cpp
nnhan0719: currentKeyboardLayout may be an ASCII fallback for UCKeyTranslate; we need the active input source to detect EVKey (no Unicode layout).

src/lib/platform/OSXKeyState.cpp
nnhan0719: Done — inlined in 426d0ce.

nnhan0719:
Re: refactor/fix in same commit — split into 9a1d934 (refactor) + 426d0ce (fix).

Re: resolving threads early — sorry, replied on all open threads now.

src/lib/platform/OSXKeyState.h
nnhan0719: Removed friend — using clearDeadKeyState() accessor instead.

src/lib/platform/OSXScreen.mm
nnhan0719: Local EVKey typed text (U+202F inserted), not Deskflow internal state.

src/lib/platform/OSXScreen.mm
nnhan0719: Done: 9a1d934 refactor, 426d0ce fix.

src/lib/platform/OSXScreen.mm
nnhan0719: Agreed — removed onKey guard. Kept clearDeadKeyState() on enter/leave for remote + EVKey.

src/lib/platform/OSXScreen.mm
nnhan0719: No — EVKey breaks when the tap listens for keyboard, not because of onKey().

src/lib/platform/OSXScreen.mm
nnhan0719: Updated in 426d0ce — removed the onKey guard; keyboard-exclusive mask is the fix.

src/lib/platform/OSXScreen.mm
amagus: This is confusing to me. Why would the event tap corrupt the keyboard state? If this is the case, then it's a bug on Apple's code itself which should be reported to them. We should not try to workaround upstream issues.

Are you sure that it's the event tap itself that creates the problem? 

I would suggest you to create a small reproducer with an event tap that just returns the event to confirm that having an event tap corrupts the functionality of this keyboard feature.
If the event tap works fine, we need to figure out what deskflow is doing differently. If it doesn't work, then you confirmed the upstream issue and you should be opening a bug report to Apple.

amagus: And it seems that EVKey is a non official keyboard layout/engine/plugin (https://github.com/lamquangminh/EVKey). Perhaps you should report the issue to them instead of Apple. They likely don't support event taps. Deskflow may not work correctly in these situations.

nnhan0719: Bug only when Deskflow server is running.

amagus: Yes, EVKey, based on what I could translate, also consumes and uses the event tap (since they also need accessibility permissions). 
Based on your report they misbehave if there's another event tap enabled for keyboard. You should contact their developers and ask them to investigate the issue and to add support to additional event taps on the same mac. 

We should not be recreating the event tap that often in deskflow just to workaround a bug on some software.

PR #10111, opened 2026-08-29
Support a clean exit of the client

KoljaFrahm:
## Description
Currently, when we click on stop to stop the server, we just kill it without letting it finish or destroy. This commit makes it, so the server can exit gracefully.

## AI tools used for this PR
I used Claude for some advice on how to let the server gracefully exit. I also in generall asked it for feedback.

## Fixed/related issues
-
## How has this been tested?
Tested on a wayland server. I clicked the stop icon, and now it exits with code 0 instead of 9 like before.

### DIFF
--- a/src/lib/gui/core/CoreProcess.cpp
+++ b/src/lib/gui/core/CoreProcess.cpp
@@ -269,7 +269,7 @@ void CoreProcess::startProcessFromDaemon()
   }
 }
 
-void CoreProcess::stopForegroundProcess() const
+void CoreProcess::stopForegroundProcess()
 {
   if (m_processState != ProcessState::Stopping) {
     qCritical("not stopping core desktop process, unexpected process state");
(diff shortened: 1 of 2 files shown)


src/lib/platform/EiScreen.cpp
```diff
@@ -141,26 +141,23 @@ void EiScreen::initEi()
 
 void EiScreen::cleanupEi()
 {
+  for (auto device : m_eiDevices) {
+    delete static_cast<ScrollRemainder *>(ei_device_get_user_data(device));
+    ei_device_set_user_data(device, nullptr);
+  }
+
   if (m_eiPointer) {
-    free(ei_device_get_user_data(m_eiPointer));
```
sithlord48: does the ei_device _set_user_data call do this for us now ?

KoljaFrahm: I replaced this by 
`delete static_cast<ScrollRemainder *>(ei_device_get_user_data(device));` (see two lines above). We allocate the memory with new, so we should free it with delete.

sithlord48 (changes requested):
Did a quick test of this and while it appears to work there is some odd things happening now 

1. ~~On Initial starting up two deskflow processes are made (one exits shortly after starting). This i can see in the process monitor when starting the application~~ (Ide being weird not this PR)
2. When i "Stop" the core the message ` WARNING: QProcess: Destroyed while process ("<path-of>/deskflow-core") is still running.` 
3. After stopping the core when its restarted two copies of deskflow-core are started (one exit shortly after, they have the same PID so maybe not an issue ??) 
4. When stopping the core `WARNING: failed to enable system idle sleep` is printed to the app log  if you have the "Prevent this computer from going to sleep" option enabled. But the sleep lock is removed so this message maybe in our code just happening a bit to soon. (check / fix in another PR) 

I suspect this would have issues if someone was really trying to run two instances with the `--new-instance` option we provide.

amagus (approved):
lgtm, all changes make sense, haven't tried to test it yet

KoljaFrahm:
Now put all but the last commits to https://github.com/deskflow/deskflow/pull/10116 so this can focus on the exit commit.

nbolton (approved):

sithlord48:
This should be checked on windows also

nbolton:
> This should be checked on windows also

Yep, tested, all good.

image

Still needs your approval as you requested changes.

sithlord48 (approved):

PR #10095, opened 2026-08-24
fix(macOS): preserve literal dead-key output

RomuloGatto:
## Description

Preserve literal characters produced by macOS dead keys instead of dropping them when the active layout's spacing output differs from Deskflow's generic decomposition table.

This change:

- asks `UCKeyTranslate` for each dead key's actual spacing output;
- binds that output to the exact physical/modifier `KeyItem` that produced it (important because one dead-key ID can have multiple spacing outputs in the same layout);
- adds resolved dead-key + space combinations without replacing direct mappings;
- handles multiple characters from one macOS physical event at the macOS event boundary, inserting releases between normal presses while preserving genuine autorepeat.

The generic `KeyState` path is unchanged. The only shared-code addition is a small `KeyMap` overload for installing an already-resolved key sequence; macOS needs it to retain the exact physical/modifier entry instead of resolving two different dead keys through the same canonical dead-key ID.

On Brazilian - Pro, this distinguishes:

```text
U+0301 -> U+0027 on the apostrophe key
U+0301 -> U+00B4 on Option+E
```

It fixes `I'm` becoming `Im` on a macOS client and also addresses the standalone quote path reported in #9740.

## AI tools used for this PR

Hermes Agent with OpenAI GPT-5.6 Sol was used for source tracing, test design, implementation assistance, and review. The behavior was reproduced with `UCKeyTranslate`, and the patch was built and tested locally on macOS.

## Fixed/related issues

fixes: #9740
related: #9254

## How has this been tested?

Environment: macOS 14.5 arm64, Apple Clang 16.0.0, Qt 6.11.1, Debug build.

- Verified the active `com.apple.keylayout.Brazilian-Pro` resource reports `U+0301 -> U+0027` for the apostrophe key and `U+0301 -> U+00B4` for Option+E.
- Added macOS regression tests for layout-specific spacing output, ambiguous and modifier-specific dead-key IDs, direct-mapping precedence, multi-character `I'm` transitions, and genuine autorepeat.
- Built the complete application successfully.
- All 25 non-HID-polling unit test executables pass.
- The new targeted `OSXKeyStateTests` cases pass. Three pre-existing synthetic HID polling cases (`fakePollShift`, `fakePollChar`, and `fakePollCharWithModifier`) remain blocked by local macOS input-event permissions and fail unchanged outside this patch.
- `clang-format` 20.1.0, `git diff --check`, and `reuse lint` pass.

PR #10115, opened 2026-08-30
fix(mac): prevent Option characters from becoming dead keys

nburrus:
## Description

UCKeyTranslate needs the Option modifier to resolve characters from the Option layer of the active layout. Clearing it unconditionally makes those keys translate as their base or dead-key variants, leaving composition state
pending and sending the wrong character.

For example, with the Spanish-ISO layout, Option plus the ` key produces `[`. Without Option, UCKeyTranslate sees the grave-accent dead key, returns no character, and waits for another key to complete the composition. Deskflow
therefore sends a backtick first instead of `[`. Other Option-layer punctuation, including `{`, is affected in the same way.

Restore a commented out if statement to only remove Option when Command or Control is also held, where Deskflow intentionally uses the base glyph to identify the shortcut key. Special keys are handled before layout translation and are unaffected.

**I don't know the motivations of the original commented out code**, so unclear to me if this could introduce regressions. Still it fixes the Spanish keyboard issue for me and is a tiny change, so it might be worth considering.

This changes Option-modified printable keys to follow the active layout, including its dead-key behavior. An application binding bare Option plus a printable key may therefore receive the layout-produced character instead of
the base glyph it received from the previous, incorrect translation.

## AI tools used for this PR
Codex with sol medium. I am new to the deskflow codebase so this might very well not be the sound fix, it just empirically works for my two-mac setup without triggering other issues so far.

## Fixed/related issues

related: #10095 but this patch is just re-enabling a commented out if, so it's tiny in comparison. Not sure if it fixes the same issue with Brazilian keyboards though.

## How has this been tested?
I have a mac mini with an external Apple spanish keyboard connected as the deskflow server. The client is a macbook pro with a spanish layout as well, used as the client here.

### DIFF
--- a/src/lib/platform/OSXKeyState.cpp
+++ b/src/lib/platform/OSXKeyState.cpp
@@ -352,9 +352,9 @@ KeyButton OSXKeyState::mapKeyFromEvent(KeyIDs &ids, KeyModifierMask *maskOut, CG
 
   // if we've used a command key then we want the glyph produced without
   // the option key (i.e. the base glyph).
-  // if (isCommand) {
-  modifiers &= ~optionKey;
-  //}
+  if (isCommand) {
+    modifiers &= ~optionKey;
+  }
 
   // choose action
   uint16_t action;
--- a/src/lib/platform/OSXScreen.mm
+++ b/src/lib/platform/OSXScreen.mm
@@ -1159,12 +1159,14 @@ bool OSXScreen::onKey(CGEventRef event)
     return false;
   }
 
-  // check for AltGr in mask.  if set we send neither the AltGr nor
-  // the super modifiers to clients then remove AltGr before passing
-  // the modifiers to onKey.
+  // AltGr marks a printable character that was translated using the
+  // macOS Option layer.  The KeyID already contains Option's effect, so
+  // don't also send Option (represented by KeyModifierAlt) to clients.
+  // Keep Option in the local state so its separate key-up event remains
+  // balanced.
   KeyModifierMask sendMask = (mask & ~KeyModifierAltGr);
   if ((mask & KeyModifierAltGr) != 0) {
-    sendMask &= ~KeyModifierSuper;
+    sendMask &= ~KeyModifierAlt;
   }
   mask &= ~KeyModifierAltGr;
 


sithlord48 (changes requested):
That is one hell of a commit message to just uncomment a line of code. Can you shorten the message.

sithlord48:
@amagus  Can you look at this one please I can do a quick test as a client / server with mac os to be sure "normal" layouts are still working but that is all i can check.

src/lib/platform/OSXKeyState.cpp
```diff
@@ -352,9 +352,9 @@ KeyButton OSXKeyState::mapKeyFromEvent(KeyIDs &ids, KeyModifierMask *maskOut, CG
 
   // if we've used a command key then we want the glyph produced without
   // the option key (i.e. the base glyph).
-  // if (isCommand) {
-  modifiers &= ~optionKey;
-  //}
+  if (isCommand) {
```
amagus, line 355: not sure if this is correct, there are valid keyboard bindings using command + option. 
We can't blindly ignore the option modifier as you suggested... 
But why wouldn't the fix be not touching the modifiers at all?
What exactly you tried to type that got solved with this fix?
What keyboard settings you used on the client and on the server?

amagus:
please use "fix(mac)" prefix so that the commits can be grouped better

amagus:
This made me realize that global shortcuts don't work at all. 
For example "command + shift + 4" should trigger selective screenshot on mac. 
It works on the server, but not on any macOS client... Even using US layout.

I also did some basic testing and I really couldn't find situation on why the modifiers need to get removed (including the command one, just above what this commit tries to change). Everything looks to work fine even without these (using portuguese brazilian as base).

I notice as well that the dead keys have some visual feedback on the server which are not visible on client. 
We are likely accumulating it internally and emitting the resolved key later (somewhat like how the US keyboard works - in which key modifiers have no visual feedback on the screen). This may need deeper investigation.

src/lib/platform/OSXKeyState.cpp
nburrus: well actually before the changes it was _always_ removing the optionKey, unconditionally. So the change made it only remove it in the case of a command or control combo. Not sure why we'd remove it even in that case though, but that was the preexisting code.

nburrus: I was trying to type [ and { on a Spanish Apple keyboard. Both macOS machines using the spanish layout. Those keys just don't work without the fix. For example for [ I end up getting a backstick after pressing another key. I left the default settings on the server, on the client I have Use server's keyboard languaged enabled. But it wasn't working with it disabled either.

amagus: Indeed, I can confirm the bug you mentioned on brazilian portuguese. As I mentioned down below, I wished to understand what exactly this if is trying to fix, as it seems there are way more other issues related to that. Just some lines up, we can see that we remove the command key as well and I can't understand why. We need a proper investigation.

nburrus:
> This made me realize that global shortcuts don't work at all. For example "command + shift + 4" should trigger selective screenshot on mac. It works on the server, but not on any macOS client... Even using US layout.

This works for me on the spanish layout. However just realized that shortcuts like Cmd+Alt+Space don't work, neither with or without the fix. However Cmd+Option+Esc or Cmd+Option+F5 work.

src/lib/platform/OSXKeyState.cpp
nburrus: yes sorry I can't provide more details as I'm not familiar with the codebase, I've basically just driven codex to add some logging, deploy on both machines and iterate on a fix. Agree that it looks weird to treat Option differently...

amagus (changes requested):
I've been trying this over the weekend and I just realized that this may solve the problem for pt-br and spanish layouts, but it breaks US layout.
I can't make it generate "ç" (option + c) on clients with this change. But it works without it. We may need a better engineered change, it's unfortunately not as simple as uncommenting the "if" checks.

nburrus:
> I've been trying this over the weekend and I just realized that this may solve the problem for pt-br and spanish layouts, but it breaks US layout. I can't make it generate "ç" (option + c) on clients with this change. But it works without it. We may need a better engineered change, it's unfortunately not as simple as uncommenting the "if" checks.

mm I'm trying here on macOS with both keyboards configured to US layout and Alt+c works to generate ç , not sure why. 

But asking codex to investigate that also pointed at what looks like a very old bug where detecting AltGr would remove Super instead of Alt. It wasn't such an issue before because the Alt was removed unconditionally anyway, but might be now and in any case it looked weird. Pushed that second commit to the branch in case it rings a bell. Again, take it with a grain of salt as I don't really know what I'm doing here 😅 

But fwiw I've been testing that branch with success so far with my dual-mac with spanish keyboard setup.

amagus:
I forgot to be more specific: it stops being able to generate "ç" on windows clients (configured with US macintosh keyboard)

PR #10116, opened 2026-08-30
Deletion of dead code, refactor and correcting logging messages

KoljaFrahm:
## Description
The commits all work by themselves
- refactor: don't route startNode through AppUtil
Currently we route our startNode calls through AppUtil. This is not needed.
- refactor: delete dead InputFilter::removeFilterRule
InputFilter::removeFilterRule is called nowhere. I removed it
- fix: correct wrong screen entered warning
See commit description ("In #7437, we replaced an assert with an if for logging, but forgot to negate the if clause")
- fix: don't show an error when ipc server is closed
When we stop the server, we close the ipc server. This is not an error, so don't show it as one in the logs
- fix: let inhibitScreenCall succeed when nothing was inhibited
When we close the server, inhibitScreenCall fails if nothing was inhibited. Correct this to not show an error
- fix(wayland): delete ScrollRemainder in EiScreen
ScrollRemainder is not necessarily deleted in EiScreen. Change this
- fix: delete coreApp
coreApp is created with new, but nowhere deleted.

## AI tools used for this PR
I  asked claude it for feedback.

## Fixed/related issues
-
## How has this been tested?
Tested on a wayland server

### DIFF
--- a/src/apps/deskflow-core/deskflow-core.cpp
+++ b/src/apps/deskflow-core/deskflow-core.cpp
@@ -137,6 +137,7 @@ int main(int argc, char **argv)
   if (exitCode == s_exitSuccess) {
     exitCode = coreApp->getExitCode();
   }
+  delete coreApp;
 
   LOG_DEBUG("core exited, code: %d", exitCode);
   return exitCode;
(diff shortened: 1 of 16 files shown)


sithlord48 (approved):
Looks good quick test no new errors

@nbolton I'll leave landing up to you so you can get a chance to review this.

nbolton (approved):

PR #10117, opened 2026-08-30
Only enable the screen after connecting to the client

KoljaFrahm:
## Description
To be able to react to rules in screen->activate() that are enabled in setPrimaryClient, we have to reverse the order of the two calls. The same goes for screen->deactivate()

I need this for the globalshortcuts PR, to not register each shortcut themselves, but call to the portal in EiScreen::activate(). This can only be done after reading the config and enabling the rules.

I want to push this before the globalshortcuts PR to see if there are any regressions.

## AI tools used for this PR
-

## Fixed/related issues
-

## How has this been tested?
Wayland server

### DIFF
--- a/src/lib/server/Server.cpp
+++ b/src/lib/server/Server.cpp
@@ -126,10 +126,19 @@ Server::Server(ServerConfig &config, PrimaryClient *primaryClient, deskflow::Scr
   // set initial configuration
   setConfig(config);
 
+  // If we are using libportal, we need to set the primary client first.
+  // This enables the rules of the primary client. The hotkeys are then enabled in EiScreen.cpp after the rules
+  // registered them.
+#if WINAPI_LIBPORTAL
   // enable primary client
+  m_inputFilter->setPrimaryClient(m_primaryClient);
+  m_primaryClient->enable();
+  // If we are on Mac, X11 or Windows, we need a key map, which is created in screen->enable().
+  // So we need to enable the screen first, then enable the primary client.
+#else
   m_primaryClient->enable();
   m_inputFilter->setPrimaryClient(m_primaryClient);
-
+#endif
   // Determine if scroll lock is already set. If so, lock the cursor to the
   // primary screen (unless the user has disabled lock to screen in config)
   if (!m_disableLockToScreen && (m_primaryClient->getToggleMask() & KeyModifierScrollLock)) {
@@ -175,11 +184,16 @@ Server::~Server()
     delete client;
   }
 
-  // remove input filter
-  m_inputFilter->setPrimaryClient(nullptr);
-
+  // See at Server::Server for why we do this in this order.
+#if WINAPI_LIBPORTAL
   // disable and disconnect primary client
   m_primaryClient->disable();
+  // remove input filter
+  m_inputFilter->setPrimaryClient(nullptr);
+#else
+  m_inputFilter->setPrimaryClient(nullptr);
+  m_primaryClient->disable();
+#endif
   removeClient(m_primaryClient);
 }
 


src/lib/server/Server.cpp
```diff
@@ -127,8 +127,8 @@ Server::Server(ServerConfig &config, PrimaryClient *primaryClient, deskflow::Scr
   setConfig(config);
 
   // enable primary client
-  m_primaryClient->enable();
   m_inputFilter->setPrimaryClient(m_primaryClient);
```
nbolton, line 134: At first glance, seems fine on Wayland, but I'm concerned about the impact on X11 (and maybe also Windows and macOS to a lesser degree).

`setPrimaryClient` registers hotkey rules (`m_ruleList`) at the end of the function. `registerHotKey` on X11 needs the key map. `Screen::enable()` builds the key map. When swapped, registration runs against empty maps and returns 0 (so no `XGrabKey`).

https://github.com/deskflow/deskflow/blob/8dfdf8739a29613ebae52a1e207b912c804dcb26/src/lib/server/InputFilter.cpp#L799

https://github.com/deskflow/deskflow/blob/c20870cc7a29b7e0a3376c096f044f85110fa40b/src/lib/platform/XWindowsScreen.cpp#L512

https://github.com/deskflow/deskflow/blob/8dfdf8739a29613ebae52a1e207b912c804dcb26/src/lib/deskflow/Screen.cpp#L97

KoljaFrahm: Hmm registerHotkey is the exact reson why I wanted to swap it. On wayland, it would make sense to first save all the "registred" hotkeys in a vector and then activate them all at once. 
If this is not possible, I can add a rulesActivated function to Screen but this is less pretty. Or do you think it's worth it write the implementation in XWindowsScreen.cpp?

sithlord48: as far as i know we can use portals also on X11 they just are not "enforced"

I think we have a deeper problem around X11 and that is that We have no active developers using it. Unless we find a champion for the platform its gonna end up bit-rotting away. I would say if after 1.27 we do not have an active dev daily driving X11 (client and server) then we have to look to maybe drop the platform with in a release or two after that.

nbolton: > and maybe also Windows and macOS to a lesser degree

Don't forget about these fellers... don't say you want to drop them too ;) -- yeah we all know what you think about Windows.

sithlord48: I was just addressing the situation around the X11 backend.
We have active macOS champion (@amagus) and it would be nice to have one for windows too. but for now we just all work a bit on windows fixes when we can...

sithlord48:
@amagus can you check this has no Ill effects on macOS.

amagus (approved):
LGTM, tested during some days on mac, no issues

edit: it breaks debug build on macos :(

sithlord48:
we still need to test this on windows / X11. but 1/2 them being ok with the order change is a good start. 

Edit: 
Did a quick test with an X11 client and it seams to work as expected still (i do not have any hotkeys configured)

src/lib/server/Server.cpp
nbolton: Confirmed. On macOS server, debug build crashes:
```
[2026-09-08T15:22:44.291] WARNING: can't get the active group, use the first group instead
	/Users/nick-symless/Projects/deskflow/src/lib/platform/OSXKeyState.cpp:485
Assertion failed: (group >= 0 && group < getNumGroups()), function findCompatibleKey, file KeyMap.cpp, line 328.
[2026-09-08T15:22:44.659] DEBUG: core connection state changed: 0
	/Users/nick-symless/Projects/deskflow/src/lib/gui/MainWindow.cpp:989
[2026-09-08T15:22:44.659] DEBUG: core process state changed: started -> stopped
	/Users/nick-symless/Projects/deskflow/src/lib/gui/core/CoreProcess.cpp:614
[2026-09-08T15:22:44.660] WARNING: desktop process exited with code: 6
	/Users/nick-symless/Projects/deskflow/src/lib/gui/core/CoreProcess.cpp:188
```

Edit: Also confirmed broken on Windows server. Debug build runs but hotkeys simply just don't work (nothing happens when I use a hotkey that works fine on `master`).

nbolton:
Breaks hotkeys on Windows and macOS: https://github.com/deskflow/deskflow/pull/10117#discussion_r3958946972

KoljaFrahm:
I now changed it so that the order is depended on the platform, like @sithlord48 suggested.

sithlord48 (approved):

nbolton (approved):
Tested server on Windows and macOS, no longer crashes.

PR #10122, opened 2026-09-01
fix(wayland): stop the cursor sticking to client screen edge

### LINKED ISSUE #9886: Cursor stuck to left hand side of client screen (Wayland GNOME server)
### Deskflow version info

**Server**

Deskflow: 1.26.0.288 (f3fb3191)
Qt: 6.11.1
System: Fedora Linux 44 (Workstation Edition)
Session: GNOME (wayland)

**Client**

Deskflow: 1.26.0.286 (a2aa56a1)
Qt: 6.9.2
System: macOS Tahoe (26.5)

### Build types

### Deskflow configuration

- Wayland server (single 4k landscape screen)
- macOS client (single SD portrait screen)
- Windows client (single SD landscape screen)
- Wayland client (single SD landscape screen)

Arrangement:

```
                [ Windows client ]
[ macOS ] [         Wayland server        ] [ Wayland client ]
```

Throw the cursor hard into the top-left corner of the Wayland server screen (0,0).

The cursor lands on the macOS client to the left and sticks against its edge, only movable
along that edge. This happens every time.

The GNOME Activities Overview also fires, since 0,0 is the hot corner, but its timing varies:
sometimes it opens straight away, sometimes it doesn't appear until I stop the client, as if
it had been held back. Either way the cursor sticks, so the Overview looks like a side effect
of hitting the corner rather than part of the failure.

On my layout that corner is three things at once: the GNOME hot corner, the barrier to the
macOS client (left) and the barrier to the Windows client (up).

### Log output

[server.log](https://github.com/user-attachments/files/29246284/server.log)
[client.log](https://github.com/user-attachments/files/29246285/client.log)

### Additional information

Could either be Wayland server bug or macOS client bug

I wonder if this bug could be caused by be3cc0ff7acf249b4e8ed6b91770870449ed45b6 or 5e79e49966a05d2344eddb14448bf423cbba7435 from PR:
- #9763

We should git bisect this sooner rather than later before it becomes too difficult to go back in time (those commits were already done almost a month ago now).

nbolton:
## Description

`EiScreen::onMotionEvent` cast libei's relative pointer delta straight to `int32_t`, and mutter
sends a NaN on capture activation, which casts to `INT32_MIN` and poisons the sub-pixel
accumulator for the life of the process, pinning the cursor to a client screen edge. Now rejects
non-finite and absurd deltas and resets the accumulator on each screen switch.

## AI tools used for this PR

Claude Code (Opus 5) - diagnosed from server logs, found the `INT32_MIN` poisoning, wrote the fix.

## Fixed/related issues

Fixes: #9886

## How has this been tested?

Moved the cursor off screen at 0,0 on GNOME Shell 50.1. The cursor is no longer pinned to the
client edge and moves freely, but still can't transition back to the server, so recovery is still
stopping the client.

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -852,9 +852,13 @@ void EiScreen::onMotionEvent(ei_event *event)
 
   auto dx = ei_event_pointer_get_dx(event);
   auto dy = ei_event_pointer_get_dy(event);
+  if (!std::isfinite(dx) || !std::isfinite(dy)) {
+    LOG_WARN("dropping pointer motion with non-finite delta: %g,%g", dx, dy);
+    return;
+  }
 
   if (m_isOnScreen) {
-    LOG_DEBUG("event: motion on primary x=%i y=%i)", m_cursorX, m_cursorY);
+    LOG_DEBUG("event: motion on primary x=%i y=%i", m_cursorX, m_cursorY);
     sendEvent(EventTypes::PrimaryScreenMotionOnPrimary, MotionInfo::alloc(m_cursorX, m_cursorY));
     if (m_portalInputCapture->isActive()) {
       m_portalInputCapture->release();


nbolton:
@sithlord48 I know I said I don't want to block the release, but... this seems pretty significant. Drives me nuts about once a day.

Edit: IDK if this even fixes it, so will test over a few days. Feel free to release without, as not sure if its worth the wait yet... 

Déjà vu!

sithlord48 (commented):
Does not change anything when using KDE plasma as server / client

nbolton:
OK, this should be OK to merge as it stops users falling into the trap of losing their mouse cursor control entirely, but the real fix is upstream in mutter. I'll file a bug.

nbolton:
@whot We're seeing one `motion_relative` frame with `dx = -nan` per input capture activation on
GNOME Shell 50.1. This PR drops non-finite deltas on our side.

In mutter's `meta_input_capture_session_process_event()`, `dx` and `dy` are uninitialised locals
and the return value of `clutter_event_get_relative_motion()` is ignored, so events without
`CLUTTER_EVENT_FLAG_RELATIVE_MOTION` send stack garbage.

libei doesn't check either: `eis_device_pointer_motion()` validates capability and state but not
the values, and there's no `isfinite` in `src/` -- is that by design?

Is this a mutter bug, a libei bug, or both? Want to file it in the right place.

sithlord48 (approved):
Quick test on Plasma no changes so this looks to be isolated enough for the fix

whot:
> libei doesn't check either: eis_device_pointer_motion() validates capability and state > but not the values, and there's no isfinite in src/ -- is that by design?

libei has upstream commit 67c28ba7de9a2bff8a1e2664ed0ccfa2b9b4bb17 which isn't in a release yet which will now disconnect on receiving NaN/Inf. libei doesn't do anything with the values other than passing them through, so if you're getting them then the issue is in the compositor.

sithlord48:
Thanks @whot for the input !

Edit: for the new release if we do get a NAN/INF will we then need to reconnect to the portal or this "should" happen automagicly ?

Issue #9845, opened 2026-06-10
Capture Input still shows after selection remembered (on screen wake)


nbolton:
### Before reporting a bug, please complete the sanity checks above.

- [x] I have done the sanity checks, and my issue persists.

### Deskflow version info

Deskflow: 1.26.0.236 (9b64a98a)
Qt: 6.11.1
System: Fedora Linux 44 (Workstation Edition)
Session: GNOME (wayland)

### Build types

Local developer build (built it myself)

### Deskflow configuration

GNOME server

### What steps will reproduce the problem?

1. Start server on GNOME
2. See Capture Input dialog
3. Click 'Remember this selection'
4. Accept/share
5. Sleep then wake computer

Expected: Preference is remembered and I never see dialog again

Actual: Dialog shows on screen wake

<img width="526" height="324" alt="Image" src="https://github.com/user-attachments/assets/ddea33ad-2980-4ce8-a995-c4a64a75a754" />

### Log output

To do

### Additional information

Gnome issue: https://gitlab.gnome.org/GNOME/gnome-remote-desktop/-/work_items/16
Workaround: Use the [allow-locked-remote-desktop](https://github.com/jikamens/allow-locked-remote-desktop/) extension

sithlord48:
The compositor can invalidate the token when it wants to .. Make sure we are storing it and it works in normal cases. if so maybe xdg-portal-gnome issue ?

nbolton:
I think the issue is that the validator is not ready to validate rather than the token being invalid... just a hunch. My suspicion is we need a "is portal ready to test this token" type check.  For instance, if I cancel the dialog and click start again, it then accepts the token.

I'll have a poke around.

koter84:
i have both my client and server on Fedora 44, both using Deskflow 1.26.0
on my client i get the checkbox, and it automatically works after a sleep, on my server i don't get the checkbox option, and probably because of that, it keeps asking after i lock and unlock my pc.

haven't checked logs yet, i'll try to get some logs soon

(also, just moved over to deskflow from input-leap)

[edit: i found out that i have to wait for GNOME 51 in Fedora 45]

sithlord48:
The dialog having a check or not is upto the version of portal being used by the backend. I am guess your server is using either libporal 0.9.x or the DE your using on the server is not updated to use the v2 input capture portal (that has the persistence option).

nbolton:
[...]

PR #10124, opened 2026-09-02
fix(wayland): Prevent clipboard causing portal reprompts if the backend is incorrectly reporting clipboard state of the previous token

ElCruncharino:
## Description

`PortalInputCapture` (server) and `PortalRemoteDesktop` (client) discard their saved xdg-desktop-portal restore token and force a fresh permission dialog whenever a restored session doesn't report clipboard as enabled, on the assumption the token predates clipboard support. If clipboard is never reported enabled (reported in #9845) the token gets discarded and reprompted on every single restore, forever.

Basically a one-shot "already retried" flag. It discards and retries once, then keeps the session instead of reprompting again if clipboard is still disabled.

## AI tools used for this PR

I wrote the fix. Claude (Sonnet 5) via Claude Code did code review, caught the ifdef mismatch and a memory leak in the client-path fallthrough.

## Fixed/related issues

related: #9845

## How has this been tested?

Running as my live deskflow client on my daily-driver machine: openSUSE Tumbleweed (20260830), KWin/Plasma 6.7.4, Wayland session, libportal 0.10.0, xdg-desktop-portal-kde6 6.7.4. Built from this branch and running continuously since, across normal sleep/wake and lock/unlock cycles, with no clipboard-portal reprompt recurring.

### DIFF
--- a/src/lib/common/Settings.h
+++ b/src/lib/common/Settings.h
@@ -43,6 +43,7 @@ public:
     inline static const auto XScrollScale = QStringLiteral("client/xScrollScale");
     inline static const auto LanguageSync = QStringLiteral("client/languageSync");
     inline static const auto RemoteHost = QStringLiteral("client/remoteHost");
+    inline static const auto XdpClipboardRetried = QStringLiteral("client/xdpClipboardRetried");
     inline static const auto XdpRestoreToken = QStringLiteral("client/xdpRestoreToken");
   };
   struct Core
@@ -121,6 +122,7 @@ public:
     inline static const auto SwitchDelay = QStringLiteral("server/switchDelay");
     inline static const auto SwitchDoubleTap = QStringLiteral("server/switchDoubleTap");
     inline static const auto Win32KeepForeground = QStringLiteral("server/win32KeepForeground");
+    inline static const auto XdpClipboardRetried = QStringLiteral("server/xdpClipboardRetried");
     inline static const auto XdpRestoreToken = QStringLiteral("server/xdpRestoreToken");
   };
 
@@ -334,6 +336,7 @@ private:
     , Client::DynamicConnectionRetry
     , Client::InvertYScroll
     , Client::InvertXScroll
+    , Client::XdpClipboardRetried
     , Log::ToFile
     , Log::GuiDebug
     , Server::DefaultLockToComputerState
@@ -343,6 +346,7 @@ private:
     , Server::EnableSwitchDoubleTap
     , Server::ExternalConfig
     , Server::RelativeMouseMoves
+    , Server::XdpClipboardRetried
   };
 
   // When checking the default values this list contains the ones that default to true.
@@ -362,7 +366,9 @@ private:
   // Settings saved in our State file
   inline static const QStringList m_stateKeys = {
       Gui::WindowGeometry
+    , Client::XdpClipboardRetried
     , Client::XdpRestoreToken
+    , Server::XdpClipboardRetried
     , Server::XdpRestoreToken
    };
 
--- a/src/lib/platform/PortalInputCapture.cpp
+++ b/src/lib/platform/PortalInputCapture.cpp
@@ -11,15 +11,12 @@
 #include "base/Event.h"
 #include "base/Log.h"
 #include "base/TMethodJob.h"
+#include "common/Settings.h"
 #include "deskflow/ClipboardTypes.h"
 #include "platform/EiClipboard.h"
 
 #ifdef HAVE_LIBPORTAL_CLIPBOARD
 #include "platform/PortalClipboard.h"
-#endif
-
-#ifdef HAVE_LIBPORTAL_INPUTCAPTURE_RESTORE
-#include "common/Settings.h"
 #endif
 
 #include <algorithm>
@@ -298,17 +295,20 @@ void PortalInputCapture::setupSession(XdpInputCaptureSession *session)
 
 #ifdef HAVE_LIBPORTAL_CLIPBOARD
   if (!xdp_session_is_clipboard_enabled(parentSession) && m_portalVersion > 1) {
-    // Restored sessions can pre-date clipboard support, leaving the channel
-    // disabled even though we requested it. Drop the saved token and recreate
-    // the session from scratch so the user gets a fresh permission dialog.
-    LOG_WARN("clipboard not enabled on session, discarding restore token to force a fresh session");
+    if (Settings::value(Settings::Server::XdpClipboardRetried).toBool()) {
+      // some backends never report clipboard enabled even when granted; don't loop forever
+      LOG_DEBUG("clipboard still not enabled on session after one retry, continuing without it");
+    } else {
+      LOG_WARN("clipboard not enabled on session, discarding restore token to force a fresh session");
 #ifdef HAVE_LIBPORTAL_INPUTCAPTURE_RESTORE
-    Settings::setValue(Settings::Server::XdpRestoreToken, QString());
+      Settings::setValue(Settings::Server::XdpRestoreToken, QString());
 #endif
-    g_object_unref(m_session);
-    m_session = nullptr;
-    g_idle_add([](gpointer data) { return static_cast<PortalInputCapture *>(data)->initSession(); }, this);
-    return;
+      Settings::setValue(Settings::Server::XdpClipboardRetried, true);
+      g_object_unref(m_session);
+      m_session = nullptr;
+      g_idle_add([](gpointer data) { return static_cast<PortalInputCapture *>(data)->initSession(); }, this);
+      return;
+    }
   }
 #endif
 
--- a/src/lib/platform/PortalRemoteDesktop.cpp
+++ b/src/lib/platform/PortalRemoteDesktop.cpp
@@ -100,28 +100,35 @@ void PortalRemoteDesktop::handleSessionStarted(GObject *object, GAsyncResult *re
 
 #ifdef HAVE_LIBPORTAL_CLIPBOARD
   if (!xdp_session_is_clipboard_enabled(session)) {
-    LOG_WARN("clipboard not enabled on remote desktop session, discarding restore token to force a fresh session");
-    Settings::setValue(Settings::Client::XdpRestoreToken, QString());
-    free(m_sessionRestoreToken);
-    m_sessionRestoreToken = nullptr;
-    if (m_selectionTransferSignalId) {
-      g_signal_handler_disconnect(session, m_selectionTransferSignalId);
-      m_selectionTransferSignalId = 0;
+    if (Settings::value(Settings::Client::XdpClipboardRetried).toBool()) {
+      // some backends never report clipboard enabled even when granted; don't loop forever
+      LOG_DEBUG("clipboard still not enabled on remote desktop session after one retry, continuing without it");
+    } else {
+      LOG_WARN("clipboard not enabled on remote desktop session, discarding restore token to force a fresh session");
+      Settings::setValue(Settings::Client::XdpRestoreToken, QString());
+      Settings::setValue(Settings::Client::XdpClipboardRetried, true);
+      free(m_sessionRestoreToken);
+      m_sessionRestoreToken = nullptr;
+      if (m_selectionTransferSignalId) {
+        g_signal_handler_disconnect(session, m_selectionTransferSignalId);
+        m_selectionTransferSignalId = 0;
+      }
+      if (m_selectionOwnerChangedSignalId) {
+        g_signal_handler_disconnect(session, m_selectionOwnerChangedSignalId);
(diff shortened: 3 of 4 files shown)


src/lib/platform/PortalRemoteDesktop.cpp
```diff
@@ -100,28 +100,35 @@ void PortalRemoteDesktop::handleSessionStarted(GObject *object, GAsyncResult *re
...
-      m_selectionOwnerChangedSignalId = 0;
-    }
-    if (m_sessionSignalId) {
-      g_signal_handler_disconnect(session, m_sessionSignalId);
-      m_sessionSignalId = 0;
+    if (Settings::value(Settings::Client::XdpClipboardRetried).toBool()) {
+      // some backends never report clipboard enabled even when granted; don't loop forever
+      LOG_WARN("clipboard still not enabled on remote desktop session after one retry, continuing without it");
+    } else {
+      LOG_WARN("clipboard not enabled on remote desktop session, discarding restore token to force a fresh session");
```
sithlord48, line 107: This should not be an warning It should be a debug message so its not always logged

src/lib/platform/PortalInputCapture.cpp
```diff
@@ -298,17 +295,20 @@ void PortalInputCapture::setupSession(XdpInputCaptureSession *session)
 
 #ifdef HAVE_LIBPORTAL_CLIPBOARD
   if (!xdp_session_is_clipboard_enabled(parentSession) && m_portalVersion > 1) {
-    // Restored sessions can pre-date clipboard support, leaving the channel
-    // disabled even though we requested it. Drop the saved token and recreate
-    // the session from scratch so the user gets a fresh permission dialog.
-    LOG_WARN("clipboard not enabled on session, discarding restore token to force a fresh session");
+    if (Settings::value(Settings::Server::XdpClipboardRetried).toBool()) {
+      // some backends never report clipboard enabled even when granted; don't loop forever
```
sithlord48, line 299: This sounds like a bug we should report to libportal or the backend that is not telling us it's granted.

src/lib/platform/PortalInputCapture.cpp
```diff
@@ -298,17 +295,20 @@ void PortalInputCapture::setupSession(XdpInputCaptureSession *session)
 
 #ifdef HAVE_LIBPORTAL_CLIPBOARD
   if (!xdp_session_is_clipboard_enabled(parentSession) && m_portalVersion > 1) {
-    // Restored sessions can pre-date clipboard support, leaving the channel
```
sithlord48, line 301: Instead of just doing this blindly perhaps its a better idea to instead check if we even have clipboard enabled in our settings. If settings are enabled and our token does not provide clipboard access we need to ask for a new one .. or if our token provides access and we don't have clipboard enabled we also need to ask for a new token.

src/lib/platform/PortalRemoteDesktop.cpp
ElCruncharino: Good catch on the log level, changed to LOG_DEBUG! I swear I had that configured, but with was late when I pushed 🤷

src/lib/platform/PortalInputCapture.cpp
ElCruncharino: Agreed. Will need to collect more info to do so and try to gumshoe that out.

src/lib/platform/PortalInputCapture.cpp
ElCruncharino: Gated the server side on `Settings::Server::EnableClipboard`, so it won't bother the user about clipboard if they don't have it enabled.

On the client side, there's no `Client::EnableClipboard` setting to check. The portal session negotiation in PortalRemoteDesktop.cpp happens at client startup, before any server connection exists, so I left that part as-is.

Thoughts on if you'd rather wire that up differently?

sithlord48: I think that works.

src/lib/common/Settings.h
```diff
@@ -44,6 +44,7 @@ class Settings : public QObject
     inline static const auto LanguageSync = QStringLiteral("client/languageSync");
     inline static const auto RemoteHost = QStringLiteral("client/remoteHost");
     inline static const auto XdpRestoreToken = QStringLiteral("client/xdpRestoreToken");
+    inline static const auto XdpClipboardRetried = QStringLiteral("client/xdpClipboardRetried");
```
sithlord48, line 46: I think my biggest problem with this is the need for these two additional state items.
I just don't want to add additional items to the settings when we do not need them.
Are these only because we do not trust if the return is clipboard is enabled from xdp ?

ElCruncharino: Yeah, we don't trust `xdp_session_is_clipboard_enabled() ` on backends that misreport it.

Tried an in-memory flag on EiScreen instead (it survives the PortalInputCapture recreation on EI_EVENT_DISCONNECT, so it fixes the lock/wake loop), but it resets on every restart. If a backend permanently misreports clipboard rather than just glitching once, that's one fresh reprompt per reboot, forever. So I think we need something that survives restarts to actually ask once, ever.

Open to another mechanism if you've got one.

ElCruncharino: FWIW, my issue with this was primarily on reboot, and that's the symptom I tried to address.

sithlord48: Yeah that is the upside of having those checks if we can't trust the `xdp_session_is_clipboard_enabled()`  we may not have a choice but to track it

ElCruncharino: I'm not a huge fan of it either, as your argument that it's _technically_ an upstream issue is probably the right one. Probably libportal related, but my dirty approach to the patch was really more about me just getting annoyed by the prompt more than I wanted to rabbit hole down to the source.

Will mark this thread as resolved, at least since we're on the same page, but lmk if you want me to cut this over from Draft. @nbolton open to criticism/suggestion from you as well. It's an unfortunate position to be in to decide whether workarounds belong in the project not entirely responsible for the issue. The wayland landscape is rough waters and slow moving as far as protocols and determinism goes for backends in my experience.

sithlord48, line 46: Can we move this before the other string trying to keep things mostly alphabetized. 
also in the other additions to structs

sithlord48 (changes requested):

src/lib/common/Settings.h
ElCruncharino: Done.

src/lib/platform/PortalInputCapture.cpp
```diff
@@ -297,18 +294,22 @@ void PortalInputCapture::setupSession(XdpInputCaptureSession *session)
   XdpSession *parentSession = xdp_input_capture_session_get_session(session);
 
 #ifdef HAVE_LIBPORTAL_CLIPBOARD
-  if (!xdp_session_is_clipboard_enabled(parentSession) && m_portalVersion > 1) {
-    // Restored sessions can pre-date clipboard support, leaving the channel
-    // disabled even though we requested it. Drop the saved token and recreate
-    // the session from scratch so the user gets a fresh permission dialog.
-    LOG_WARN("clipboard not enabled on session, discarding restore token to force a fresh session");
+  if (Settings::value(Settings::Server::EnableClipboard).toBool() && !xdp_session_is_clipboard_enabled(parentSession) &&
```
sithlord48: I guess also since we can't do this for clients lets just skip the check for now. sorry to ask then ask you to remove it.

ElCruncharino: No skin off my back. Can do.

src/lib/platform/PortalInputCapture.cpp
```diff
@@ -11,17 +11,14 @@
 #include "base/Event.h"
 #include "base/Log.h"
 #include "base/TMethodJob.h"
+#include "common/Settings.h"
```
sithlord48, line 14: you can remove this one too

src/lib/common/Settings.h
```diff
@@ -342,6 +344,8 @@ class Settings : public QObject
     , Server::EnableSwitchDoubleTap
     , Server::ExternalConfig
     , Server::RelativeMouseMoves
+    , Client::XdpClipboardRetried
```
sithlord48: place with other client names please for other group too 

i.e 
```
Client::bar
Client::foo
Server::bar
Server::foo
```

sithlord48 (changes requested):

src/lib/common/Settings.h
ElCruncharino: Ah, alphabetize and group, eh?

As you wish. Done.

src/lib/platform/PortalInputCapture.cpp
sithlord48: Oh sorry i forgot you added calls to settings yeah can you put that back

ElCruncharino: Hahaha, you had me convinced for a minute there was something I missed for stripping that, like a nested import bringing it in.

sithlord48 (commented):
Last question here sorry for so much for such a little change. 

When should the new settings be reset to false ? They should not stay true forever. Some reasons I would think to invalidate the token / requested bool 

1. Server Side: when the user saves the settings w/ a new settings for clipboard the server token and requested bool should be reset. 

1. If the client connects to a new server should the token be invalidated ? 

any other cases where this should be reset ?  We can add the token invalidation to another PR but i do think there we can discuss this here. (or in the matrix chat if your there)

sithlord48 (dismissed):
Tested this and working well but i think we should consider here or in a follow up when to invalidate the token / reset that we checked the clipboard bool

sithlord48:
@nbolton  leaving open for you to get a chance to review , please land after you do

nbolton (commented):
Doesn't seem to fix it on GNOME, at least not for me.  On my GNOME Wayland server I still get the dialog on wake.

I'll try again tomorrow though just in case I didn't test it right.

Idea: we could re-scope the PR to be just KDE if I confirm it's not a fix for GNOME (though I haven't tested _yet_ on KDE myself, I'm assuming @sithlord48 has).

sithlord48:
> Doesn't seem to fix it on GNOME, at least not for me. On my GNOME Wayland server I still get the dialog on wake. 

i wonder if your case is gnome invalidating the token not deskflow. 

On gnome does xdp_session_is_clipboard_enabled(parentSession) return true as it "should" . it appears for kde it may not (or a libportal issue ? )

ElCruncharino:
I had started looking into this earlier.

Looks like a [libportal](https://github.com/flatpak/libportal/blob/main/libportal/clipboard.c) race.

`xdp_session_request_clipboard()` is fire-and-forget:
```c

/**
 * xdp_session_request_clipboard:
 * @session: a [class@Session] in initial state
 *
 * Requests clipboard integration on the session.
 */
void
xdp_session_request_clipboard (XdpSession *session)
{
  GVariantBuilder options;

  g_variant_builder_init (&options, G_VARIANT_TYPE_VARDICT);
  g_dbus_connection_call (session->portal->bus,
                          PORTAL_BUS_NAME,
                          PORTAL_OBJECT_PATH,
                          "org.freedesktop.portal.Clipboard",
                          "RequestClipboard",
                          g_variant_new ("(oa{sv})",
                                         session->id,
                                         &options),
                          NULL,
                          G_DBUS_CALL_FLAGS_NONE,
                          -1, NULL, NULL, NULL);

  ensure_signal_listeners (session->portal);
}
```

No callback, so it returns before the backend processes it. We call this then immediately call `Start` right after, nothing waits. If `Start` lands first, `clipboard_enabled` comes back false even though we asked correctly.

Over in #9845 to quote @nbolton:

> I think the issue is that the validator is not ready to validate rather than the token being invalid... just a hunch. My suspicion is we need a "is portal ready to test this token" type check. For instance, if I cancel the dialog and click start again, it then accepts the token.

I'll concur with that hunch, at least having seen that snippet above. Maybe I'm missing another piece or so.

Still reviewing, so take this with a grain of salt but for GNOME it looks like a restored session replays `clipboard_enabled` straight from the old token blob instead of re-checking, so a pre-clipboard token stays `no` forever. KDE appears to fall back to a fresh dialog on that mismatch instead, but in theory either of them would hit the race. KDE just handles it a little nicer.

So: plausibly libportal's bug, not a KDE/GNOME difference.

sithlord48 (approved):
Looks ok we should report and like related libportal bugs to this pr. 

@nbolton leaving landing to you

nbolton:
> fixes: #9845

If this can only be fixed upstream [for GNOME], then perhaps this should say related? Also, if this is now KDE only, I think the title should reflect that.

ElCruncharino:
Issue filed: https://github.com/flatpak/libportal/issues/231

nbolton (approved):

PR #10126, opened 2026-09-02
hack(wayland): avoid sharing a pointer barrier endpoint at screen corners

nbolton:
## Description

This is a hack, real fix is upstream.

This prevents the cursor from crossing at 0,0 but should not be landed because it will also impact non-GNOME users.

## AI tools used for this PR

Claude Code (Sonnet 5) 

## Fixed/related issues

Requires PR #10122
Related: #9886

## How has this been tested?

Moving mouse from 0,0 on Wayland server

### DIFF
--- a/src/lib/platform/PortalInputCapture.cpp
+++ b/src/lib/platform/PortalInputCapture.cpp
@@ -524,6 +524,9 @@ void PortalInputCapture::addBarrier(
   gint y1 = 0;
   gint y2 = 0;
 
+  // a horizontal barrier must not share an end point with a vertical one: mutter's release only
+  // clears the held barrier and re-runs the warp from the trapped point, so a pointer sitting on
+  // both lines is caught again by the other before it moves and never leaves the corner
   using enum BarrierSide;
   switch (side) {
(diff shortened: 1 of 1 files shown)


nbolton:
Closing, useful as evidence for the mutter bug but we shouldn't merge it.

nbolton:
@whot Separate one from the NaN issue. When the cursor hits the corner at 0,0 it gets stuck on the
client. We release the capture but it stays put.

Our left and top barriers both cover that corner pixel. Trimming them so they don't overlap there
fixes it, which is what this PR does, but it felt like a hack so I closed it.

Any idea what's going on? Is it a bug in mutter or something we're doing wrong?

whot:
the portal spec says the barrier is on the left or top side of the pixel so 0/0 on both barriers is explicitly supported. I'm guessing this is a bug in the mutter implementation, *maybe* related to mutter using that corner for the activities overview?

Realistically: you're not in charge of the cursor so if you set up the barrier correctly and you release it when required, anything else is not something you can fix.

Issue #5007, opened 2015-09-07
Dead keys don't work when writing quickly


DiThi:
- Have a windows machine with the server active.
- Configure the keyboard layout to one with dead keys such as Spanish or French.
- Write a letter with accent quickly, e.g. "éééééééé" by alternating the accent key with the letter E.
- You'll be writing "éeeeeeee" unless you do it slow enough. It works correctly when the Synergy server is stopped.

It seems dead keys don't work when they are pressed while the previous key wasn't released yet.

DiThi:
On client machines the issue is even worse, sometimes putting the accent twice, e.g. "c´ámara" instead of "cámara".

juanmagl:
I confirm this behaviour. This has been happening to me since forever, at several machines.
I confirm that killing the server stops the wrong behaviour. Tried in Windows XP and 7.
This is what you get when repeatedly writting an accented letter: "ááá´´aáá".
It is very frustrating when you type at a considerable speed and you get disrupted by these glitches in normal text writing that make you need to correct your typing.

GViktorsson:
I am experiencing this same problem. My language commonly uses an acute accent on all vowels which is produced using a dead key. With the Synergy server enabled I routinely "lose" the accent character while typing fast and just get the vowel without an accent. Disabling the Synergy server fixes the problem right away. This makes it impossible for me to have Synergy constantly running so I've resorted to just starting the server whenever I need to use a client.

The problem can be easily reproduced as outlined by mbargiel in #5079 and doesn't even require one to type quickly. mbargiel's post perfectly summarizes the problem, which I'll restate here for the benefit of other readers:

"The exact problem appears to be when an accent key is pressed while another key is already down, and that key is released before the accent is "consumed". In that case, Synergy "swallows" the accent key."

Note that by turning the Synergy server off I immediately get the expected behaviour: dead key presses are never "lost". Note also that this happens without any client connected, just running the Synergy server is enough.

Synergy 1.8.8-stable-25a8cb2 (Build March 16th, 2017, tried both Basic and Pro license)
Windows 7 SP1 Professional
Keyboard layout: Icelandic (IS)

DiThi:
This still happens, and it's THE reason we don't use Synergy.

AlanFPS:
Can this be`reopened? It won't stop adding random commas and tildes to words when using spanish keyboàrd.

sithlord48:
I think this is just #6599

PR #10128, opened 2026-09-03
Fix Windows dead-key composition when typing quickly in RELAY mode

agvillaca:
## Description
Fixes garbled/lost dead keys on a Windows **server** while the cursor is on a client (RELAY mode), especially with Brazilian ABNT2 (`~` + `a` → `ã`).

When typing quickly, the previous letter’s key-up (e.g. `ç`) was still calling `ToUnicode` while a dead key was pending. That flushed the dead key as a spacing `~` even if the combining letter was never pressed. Restoring the Windows dead-key state in RELAY mode could also create phantom compositions.

Changes in `MSWindowsHook.cpp`:
- Skip `ToUnicode` on key-up; reuse the last mapped character for the release event
- Isolate `ToUnicode` key state to the current key + modifiers (ignore rollover)
- Do not put the dead key back into the Windows keyboard state in RELAY mode
- Suppress the dead key’s UP after a successful composition

Validated on Input Leap (same hook logic) with Windows 11 server + ABNT2, then ported here. Input Leap itself is archived, so this PR targets Deskflow as the active upstream.

## Fixed/related issues
related: #9254
related: #5007

## How has this been tested?
- Environment: Windows 11 server, Brazilian ABNT2 keyboard, Input Leap 3.x with this hook change installed (server binary), controlling a secondary client screen (RELAY mode).
- `~` then `a` typed quickly → `ã` only (no stray `~` / phantom `õ`)
- `ç` then `~` typed quickly and stop → only `ç` (dead key stays pending; no spacing `~`)
- Then `a` → `çã`
- `çã` typed quickly as one sequence works
- Slow typing still works
- Deskflow source change is the equivalent port; binary retest on Deskflow itself was not repeated after the port.


sithlord48 (changes requested):
From our hacking guide. 

> Do not add co-author tags for LLM models.

agvillaca:
Removed the co-author tag, thanks.

nbolton:
Thanks for the fix, I want to trace the dead-release handling against a couple of other layouts before merging, will get back to you shortly.

nslucas:
Thanks for fixing this. Is this PR expected to be included in the next official release, or will it appear in a continuous build first? I’m currently using a local build and would like to know which version will contain the fix.

sithlord48:
anything landed will be in continuous first.

nbolton:
Repro'd bug on `master`: on Windows server with ABNT2 layout, if `ç` (`;`/`;` on GB) and `~` (`'`/`@` on GB) then `a` is typed too quickly, it gives `ç~a` not `çã` on the client (macOS with GBPC layout, in my case).

On this PR, `agvillaca:fix-win-abnt2-dead-keys` the bug is fixed, I can type  `ç` + `~` + `a` quickly and no `~` char is typed.

Also tested on the client: fast typing of ordinary words drops no characters, US-International (where ~ needs shift) composes nã with no stuck shift afterwards, and ~q still gives a plain tilde.

Edit: I'll do the code review now.

src/lib/platform/MSWindowsHook.cpp
```diff
@@ -311,99 +345,117 @@ static bool keyboardHookHandler(WPARAM wParam, LPARAM lParam)
...
   }
 
-  // map the key event to a character.  we have to put the dead
-  // key back first and this has the side effect of removing it.
-  WCHAR wc[] = {0, 0};
-  setDeadKey(wc, 2, flags);
-
-  UINT scanCode = ((lParam & 0x10ff0000u) >> 16);
-  int n = ToUnicode((UINT)wParam, scanCode, keys, wc, 2, flags);
+  isolateKeyState(keys, vkCode);
```
nbolton, line 348: I removed this, and the fix still holds (`ç` + `~` + `a` composes fine without it).
Why is the `isolateKeyState` function needed?

src/lib/platform/MSWindowsHook.cpp
```diff
@@ -311,99 +345,117 @@ static bool keyboardHookHandler(WPARAM wParam, LPARAM lParam)
...
-    for (size_t i = 0; i < sizeof(keys) / sizeof(keys[0]); ++i) {
-      g_deadKeyState[i] = keys[i];
+  // Key-up must not call ToUnicode: with a pending dead key (e.g. ~ after
+  // ç on ABNT2) the previous letter's UP would flush the dead key as a
+  // spacing '~' even if the user never typed the combining letter.
+  if (kf_up) {
+    WCHAR prev = (vkCode < 256) ? g_lastChars[vkCode] : 0;
+    noAltGr = (vkCode < 256) && (g_lastNoAltGr[vkCode] != 0);
+    charAndVirtKey = makeKeyMsg((UINT)wParam, prev, noAltGr);
+  } else {
```
nbolton, line 362: @agvillaca The reason this diff is hairy to read is the indentation: about 90 lines moved right by two spaces without otherwise changing, so git reports them all as new. 

Edit: This is part of the reason why it took me so long to grok what had actually changed (that's my excuse and I'm sticking with it).

For a change this size it might have been more sensible to make the function more readable and pull the key-down path out into its own function.

The if/else would then look something like...

```cpp
  if (kf_up) {
    WCHAR prev = (vkCode < 256) ? g_lastChars[vkCode] : 0;
    bool noAltGr = (vkCode < 256) && (g_lastNoAltGr[vkCode] != 0);
    charAndVirtKey = makeKeyMsg((UINT)wParam, prev, noAltGr);
  } else {
    charAndVirtKey = mapKeyDown(wParam, lParam, keys, control, menu);
  }
```

nbolton (changes requested):
I have something stashed that works, so if you're busy I can just change your PR.

PR #10132, opened 2026-09-06
fix(mac): OS events memory leak

### LINKED ISSUE #8653: Memory leak when Deskflow app runs for a long time
### Project

Deskflow

### Deskflow version info

Deskflow: 1.22.0
System: MacOSX Sequoia 15.5

### Build type

### Operating systems (OS)

- [ ] Windows
- [x] macOS
- [ ] Linux (X11)
- [ ] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### Wayland on Linux

### Signing on macOS

- [ ] I am not using macOS

### Continuous build

### OS versions/distros

MacOSX Sequoia 15.5

### Deskflow configuration

two computers with MacOSX, pretty simple setup

1. Start Deskflow
2. Work every day, periodically putting computers to sleep
3. See in a week or more, memory footprint is growing bigger and bigger (1 Gb in my case)

### Log output

```shell

```

### Additional information

_No response_

pedrorambo:
## Description

Fixes a memory leak when using Deskflow on MacOS as a client. Every time it posted an OS event like mouse movement, scroll, key events or clicks, CGEventPost autoreleased a CFData internally, but the thread running the event loop was never draining that autorelease pool, so those objects were never being released. 

The solution was to push and pop the autorelease pool on each event dispatch, so that the pool drains on every iteration instead of once when the process ends.

This should not risk freeing objects incorrectly. Draining a pool releases objects that were already autoreleased. 

## AI tools used for this PR

Used Claude Code Opus 5 to diagnose the issue and implement the fix. Not used to create the description of this PR.

Note: I don't program in C++, so I just let Claude Code investigate the problem and then I made my way through the findings to understand the issue and let it write the fix and regression tests. The manual tests I did by myself and confirmed the issue was fixed.

## Fixed/related issues
fixes: #8653 

## How has this been tested?

I hit this in daily use before I went looking for it. If I let the process running for over 15+ minutes using the client mode, before the patch I noticed the allocation of CFData objects just growing (until it reached 659MB if running for more than 20h in previous tests). Then I let Claude investigate it further by reading the process memory dump and the source code. It found that it was caused by the issue mentioned in the description. Then after the fix, I did the same thing and the memory was kept stable.

Amount of allocations of CFData before the fix (pulled from the heap command on MacOS, one line per 5min interval): 

1. ~10,000 CFData allocations
2. ~12,000
3. ~16,000
4. ~20,000

Then after the fix:

1. 457 CFData allocations
2. 502
3. 502
4. 502

Environment used for all of the tests: Macbook Pro M1 running MacOS 15.7.9


src/lib/arch/ScopedAutoreleasePoolOSX.cpp
amagus, line 1: why this is a .cpp file? Shouldn't it be a .m? 
Why not name it OSX something like all the other mac exclusive files?

src/lib/base/EventQueue.cpp
```diff
@@ -166,6 +167,10 @@ bool EventQueue::getEvent(Event &event, double timeout)
 
 bool EventQueue::dispatchEvent(const Event &event)
 {
+  // The event loop thread has no run loop, so nothing else drains what Apple
+  // frameworks autorelease while handlers run.
+  deskflow::ScopedAutoreleasePool pool;
```
amagus, line 172: Can't we put the if (APPLE) here instead? Why do we need to create an empty object if we don't need it?

src/lib/arch/ScopedAutoreleasePoolOSX.cpp
```diff
@@ -0,0 +1,29 @@
...
+ * SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
+ * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
+ */
+
+#include "arch/ScopedAutoreleasePool.h"
+
+#if defined(__APPLE__)
+
+// Declared rather than included so this stays a plain C++ translation unit.
+extern "C" void *objc_autoreleasePoolPush(void);
```
amagus, line 12: This is somewhat a private API, given we now notarize our app, this can pose a problem.
Can't we create some wrapper in objc and use an @autoreleasepool instead?

src/lib/arch/ScopedAutoreleasePoolOSX.cpp
pedrorambo: I think it doesn't need to be a .m or .mm file because it doesn't contain any Objective-C syntax. It just calls C functions.

I'll update the file name.

amagus: Anyway, now that we know the cause of the issue, thanks a lot for that.
I'm preparing a substitute commit that tries to avoid calling private/undisclosed apis as well.
I will open a PR and ask for your validation shortly.

pedrorambo: Thank you so much for the feedback and review.

@amagus About the private APIs and notarization: It seems like Deskflow already imports private symbols through Qt.

amagus: We import the symbols that we want, but if we call them directly it may become a problem. 
We should only use APIs that are declared explicitly on the public header files. Using "external" symbols may be problematic.

Anyway, as I said, I was able to create a objc++ simple wrapper that provides everything that I think we would need.
I'm just waiting for CI validation to ensure it doesn't break other operating systems to submit a PR to the main repo.

pedrorambo: Understood. Thank you! You can ping me to test it here on my Mac to see if the issue is fixed.

amagus: Please try #10133 and let me know.

pedrorambo:
Replaced by #10133

PR #10133, opened 2026-09-06
fix(mac): run event queue events in autoreleasepool

### LINKED ISSUE #8653: Memory leak when Deskflow app runs for a long time

amagus:
## Description
The event loop thread was not properly being drained leaving a huge memory footprint if deskflow was used for a very long period unnatended.

It enforces events dispatched to run throughout an autoreleasepool.

Thanks to @pedrorambo help on doing diagnosing this issue.  

## AI tools used for this PR
No AI tool used for the implementation.

Anyway, this was done based on the changes suggested in #10132 by @pedrorambo which used Claude Code Opus 5 to diagnose the issue and implement the initial fix.

## Fixed/related issues
fixes: #8653
replaces: #10132 

## How has this been tested?

Verified that it caused no issue on my local mac.
Ran CI in my repo to verify builds passes on all other OS targets.

Pending validation by @pedrorambo (will move from draft after the fix is confirmed to be working)

### DIFF
--- a/src/lib/base/CMakeLists.txt
+++ b/src/lib/base/CMakeLists.txt
@@ -1,9 +1,17 @@
+# SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
 # SPDX-FileCopyrightText: (C) 2024 Chris Rizzitello <sithlord48@gmail.com>
 # SPDX-FileCopyrightText: (C) 2012 - 2024 Symless Ltd
 # SPDX-FileCopyrightText: (C) 2009 - 2012 Nick Bolton
 # SPDX-License-Identifier: MIT
 
-add_library(base STATIC
+if(APPLE)
+  set(PLATFORM_CODE
+    OSXAutoReleasePool.h
+    OSXAutoReleasePool.mm
+  )
+endif()
+
+add_library(base STATIC ${PLATFORM_CODE}
   BaseException.cpp
   BaseException.h
   DirectionTypes.h
--- a/src/lib/base/EventQueue.cpp
+++ b/src/lib/base/EventQueue.cpp
@@ -18,6 +18,10 @@
 
 #include <stdexcept>
 
+#if defined(__APPLE__)
+#include "base/OSXAutoReleasePool.h"
+#endif
+
 // interrupt handler.  this just adds a quit event to the queue.
 static void interrupt(Arch::ThreadSignal, void *data)
 {
@@ -166,16 +170,23 @@ bool EventQueue::getEvent(Event &event, double timeout)
 
 bool EventQueue::dispatchEvent(const Event &event)
 {
-  void *target = event.getTarget();
-  if (auto typeHandler = getHandler(event.getType(), target); typeHandler.has_value()) {
-    (*typeHandler)(event);
-    return true;
-  }
-  if (auto anyHandler = getHandler(EventTypes::Unknown, target); anyHandler.has_value()) {
-    (*anyHandler)(event);
-    return true;
-  }
-  return false;
+#if defined(__APPLE__)
+  auto dispatch = [&] {
+#endif
+    void *target = event.getTarget();
+    if (auto typeHandler = getHandler(event.getType(), target); typeHandler.has_value()) {
+      (*typeHandler)(event);
+      return true;
+    }
+    if (auto anyHandler = getHandler(EventTypes::Unknown, target); anyHandler.has_value()) {
+      (*anyHandler)(event);
+      return true;
+    }
+    return false;
+#if defined(__APPLE__)
+  };
+  return deskflow::runInAutoReleasePool(dispatch);
+#endif
 }
 
 void EventQueue::addEvent(Event &&event)
--- a/src/lib/base/OSXAutoReleasePool.h
+++ b/src/lib/base/OSXAutoReleasePool.h
@@ -1,0 +1,16 @@
+/*
+ * Deskflow -- mouse and keyboard sharing utility
+ * SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
+ * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
+ */
+
+#pragma once
+
+#include <functional>
+
+namespace deskflow {
+
+//! Run \p fn inside an Objective-C \c \@autoreleasepool block.
+bool runInAutoReleasePool(const std::function<bool()> &fn);
+
+} // namespace deskflow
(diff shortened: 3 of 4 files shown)


pedrorambo:
Validating locally...

pedrorambo:
I can confirm that the issue was fixed. I ran it again on master and the issue is still there (allocations of CFData always growing). On this branch the issue was fixed.

Thank you so much @amagus

I'm noticing stutters/freezes of ~1s every minute or so, but that's also happening on the current master. It's probably my local setup.

src/lib/base/CMakeLists.txt
```diff
@@ -37,6 +37,12 @@ add_library(base STATIC
   Unicode.h
 )
 
+if(APPLE)
```
sithlord48: Perhaps this would be better suited for the "arch" library.

amagus:
You are very welcome @pedrorambo ! Thanks a lot for taking your time to investigate this issue and find out where it was coming from.

src/lib/base/CMakeLists.txt
amagus: Indeed. I was just willing to have a quickfix and ended up taking shortcuts. Will adjust.

amagus: Fixed. Please check if the new organization looks good for you.

src/lib/base/OSXAutoReleasePool.h
```diff
@@ -0,0 +1,16 @@
+/*
```
sithlord48, line 1: Why not just `AutoReleasePool` ?

amagus: all files in arch folder have “Arch” in their name. I just followed the pattern… Can be renamed if you think it should.

sithlord48: Oh, well lets not do that at this time.

src/lib/base/OSXAutoReleasePool.h
nbolton, line 1: By convention, in our codebase "arch" means either Unix or Win32 and "platform" is for the next level up (Windows, macOS or "OS X", Linux as X11 or Wayland). 

For consistency I suggest this file is moved...
From: `src/lib/arch/mac/ArchAutoReleasePool.h` 
To: `src/lib/platform/OSXAutoReleasePool.h`

The OSX prefix is historical and of course no longer accurate, but rather than changing this convention piecemeal (which will likely cause confusion), we should stick to it and then change it in one clean and deliberate motion.

Edit: I see you got redirected to arch, but I stand by the above (platform is the convention).

src/lib/base/EventQueue.cpp
```diff
@@ -166,16 +170,23 @@ bool EventQueue::getEvent(Event &event, double timeout)
...
+    void *target = event.getTarget();
+    if (auto typeHandler = getHandler(event.getType(), target); typeHandler.has_value()) {
+      (*typeHandler)(event);
+      return true;
+    }
+    if (auto anyHandler = getHandler(EventTypes::Unknown, target); anyHandler.has_value()) {
+      (*anyHandler)(event);
+      return true;
+    }
+    return false;
```
nbolton, line 185: Might be slightly more readable to pull this out into its own function and do something like...

```cpp
#if defined(__APPLE__)
return deskflow::runInAutoReleasePool(/* lambda to someFunction */);
#else 
return someFunction();
#endif
```

amagus: I did it this way to avoid changes on other platforms, but looks like you are ok with it, so I will pursue it that way

src/lib/base/OSXAutoReleasePool.h
sithlord48: base is not linked to platform only arch..

src/lib/base/EventQueue.cpp
sithlord48: I also agree that the ifdef would be better

src/lib/base/OSXAutoReleasePool.h
amagus: This is exactly what I am seeing: there's some cyclic dependency: platform links to base which now links to platform back.
What should I do?

sithlord48: don't link platform to base. 
do put it in #if def code inside the event queue (for the second comment)

If the method is only used  in the one place maybe we should define it here but only for mac os ?

amagus: Ok, I got this to work, but then I had to add platform to the dependencies of base unit tests (for APPLE)

src/lib/base/OSXAutoReleasePool.mm
```diff
@@ -0,0 +1,20 @@
...
+ * SPDX-License-Identifier: GPL-2.0-only WITH LicenseRef-OpenSSL-Exception
+ */
+
+#import "platform/OSXAutoReleasePool.h"
+
+#import <Foundation/Foundation.h>
+
+namespace deskflow {
+
+bool runInAutoReleasePool(const std::function<bool()> &fn)
```
sithlord48, line 13: can't this be a mac only method in the event queue object ?

amagus: yes, but then I would need to either:

1) duplicate code (the same code that we have in other OS will need to be part of this mac only method) or
2) go back to having some #if defined(APPLE) block around the definitions of functions and/or lambdas (which you and nick already said it was not very readable)

Which one do you think would be best?

sithlord48: I perfer option 2, over duplicate code.

src/lib/base/OSXAutoReleasePool.h
nbolton: > don't link platform to base.

Good point. `base` should not depend on `platform`.

src/lib/base/EventQueue.cpp
amagus: please take a look on the next comment to understand why I went back to the ifdef with lambda only for mac (avoiding to create one extra function for non affected platforms)

sithlord48 (approved):

nbolton (approved):

PR #7850, opened 2024-11-02
build: Add Nix Flake & derivation scaffolding

shymega:
Blocked by: #7858

build: Add Nix Flake & derivation scaffolding
Also added a CI workflow for the Flake and derivation checks. Will be
upstreaming the derivation to Nixpkgs, but only on Deskflow releases.

Having the Flake & derivation in this repo allows people to build from
Git, or enter a development environment.

You will see `/.envrc` - this is for `direnv`, and if people have Nix
installed, it will automatically generate a developer environment,
Nixified.

One big benefit is that people can simply run the command:

`nix run github:deskflow/deskflow` - and Nix will fetch the latest
commit, build, and run Deskflow in one hit.

Later on, we should add 'Cachix' support, which means binaries will be
built and hosted on a Nix cache. This is free of charge.

Signed-off-by: Dom Rodriguez <shymega@shymega.org.uk>

PR #10140, opened 2026-09-08
build: add Nix flake for source-build package manager install

levonk:
## Description

Adds a Nix flake that builds Deskflow from source, enabling installation via `nix run` and `nix profile install` on Linux and macOS.

### Outputs

| Output | Description | Platforms |
|--------|-------------|-----------|
| `#default` / `#deskflow` / `#source` | From-source build | All 4 systems |
| `#nixpkgs` | Wraps the nixpkgs-packaged version | Linux only |
| `devShells.default` | Development shell with all build deps | All 4 systems |

### Supported systems

- `x86_64-linux`
- `aarch64-linux`
- `x86_64-darwin` (pinned to `nixpkgs-26.05-darwin` — nixpkgs-unstable dropped x86_64-darwin in 26.11)
- `aarch64-darwin`

### Build patches

The source build adapts the nixpkgs derivation with patches for:
- Qt translation file path (`qttranslations` is a separate package in nixpkgs)
- `QFile::filesystemFileName()` (nixpkgs Qt6 disables `cxx17_filesystem`)
- Static OpenSSL request on macOS (nixpkgs OpenSSL is shared only)
- XKB `evdev.xml` path substitution on Linux
- `macdeployqt` PATH on macOS (CMake checks for it unconditionally)

### Usage

```bash
# Run from source
nix run github:deskflow/deskflow

# Install from source
nix profile install github:deskflow/deskflow

# Use the nixpkgs-packaged version (Linux only)
nix run github:deskflow/deskflow#nixpkgs

# Development shell
nix develop github:deskflow/deskflow
```

### CI

Adds `.github/workflows/nix.yml` with:
- `flake check --all-systems --no-build` on Ubuntu
- Build on `x86_64-linux` and `aarch64-linux`
- Build and `nix run -- --help` on macOS

## AI tools used for this PR

- Devin (GLM-5.2 High) was used to generate the Nix flake, CI workflow, and documentation.

## Fixed/related issues

Related to the closed PR #7850 which added initial Nix scaffolding. This PR takes a different approach by building from source with a complete flake that exposes installable package outputs.

## How has this been tested?

- `nix flake check --all-systems --no-build` passes on all 4 systems
- `nix build .#deskflow` succeeds on `x86_64-darwin`
- `nix run .#deskflow -- --help` works on `x86_64-darwin`
- `statix check` and `deadnix` pass clean
- `nixfmt` formatting applied

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,74 @@
+# SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
+# SPDX-License-Identifier: MIT
+name: Nix
+
+on:
+  push:
+    branches: [master]
+    paths:
+      - flake.nix
+      - flake.lock
+      - deploy/nix/**
+      - .github/workflows/nix.yml
+  pull_request:
+    paths:
+      - flake.nix
+      - flake.lock
+      - deploy/nix/**
+      - .github/workflows/nix.yml
+
+permissions:
+  contents: read
+
+concurrency:
+  group: nix-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  flake-check:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+        with:
+          persist-credentials: false
+      - uses: cachix/install-nix-action@ba0dd844c9180cbf77aa72a116d6fbc515d0e87b # v27
+        with:
+          nix_path: nixpkgs=channel:nixpkgs-unstable
+      - run: nix flake check --all-systems --no-build
+
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - system: x86_64-linux
+            runner: ubuntu-latest
+          - system: aarch64-linux
+            runner: ubuntu-24.04-arm
+          - system: x86_64-darwin
+            runner: macos-13
+          - system: aarch64-darwin
+            runner: macos-latest
+    runs-on: ${{ matrix.runner }}
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+        with:
+          persist-credentials: false
+      - uses: cachix/install-nix-action@ba0dd844c9180cbf77aa72a116d6fbc515d0e87b # v27
+        with:
+          nix_path: nixpkgs=channel:nixpkgs-unstable
+      - uses: cachix/cachix-action@ad2ddac53f961de1989924296a1f236fcfbaa4fc # v15
+        with:
+          name: deskflow
+          # Create the cache at https://app.cachix.org and add CACHIX_AUTH_TOKEN
+          # to GitHub secrets, then remove skipPush to enable binary caching.
+          skipPush: true
+      - run: nix build .#deskflow --print-out-paths
+      # SECURITY: nix run executes the built binary outside the Nix sandbox on
+      # the runner, where it can reach GITHUB_TOKEN. On a PR, that binary is
+      # PR-controlled code. Drop this guard only if the workflow has no
+      # pull_request trigger.
+      - run: nix run .#deskflow -- --version
+        if: github.event_name != 'pull_request'
--- a/deploy/nix/deskflow.nix
+++ b/deploy/nix/deskflow.nix
@@ -1,0 +1,111 @@
+# SPDX-FileCopyrightText: (C) 2026 Deskflow Developers
+# SPDX-License-Identifier: MIT
+#
+# Deskflow derivation — builds from source using CMake/Qt6.
+# Imported by the root flake.nix.
+{
+  lib,
+  pkgs,
+  self,
+}:
+let
+  isLinux = pkgs.stdenv.hostPlatform.isLinux;
+  isDarwin = pkgs.stdenv.hostPlatform.isDarwin;
+in
+pkgs.stdenv.mkDerivation {
+  pname = "deskflow";
+  # Derive from the flake's git revision — CMake also reads git tags for the
+  # binary's internal version, so this stays honest with the actual source.
+  version = "unstable-${self.shortRev or "dirty"}";
+
+  src = self;
+
+  nativeBuildInputs = [
+    pkgs.cmake
+    pkgs.ninja
+    pkgs.pkg-config
+    pkgs.qt6.wrapQtAppsHook
+    pkgs.doxygen
+  ];
+
+  buildInputs = [
+    pkgs.qt6.qtbase
+    pkgs.qt6.qttools
+    pkgs.qt6.qttranslations
+    pkgs.openssl
+    pkgs.pugixml
+    pkgs.python3
+  ]
+  ++ lib.optionals isLinux [
+    pkgs.gtest
+    pkgs.libei
+    pkgs.libportal
+    pkgs.libx11
+    pkgs.libxkbfile
+    pkgs.libxinerama
+    pkgs.libxi
+    pkgs.libxrandr
+    pkgs.libxtst
+    pkgs.libxkbcommon
+    pkgs.gdk-pixbuf
+    pkgs.libnotify
+    pkgs.qt6.qtwayland
+    pkgs.qt6.qtdeclarative
+    pkgs.wayland
+    pkgs.wayland-protocols
+    pkgs.libsysprof-capture
+    pkgs.lerc
+  ];
+
+  cmakeFlags = [
+    "-DCMAKE_SKIP_RPATH=ON"
+    "-DSKIP_BUILD_TESTS=ON"
+    "-DBUILD_INSTALLER=OFF"
+  ];
+
+  preConfigure = lib.optionalString isDarwin ''
+    # CMake unconditionally checks for macdeployqt on macOS (even with
+    # BUILD_INSTALLER=OFF). The binary is in qtbase's bin directory.
+    export PATH="${pkgs.qt6.qtbase}/bin:$PATH"
+  '';
+
+  postPatch = ''
+    # Qt translation files are in a separate qttranslations package in nixpkgs.
+    substituteInPlace translations/CMakeLists.txt \
+      --replace-fail 'PATHS ''${QT_ROOT_DIR} PATH_SUFFIXES "translations" "share/qt/translations"' 'PATHS "${pkgs.qt6.qttranslations}/translations"'
+    # nixpkgs Qt6 disables the cxx17_filesystem feature, so
+    # QFile::filesystemFileName() is unavailable. Use the equivalent
+    # conversion via QString::toStdString().
+    substituteInPlace src/lib/net/SecureUtils.cpp \
+      --replace-fail "file.filesystemFileName()" "std::filesystem::path(file.fileName().toStdString())"
+  ''
+  + lib.optionalString isLinux ''
+    substituteInPlace src/lib/deskflow/unix/AppUtilUnix.cpp \
+      --replace-fail "/usr/share/X11/xkb/rules/evdev.xml" "${pkgs.xkeyboard_config}/share/X11/xkb/rules/evdev.xml"
+  ''
+  + lib.optionalString isDarwin ''
+    # The project requests static OpenSSL on macOS to avoid Apple's
+    # system OpenSSL. With nixpkgs OpenSSL (shared libs only), this is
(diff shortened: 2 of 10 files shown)


sithlord48:
I do not run NIX myself and have not idea how this should work. I wonder if @shymega would be so kind to look at this for us. I do recall at some point he was looking at add this support as well.

sithlord48 (commented):
Can you please amend the second commit into the first one, we do not allow for "fixing" of proposed commits with new ones. (if not we will fix it pre landing)

flake.lock
```diff
@@ -0,0 +1,78 @@
+{
```
sithlord48, line 1: This has the same problems as before can you put the files somewhere not in the root of the repo? 

Ideally these files would go in `deploy/nix`

sithlord48 (changes requested):
The new files are also missing copyright info they will cause the REUSE check to fail.

.github/workflows/nix.yml
```diff
@@ -0,0 +1,53 @@
+name: Nix
```
sithlord48, line 3: This might be an issue if nix artifacts need to be released.

sithlord48:
We do not allow for new files in the root of the project.

deploy/nix/deskflow.nix
```diff
@@ -0,0 +1,109 @@
...
+  pkgs,
+  self,
+}:
+let
+  isLinux = pkgs.stdenv.hostPlatform.isLinux;
+  isDarwin = pkgs.stdenv.hostPlatform.isDarwin;
+in
+pkgs.stdenv.mkDerivation {
+  pname = "deskflow";
+  version = "1.26.0";
```
sithlord48: This version is wrong. you will need to configure this file or otherwise inject the correct version that CMake is going to assign the built application.

shymega:
I seem to recall my previous PR adding Nix support was superseded by it already existing in Nixpkgs.

My previous PR - https://github.com/deskflow/deskflow/pull/7850 - is simpler than this one, as all it does is change the `src` attribute by overriding the upstream package to the root path of this repo, at any particular commit after landing it.

~~Also, this PR seems to be AI-generated, and is not disclosed. That to me is a red flag.~~

This PR is also AI-generated, which is worth considering.

As for adding a dependency for `x86_64-darwin`, this might be beneficial, but I don't like the behaviour in the Flake.

@sithlord48 I think its worth reopening https://github.com/deskflow/deskflow/pull/7850, because a) it was raised first, and has had more testing and b) uses the upstream Nixpkgs derivation for the package, ensuring two layers of testing.

levonk:
Thanks for the review, @sithlord48! I've addressed all the requested changes in a single squashed commit (force-pushed).

## 1. Commit squash

Both commits have been squashed into a single commit as requested.

## 2. REUSE copyright headers

Added SPDX headers to all new files:
- `flake.nix` — inline `SPDX-FileCopyrightText` / `SPDX-License-Identifier`
- `deploy/nix/deskflow.nix` — same
- `deploy/nix/outputs.nix` — same
- `.envrc` — same
- `.github/workflows/nix.yml` — same
- `flake.lock` — added to `REUSE.toml` (can't have inline comments in JSON)

## 3. Files moved to `deploy/nix/`

The flake implementation now lives in `deploy/nix/`:
- `deploy/nix/deskflow.nix` — the derivation (build inputs, patches, cmake flags)
- `deploy/nix/outputs.nix` — per-system outputs (packages, apps, devShells)

`flake.nix` remains at the repository root as a thin shim (~47 lines) that just declares inputs and delegates to `deploy/nix/`. This is a hard Nix requirement — `nix run github:deskflow/deskflow` only works if `flake.nix` is at the repo root. There's no way around this; it's how the Nix flake protocol discovers the flake entry point. The root `flake.nix` contains no build logic — just the input declarations and a one-line `import ./deploy/nix/outputs.nix`.

`flake.lock` also must stay at the root (Nix requires it next to `flake.nix`), but it's a generated file with no hand-written content.

## 4. Release artifacts concern

The flake does not produce or require release artifacts. It's a build recipe — when a user runs `nix run github:deskflow/deskflow`, Nix fetches the source and builds it on their machine. It doesn't add anything to the GitHub release process, the `deploy/` packaging workflow, or the CI build matrix.

The `#nixpkgs` output simply wraps the existing [nixpkgs Deskflow package](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/de/deskflow/package.nix) — it doesn't duplicate or conflict with it. It's exposed as a convenience so users can try the nixpkgs-packaged version (which may include distribution patches) via the same flake. The nixpkgs package has its own release cycle independent of this repo.

The CI workflow (`.github/workflows/nix.yml`) only runs `nix flake check` and `nix build` — it doesn't upload artifacts or interact with the release process. It's scoped to only trigger on changes to `flake.nix`, `flake.lock`, `deploy/nix/**`, and the workflow itself, so it won't run on unrelated changes.

Happy to make further adjustments if needed. Also happy to hear any feedback from @shymega if they have time to take a look.

shymega:
@sithlord48 The response above was purely written by a LLM. Can we be sure this Nix contribution will be actively maintained by a human, or at least with human oversight? I have my doubts...

levonk:
@shymega Thanks for taking a look — I appreciate the Nix expertise.

You're right that #7850 is simpler, and the `overrideAttrs` approach of just swapping the `src` attribute on the nixpkgs package is cleaner where it works. I'd be happy to adopt that approach for Linux if it means inheriting nixpkgs's fixup logic and testing.

The blocker I ran into is that the nixpkgs Deskflow package is Linux-only — `meta.platforms` contains no Darwin entries:

```
$ nix eval nixpkgs#deskflow.meta.platforms --json
["aarch64-linux","x86_64-linux",...]  # no darwin
```

So the override approach from #7850 doesn't cover macOS at all — neither Intel nor Apple Silicon. That's the gap this PR tries to fill with the from-source build on Darwin.

A hybrid approach might be the best of both: use `overrideAttrs` on Linux (inheriting nixpkgs's packaging and testing) and fall back to the from-source build on Darwin (the only option, since nixpkgs doesn't build Deskflow for macOS). I'm happy to restructure the PR that way if it's preferable.

On the AI question — yes, I used AI tooling to help generate the flake. I find it useful for the kind of iterative build-fix-debug loop that Nix packaging involves. The PR discloses this in the description. I'm here to maintain it and respond to feedback, and the CI workflow catches breakage on changes to the flake or its inputs. If there's a specific concern about the code quality or correctness, I'm glad to address it.

sithlord48:
@shymega  agreed my main issue w/ the previous is that i would like the nix files to be in deploy/nix . If your looking to update that PR Id be willing to look at landing. We should be releasing 1.27.0 soon and that maybe good for having as a base version for nix. We can talk here or on matrix if your around.

levonk:
@shymega @sithlord48 I looked at #7850 to see what it would take to update it. The main blocker is that it doesn't work on macOS — neither Intel nor Apple Silicon — and that's the #1 reason I opened this PR.

Specifically:

**`x86_64-darwin`**: #7850 uses `nix-systems/triplet` which only lists 3 systems (`aarch64-darwin`, `aarch64-linux`, `x86_64-linux`). There's no `x86_64-darwin` output at all — `nix eval` returns "does not provide attribute `packages.x86_64-darwin`".

**`aarch64-darwin`**: Listed in the triplet but broken. It does `pkgs.deskflow.overrideAttrs` on the nixpkgs package, but the nixpkgs Deskflow derivation has `meta.platforms = lib.platforms.linux` and inherits Linux-only buildInputs (`libei`, `libportal`, `libx11`, `libxkbfile`, `wayland`, `libsysprof-capture`, `lerc`, `gdk-pixbuf`, `libnotify`, etc.). These don't build on Darwin. The derivation evaluates but `nix build` would fail.

The override approach only works where nixpkgs already builds the package — and nixpkgs only builds Deskflow for Linux. That's the fundamental gap.

For context: #7850 was closed in November 2024, well before this PR was opened. I opened this PR because there was no active Nix work and macOS users had no path forward.

This PR already has the Nix files in `deploy/nix/` (per earlier feedback from sithlord48 — #7850 placed them in the repo root), and the latest push adopts the hybrid approach: `overrideAttrs` on Linux (inheriting nixpkgs' fixup logic and testing) and from-source on Darwin (the only option since nixpkgs doesn't cover macOS). Only `flake.nix` and `flake.lock` remain at the root, which is a hard Nix requirement — the flake protocol only discovers entry points at the repo root.

On the version — the flake derives it from the git revision (`unstable-<shortRev>`), so it always matches the source. I can pin it to a release tag if maintainers prefer.

Happy to talk on matrix if it's easier to work through the details.

<details>
<summary>Full gap comparison: #7850 vs this PR</summary>

| Gap | #7850 | This PR |
|-----|-------|---------|
| Nix files in `deploy/nix/` | No (files in repo root) | Yes (per maintainer request; only `flake.nix`/`flake.lock` at root — a hard Nix requirement) |
| `x86_64-darwin` support | No (not in system list) | Yes (legacy nixpkgs pin) |
| `aarch64-darwin` support | Broken (Linux-only deps inherited) | Yes (from-source build) |
| `x86_64-linux` support | Yes (`overrideAttrs` on nixpkgs) | Yes (`overrideAttrs` — inherits nixpkgs packaging, can customize) |
| `aarch64-linux` support | Yes (`overrideAttrs` on nixpkgs) | Yes (`overrideAttrs` — inherits nixpkgs packaging, can customize) |
| `apps` outputs | No | Yes |
| `#source` output | No | Yes |
| `#nixpkgs` output | No | Yes (Linux) |
| Nix package version | `"git"` (honest but not parseable) | `unstable-<shortRev>` (derived from git, always matches the source) |
| macOS binary path handling | No | Yes (`.app/Contents/MacOS/`) |
| README docs | No | Yes |
| REUSE/SPDX for `flake.lock` | No | Yes |
| `.gitignore` for Nix | No | Yes |
| macOS CI | No | Yes |
| `--no-build` all-systems check | No | Yes |
| Dev shell tooling | `devenv` (extra flake input, lock entries, and `nixConfig` for its cache — all for the same result `mkShell` gives in one line) | Plain `mkShell` (same dev environment, no extra deps or complexity) |
| Cachix / binary cache | `nixConfig` for upstream devenv cache (only needed because devenv was pulled in; doesn't cache Deskflow builds) | `cachix-action` in CI for Deskflow builds (optional, `skipPush` until cache created) |

</details>

levonk:
Also @shymega I responded to your PR inviting support to get Darwin apple silicon back on track. I think my suggestions may help your PR there: https://github.com/NixOS/nixpkgs/pull/535029#issuecomment-5601270675

sithlord48:
This is just worse and worse every time you add stuff. We do not want ANY new files in the root of the project. I am gonna close this. when you remove all the new files from the project root and we have someone to maintain it we can look at it again.

levonk:
Sorry @sithlord48 .envrc leeched out from my build process, I didn't intend to add it. I was actively working on an improvement in the CI process. Just to confirm, does that include flake.nix? That wouldn't be possible with the Nix package manager, so there is no redemption for any of our pull requests. I believe flake.lock can technically be skipped but it sacrifices discipline. Should I open a new PR with just flake.nix or just give up?

Either way, thanks for your consideration.

sithlord48:
you added 3 files to the root. Id rather someone who knows about how nix packages work make the nix package.
I am unsure if nix packaging should even be here we don't have a debian folder either because it should happen downstream in the distro. I would suspect nix is the same. We only make a debian package in repo because cmake can do it with ease for nix i would think we want to do something like what is done for arch in the repo but if the nix build is not producing artifacts then i don't see the point of even having that stuff here.. 

Edit: isn't this is upstream package for nix ? https://search.nixos.org/packages?channel=26.05&query=deskflow#show=deskflow

levonk:
@sithlord48 Fair point on the root files — `flake.nix`/`flake.lock` have to sit at the repo root for the flake protocol to resolve `github:deskflow/deskflow`, so they can't move under `deploy/nix/`. I understand if that's a dealbreaker.

On the downstream point, you're right that packaging belongs there, and that's where I'm putting the effort — I'm helping get Apple Silicon support into the central nixpkgs package (NixOS/nixpkgs#535029). But x86_64-darwin can't live there; nixpkgs is dropping the platform entirely. So just to be explicit about the tradeoff: closing this means Intel Macs have no Nix path at all, and there's no way to `nix run` the latest source — nixpkgs only ever ships tagged releases. If either of those matter to the project, the flake is the only place they can exist, and I'm committed to maintaining it.

Regardless of what you decide on hosting Nix files here, I appreciate the consideration and the time you put into reviewing this.

sithlord48:
Its fine if there is not a x86_64 darwin package via nix. The affected users can just use the app we provide from our normal download page.

levonk:
Understood. Thanks for the consideration, and your work on Deskflow.

PR #10146, opened 2026-09-10
fix(wayland): track held mouse buttons in EiScreen

drizzt:
## Description
`EiScreen::isAnyMouseButtonDown()` always returned false, so the server never saw the screen as locked and could switch away in the middle of a drag.

A drag started on the server screen still cannot cross the barrier under KWin (I didn't tested GNOME), which ignores pointer barriers while a button is held. That side is being fixed in KWin itself: https://invent.kde.org/plasma/kwin/-/merge_requests/9891 ("plugins/eis: Fix input capture barrier activation"). This PR only fixes the other direction: the server no longer switches away while a button is down.

## AI tools used for this PR
No AI tool used for the code implementation. It was just used for commit message.

## How has this been tested?
Builds on Linux (Wayland, GCC, C++20). Full test suite passes, 27/27.
Tested the build by keeping the left/right button (I did 2 tests) pressed while in another screen and the cursor correctly doesn't exit the screen.

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -22,6 +22,7 @@
 #include "platform/PortalRemoteDesktop.h"
 
 #include <algorithm>
+#include <bit>
 #include <cmath>
 #include <cstdlib>
 #include <cstring>
@@ -261,9 +262,13 @@ std::int32_t EiScreen::getJumpZoneSize() const
   return 1;
 }
 
-bool EiScreen::isAnyMouseButtonDown(uint32_t &) const
+bool EiScreen::isAnyMouseButtonDown(uint32_t &buttonID) const
 {
-  return false;
+  if (m_buttons.none())
+    return false;
+
+  buttonID = std::countr_zero(m_buttons.to_ulong());
+  return true;
 }
 
 void EiScreen::getCursorCenter(int32_t &x, int32_t &y) const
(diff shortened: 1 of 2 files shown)


sithlord48:

sithlord48:
I did a quick test here and i do not see any change in the behavior if i am dragging i can still move to the other screen while dragging...

drizzt:
> I did a quick test here and i do not see any change in the behavior if i am dragging i can still move to the other screen while dragging...

what do you mean? with my patch, when I'm on another (non server) screen and I keep a button pressed I CAN'T change screen (and in debug log I correctly see):

```
[2026-09-10T22:50:39.487] DEBUG: locked by mouse buttonID: 1
        /home/drizzt/dev/sources/deskflow/src/lib/deskflow/Screen.cpp:361
```

without my patch I don't see the DEBUG log at all and of course the cursor is not locked to the screen and came back to server

sithlord48:
i was testing going from server to client w/ a held button.  i will try from client -> server also

drizzt:
> i was testing going from server to client w/ a held button. i will try from client -> server also

for server to client, it must be fixed in kwin and in any other server (it's not possible to fix it in barrier itself)

sithlord48 (approved):
Confirmed when leaving a wayland client holding a mouse button you can not change screens.

drizzt:
I sent also the [MR](https://invent.kde.org/plasma/kwin/-/merge_requests/9891) for the other side (kwin) of the issue and updated the description of this PR with it.

nbolton (approved):

Issue #7216, opened 2022-08-26
Add a keystroke/shortcut blacklist


robmonte:
### Problem ###

Sometimes when I am controlling a client device from my server, I want to hit a certain key or a key combination that will still trigger something on the server device.

### Solution ###

A blacklist of keys or key combinations that, when pressed, will always be sent to the server device.

### More Info ###

Example:
If I set Shift+F12 to Discord Mute on my server PC and I add "Shift+F12" to the blacklist, then when controlling my Mac I can still press Shift+F12 to mute Discord without requiring me to first move the cursor back to the PC.

robmonte:
Still a useful feature request.

nbolton:
Thanks for the feature request. I edited the title and description. Does it still represent your original idea?

bjsvedin:
I am not the original poster, but this is a feature request i agree with. The main use for this I have is the media buttons, pause, play, volume up, volume down, and skip. The server computer will be playing music, and I would like to control the music with key strokes no matter what machine is currently in focus. The current title and description is exactly the feature I am looking for.

robmonte:
I missed the original follow-up question, but yes your edits represent my original ask perfectly.

sithlord48:
This should already work, Try setting a hotkey / action on the server config so that the keys / combos you want are only sent to the system you want. .

bjsvedin:
I tried this today on version 1.24.0.0 (56a1bd68), and the hotkey configuration would not accept the media play/pause  key as an input. 

I also tried a couple other key combo variants to get around that, but i kept running into a "*** buffer overflow detected ***: terminated" error when starting up.

I created one hot key with an action of "Press and release key(s)" and left only my server computer as a receiver.

sithlord48:
1.24.0.0 has some issues you should try again with 1.24.0.343+

bjsvedin:
Sorry for the late response. I am currently on 1.26.0 of Deskflow. I'm on a KDE Plasma 6.6.2 Wayland environment, and I am still unable to set an audio key has hotkey. Is there a way to manually set these keys in the config file?

sithlord48:
I think but also i don't expect it to work on wayland as hotkeys on wayland are not supported (since libportal does not support them yet )

johnsynnott:
I'm not sure if this is the correct place to put this, so please let me know if I should open a new issue instead.

[...]

PR #10150, opened 2026-09-14
feat(server): add configurable local-only key codes (localKeys)

dhalem:
## Summary

Adds a new `localKeys` server option that lists virtual key codes to be
**always delivered locally to the host system** instead of being forwarded to
the client, regardless of which screen has focus.

This makes the previously hardcoded macOS key handling configurable, and
provides a cross-platform hook for "local-only" keys.

Refs #7216

## Design

On `master`, global server options are read from the Qt `QSettings` store
(see `Settings::m_validKeys`), **not** from the `[options]` section of the
config file (that section is now parsed only for `keystroke`/`mousepress`
filter rules). This change follows the master architecture:

- `Settings.h`: registers a new `server/localKeys` key in `struct Server` and
  adds it to the `m_validKeys` whitelist (required — `cleanSettings()` strips
  any unregistered key on load).
- `Config.cpp` / `Config.h`: parses the comma/space-separated list into
  `std::vector<uint32_t> m_localKeys`, exposed via `getLocalKeys()`.
- `IPlatformScreen.h`: adds a non-pure virtual `setLocalKeys()` (default
  no-op), so existing platform implementations are unaffected.
- `OSXScreen.h` / `OSXScreen.mm`: implements `setLocalKeys()` / `isLocalKey()`
  and, in the `CGEventTap` key handler, intercepts local-only keys so they are
  delivered to the host and **never forwarded** to the client.
- `Server.cpp`: applies the list to the primary screen's platform screen in
  `processOptions()` (runs on startup and on config reload).

### Usage

Set the key codes in the Deskflow settings store (QSettings), e.g.:

```
server/localKeys=64,79,80,90
```

## Scope

Minimal: Settings key + parsing + core feature + macOS implementation. A GUI
field for `localKeys` is intentionally **not** included in this PR (follow-up).

## Testing

- Builds cleanly on macOS arm64 against `master` (no new warnings).
- `deskflow-core` links and runs (version smoke test passes).
- Verified the local-only key path: keys in the list are delivered to the host
  and not forwarded to the client.

## Notes

- Based on `master`, not `v1.26.0` (the two have diverged significantly;
  `master` reads global options from QSettings).
- An unrelated local-only build fix in `cmake/Libraries.cmake` (pthread) is
  intentionally **excluded** from this PR and will be proposed separately.


sithlord48:

Issue #8909, opened 2025-09-01
Repeated 'failed to initialize remote desktop session' error while locked


nbolton:
### Project

Deskflow

### Deskflow version info

Deskflow: 1.23.0.162 (39cb3dc0)
Qt: 6.8.2
System: Fedora Linux 41 (Workstation Edition)
Session: GNOME (wayland)

### Build type

Local developer build (built it myself)

### Operating systems (OS)

- [ ] Windows
- [ ] macOS
- [ ] Linux (X11)
- [x] Linux (Wayland)
- [ ] BSD-derived
- [ ] Other (please specify)

### OS versions/distros

Fedora 41 GNOME Wayland 

### Deskflow configuration

Client: Fedora 41

### What steps will reproduce the problem?

1. Start as client on Wayland
2. Lock screen

Expected: Remote desktop portal isn't attempted while computer is locked
Actual: Continuous portal errors in log (unnecessarily noisy)

### Log output

```shell
[2025-09-01T07:52:49] ERROR: failed to initialize remote desktop session: CreateSession failed
	/home/nick/Projects/deskflow/src/lib/platform/PortalRemoteDesktop.cpp:119
[2025-09-01T07:52:50] ERROR: failed to initialize remote desktop session: CreateSession failed
	/home/nick/Projects/deskflow/src/lib/platform/PortalRemoteDesktop.cpp:119
[2025-09-01T07:52:51] ERROR: failed to initialize remote desktop session: CreateSession failed
	/home/nick/Projects/deskflow/src/lib/platform/PortalRemoteDesktop.cpp:119
[2025-09-01T07:52:52] ERROR: failed to initialize remote desktop session: CreateSession failed
	/home/nick/Projects/deskflow/src/lib/platform/PortalRemoteDesktop.cpp:119
[2025-09-01T07:52:53] ERROR: failed to initialize remote desktop session: CreateSession failed
	/home/nick/Projects/deskflow/src/lib/platform/PortalRemoteDesktop.cpp:119
[2025-09-01T07:52:54] ERROR: failed to initialize remote desktop session: CreateSession failed
	/home/nick/Projects/deskflow/src/lib/platform/PortalRemoteDesktop.cpp:119
[2025-09-01T07:52:55] ERROR: failed to initialize remote desktop session: CreateSession failed
	/home/nick/Projects/deskflow/src/lib/platform/PortalRemoteDesktop.cpp:119
[...]

PR #10159, opened 2026-09-18
fix(wayland): wait for the desktop to unlock before restoring portal sessions

### LINKED ISSUE #9845: Capture Input still shows after selection remembered (on screen wake)

nbolton:
## Description

We were asking the portal for input capture while the screen was locked, which the compositor rightly refuses, leading to the re-prompt. Now we wait until the desktop is unlocked and awake, and only then ask for permission. Probably unnoticed on KDE, which perhaps doesn't close the session on screen lock.

## AI tools used for this PR

Claude Code (Opus 5) - found bug was in asking for permission when it can't be granted, fixed by asking for permission when unlocked.

## Fixed/related issues

fixes: #9845

## How has this been tested?

Locked the screen on the GNOME Wayland server, then wait a few seconds and unlock.

Cursor successfully crossed to the client after unlock with no permission prompt.


src/lib/platform/XDGSessionMonitor.cpp
```diff
@@ -0,0 +1,174 @@
+/*
```
sithlord48, line 1: Why are you not using Qt for  any of this ? Also maybe this should be part of the XDGPower manager since that just handles the sleep service.. perhaps we should extend it to handle both since it has all the dbus parts all ready

sithlord48 (commented):
This is alot of code to fix a GNOME specific issue have we tried talking to the gnome team about the behavior ? 

Why not just suggest another extension for gnome it has to be safer and easier for us to maintain then trying to fix every design choice SOME DE makes ...

https://github.com/deskflow/deskflow/issues/8909#issuecomment-5253066706
> With this extension, you don't get this error message and you can remote control a locked client: https://github.com/jikamens/allow-locked-remote-desktop/ Gnome issue: https://gitlab.gnome.org/GNOME/gnome-remote-desktop/-/work_items/16

nbolton:
@sithlord48 From looking at how GNOME behaves, rejecting the request when it's not possible to allow it (ie screen locked), seems sensible... and us checking to see if the system is in a state where it could be accepted seems rational to me -- but if you're against it, I understand. We can close this if you're against it, happy to maintain it downstream instead.

sithlord48:
> @sithlord48 From looking at how GNOME behaves, rejecting the request when it's not possible to allow it (ie screen locked), seems sensible... and us checking to see if the system is in a state where it could be accepted seems rational to me -- but if you're against it, I understand. We can close this if you're against it, happy to maintain it downstream instead.

This should be fixed in GNOME not here. There is an open work item to [fix this](https://gitlab.gnome.org/GNOME/gnome-remote-desktop/-/work_items/16) and currently a [GNOME extension](https://github.com/jikamens/allow-locked-remote-desktop/) for users who want to work around this issue. IMHO this will be resolved in future GNOME and or just like we do with the tray icon by telling the user about the extension.

Issue #9959, opened 2026-07-11
Port Screen Configs to General config file


sithlord48:
Move the options for the screens to the general config

Done: 
 - aliases #9880 

Todo: 
 - Name 
 - Duplex options
 - corners

PR #10161, opened 2026-09-19
Move more screen options to the general config

sithlord48:
## Description

Continue to work on getting the configurations unified. Along with some support items the following notable things have been changed.
 - Move `screenName` to general config (not removed from old yet as it used to read other keys) 
 - Move `HalfDuplexCapsLock`, `HalfDuplexNumLock`, `HalfDuplexScrollLock` and `xTestIsXiramanUnaware` to general config
 - Move `switchCornerSize` to the general config
 - Move unused ? `perserveFocus` (fix exposure later) 
 - some clean up in the config for unused method
## AI tools used for this PR

 None

## Fixed/related issues

Part of: #9959 
Maybe: https://github.com/deskflow/deskflow/issues/10162 
Unified config

## How has this been tested?

Locally run


src/lib/server/Config.cpp
```diff
@@ -547,6 +546,51 @@ void Config::readSectionOptions(ConfigReadContext &s)
 
 void Config::readSectionScreens(ConfigReadContext &s)
 {
+  const auto screens = Settings::knownScreens();
+  for (const auto &screen : screens) {
```
nbolton, line 580: I need to test the dead corners settings to make sure they're working, does the loop run before anything is in `m_map`?

sithlord48: m_map is filled just after this i was planning to move that here when the modifiers were done. Does it need to be set before the corners work as expected ?

sithlord48: Updated to move the add screen call into this loop.

nbolton (dismissed):
Since you're in a rush: LGTM, I'm swamped today and don't have time to test.

Edit: One small thing, if a core won't start issue comes in from someone using an external `deskflow-server.conf`, it could be `Config.cpp:623` and the log line will be unknown screen name.

sithlord48:
>  One small thing, if a core won't start issue comes in from someone using an external deskflow-server.conf, it could be Config.cpp:623 and the log line will be unknown screen name.

Good point, We can have a better message in this case. ( I've added that to the PR)

I was thinking if were to move the seeding of options to addScreen we can call that for our upper loop and when parsing the server config if the external config is being used we can call addScreen if one exists in the server but not general config . This will add it to the general config so any options could be set on that screen.  Can you see any reason for us not to do this ?

sithlord48:
@nbolton  This should be good now for either config source. we now set these in the addscreen method.

sithlord48:
Please note the sonar warning will be resolved when we move the modifiers and thus the need to check within the old screen options

nbolton (approved):

